PR #16, opened 2024-05-04
add examples

brianjking:


Blaizzy:
Hey @brianjking 

Thanks for this PR!

README.md
```diff
@@ -17,4 +17,19 @@ pip install mlx-vlm
...
+* Idefics 2 w/ Local Path
+
+```sh
+python -m mlx_vlm.generate --model mlx-community/idefics2-8b-4bit --max-tokens 1000 --temp 0.0  --prompt "Describe this image" --image /path/to/file.png
+```
+
+* Idefics 2 w/ Local Path
+
+```sh
+python -m mlx_vlm.generate --model mlx-community/idefics2-8b-4bit --max-tokens 1000 --temp 0.0  --prompt "Is this a logo?" --image https://pypi.org/static/images/logo-small.8998e9d1.svg
```
Blaizzy, line 33: I really like the direction!

How about we compress these examples. 

For instance, hve a single example and then explain that what we can pass to the --image and --model. Since both are quite similar

Blaizzy (changes requested):

README.md
Blaizzy: Also, there are some changes you might want to check in #17

PR #59, opened 2024-09-16
Add support for Qwen2-VL

Blaizzy:

PR #64, opened 2024-09-29
fix: Restore Compatibility with models other than Qwen2-VL

benzimring:
https://github.com/Blaizzy/mlx-vlm/pull/59 added `image_grid_thw` temporal height and width image feature shape input to support Qwen2-VL, however this broke inference using other models:
```
UnboundLocalError: cannot access local variable 'image_grid_thw' where it is not associated with a value
```

The changes assign a value of `None` to the variable, as well as conditionally guard its assignment in the `kwargs` dictionary.

inference with `Qwen/Qwen2-VL-2B-Instruct` as well as `qnguyen3/nanoLLaVA` (and others) is again functional:
Screenshot 2024-09-28 at 10 28 44 PM

### DIFF
--- a/mlx_vlm/utils.py
+++ b/mlx_vlm/utils.py
@@ -708,6 +708,7 @@ def prepare_inputs(image_processor, processor, image, prompt, image_token_index)
     from transformers.image_utils import load_image
 
     mask = None
+    image_grid_thw = None
     if isinstance(image, str):
         image = load_image(image)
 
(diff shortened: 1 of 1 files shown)


Blaizzy:
Yeah, I noticed that at 3AM 😅 

I fix it by adding kwargs to all models. I think there will be more kwargs with new models.

The fix is included in the Pixtral PR I'm working on.

However, If you add this fix here we can merge it :)

Blaizzy:
Thanks for catching that!

Blaizzy:
Let me fix this issue here: #67

PR #77, opened 2024-10-09
fix issues handling multiple images

hiima234:
1. [IMG_BREAK] and [IMG_END] are lost after embedding
     2. image position encode should be done per image base
        https://github.com/mistralai/mistral-inference/blob/main/src/mistral_inference/vision_encoder.py#L85
        https://github.com/huggingface/transformers/blob/main/src/transformers/models/pixtral/modeling_pixtral.py#L492


hiima234:
close for editing

PR #78, opened 2024-10-09
Fix issues with multiple image handling

hiima234:
1. [IMG_BREAK] and [IMG_END] are lost after embedding
 2. image position encode should be done per image base https://github.com/mistralai/mistral-inference/blob/main/src/mistral_inference/vision_encoder.py#L85 https://github.com/huggingface/transformers/blob/main/src/transformers/models/pixtral/modeling_pixtral.py#L492


Blaizzy:
@hiima234 thank you very much!

Blaizzy:
Really nice catch! 🔥

Blaizzy:
It more consistent now!

Especially with quantized model!

Screenshot 2024-10-11 at 6 13 56 PM

Blaizzy:
What is your X account handle?

PR #99, opened 2024-10-20
fix pixtral image prompt order for doc VQA

### LINKED ISSUE #98: Pixtral text/image prompt order
The current code in the prompt_utils puts image references at the end of the prompt. Pixtral seems sensitive to text/image order: in a Document Visual Q/A use case with a German letter, the current order of `<s>[INST]Who...[IMG]` just produces a translation, the proposed order of `<s>[INST][IMG]Who...` correctly answers the question. Test image attached, prompts below.

I have reproduced this behaviour with both mistral-community/pixtral-12b and mistralai/Pixtral-12B-2409 on Nvidia H100. Because ordering is transparent to API users there, I suppose the behaviour is less of a surprise there as the API conveys that the text prompt has to include a textual reference like "following", e.g. "Describe the following image". That's currently not true for MLX_VLM.

On the other hand, if the prompt template were changed, text prompts taken from elsewhere may not be portable as-is, which could also cause surprises. I'm preparing a PR nonetheless, but an API overhaul may be right way to address this.

# Reproduction
## Test code
```
import mlx.core as mx
from PIL import Image

# Load the model
model_path = "mlx-community/pixtral-12b-bf16"
model, processor = load(model_path)
config = load_config(model_path)

# Prepare input
image = [Image.open("sample-vo.png")]
prompt = "Who's the recipient?"

# Apply chat template
    processor, config, prompt, num_images=len(image)
)

## Old prompt:
## <s>[INST]Who's the recipient?[IMG][/INST]
## New prompt after PR applied:
## <s>[INST][IMG]Who's the recipient?[/INST]

print(formatted_prompt)

# Generate output
output = generate(model, processor, image, formatted_prompt, verbose=False)
print(output)
```

## Sample file used in the code
sample-vo

ndurner:
Puts image references at the beginning of the prompt to Pixtral to make prompting more natural - without using "following" which the API user is likely not aware of. Fixes #98.

### DIFF
--- a/mlx_vlm/prompt_utils.py
+++ b/mlx_vlm/prompt_utils.py
@@ -25,7 +25,9 @@ def get_message_json(
         if role == "user" and not skip_image_token:
             if isinstance(message["content"], list):
                 if model_name == "pixtral":
-                    message["content"] = [{"type": "image"}] * num_images + message["content"]
+                    message["content"] = [{"type": "image"}] * num_images + message[
+                        "content"
+                    ]
                 else:
                     message["content"].extend([{"type": "image"}] * num_images)
             else:
--- a/mlx_vlm/prompt_utils.py
+++ b/mlx_vlm/prompt_utils.py
@@ -24,7 +24,10 @@ def get_message_json(
             return message
         if role == "user" and not skip_image_token:
             if isinstance(message["content"], list):
-                message["content"].extend([{"type": "image"}] * num_images)
+                if model_name == "pixtral":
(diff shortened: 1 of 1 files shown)


Blaizzy:
Thanks a lot, @ndurner!

This is a great catch 🔥

Blaizzy:
Could you write pre-commit to clear the tests??

Blaizzy:
Thanks a lot!

I will merge it. 

The tests now are failing because of a bug with Github test env installing MLX and not the code.

PR #112, opened 2024-11-13
Add support for Molmo

Blaizzy:
Closes #70

PR #114, opened 2024-11-14
Faster / more memory efficient Qwen VL

awni:
- Remove the specialized RoPE in the language model as it is equivalent to a regular RoPE in that case. Using the MLX fast RoPE is also faster
- Avoid accidentally upcasting everything to fp32. This can make a big difference in memory use, especially for lots of images or a videos.

### DIFF
--- a/mlx_vlm/models/qwen2_vl/qwen2_vl.py
+++ b/mlx_vlm/models/qwen2_vl/qwen2_vl.py
@@ -92,6 +92,10 @@ class Model(nn.Module):
     ):
         image_grid_thw = kwargs.pop("image_grid_thw", None)
         image_grid_thw = mx.array(image_grid_thw)
+
+        dtype = self.vision_tower.patch_embed.proj.weight.dtype
+        pixel_values = pixel_values.astype(dtype)
+
         input_embddings = self.get_input_embeddings(
             input_ids, pixel_values, image_grid_thw
         )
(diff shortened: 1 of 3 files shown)


mlx_vlm/models/qwen2_vl/qwen2_vl.py
Blaizzy, line 1: Why do we cast the type of the pixel values here?

Blaizzy (approved):
LGTM!

Blaizzy:
Thanks a lot @awni!

This brings another significant boost in performance.

I will review other models as soon as I finish Molmo #112

mlx_vlm/models/qwen2_vl/qwen2_vl.py
awni: The pixel values are input to the model in `float32` from the image / video preprocessor. The model weights are in `fp16` (or some other precision). If you don't cast the pixel values then all the computation will be done in fp32. (The weights will be upcast to the type of the input if it's in a higher precision).

Blaizzy: Got it, I didn't realize that.

I made a lot of assumptions about how much MLX handles types.

I will keep that in mind.

awni: MLX handles types very similar to how both NumPy and Jax handle types. You almost always get cast up to avoid losing precision.

fp16 op fp32 = fp32
int16 op int32 = int32
int op float = float

Issue #110, opened 2024-11-13
[Performance] Speed up Qwen 2 VL


awni:
This pattern of mixing in numpy and MLX inside the model forward will really slow things down. It forces a synchronization at each layer and breaks asynchronous evaluation:

https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/qwen2_vl/language.py#L184-L185

Try to keep everything in MLX without converting to numpy in the `apply_multimodal_rotary_pos_emb` with only a small modification

```python
        if cache is not None:
            kv_seq_len += cache.offset + 1
            position_ids = mx.arange(cache.offset, cache.offset + L)
        else:
            position_ids = mx.arange(0, L)

        position_ids = mx.expand_dims(position_ids, axis=0)
        position_ids = mx.tile(position_ids, (3, 1, 1))
```

And in `apply_multimodal_rotary_pos_emb`, change:

```python
    mrope_section = np.cumsum(mrope_section * 2)[:-1].tolist()

    # position_ids = position_ids.tolist() <-- no need for that!
    cos = cos[position_ids]
    sin = sin[position_ids]

    cos = mx.concatenate(
        [m[i % 3] for i, m in enumerate(mx.split(cos, mrope_section, axis=-1))], axis=-1
    )[
        :, None, :, :
    ]  # unsqueeze dim 1
    sin = mx.concatenate(
        [m[i % 3] for i, m in enumerate(mx.split(sin, mrope_section, axis=-1))], axis=-1
    )[:, None, :, :]

    # Apply rotary embedding
    q_embed = (q * cos) + (rotate_half(q) * sin)
    k_embed = (k * cos) + (rotate_half(k) * sin)

    return q_embed, k_embed # <-- no need to cast to `mx.array` as they are already `mx.array`
```

awni:
Just FYI this change speeds up generation by 40% on an M2 Ultra:

```
mlx_vlm.generate --model mlx-community/Qwen2-VL-7B-Instruct-4bit --image ~/mlx-examples/stable_diffusion/still-life.png --prompt "Describe this please" --max-tokens 512
```

Pre: Generation: 55.160 tokens-per-sec

Post: Generation: 90.429 tokens-per-sec

Blaizzy:
Thanks, will make the changes!

Btw I have been meaning to ask.

I can't reproduce this with `mx.where`. 

```python 
 def _merge_input_ids_with_image_features(
[...]

PR #124, opened 2024-11-26
Idefics 3 support

pcuenca:
I didn't find anyone working on it yet, so decided to give it a go.

### DIFF
--- a/mlx_vlm/models/idefics3/language.py
+++ b/mlx_vlm/models/idefics3/language.py
@@ -1,0 +1,185 @@
+import inspect
+import re
+from dataclasses import dataclass
+from typing import Dict, Optional, Tuple, Union
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from ..base import KVCache, LanguageModelOutput, create_attention_mask
+
+
+@dataclass
+class TextConfig:
+    model_type: str
+    hidden_size: int
+    num_hidden_layers: int
+    intermediate_size: int
+    num_attention_heads: int
+    rms_norm_eps: float
+    vocab_size: int
+    num_key_value_heads: int
+    rope_theta: float = 1000000.0
+    rope_traditional: bool = False
+    tie_word_embeddings: bool = False
+
+    @classmethod
+    def from_dict(cls, params):
+        return cls(
+            **{
+                k: v
+                for k, v in params.items()
+                if k in inspect.signature(cls).parameters
+            }
+        )
+
+    def __post_init__(self):
+        if self.num_key_value_heads is None:
+            self.num_key_value_heads = self.num_attention_heads
+
+
+class Attention(nn.Module):
+    def __init__(self, config: TextConfig):
+        super().__init__()
+
+        dim = config.hidden_size
+        self.n_heads = n_heads = config.num_attention_heads
+        self.n_kv_heads = n_kv_heads = config.num_key_value_heads
+
+        head_dim = config.hidden_size // n_heads
+        self.scale = head_dim**-0.5
+
+        self.q_proj = nn.Linear(dim, n_heads * head_dim, bias=False)
+        self.k_proj = nn.Linear(dim, n_kv_heads * head_dim, bias=False)
+        self.v_proj = nn.Linear(dim, n_kv_heads * head_dim, bias=False)
+        self.o_proj = nn.Linear(n_heads * head_dim, dim, bias=False)
+
+        self.rope = nn.RoPE(
+            head_dim,
+            traditional=config.rope_traditional,
+            base=config.rope_theta,
+        )
+
+    def __call__(
+        self,
+        x: mx.array,
+        mask: Optional[mx.array] = None,
+        cache: Optional[KVCache] = None,
+    ) -> mx.array:
+        B, L, D = x.shape
+
+        queries, keys, values = self.q_proj(x), self.k_proj(x), self.v_proj(x)
+
+        # Prepare the queries, keys and values for the attention computation
+        queries = queries.reshape(B, L, self.n_heads, -1).transpose(0, 2, 1, 3)
+        keys = keys.reshape(B, L, self.n_kv_heads, -1).transpose(0, 2, 1, 3)
+        values = values.reshape(B, L, self.n_kv_heads, -1).transpose(0, 2, 1, 3)
+
+        if cache is not None:
+            queries = self.rope(queries, offset=cache.offset)
+            keys = self.rope(keys, offset=cache.offset)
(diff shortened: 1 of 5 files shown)


mlx_vlm/models/idefics3/vision.py
```diff
@@ -0,0 +1,240 @@
...
+        super().__init__()
+        self.activation_fn = FastGELUActivation()
+        self.fc1 = nn.Linear(config.hidden_size, config.intermediate_size, bias=True)
+        self.fc2 = nn.Linear(config.intermediate_size, config.hidden_size, bias=True)
+
+    def __call__(self, x: mx.array) -> mx.array:
+        x = self.fc1(x)
+        x = self.activation_fn(x)
+        x = self.fc2(x)
+        return x
```
Blaizzy: Nit:  MLX has this [Gelu activation](https://ml-explore.github.io/mlx/build/html/python/nn/_autosummary/mlx.nn.GELU.html#mlx.nn.GELU). 
```suggestion
class MLP(nn.Module):
    def __init__(self, config: VisionConfig):
        super().__init__()
        self.activation_fn = nn.GELU(approx='precise')
        self.fc1 = nn.Linear(config.hidden_size, config.intermediate_size, bias=True)
        self.fc2 = nn.Linear(config.intermediate_size, config.hidden_size, bias=True)

    def __call__(self, x: mx.array) -> mx.array:
        x = self.fc1(x)
        x = self.activation_fn(x)
        x = self.fc2(x)
        return x
```

Blaizzy (changes requested):
Thank you very much for your contribution @pcuenca!

Idefics 3 is an awesome model that has been on my radar.

Just 3 tiny changes before we merge:
1. `pre-commit run --all-files` (to fix style tests)
2. Change GELU to default 
3. Add model tests [here](https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/tests/test_models.py)

Blaizzy:
My pleasure!

I noticed, and am very happy to see you contributing. It means a lot to me! 🎉❤️

mlx_vlm/models/idefics3/vision.py
```diff
@@ -0,0 +1,240 @@
...
+        )
+
+        self.num_patches = (self.image_size // self.patch_size) ** 2
+        self.num_positions = self.num_patches
+        self.position_embedding = nn.Embedding(self.num_positions, self.embed_dim)
+
+    def __call__(self, x: mx.array) -> mx.array:
+        patch_embeddings = self.patch_embedding(x)
+        patch_embeddings = mx.flatten(patch_embeddings, start_axis=1, end_axis=2)
+        position_ids = mx.array(np.arange(self.num_positions)[None, :])
```
Blaizzy: Nit: 

MLX has [mx.arange](https://ml-explore.github.io/mlx/build/html/python/_autosummary/mlx.core.arange.html#mlx.core.arange) built-in.

Mixing numpy/torch in the graph execution leads to significant performance slow down.

Read more [here](https://github.com/Blaizzy/mlx-vlm/issues/110#issuecomment-2474931063).

```suggestion
        position_ids = mx.array(mx.arange(self.num_positions)[None, :])
```

Blaizzy (changes requested):

mlx_vlm/models/idefics3/vision.py
Blaizzy: This change and the GELU might get you a few extra tokens-per-sec (+3).

mlx_vlm/models/idefics3/language.py
```diff
@@ -0,0 +1,185 @@
...
+        inputs: mx.array,
+        cache=None,
+        inputs_embeds=None,
+        mask: Optional[mx.array] = None,
+    ):
+        # for passing merged input embeddings
+        if inputs_embeds is None:
+            h = self.embed_tokens(inputs)
+        else:
+            h = inputs_embeds
```
awni: I recommend setting the `inputs_embeds` to the model dtype here. It will be faster since you will avoid unexpectedly upcasting weights to a higher precision.
```suggestion
            h = inputs_embeds.astype(self.norm.weight.dtype)
```

awni: (With this change + fixing the GeLU I get a nice speedup)

awni:
I was playing with this model locally and it works quite well (and is even faster with some of the suggested changes).

One odd thing I noticed is the vision encoder is quantized with type `float32` whereas the LM is quantized with `float16`. I traced it to the padding operation which I think inadvertently [always uses `float32`](https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/utils.py#L634-L635). Those `mx.zeros` will default to `float32`. You can fix that by doing:

```python
    new_weight = mx.zeros((new_out, new_in), weight.dtype)
    new_bias = mx.zeros((new_out), weight.dtype)
```

mlx_vlm/models/idefics3/vision.py
```diff
@@ -0,0 +1,240 @@
...
+        self.encoder = Encoder(config)
+        self.post_layernorm = nn.LayerNorm(config.hidden_size)
+
+    def __call__(
+        self,
+        x: mx.array,
+        output_hidden_states: Optional[bool] = None,
+    ) -> mx.array:
+        x = self.embeddings(x)
+        encoder_outputs = self.encoder(
```
awni, line 204: Probably also good to cast `x` to the model type here. Right now it doesn't matter much since the model type is `fp32` due to what I think is a bug in the quantization step. But if that's fixed this will matter a lot (will use less RAM and avoid upcasting the model).

```suggestion
        x = x.astype(self.embeddings.patch_embedding.weight.dtype)
        encoder_outputs = self.encoder(
```

Blaizzy:
Thanks a lot @awni!  

I will fix it as soon as I get back home.

mlx_vlm/models/idefics3/vision.py
pcuenca: oh nice, this was an oversight, my bad!

pcuenca:
~114 tok/s with your fixes @Blaizzy @awni 🤯 🙌

Blaizzy:
Awesome, that’s the power of pure MLX 🚀😎

I’m about to merge the utils, then rebase this PR and merge.

Blaizzy:
@pcuenca have you uploaded the models to the HUB?

If not, where do you want the the weights to live? 

Is it ok to put them in the MLX-Community repo?

pcuenca:
> Is it ok to put them in the MLX-Community repo?

Of course!

Blaizzy:
Thanks, uploading the models right now :)

PR #132, opened 2024-11-28
Fix broken stream generate for SmolVLM and others

andimarafioti:
This breaks stream generation if the model doesn't have image_token_index

### DIFF
--- a/mlx_vlm/utils.py
+++ b/mlx_vlm/utils.py
@@ -1095,7 +1095,10 @@ def stream_generate(
         tokenizer = processor.tokenizer
 
     resize_shape = kwargs.pop("resize_shape", None)
-    image_token_index = model.config.image_token_index
+    if hasattr(model.config, "image_token_index"):
+        image_token_index = model.config.image_token_index
+    else:
+        image_token_index = None
 
     # Prepare inputs
     inputs = prepare_inputs(


Blaizzy:
Hey @andimarafioti 

Thanks for the PR!

To clarify, all models should have `image_token_index`.

I see that idefics3 uses instead `image_token_id`.

So the fix should be renaming it to `image_token_index`.

Blaizzy:
Could you make the change above?

I will review and merge it as soon as you do.

andimarafioti:
Why should the model have `image_token_index`? The idefics3 PR was reviewed quite intensely and no one mentioned this.

andimarafioti:
Also, there is another part of this same file where this check is used to avoid other models failing :/

Blaizzy:
> Why should the model have image_token_index? The idefics3 PR was reviewed quite intensely and no one mentioned this.

It was a mistake we all missed,

I also found another bug just now with the positioning of the image token from that PR.

Blaizzy:
> Also, there is another part of this same file where this check is used to avoid other models failing :/

Ok, you have a point. I will re-open and merge your PR.

I'm about to refactor the entire codebase to address this cases because all model should have this even if it's zero/None.

Blaizzy:
Thank you very much for bringing this up @andimarafioti!

PR #169, opened 2025-01-02
Corrected order of parameters on example

bbaaxx:
The example on the readme does not run because the order of parameters is incorrect resulting in:

`TypeError: TextEncodeInput must be Union[TextInputSequence, Tuple[InputSequence, InputSequence]]`

Fixing example on Readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ formatted_prompt = apply_chat_template(
 )
 
 # Generate output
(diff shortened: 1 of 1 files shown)


Blaizzy:
Thank you very much, @bbaaxx! 

It like you read my mind.

I noticed it late last night.

Will merge as soon as tests pass.

PR #202, opened 2025-02-07
Add stream_generate and GenerationResult to __init__.py

asmeurer:
This slightly lower level API function allows extracting things like performance metrics from the generation programmatically.


Blaizzy:
Thanks a lot @asmeurer!

Sorry, I missed your PR.

PR #221, opened 2025-03-01
Fix typo and local path

lucataco:
This should fix the local path to the audio file

I think you meant to use this model? I see others under [mlx-community](https://huggingface.co/mlx-community) like 8/6/4bit
mlx-community/ShowUI-2B-bf16-bf16


Blaizzy:
Thanks a lot @lucataco, great looking out! 

The model name was a typo. I just fixed it.

It should be [mlx-community/ShowUI-2B-bf16](https://huggingface.co/mlx-community/ShowUI-2B-bf16)

So we can revert that change and merge the rest.

lucataco:
Updated!

Blaizzy:
Merged 🚀

Issue #253, opened 2025-03-18
Remove scipy dependency


r-bit-rry:
Its a heavy dependency which requires external libraries and tooling installed.
We are only using it for the scipy.ndimage.zoom method which is provided both by Pillow and OpenCV which are already a depepndency for this project.

Of the top of my head these two implementations can replace zoom method, note that the opencv one will be the best performant:
```
from PIL import Image
import numpy as np

def pillow_zoom(input_array, zoom_factor):
    """Resize using Pillow with similar functionality to scipy.ndimage.zoom"""
    if input_array.ndim == 2:
        # Handle 2D arrays (single-channel images)
        img = Image.fromarray(input_array)
        new_size = tuple(int(s * zoom_factor) for s in input_array.shape)
        # Swap dimensions for PIL (width, height)
        resized = img.resize((new_size[1], new_size[0]), Image.Resampling.BICUBIC)
        return np.array(resized)
    elif input_array.ndim == 3:
        # Handle multi-channel images
        results = []
        for i in range(input_array.shape[0]):
            results.append(pillow_zoom(input_array[i], zoom_factor))
        return np.stack(results)
    else:
        raise ValueError("Unsupported array dimension")
```

or

```
import cv2
import numpy as np

def opencv_zoom(input_array, zoom_factor):
    """Resize using OpenCV with similar functionality to scipy.ndimage.zoom"""
    if input_array.ndim == 2:
        # Single channel 2D array
        new_shape = tuple(int(s * zoom_factor) for s in input_array.shape)
        # Note: cv2.resize takes (width, height) which is opposite of numpy's (height, width)
        return cv2.resize(input_array, (new_shape[1], new_shape[0]), interpolation=cv2.INTER_CUBIC)
    elif input_array.ndim == 3:
        # Multi-channel or 3D array
        results = []
        for i in range(input_array.shape[0]):
            results.append(opencv_zoom(input_array[i], zoom_factor))
        return np.stack(results)
    else:
        raise ValueError("Unsupported array dimension")
```

Blaizzy:
I agree with you!

Could you send a PR and add some metrics like performance (acc and performance)?

asmeurer:
Also scipy is pinned in the requirements, which makes it harder to install mlx-vlm in an environment with other packages.

r-bit-rry:
on it

r-bit-rry:
https://github.com/Blaizzy/mlx-vlm/pull/268/

[...]

PR #268, opened 2025-03-21
refactor: update image resizing implementation to use OpenCV and add …

r-bit-rry:
closing this issue:
https://github.com/Blaizzy/mlx-vlm/issues/253

removing scipy dependency
chaging resize_image from scipy.ndimage.zoom to opencv resize
adding unittest to avoid regression

### DIFF
--- a/mlx_vlm/models/gemma3/language.py
+++ b/mlx_vlm/models/gemma3/language.py
@@ -196,17 +196,18 @@ class Gemma3Model(nn.Module):
             sliding_window_mask = create_attention_mask(h, cache)
 
         for i, (layer, c) in enumerate(zip(self.layers, cache)):
-            is_sliding = (
+            is_global = (
                 i % self.config.sliding_window_pattern
                 == self.config.sliding_window_pattern - 1
             )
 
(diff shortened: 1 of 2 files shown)


maxlund:
This might be worth taking a look at
https://zuru.tech/blog/the-dangers-behind-image-resizing

image

Blaizzy:
@r-bit-rry could you share some benchmark results with the old and new method?

Blaizzy:
Also, why can't we use PIL for this?

r-bit-rry:
I guess it doesn't really matter.
I chose opencv for the speed (at least the documented one)
I don't even know if the scipy method zoom that was used before was equivalent to the quality by PIL

r-bit-rry:
> @r-bit-rry could you share some benchmark results with the old and new method?

checked for regression and added a test, not for benchmark.
benchmarking can be seen here but obviously it was not made for this specific usecase:
https://github.com/libvips/libvips/wiki/Speed-and-memory-use

generally speaking I'm not sure that now is the time to optimize performance, especially when the bottleneck is most likely the mlx model itself

PR #275, opened 2025-03-24
feat: allow `PIL.Image.Image` as input

SauravMaheshkar:
simple patch to allow passing PIL Images as input to the model. I tested this locally by a modified version of the script in the README.

```python
from PIL import Image

from mlx_vlm import load, generate
from mlx_vlm.prompt_utils import apply_chat_template
from mlx_vlm.utils import load_config

# Load the model
model_path = "mlx-community/Qwen2-VL-2B-Instruct-4bit"
model, processor = load(model_path)
config = load_config(model_path)

# Prepare input
image = [Image.open("snoopy.png")]
prompt = "Describe this image."

# Apply chat template
formatted_prompt = apply_chat_template(
    processor, config, prompt, num_images=len(image)
)

# Generate output
output = generate(model, processor, formatted_prompt, image, verbose=False)
print(output)
```

PR #276, opened 2025-03-25
docs: add note on PIL

SauravMaheshkar:
ref: #275


Blaizzy:
Thank you very much @SauravMaheshkar!

PR #289, opened 2025-04-06
chore: improve dependency management

tmc:
I'm an old head and have not gotten on to the uv game just yet but this PR adds a little convenience and security for developers/maintainers.

I noticed a somewhat old scipy was being pulled which was needing me to install gcc on a new mac so I took pause.
I'm still waiting on weights so I have not exercised this heavily.

I normally would add a test target but the test_model_loading, test_generation smoke tests need some attention.

No worries if these tools aren't the ones you want to use here.


Blaizzy:
I love how the new requirements.txt is well described

But wonder if there isn't a simpler way

PR #290, opened 2025-04-07
fixed mx.metal.get_peak_memory() depreciate

vovw:


vovw:
`mx.metal.get_peak_memory()` is being depreciated, with `mx.get_peak_memory()` instead

vovw:
@Blaizzy wdyt ?

Blaizzy:
Thanks @vovw!

Could you bump and pin the MLX version in requirements txt?

I believe this changed recently.

PR #307, opened 2025-04-18
Add Stopping criteria

Blaizzy:


banteg:
a more general solution here https://github.com/Blaizzy/mlx-vlm/pull/309

PR #309, opened 2025-04-19
fix(utils): stop generation on every EOS id declared in model config

banteg:
###   Problem
Some newer checkpoints (e.g. **`mlx‑community/gemma‑3‑27b‑it‑qat‑4bit`**) ship a `generation_config.json` where **`eos_token_id` is a *list***:

```jsonc
"eos_token_id": [1, 106]          // 1 = <eos>, 106 = <end_of_turn>
```

`stream_generate()` in **mlx‑vlm** only stopped on the single value returned by  
`tokenizer.eos_token_id` (→ `1`).  
During sampling the model frequently emitted the *second* ID (`106`), so the loop never broke and the program kept printing `<end_of_turn>` until `max_tokens` was reached.

###   🛠️ What this patch does
1. **Plumbs the model config into the generation helpers**  
   * `generate.py` now passes `config` to `generate()`  
   * `generate()` forwards it to `stream_generate()`
2. **Builds a unified `eos_token_ids` set** inside `stream_generate()`  
   * always includes the tokenizer’s canonical ID  
   * merges *every* ID found in `config["eos_token_id"]` (int **or** list)  
3. **Breaks the sampling loop on *any* of those IDs**

```python
eos_token_ids = {processor.tokenizer.eos_token_id}
cfg_ids = config.get("eos_token_id")
if cfg_ids is not None:
    eos_token_ids.update(cfg_ids if isinstance(cfg_ids, list) else [cfg_ids])

...
if token in eos_token_ids:        # ← stops immediately
    break
```

###   Results
* Gemma‑3 and other multi‑EOS models stop cleanly at `<end_of_turn>` instead of running to the token limit.
* Behaviour for models with a single `eos_token_id` is unchanged (the set just contains one element).
* Zero performance overhead (constant‑time set lookup) and no extra dependencies.

###   🔍 Reproduction (before the patch)

```bash
python -m mlx_vlm.generate \
       --model mlx-community/gemma-3-27b-it-qat-4bit \
       --prompt "say hello and hang up"
# prints "<end_of_turn>" over and over until max_tokens
```

###   After the patch

The same command stops after the first `<end_of_turn>` and returns the response promptly.

---

Let me know if you’d like additional test coverage or stylistic tweaks – happy to iterate!

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -142,6 +142,7 @@ def main():
             model,
             processor,
             prompt,
+            config=config,
             image=args.image,
             temperature=args.temperature,
             max_tokens=args.max_tokens,
--- a/mlx_vlm/utils.py
+++ b/mlx_vlm/utils.py
@@ -1060,6 +1060,7 @@ def stream_generate(
     processor: PreTrainedTokenizer,
     prompt: str,
     image: Union[str, List[str]] = None,
+    config: Dict = None,
     **kwargs,
 ) -> Union[str, Generator[str, None, None]]:
     """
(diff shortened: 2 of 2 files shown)


Blaizzy:
Hi Bateng

I fixed it and tested it extensivel here #307.

Normal config EOS
Screenshot 2025-04-19 at 2 51 12 PM

Custom EOS ("and")
Screenshot 2025-04-19 at 2 52 42 PM

Blaizzy:
Basically does the same but also accounts for cases where config doesn't hold the EOS tokens and gives users a way to pass custom EOS tokens too.

banteg:
great! i'll close mine then.

Blaizzy:
Thanks for the contribution! 

We have lots of open issues and I would love your help there.

Issue #229, opened 2025-03-06
Add Support for OS-Copilot/OS-Atlas-Base-7B


prncvrm:
@Blaizzy Can you help me run OS-Copilot/OS-Atlas-Base-7B, i tried converting to mlx, 8bit but unable to get the same accuracy as the original model here https://huggingface.co/spaces/maxiw/OS-ATLAS
What could i be doing wrong?
used command `mlx_vlm.convert --hf-path OS-Copilot/OS-Atlas-Base-7B -q  --q-bits 8`

prncvrm:
more details here https://github.com/OS-Copilot/OS-Atlas/issues/51

Blaizzy:
Hey,

I just tried it. 

It works well on demo samples but fails with custom UIs 

Check the screen resolution they are using and the prompting strategy

prncvrm:
i checked that, they're resizing, but still that doesn't helps me @Blaizzy 
any guidance you could provide with?

also it fails on the same https://maxiw-os-atlas.hf.space/gradio_api/file=/tmp/gradio/fc0a8ef05b952a970924913eea2b89a8d626c92031f94ff5f3e6ccaf8dd23a4e/web_6f93090a-81f6-489e-bb35-1a2838b18c01.png attached here
The y index specifically

Blaizzy:
Qwen2vl needs to normalise their bbox to 1000

prncvrm:
i've been doing that @Blaizzy 
i do get the x axis correct, just the y axis is wrong

prncvrm:
also when you say you've tried on the demo samples here https://github.com/Blaizzy/mlx-vlm/issues/229#issuecomment-2706806607, did you mean you've used the original hugging face model, or a converted via mlx one?

prncvrm:
![Image](https://github.com/user-attachments/assets/3759b0f9-9c0a-49fe-a047-09eda9330aff) this is from hugging face spaces logs
vs 

![Image](https://github.com/user-attachments/assets/f5ebd0aa-d73a-438b-bcaa-28d8a4fdb48a)this is using mlx 

can you see if you can help by any chance?

prncvrm:
btw, i've kept the original size of image on huggingface spaces, modified the code to keep the original image
you can see it in the logs

prncvrm:
@Blaizzy any directions you could help me with?

prncvrm:
[...]

PR #319, opened 2025-04-21
Fixes Qwen2 VL Position Id

prncvrm:
computes position id using `get_rope_index` function and replaces with the cache offset generated,
results are as follows 
image

----WITH THE FIX----

image

----WITHOUT THE FIX----

image

----SAME MODEL USING CUDA/PYTORCH----

Related Issue https://github.com/Blaizzy/mlx-vlm/issues/229
can be verified at https://huggingface.co/spaces/maxiw/OS-ATLAS


Blaizzy:
Could you share the original images and the bbox overlay ?

prncvrm:
https://huggingface.co/spaces/maxiw/OS-ATLAS
you can use to reference the same

Blaizzy:
Ok, got it.

Thanks @prncvrm for digging into this!

I will give it a test and review your PR

prncvrm:
sure thanks

Blaizzy:
Hey @prncvrm 

After some testing I found that it only works sometimes. If you ask to click for example the shopping cart text in blue it fails whilst the torch works every single time.

Could you address these and ensure it's 1:1 with the torch version before we merge?

Blaizzy:
Same as yours just try to locate different items.

prncvrm:
@Blaizzy i did check
`click shopping cart text in blue`
image
image

`click view and edit cart`
image
image

and for the results MLX generates, the IoU is more than 90%(for few of cases only) while for rest, its more than 95%, which i believe is fine and is far better than what we previously had ~60% https://github.com/Blaizzy/mlx-vlm/issues/229#issuecomment-2717591772

let me know your thoughts here

prncvrm:
@Blaizzy here's the result you requested for, 
the plotting and the required bbox on the image 

Except `Grocery & Gourmet Food` (which is truncated), everything else seems to be same place
I hope this is acceptable range and error threshold 
MLX ::
shopping-bbox-mlx
Pytorch ::
shopping-bbox-pytorch
Plotting ::
shopping-bbox-graph

prncvrm:
inputs were (add prefix of `click on` for each input)
``` "Proceed to Checkout",
  "shopping cart icon",
  "Carnation Breakfast Essentials Nutritional Drink (Pack of 8)",
  "Pre-baked Gingerbread House Kit Value Pack, 17 oz., Pack of 2, Total 34 oz.",
  "View and Edit Cart",
  "My Account",
  "My Wish List 55 items",
  "Sign Out",
  "Beauty & Personal Care",
  "Sports & Outdoors",
  "Clothing, Shoes & Jewelry",
  "Home & Kitchen",
  "Office Products",
  "Health & Household",
  "Patio, Lawn & Garden",
  "Electronics",
  "Cell Phones & Accessories",
  "Grocery & Gourmet Food",
  "One Stop Market",
  "Welcome, Emma Lopez!",
  "7 Items in Cart",
  "Cart Subtotal:",
  "$369.98",
  "Product Showcases",
  "V8 Energy cans",
  "orange vanilla caffeine-free black pouches",
  "shopping cart text in blue link"

**Comment by Blaizzy** [comment]
Could you share your run script? I wanna see something.

**Comment by Blaizzy** [comment]
I found the bug.

The current implementation doesn't work with multi-turn.

It get the first item right but fails after that unless you restart.

**Comment by Blaizzy** [comment]
However, I still note that it fails with "View and Edit Cart" 

image

**Comment by prncvrm** [comment]
okay, got it
will check
thanks for pointing out

**Comment by prncvrm** [comment]
@Blaizzy https://github.com/Blaizzy/mlx-vlm/pull/319/commits/b355cf1a85170957f626fa58c4c8dd205c92dd64 fixed the same can you check and review now

File: mlx_vlm/utils.py [hunk 1]
```diff
@@ -1027,6 +1027,12 @@ def _step(y, **kwargs):
                     cache=cache,
                     **kwargs,
                 )
+            elif model.config.model_type == "qwen2_vl":
+                outputs = model(
+                    y[None],
+                    cache=cache,
+                    **kwargs,
+                )
```
Blaizzy: Could you explain why do this? Being that `y` will land in language model?

I would prefer we move the position ids logic to the language model module and revert this part.

Blaizzy:
Cool, thanks @prncvrm! 
I left a small nit comment, and will test the model early tomorrow.

Blaizzy (changes requested):
Small nit

mlx_vlm/utils.py
prncvrm: @Blaizzy if i were to move the position id logic inside the language model module, it would also require me to pass other params like `image_grid_thw` , `video_grid_thw`
the implementation of transformers actually invokes the original model, not just language and hence tried to use the same pattern here

this position_id information is to be kept at the highest level(model) as seen in transformers hence this is the case, i m pretty sure, the same is with Qwen2.5 too raised [here](https://github.com/Blaizzy/mlx-vlm/issues/330) and maybe for other foundational models too

Let me know your thoughts on the same

Blaizzy: I don't mind :) 

You can just pass them as kwargs and index inside the LM. 

It makes the logic much cleaner and no need for custom inference code for 2 models only.

prncvrm: @Blaizzy lmk if you want me to move it, 
IMO, this would be better as it resemble the interface of transformers which may help other contributors to match and debug any other issues

Blaizzy: Yes, please do move it.

It's not different from transformers, it's just in another file. Transformers uses a single modelling file I prefer to break it down into modules.

prncvrm: @Blaizzy i've done the required changes, and ran the same set i'd done earlier, works as expected

Blaizzy:
@prncvrm thank you very much, you did phenomenal work!

Multi-turn works really well now 🚀

image

PR #320, opened 2025-04-21
feat: Resume Lora fine-tuning from saved adapter file

breynolds007:
Adds the ability to resume Lora training from previously saved weights by passing:

--resume-adapter-file <path_to_adapters.safetensors>

It's the same syntax as used in mlx-lm. 

- Updated trainer.utils and added a simple unit test.
- Black is happy with the formatting; it forced me to change **get_peft_model()** to have one argument on each line.


breynolds007:
After spending more time with it, I don't like the way this PR was implemented. Will close this PR and create a much simpler version.

PR #323, opened 2025-04-22
When converting models, default dtype to config.json's torch_dtype

neilmehta24:
Check the `torch_dtype` in the hf-model before starting conversion. We need this because precision is lost when casting from `bfloat16` to `float16`
- Remove the `ignored if -q is given` from the help text, since it's not true and it does indeed affect the quantization quality
- Check if the `torch_dtype` top-level field in the hf-model's config.json matches one of our three dtypes. If it does, and the user has not specified a dtype, use `torch_dtype`
- If `torch_dtype` isn't found, or if the value isn't one we expect, we default to `bfloat16`. More and more models need `bfloat16` to work well, so this is a better default moving forward

### DIFF
--- a/mlx_vlm/convert.py
+++ b/mlx_vlm/convert.py
@@ -31,7 +31,7 @@ def configure_parser() -> argparse.ArgumentParser:
     )
     parser.add_argument(
         "--dtype",
-        help="Type to save the parameter. Defaults to config.json's `torch_dtype` or bfloat16",
+        help="Type to save the parameter. Defaults to config.json's `torch_dtype` or the current model weights dtype",
         type=str,
         choices=MODEL_CONVERSION_DTYPES,
         default=None,
--- a/mlx_vlm/convert.py
+++ b/mlx_vlm/convert.py
@@ -2,7 +2,7 @@
 
 import argparse
 
-from .utils import convert
+from .utils import MODEL_CONVERSION_DTYPES, convert
 
 
 def configure_parser() -> argparse.ArgumentParser:
--- a/mlx_vlm/convert.py
+++ b/mlx_vlm/convert.py
@@ -31,10 +31,10 @@ def configure_parser() -> argparse.ArgumentParser:
     )
     parser.add_argument(
         "--dtype",
-        help="Type to save the parameters, ignored if -q is given.",
+        help="Type to save the parameter. Defaults to config.json's `torch_dtype` or bfloat16",
         type=str,
-        choices=["float16", "bfloat16", "float32"],
-        default="float16",
+        choices=MODEL_CONVERSION_DTYPES,
+        default=None,
     )
     parser.add_argument(
         "--upload-repo",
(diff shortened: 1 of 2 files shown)


mlx_vlm/utils.py
```diff
@@ -743,8 +745,14 @@ def convert(
...
-    weights = dict(tree_flatten(model.parameters()))
+    if dtype is None:
+        dtype = config.get("torch_dtype", None)
+    if dtype not in MODEL_CONVERSION_DTYPES:
+        dtype = "bfloat16"
+    print("[INFO] Using dtype:", dtype)
     dtype = getattr(mx, dtype)
+
+    weights = dict(tree_flatten(model.parameters()))
     weights = {k: v.astype(dtype) for k, v in weights.items()}
```
awni: I might just not set the dtype if it isn't specified. So concretely:
```suggestion
    weights = dict(tree_flatten(model.parameters()))
    if dtype in MODEL_CONVERSION_DTYPES:
        print("[INFO] Using dtype:", dtype)
        dtype = getattr(mx, dtype) 
        weights = {k: v.astype(dtype) for k, v in weights.items()}
```

awni: The rational is if the dtype is not set or specified use whatever type the model is saved in as we assume that's the best precision for the model. 

fp16 can be better than bf16 and vice versa depending on how the model is trained, so rather than defaulting to some fixed type it can be better to use the type of the saved model.

Blaizzy: > I might just not set the dtype if it isn't specified. So concretely:

I agree, this makes sense 👌🏽

Blaizzy:
Thanks a lot @neilmehta24 and @awni! 

Yes, indeed this has been happening more frequently with latest releases of Gemma in particular. So this surely solves the problem for models alike. 

However, I haven’t seen any other model with this particular precision sensitivity. Most work fine with FP16 even if the defined type is bf16 (usually the best precision choice when training)

mlx_vlm/convert.py
```diff
@@ -31,10 +31,10 @@ def configure_parser() -> argparse.ArgumentParser:
     )
     parser.add_argument(
         "--dtype",
-        help="Type to save the parameters, ignored if -q is given.",
+        help="Type to save the parameter. Defaults to config.json's `torch_dtype` or bfloat16",
```
Blaizzy:
```suggestion
        help="Type to save the parameter. Defaults to config.json's `torch_dtype` or the current model weights dtype",
```

Blaizzy (approved):
LGTM!

neilmehta24:
Seeing this on github
```
Merging is blocked
New changes require approval from someone other than Blaizzy because they were the last pusher.
Cannot change this locked branch
```
@Blaizzy If you could merge this in it would be appreciated!
Thanks!

awni:
> However, I haven’t seen any other model with this particular precision sensitivity

FWIW I've seen it with several other models including GLM4, DSV3 and at least a couple others.

Blaizzy:
Thanks, for letting me know @awni 👌🏽

I haven't had the chance to work with those models yet.

Blaizzy:
Absolutely @neilmehta24, it's done!

I had approved it so probably just needed a refresh to show the merge button

PR #327, opened 2025-04-23
feat: Resume training from saved adapter files

breynolds007:
Allow Lora training to be resumed from a previously saved adapter path, with fewer changes than previous version:

- Added `--resume-adapter-path` parameter where we expect to find saved `adapter_config.json` and `adapters.safetensors` files from previous training
- Fixed docs so the command-line parameters are formatted correctly
- Tests pass
- Black approves of Python formatting
- Should be an easy merge with PR for general fine-tuning overhaul


Blaizzy (changes requested):
Hey @breynolds007 

This looks awesome!

One tiny nit: Let's use adapter-path instead of resume-adapter path

After this change we can merge!

breynolds007:
> Hey @breynolds007
> 
> This looks awesome!
> 
> One tiny nit: Let's use adapter-path instead of resume-adapter path
> 
> After this change we can merge!

Awesome! I made the change as you suggested. Thanks for taking the time to check it, and for the great tool. :)

breynolds007:
> My pleasure!
> 
> Could you run
> 
> > pre-commit run --all

@Blaizzy my apologies, fixed the import order and all is now green locally.

Blaizzy (approved):
LGTM!

PR #329, opened 2025-04-25
[phi3_v] Remove logit dtype conversion from float32 to uint32

neilmehta24:
The dtype of `self.lm_head.weight.dtype` is `uint32`, since it's of type `QuantizedLinear`. `self.lm_head(out)` outputs the correct dtype `float32`.

This change is needed because logits processing was breaking in mlx-engine due to the data loss from this conversion.


neilmehta24:
I am not sure what the best way is to deal with the test failure here 🤔

Blaizzy:
Hey @neilmehta24 

Thanks!

> The dtype of self.lm_head.weight.dtype is uint32

Upon further investigation I found that `self.lm_head.weight.dtype` is usually set to model dtype (float16, float32 and bfloat16) and not `uint32`. That dtype shouldn't be possible since the model weights are floats. So I suspect there is bug in the LM studio mlx-engine causing this.

Current setup ✅: 
Screenshot 2025-04-28 at 2 41 21 PM


With your change ❌:
Screenshot 2025-04-28 at 2 39 11 PM

That's also the reason why the tests are failing, because the logits dtype are now not matching the model dtype.

neilmehta24:
The quantized phi3_v model has the dtype set as `U32` for `lm_head.weight`
```
 head -c 1000 mlx-community/Phi-3.5-vision-instruct-4bit/model.safetensors
9{"__metadata__":{"format":"mlx"},"lm_head.biases":{"data_offsets":[0,3078144],"dtype":"F16","shape":[32064,48]},"lm_head.scales":{"data_offsets":[3078144,6156288],"dtype":"F16","shape":[32064,48]},"lm_head.weight":{"data_offsets":[368312320,417562624],"dtype":"U32","shape":[32064,384]},"
```

The test seems to only use float32/16 as the model weight dtypes. https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/tests/test_models.py#L16-L17

Is is the intended behavior for the logits to be `U32` for the 4-bit model? If so I can dequantize to float for the cases where ints are returned, and proceed thereafter

mlx_vlm/models/phi3_v/phi3_v.py
```diff
@@ -217,7 +217,7 @@ def __call__(
         **kwargs,
     ):
         out = self.model(inputs, pixel_values, image_sizes, mask=mask, cache=cache)
-        logits = self.lm_head(out).astype(self.lm_head.weight.dtype)
+        logits = self.lm_head(out)
```
Blaizzy, line 220:
```suggestion
        logits = self.lm_head(out)
        if self.lm_head.weight.dtype in [mx.float16, mx.bfloat16, mx.float32]:
            logits = logits.astype(self.lm_head.weight.dtype)
```

Blaizzy:
Just took another look and found a solution :)

PR #331, opened 2025-04-30
docs: Fix command-line params in LORA.md to use dash separators

breynolds007:
Fixing an inconsistency with the docs; the use of dash (-) instead of underscore (_) for the LORA.md command-line parameters should be consistently applied.


Blaizzy (approved):
LGTM

Blaizzy:
Thanks @breynolds007 !

PR #341, opened 2025-05-05
Improve api server: /chat -> /chat/completions; +port customize

e1732a364fed:


e1732a364fed:
1. by using argparse, port can also be assigned when calling `python -m mlx_vlm.server`.
2. trying to make /chat to /chat/completions to make it openai compatible. However I overlooked the implementation.

e1732a364fed:
Damn, I find that the code is completely incompatible with openai api! why! But I'd better close this PR. 

It would be so much easier to implement openai api directly.

PR #385, opened 2025-06-22
Add option to save LoRA adapter after each training epoch

breynolds007:
This PR adds the ability to save the LoRA adapter (safetensors file) after each training epoch. This is really helpful with long training runs, so if the training is interrupted or the model performance starts to deteriorate during the run, it is possible to recover the adapter state from the end of any completed training epoch.

See screenshot showing output.

- Added parameter `--save-after-epoch` to lora.py
- Corrected type and description of parameter `--lora-alpha` in lora.py

Bildschirmfoto 2025-06-22 um 12 38 21


Blaizzy:
Thanks @breynolds007 

This is an awesome addition! ❤️

PR #397, opened 2025-06-27
[gemma3n] Correctly scale text embeddings for quantized gemma3n conversions

neilmehta24:
When quantizing the `Gemma3nTextScaledWordEmbedding` module, it follows this code path:
- https://github.com/ml-explore/mlx/blob/2c11d10/python/mlx/nn/layers/embedding.py#L42
- https://github.com/ml-explore/mlx/blob/2c11d10/python/mlx/nn/layers/quantized.py#L138

Which ends up overwriting the `Gemma3nTextScaledWordEmbedding` instance with `QuantizedEmbedding`. This means that `Gemma3nTextScaledWordEmbedding.__call__` never gets called when the model is quantized.

Fix this issue by moving the scaling outside of `Gemma3nTextScaledWordEmbedding` and deleting `Gemma3nTextScaledWordEmbedding`.

@awni : I would be curious to hear your thoughts on how to enable the use case of inheriting from `nn.Embeddings` while correctly quantizing the underlying weights in the base class.

### DIFF
--- a/mlx_vlm/models/gemma3n/language.py
+++ b/mlx_vlm/models/gemma3n/language.py
@@ -559,24 +559,6 @@ class Gemma3nDecoderLayer(nn.Module):
         return corrected_predictions
 
 
-class Gemma3nTextScaledWordEmbedding(nn.Embedding):
-    """This module overrides nn.Embeddings' forward by multiplying with embeddings scale."""
-
-    def __init__(
-        self,
-        num_embeddings: int,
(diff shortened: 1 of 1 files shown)


awni:
Good catch. I think the fix you have is probably the simplest and perfectly reasonable. The other option would be to make the custom embedding layer quantizable and implement the quantized version of it. That's something I would do if the modularity was more useful.. but I don't think this case justifies it.

PR #414, opened 2025-07-04
Fix gemma3n Vision OCR + LM only reponses

Blaizzy:


mlx_vlm/models/gemma3n/vision.py
adlai, line 1: this _appears_ to be the significant "fix"

mlx_vlm/models/gemma3n/gemma3n.py
adlai, line 1: the unconditional lift patched upon this file could well also be a critical component of the "fix"; never take my word for anything, especially not if only reading single patch comments from reviews on code I didn't write nor read entirely alone.

mlx_vlm/convert.py
adlai, line 1: see the definition of `save_weights()` in `./utils.py`

PR #454, opened 2025-08-05
Update generate.py | Don't cut the stop tokens

yukiarimo:
This fixes if your stop token is "</yuna>", for example, the resulting output will now include the full stoping token as "</yuna>" instead of "</yuna"


Blaizzy:
Unfortunetly, I'm closing this PR.

This issue is not with the stopping criteria.

PR #485, opened 2025-09-03
Bump minimum required Python version to 3.10

dokterbob:
Closes #484.

PR #486, opened 2025-09-03
Add uv lockfile

dokterbob:
For much, much faster installs, reproducible builds, automatic environment management and dependency deduplication.

Ref: https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile

Requires #485 (hence still draft).


Blaizzy:
Thank you so much for your awesome PRs! I really appreciate the time and effort you've put into this.

However, I've decided not to merge this particular PR at this time. The main concern is that it adds around 5,000 lines to the requirements, which would significantly increase our maintenance burden going forward. Additionally, I believe UV users should be able to work effectively without this file.

Hope you understand this is just about maintainability – your contributions are always welcome and appreciated!

PR #510, opened 2025-09-23
feat(dots-ocr): MLX integration + CLI + projector discovery

Garry-TI:
@Blaizzy , I give up on this. Someone smarter needs to take a look at it here is where I made it with current commits:

Implemented the full dots.ocr generation path on MLX:

Extended MLXQwen with vision token detection, blended <image> row overrides, optional logits bans, MLX sampler integration, and identity-preserving fallbacks so generation survives checkpoints without an exposed head.

Reworked the MLX CLI (dots_ocr_generate.py) to normalize prompts, enforce a single vision placeholder, append safety guards, run runtime diagnostics, and clean model output. It now auto-discovers the tokenizer’s vision token, projects/pads vision features to match Qwen’s hidden size, and bans raw vision tokens from the decoded text.

Added an identity “projector” path in DotsOCRForCausalLM_MLX so the full [T_v, 1536] vision sequence is preserved for prefix embedding work.

Made mlx_vlm.utils a package-compatible shim that re-exports the legacy module plus pdf_to_images, and created mlx_vlm.io as a stable import surface for the PDF helper.

I withheld the weight files since they were so large. if you can't recreate them let me know.

Garry

### DIFF
--- a/mlx_vlm/cli/dots_ocr_generate.py
+++ b/mlx_vlm/cli/dots_ocr_generate.py
@@ -117,6 +117,18 @@ def main():
     guard = " Do not include the literal token <image> in your answer. Output only the extracted text."
     prompt = prompt + guard
 
+    print(f"[diag] vision token: {tok_str} (id={tok_id})")
+    try:
+        with np.load(args.proj_npz) as diag_npz:
+            keys = list(diag_npz.keys())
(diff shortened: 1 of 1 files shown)


Blaizzy:
Thanks for the awesome work @Garry-TI!

I'm back from holidays and will take it from here.

I will close the PR and make a new one.

PR #521, opened 2025-10-03
changed image parameter instead of files in stream_generate

Manikandan-t:
The core issue was that the model wasn't receiving the image in the correct format, and the prompt wasn't being formatted with the necessary image tokens that vision models need to process images.

1. Named parameter: Changed from positional files to named parameter image=image_file
2. Image counting: Added num_images to tell the prompt formatter how many images to expect
3. Smart token skipping: Only include image tokens in the message that actually has the image attached (the last user message)
4. History handling: Handle both text-only and multimodal message formats from Gradio


Blaizzy (approved):
LGTM!

PR #565, opened 2025-11-07
Fix Qwen3 VL token-image feature mismatch

codewithsathya:
### Problem

When sending multiple images of different resolutions in a single prompt, the Qwen3 VL model sometimes raises the following error:

```
ValueError: Image features and image tokens do not match: tokens: 4528, features 6978
```

This is caused by incorrect usage of `mx.split` in the model code. The function was used with split sizes, but it actually expects split indices. Additionally, the splitting and concatenation logic is redundant and unnecessary.

### Solution

- Removed the redundant `mx.split` and concatenation logic from the Qwen3 VL model.
- Ensured that the number of image features and image tokens now match correctly, preventing the above error when handling multiple images of varying resolutions.

### Impact

- Fixes token-image feature alignment for multi-image prompts.
- Simplifies the code and removes unnecessary operations.

### DIFF
--- a/mlx_vlm/models/qwen2_5_vl/language.py
+++ b/mlx_vlm/models/qwen2_5_vl/language.py
@@ -14,7 +14,9 @@ from .config import ModelConfig, TextConfig
 
 
 class Qwen2RotaryEmbedding:
-    def __init__(self, dim, max_position_embeddings=2048, base=10000):
+    def __init__(
+        self, dim, max_position_embeddings=2048, base=10000, rope_scaling=None
+    ):
         self.dim = dim
         self.max_position_embeddings = max_position_embeddings
         self.base = base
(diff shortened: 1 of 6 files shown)


Blaizzy:
Hi @codewithsathya 

Thanks for the awesome PR!

Did you evaluate the results of this approach? 

Also, the split are a part of the original implementation. https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_vl_moe/modeling_qwen3_vl_moe.py#L1226

codewithsathya:
Hi @Blaizzy 

I first resolved the problem by ensuring the correct usage of `mx.split` with the code below.

```python
print("Hidden states shape from vision tower: ", hidden_states.shape)
split_sizes = (
    image_grid_thw.prod(-1) // self.vision_tower.spatial_merge_size**2
).tolist()
print("split_sizes: ", split_sizes)

# Convert sizes to cumulative indices for mx.split
split_indices = []
current_index = 0
for size in split_sizes[:-1]:  # Exclude the last one
    current_index += size
    split_indices.append(current_index)

print("split_indices: ", split_indices)
hidden_states = mx.split(hidden_states, split_indices, axis=0)

print("Hidden states after split: ", [hs.shape for hs in hidden_states])
hidden_states = mx.concatenate(hidden_states, axis=0).astype(
    hidden_states[0].dtype
)
```

Later, I realized this logic was redundant and removed it. I have evaluated the results, and the fix is still working correctly without the extra code.

PR #568, opened 2025-11-09
host and port params for server

mguella:
Allow the user to specify host and port when starting the server 

e.g.
```
mlx_vlm.server --host 0.0.0.0 --port 32100
```


Blaizzy (approved):
LGTM! 

Thank you very much @mguella :)

PR #576, opened 2025-11-11
[bugfix] fix mrope in qwen2vl and qwen2.5vl

JJJYmmm:
Hi! I noticed that you've implemented the interleaved mrope for qwen3vl in #564 ! However, qwen2vl/qwen2.5vl still use original mrope, so this pr aims to corret them! 🤗


JJJYmmm:
Got it, I’ll take a closer look these days. 🫡

PR #593, opened 2025-11-19
[WIP] Reduce deps core 

Blaizzy:
**Summary:**
 This PR removes the dependency on `torch`, `torchvision`, and `transformers` by porting the necessary processors directly into `mlx-vlm`. It also restructures `pyproject.toml` to support optional installations.

 **Changes:**
 * **Removed Dependencies:** Core installation no longer requires Torch or Transformers.
 * **New Extras:** Added optional flags for `[trainer]`, `[server]`, and `[audio]`.
 * **Refactoring:**
     * Replaced `mlx-audio` with `soundfile`.
     * Moved audio imports to be lazy-loaded within functions to avoid crashes for users without audio dependencies.
     * Cleaned up redundant imports in `utils.py`.
 * **Docs:** Added installation instructions for optional dependencies to the README.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ dependencies = {file = ["requirements.txt"]}
 [project.optional-dependencies]
 ui = ["gradio>=5.19.0"]
 trainer = ["datasets>=2.19.1", "tqdm>=4.66.2"]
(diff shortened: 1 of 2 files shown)


altaic:
Sort of related, have you considered replacing `py-opencv` which pulls in a rather hefty set of deps (120+)? It looks like it's currently only used to load and resize the frames of videos.

PR #604, opened 2025-11-26
Add Support for HunyuanOCR (hunyuan_vl)

voxmenthe:
## Summary
This PR adds support for **HunyuanOCR** (`hunyuan_vl`), a specialized multimodal model designed for high-resolution Optical Character Recognition. The architecture combines a 27-layer ViT vision encoder with a 24-layer language decoder, featuring custom "xdrope" 4D rotary embeddings and a sophisticated patch merging strategy.

This implementation was tested and achieves **effective parity** with the HuggingFace reference implementation, verified via logit comparison and end-to-end generation tests.

### 1. Architecture
-   **Vision Tower**: 27-layer ViT (1152 hidden) with a "Perceive" module (Patch Merger) that compresses spatial tokens and injects `image_begin`, `image_end`, and `image_newline` tokens.
-   **Language Model**: 24-layer decoder (1024 hidden) using Grouped Query Attention (16 heads, 8 KV heads) and SwiGLU MLPs.
-   **Position Embeddings**: Custom **XD-RoPE** (4D Rotary Position Embeddings) that encodes `[base, width, height, time]` dimensions into the rotary frequencies.

### 2. Key Design Decisions
-   **Processor Inheritance**: The `HunYuanVLImageProcessor` inherits from `ImageProcessingMixin` rather than `BaseImageProcessor`. This avoids automatic validation in `prepare_inputs` that incorrectly flags missing `eos_token` or `pad_token` attributes, which are handled differently in this model.
-   **Flattened Pixel Values**: Unlike most VLM implementations that pass `(B, H, W, C)` tensors, this model expects flattened patches `(num_patches, C * patch_size * patch_size)`. We maintained this format to align exactly with the HF reference inputs, performing the reshaping internally in `VisionEmbeddings`.
-   **Sanitization**:
    -   Maps `vit.*` to `vision_tower.*`.
    -   Transposes Conv2d weights from PyTorch `[out, in, kH, kW]` to MLX `[out, kH, kW, in]`.

## Gotchas & Things to Watch Out For

Getting parity for this model was non-trivial. Here are the critical technical nuances to be aware of:

### 1. The "Alpha" Scaling
The `config.json` specifies `rope_scaling: {type: "linear", alpha: 1000.0}`.
-   **Intuition**: You might expect standard linear scaling (`inv_freq /= alpha`) or base scaling (`base *= alpha`).
-   **Reality**: Neither works. The model produces garbage text or repeats characters.
-   **The Fix**: It uses **NTK-aware scaling** logic applied to the base frequency:
    ```python
    # language.py
    base = theta * alpha ^ (dim / (dim - 2))
    ```
    This shifts the base frequency from `10,000` to `~1.1e7`, which is essential for the model to attend to long sequences and high-resolution image features correctly.

### 2. Position ID Stack Order
XD-RoPE requires 4D position IDs. The stacking order is **strictly** `[base, width, height, time]`.
-   Early attempts using `[base, time, height, width]` caused subtle degradation in image descriptions.
-   We verified this order against the HF `processing_hunyuan_vl.py` logic (specifically how `apply_rotary_pos_emb_xdrope` slices the dimensions).

### 3. Vision Interpolation Hack
The model uses a learned position embedding in the vision tower that is interpolated to the image grid size.
-   **Standard**: `scale = src_size / dst_size`
-   **Hunyuan**: `scale = src_size / (dst_size + 0.1)`
-   This specific `+0.1` offset in the denominator is present in the HF implementation. Omitting it causes a massive drift in vision feature logits (L2 distance ~350 vs ~90). We implemented a custom manual bilinear interpolation in `VisionEmbeddings` to reproduce this exact behavior.

### 4. QK Norm Order
HunyuanOCR applies RMSNorm to Queries and Keys **after** the RoPE rotation.
-   Flow: `Projection -> RoPE -> QK Norm -> Attention`
-   This is implemented in the `Attention` class. Getting this order wrong results in significant generation quality loss.

### Logit Parity
We compared the MLX implementation against the HF reference (`transformers` commit `82a06db`) using `bfloat16`.

-   **Text-Only**: Exact match.
    -   Prompt: "Hello world"
    -   Top Token: `倦` (Logit: ~9.875) matches HF exactly.
-   **Image + Text**: High semantic alignment.
    -   Image: `menu.webp`
    -   HF Top 3: `<think>`, `刘邦`, `Unsaturated`
    -   MLX Top 3: `Unsaturated`, `刘邦`, `<think>`
    -   *Note*: The slight reordering is attributed to `bfloat16` vs `float16` accumulation differences in the deep vision tower (27 layers). The semantic set of top tokens is identical.

### End-to-End Generation
-   **OCR**: Correctly identifies text in `examples/images/menu.webp`.
-   **Description**: Generates detailed, accurate descriptions for `examples/images/paper.png`.

## 📋 Checklist
- [x] Model implementation (`language.py`, `vision.py`, `hunyuan_vl.py`)
- [x] Custom Processor (`processing_hunyuan_vl.py`)
- [x] Config mapping
- [x] Weight sanitization
- [x] Integration tests (`test_integration.py`)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,10 +9,3 @@ repos:
     -   id: isort
         args:
             - --profile=black
--   repo: https://github.com/PyCQA/autoflake
-    rev: v2.2.1
-    hooks:
-    -   id: autoflake
-        args:
-            - --in-place
-            - --remove-all-unused-imports


Blaizzy:
Thanks for the update @srjoglekar246! 

I will check what's blowing up the memory.

mlx_vlm/models/hunyuan_vl/vision.py
```diff
@@ -0,0 +1,342 @@
...
+        # pos_embed shape: (1, src_h, src_w, embed_dim)
+        src_h = pos_embed.shape[1]
+        src_w = pos_embed.shape[2]
+        embed_dim = pos_embed.shape[3]
+
+        if src_h == target_h and src_w == target_w:
+            return pos_embed
+
+        # Create interpolation grid
+        h_ratio = src_h / target_h
```
srjoglekar246: This needs to be
```
h_ratio = src_h / (target_h + 0.1)
```
and then 
```
src_i = (i + 0.5) * h_ratio - 0.5
```
(same for width)
otherwise quality is broken :-)

mlx_vlm/models/hunyuan_vl/vision.py
```diff
@@ -0,0 +1,342 @@
...
+            pos_embed = self._interpolate_pos_embed(base_pos_embed, target_h, target_w)
+            pos_embed = pos_embed.reshape(1, -1, self.embed_dim)
+            patch_pos_embed_list.append(pos_embed)
+
+        patch_pos_embed = mx.concatenate(patch_pos_embed_list, axis=1)
+        embeddings = patch_embeds + patch_pos_embed
+
+        return embeddings
+
+    def _interpolate_pos_embed(
```
srjoglekar246, line 157: Out of curiosity, why the manual looping over pixels instead of using the vectorized version? It causes about a ~20% perf diff on my M1 Mac.

mlx_vlm/models/hunyuan_vl/vision.py
Blaizzy: Thanks, will update and test 👌🏽

mlx_vlm/models/hunyuan_vl/vision.py
Blaizzy: I will optimize later. At the moment I'm trying to approach the source. 

But open to suggestions if you have a better solution.

srjoglekar246: No worries, makes sense! From some fiddling on my machine, @voxmenthe 's original vectorized implementation was more performant...but yeah I haven't verified the quality of that vs your implementation, so there might be a diff there. Thanks for the quick work!

srjoglekar246:
After a lot of digging I think I know whats causing the memory to blow up...

As I mentioned before, the issue was this loop that goes over all the transformer blocks in the `VisionModel`:
```
for layer in self.layers:
            hidden_states = layer(hidden_states)
```

After staring at the dot versions of a *single* block with & without compilation, I realized the primary difference apart from the weights initialization (which doesn't change anything), is the application of the scale onto the query in MLX's SDPA fallback implementation - in eager mode, this is the code that runs from MLX's [fast implementation](https://github.com/ml-explore/mlx/blob/a6d671718143a0d96715fa0d34962e2530599bdb/mlx/fast.cpp#L707):
```
auto q = multiply(array(scale, inputs[0].dtype()), inputs[0], s);
```
As you can see, its creating a new array of the same shape as the `query` for every call to SDPA - hence blowing up the memory. Compiling the above yields a fused `CompiledBroadcastMultiply` op that doesn't need to materialize a whole array just for the scale.

@Blaizzy / @voxmenthe  how about we do this in `VisionAttention`:
```
def compiled_sdpa(queries, keys, values, scale):
            return mx.fast.scaled_dot_product_attention(
                queries, keys, values, scale=scale
            )
        self.compiled_sdpa = mx.compile(compiled_sdpa)
```

Blaizzy:
Interesting! 

We could do that. What is the memory savings? 

In my investigation I found the feature size to be the issue. This model creates a large number of features of each images.

Check features in and out of embeddings layer in vision.py

Blaizzy:
Yes, that's what I meant. 

And yes, that's one of the problems I want to address as well. 

My intuition says that's also affecting LM accuracy.

mlx_vlm/models/hunyuan_vl/vision.py
Blaizzy: Could you please send a diff for this?

mlx_vlm/models/hunyuan_vl/vision.py
```diff
@@ -0,0 +1,334 @@
...
+        self, pos_embed: mx.array, target_h: int, target_w: int
+    ) -> mx.array:
+        src_h, src_w = pos_embed.shape[1], pos_embed.shape[2]
+
+        if src_h == target_h and src_w == target_w:
+            return pos_embed
+
+        # Create coordinate grids
+        h_coords = mx.arange(target_h) * (src_h / target_h)
+        w_coords = mx.arange(target_w) * (src_w / target_w)
```
srjoglekar246:
```suggestion
        # Create coordinate grids
        h_scale = src_h / (target_h + 0.1)
        w_scale = src_w / (target_w + 0.1)
        h_coords = (mx.arange(target_h) + 0.5) * h_scale - 0.5
        w_coords = (mx.arange(target_w) + 0.5) * w_scale - 0.5
```

mlx_vlm/models/hunyuan_vl/vision.py
srjoglekar246: Done

mlx_vlm/models/hunyuan_vl/vision.py
Blaizzy: Reduced memory usage in half using chunked attention!

From 15 down to 8GB 🚀

mlx_vlm/models/hunyuan_vl/vision.py
Blaizzy: Thank you very much!

mlx_vlm/models/hunyuan_vl/language.py
```diff
@@ -0,0 +1,479 @@
...
+
+                token_num = (llm_grid_w + 1) * llm_grid_h
+
+                # Ensure we don't go out of bounds
+                end_pos = min(pos + token_num, seq_len)
+                actual_token_num = end_pos - pos
+
+                if actual_token_num > 0:
+                    # w_index: [0, 1, ..., grid_w, 0, 1, ..., grid_w, ...] repeated for each row
+                    w_pattern = np.tile(np.arange(llm_grid_w + 1), llm_grid_h)[
```
srjoglekar246, line 379: @Blaizzy  this part is causing issues when I run a larger input (the paper) after a smaller one (the handwritten example) through the same model instance. Thats an unintended behavior right?

```
  File "/Users/sachinjoglekar/projects/odml/ocr/ocr_test.py", line 95, in <module>
    result = ocr_pipeline.run(image_path)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/sachinjoglekar/projects/odml/ocr/ocr_test.py", line 58, in run
    output = generate(
             ^^^^^^^^^
  File "/Users/sachinjoglekar/projects/odml/ocr/mlx-vlm/mlx_vlm/generate.py", line 539, in generate
    for response in stream_generate(model, processor, prompt, image, audio, **kwargs):
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/sachinjoglekar/projects/odml/ocr/mlx-vlm/mlx_vlm/generate.py", line 429, in stream_generate
    for n, (token, logprobs) in enumerate(
                                ^^^^^^^^^^
  File "/Users/sachinjoglekar/projects/odml/ocr/mlx-vlm/mlx_vlm/generate.py", line 319, in generate_step
    outputs = model(input_ids, pixel_values, cache=prompt_cache, mask=mask, **kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/sachinjoglekar/projects/odml/ocr/mlx-vlm/mlx_vlm/models/hunyuan_vl/hunyuan_vl.py", line 146, in __call__
    return self.language_model(
           ^^^^^^^^^^^^^^^^^^^^
  File "/Users/sachinjoglekar/projects/odml/ocr/mlx-vlm/mlx_vlm/models/hunyuan_vl/language.py", line 449, in __call__
    position_ids = self.get_xdrope_input_positions(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/sachinjoglekar/projects/odml/ocr/mlx-vlm/mlx_vlm/models/hunyuan_vl/language.py", line 387, in get_xdrope_input_positions
    w_pattern = np.tile(np.arange(llm_grid_w + 1), llm_grid_h)[
                        ^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: array is too big; `arr.size * arr.dtype.itemsize` is larger than the maximum possible size.
```

I see that the `t, h, w` seem to be garbage: `4721269296875711272, 4743355912799685949, 4760621870313309415`. Are we evaluation/allocating arrays based on some tracer or un-evaled arrays?

Blaizzy: I found and fixed this issue last night when I ran evals 

Will push the changes

Blaizzy:
# Evals:
Commit: [a8726d8](https://github.com/Blaizzy/mlx-vlm/pull/604/commits/a8726d87a79f6ca9c29e808ca88c8877dfbe2b5f)
tencent/HunyuanOCR 

Dataset | Correct | Total Samples | Score | Split | Reported | Diff
-- | -- | -- | -- | -- | -- | -- |
OCRBench | 791 | 1000 | 79.10% | test | 86.0% | ~6.9%

P-PPPP:
Thank you! That's awesome and carefully job!

PR #608, opened 2025-11-30
Fix issue 607: qwen3_vl ValueError: Image features and image tokens do not match

ziya32:
The vision model already returns concatenated features in the correct order.  No need for splitting and re-concatenating again, which is also error prone (with '//').  Remove.

Closes #670


Blaizzy:
Please run

> pre-commit run --all

Blaizzy (changes requested):
Pre-commit

PR #637, opened 2025-12-17
Add thinking budget parameter to limit reasoning tokens

tasercake:
Hello! I wanted to place a hard cap on the number of tokens used for reasoning (inside a `<think>...</think>` block from models like Qwen3) because the models would often spend way too long thinking.

So this a thing I built for my own use and figured there might be some interest in having this upstream as well. Happy to iterate on this if needed :)

Here's a summary of changes:

- Add `thinking_budget` parameter to limit tokens in `<think>...</think>` blocks
- Force-inserts `</think>` when budget is exceeded to end reasoning phase
- Configurable think tags via `thinking_start_token` and `thinking_end_token`
- CLI support via `--thinking-budget`, `--thinking-start-token`, `--thinking-end-token`

## Usage

```bash
# Limit thinking to 100 tokens
mlx_vlm generate --model <model> --prompt "..." --thinking-budget 100

# Custom thinking tokens (for models with different tags)
mlx_vlm generate --model <model> --prompt "..." --thinking-budget 100 \
    --thinking-start-token "<reasoning>" --thinking-end-token "</reasoning>"
```

## Testing

Added a `TestThinkingBudget` class to cover a few cases.

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -425,9 +425,7 @@ def generate_step(
                                 -float("inf"),
                                 dtype=next_logprobs.dtype,
                             )
-                            next_logprobs = next_logprobs.at[
-                                thinking_end_token_id
-                            ].set(0.0)
+                            next_logprobs[thinking_end_token_id] = 0.0
                             in_thinking = False
 
             mx.async_eval(next_y)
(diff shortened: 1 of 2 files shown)


tasercake:
@Blaizzy agree that something of this sort would be nice! I'd been thinking of doing something like this, but extending the StoppingCriteria class instead – I think your suggestion is much cleaner though.
I'll take a crack at building this out over the next few days.

PR #641, opened 2025-12-20
tfms v5 prep changes

Blaizzy:


jrp2014:
rc3 is here.  A possible snag is overlapping requirements (eg, mlx-lm pinned to one version of transformers, while there is another here)

PR #648, opened 2025-12-26
Fix: fixes mlx-vlm.generate --chat when prompted with no imgages

Deekshith-Dade:
Chat errors out with "TypeError: object of type 'NoneType' has no len()", this fixes the error.


Blaizzy:
Awesome, thanks for the fix @Deekshith-Dade!

Could you please remove the uv lock file then we can merge 👌🏽

Blaizzy:
LGTM, thanks!

PR #659, opened 2026-01-16
Implement Joycaption as a custom Llava model

nArn0:
Implement Joycaption as a specific llava model. All information and quantized Joycaption model available here: https://huggingface.co/n-Arno/joycaption-mlx-mxfp4

I know this PR may not fit perfectly in mlx-vlm, but just in case, i prefer proposing the PR even if it may be rejected.

Thanks for your work on mlx-vlm!

### DIFF
--- a/examples/test_joycaption.py
+++ b/examples/test_joycaption.py
@@ -1,0 +1,23 @@
+import mlx.core as mx
+from mlx_vlm import load, generate
+from mlx_vlm.prompt_utils import apply_chat_template
+from mlx_vlm.utils import load_config
+
+# Load the model
+model_path = "n-Arno/joycaption-mlx-mxfp4"
+model, processor = load(model_path)
+config = load_config(model_path)
+
+# Prepare input
+image = ["http://images.cocodataset.org/val2017/000000039769.jpg"]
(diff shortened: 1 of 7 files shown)


Blaizzy:
Hey @nArn0 
This is perfect fit actually, thanks for the contributions!

A could nits:
1. If the model uses llava arch, then in this PR you can just add a model mapping key, mapping llava_joycaption to the existing llava. https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/utils.py#L26-L33
2. This can be a notebook with more details, you can check the existing notebooks: [examples/test_joycaption.py](https://github.com/Blaizzy/mlx-vlm/pull/659/files#diff-c38c3a275d4014a4c173903f731b39463fde9c2ba4e7665a97fac441e2e7990f)

nArn0:
Oh! Thanks for your feedback! I'll look into it, JoyCaption looks like a classic llava but since it uses Siglip2, i had to get your implementation from mlx-embeddings for the vision part which didn't fit the classic CLIP llava used.

Indeed, a notebook would help for the example.

My only issue i really didn't find how to fix is that if torchvision is present, torch.nn.functional.interpolate fails with a cryptic error.

PR #663, opened 2026-01-16
Pass kwargs through to snapshot_download in get_model_path

Deekshith-Dade:
Allow passing additional arguments like local_files_only, cache_dir, token, etc. to snapshot_download via kwargs. This enables users to control download behavior when using load() and related functions.


Blaizzy:
Hey @Deekshith-Dade 

Thanks for the awesome PR! 

It makes sense, overall one of the things that I would implement this weekend was check local files first and only download when it's not available 

That way you wouldn't need the flags

PR #672, opened 2026-01-21
Fix: Correct position embedding for LLM in PaddleOCR-VL

zhang-prog:


zhang-prog:
@Blaizzy Hi, this PR addresses a bug in the position embedding calculation for the LLM within the PaddleOCR-VL. This fix is expected to improve the model’s accuracy.

PTAL.

PR #673, opened 2026-01-21
[Model] Add Voxtral Tekken adapter and text-only support

mzau:
** Add Voxtral Tekken adapter and text-only support**

Summary:
- add Voxtral text-only model scaffold/config
- add Tekken tokenizer adapter/processor via optional `mistral-common` extra (required for Voxtral)
- route Voxtral generate/load through Tekken instruct encoding and skip chat templates
- add Voxtral README note during conversion and a minimal language unit test

Testing Requirements:
- Voxtral weights: mistralai/Voxtral-Mini-3B-2507

Testing:
- python -m unittest mlx_vlm/tests/test_models.py
- python -m unittest mlx_vlm/tests/test_models.py -v
- Manual: python -m mlx_vlm.convert --hf-path dev/voxtral/hf_mirror/Voxtral-Mini-3B-2507 --mlx-path /tmp/voxtral-mini-text-mlx
- Manual: python -m mlx_vlm.generate --model /tmp/voxtral-mini-text-mlx --prompt "Three sentences about Helsinki" --max-tokens 3200

### DIFF
--- a/mlx_vlm/convert.py
+++ b/mlx_vlm/convert.py
@@ -1,8 +1,8 @@
 import argparse
 import glob
 import shutil
-from textwrap import dedent
 from pathlib import Path
+from textwrap import dedent
 from typing import Callable, Optional, Union
 
 import mlx.core as mx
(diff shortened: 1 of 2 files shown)


Blaizzy:
Hmm interesting! 

Could you open an issue with reproducible examples?

I would love to avoid duplication, but If we find out through your issue that the only way is to add voxtral here, then that's ok.

PR #685, opened 2026-01-26
fix: prevent base64 image data from being tokenized as text

ne2030:
# Fix: Prevent base64 image data from being tokenized as text in multimodal messages

## Problem

When using the OpenAI-compatible multimodal API format with base64-encoded images, the entire base64 string was being passed to the text tokenizer, causing:

- **Token explosion**: A 428x1000 pixel image (~570KB base64) generates ~422,000 tokens instead of the expected ~500 tokens
- **Memory exhaustion**: GPU runs out of memory trying to process 422K tokens (exceeds typical 262,144 token limit)
- **System crashes**: On macOS, this causes the system to become unresponsive due to memory pressure

### Root Cause

The `apply_chat_template` function in `prompt_utils.py` extracts message content and passes it directly to `get_message_json`. When using the OpenAI-compatible multimodal format:

```python
# Client sends:
{
  "content": [
    {"type": "text", "text": "Describe this image"},
    {"type": "image_url", "image_url": {"url": "data:image/png;base64,iVBORw0KGgo..."}}
  ]
}
```

The entire `content` list (including the base64 URL) was passed to `get_message_json`, which then called `MessageBuilder.text_message(content)`. Since `content` was a list (not a string), it got converted to a string representation that included the massive base64 data, which was then tokenized as text.

### Flow of the Bug

1. `server.py` correctly extracts images from content into a separate `images` list
2. `server.py` calls `apply_chat_template(chat_messages, ...)` with the original messages (still containing base64)
3. `apply_chat_template` gets `content = p.get("content")` - the full list with base64
4. `get_message_json(content, ...)` is called with the list
5. `MessageBuilder.text_message(content)` creates `{"type": "text", "text": [list with base64]}`
6. This gets tokenized → 422K tokens

## Solution

Added `extract_text_from_content()` helper function that:
1. Passes through string content unchanged
2. For list content (multimodal format), extracts only text from `type: "text"` and `type: "input_text"` items
3. Ignores `image_url`, `input_image`, and `input_audio` items (already handled separately by `server.py`)

The function is called in `apply_chat_template` at two code paths:
1. Single dict prompt: `isinstance(prompt, dict)`
2. List of prompts: `isinstance(prompt, list)`

This ensures only text is passed to the tokenizer regardless of how the prompt is structured.

## Changes

- **Added**: `extract_text_from_content()` function in `prompt_utils.py`
- **Modified**: `apply_chat_template()` to call the helper in both single dict and list prompt paths
- **Added**: Comprehensive test suite in `tests/test_prompt_utils.py` with 20 test cases

## Test Results

```
20 passed in 2.21s
```

Tests cover:
- String passthrough
- Multimodal content with text and images
- Multiple text parts
- Different content type variants (`text`, `input_text`, `image_url`, `input_image`, `input_audio`)
- Edge cases (empty content, malformed items, etc.)
- Critical case: Large base64 data NOT included in output
- **Integration tests**: Verify fix works in `apply_chat_template` for both dict and list prompt formats
- **BaseModel support**: Verify Pydantic model messages are handled correctly

## Impact

| Metric | Before | After |
|--------|--------|-------|
| Token count (428x1000 image) | 422,518 | ~509 |
| Memory usage | OOM crash | Normal |
| Response time | N/A (crash) | ~9.5s |

## Backward Compatibility

This change is fully backward compatible:
- String content works exactly as before
- Only affects multimodal list content, extracting the same text that would be expected
- Images are still processed via the existing `images` parameter (unchanged)


Blaizzy (approved):
LGTM, thanks!

PR #689, opened 2026-01-27
Distributed inference for Qwen3, Kimi K2.5 and K2.6

pcuenca:

PR #692, opened 2026-01-28
Honor quantization_config

pcuenca:
This makes it possible to load transformers quantized weights.

This is another mismatch I found while working on #689. I was puzzled that I had to use `QuantizedSwitchLinear` explicitly in order to be able to load the weights, whereas this was not necessary in mlx_lm. The quantization process immediately performed after sanitization is driven by the existence of quantization_config, and it adapts the weights accordingly so they have `.scales` and `.biases`.

Extracting as a separate PR for discussion, maybe I missed some side effects.


mlx_vlm/utils.py
```diff
@@ -214,6 +214,30 @@ def load_model(model_path: Path, lazy: bool = False, **kwargs) -> nn.Module:
...
+                logging.warning(
+                    "Quantization method %s is not supported in mlx_vlm.load_model()",
+                    quant_method,
+                )
+
+            if quantization is not None:
+                config["quantization"] = quantization
+                config["quantization_config"] = quantization
+
     if (quantization := config.get("quantization", None)) is not None:
```
Blaizzy, line 240: I believe the new code above should be here, right?

pcuenca: It's a bit confusing 😅. `quantization` is first read in L140 (which was already there). If it exists, it's a MLX-style quantization definition. If it doesn't, then we check if `quantization_config` exists, convert it to `quantization` format and store it in the config object. Perhaps we could remove L140 and just check if config has the `quantization` attribute in line 217.

Blaizzy: Let's refactor then 😎

I think this could start at line 241. 

So all the quantisation logic is clear in one place

pcuenca: Deleted the distant quantization line from L140 and simplified slightly.

PR #704, opened 2026-02-02
Fix incorrect attention mask in Qwen2-VL Vision

hturbe:
This PR addresses two issues in the Qwen2-VL implementation to enable correct behavior during batch processing.

### 1. Fix Vision Attention Mask Logic
The previous implementation of the attention mask in `vision.py` used an additive mask initialized with ones, where valid regions were set to zero. This is incorrect for additive masking, which requires adding `-inf` to masked positions.

* **Change:** Switched to a standard Boolean mask (supported by `mx.fast.scaled_dot_product_attention`) where `True` indicates valid attention and `False` indicates masked regions.
* **Impact:** This ensures correct isolation between images and prevents cross-sequence contamination in packed batches.

### 2. Fix Cache Handling in `create_attention_mask`
In `language.py` a list of cache objects was previously passed to `create_attention_mask` instead of the individual cache object. This prevented the method from invoking the `make_mask` method on the cache instance.

* **Change:** Passed individual cache objects to `create_attention_mask`.
* **Impact:** Ensures `make_mask` is correctly invoked, allowing for proper dynamic mask generation.

### DIFF
--- a/mlx_vlm/models/qwen2_vl/language.py
+++ b/mlx_vlm/models/qwen2_vl/language.py
@@ -46,7 +46,6 @@ class Qwen2RotaryEmbedding:
         return freqs_t
 
     def __call__(self, x, position_ids):
-
         # In contrast to other models, Qwen3VL has different position ids for the grids
         # So we expand the inv_freq to shape (3, ...)
         if position_ids.ndim == 2:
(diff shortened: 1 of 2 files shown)


Blaizzy (approved):
LGTM, thanks!

PR #709, opened 2026-02-03
fix: Log underlying ImportError when model loading fails

antonvice:
When a model fails to import due to missing dependencies, the current code swallows the ImportError and raises a generic 'Model type not supported' message. This makes debugging difficult.

This change captures and includes the underlying ImportError in the error message, making it much easier to diagnose missing dependencies (e.g., missing \`SwitchGLU\` from \`mlx-lm\` for \`deepseekocr_2\`).

Also adds \`get_class_predicate\` as a module-level factory function to support existing tests.

### DIFF
--- a/mlx_vlm/utils.py
+++ b/mlx_vlm/utils.py
@@ -74,8 +74,8 @@ def get_model_and_args(config: dict):
 
     try:
         arch = importlib.import_module(f"mlx_vlm.models.{model_type}")
-    except ImportError:
-        msg = f"Model type {model_type} not supported."
+    except ImportError as e:
+        msg = f"Model type {model_type} not supported. Error: {e}"
         logging.error(msg)
         raise ValueError(msg)
 


mlx_vlm/utils.py
```diff
@@ -838,7 +838,7 @@ def prepare_inputs(
     **kwargs,
 ):
 
-    if not images and not audio:
```
Blaizzy, line 841: Please revert this as well.

mlx_vlm/tests/test_utils.py
```diff
@@ -427,4 +462,4 @@ def test_load_passes_revision():
 
         assert model is model_mock
         assert processor is processor_mock
-        mock_get_model_path.assert_called_with("repo", revision="abc")
+        mock_get_model_path.assert_called_with("repo", force_download=False, revision="abc")
```
Blaizzy: Please revert this

mlx_vlm/tests/test_utils.py
```diff
@@ -8,14 +8,35 @@
...
-    get_class_predicate,
     load,
     prepare_inputs,
     process_inputs_with_fallback,
     sanitize_weights,
+    skip_multimodal_module,
     update_module_configs,
 )
 
+def get_class_predicate(skip_vision: bool, config: dict = None, weights: dict = None):
```
Blaizzy: You can mock this function rather than duplicating.

mlx_vlm/tests/test_utils.py
Blaizzy, line 1: I'm having trouble understanding how the changes to these files relate to this PR. Could you help me see the connection?

If there is none, please revert them.

mlx_vlm/utils.py
```diff
@@ -74,8 +74,8 @@ def get_model_and_args(config: dict):
 
     try:
         arch = importlib.import_module(f"mlx_vlm.models.{model_type}")
-    except ImportError:
-        msg = f"Model type {model_type} not supported."
+    except ImportError as e:
```
Blaizzy, line 77: This is the only change needed for this PR. Revert everything else or close this and open a new PR.

If tests break, ping me I will fix it.

antonvice: sorry

Blaizzy: No worries, you don't have to be sorry. You successfully identified and solve a critical problem, that's great! 😉 

I'm just helping you shape your contribution and setting you up for more successful contributions.

The rule of thumb is small atomic changes. 

Can't wait for your new PR with the fix!

Blaizzy:
Perfect, then I'm excited because it can be the first of many! 

Mine was 10 years ago, and I never stopped since.

PR #762, opened 2026-02-25
fix(qwen3_5): restore thinking prefix, forward kwargs, ensure 2D inpu…

ziya32:
…t_ids

- Restore <think> prefix stripped during prompt tokenization in stream_generate
- Forward **kwargs in apply_chat_template so enable_thinking takes effect
- Pass --processor-kwargs through to apply_chat_template in main/chat mode
- Ensure input_ids is 2D (batch, seq) in prepare_inputs for Qwen3.5

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -157,7 +157,7 @@ def parse_arguments():
         "--processor-kwargs",
         type=json.loads,
         default={},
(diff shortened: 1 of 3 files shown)


Blaizzy:
Indeed, 

This is a chat_template.jinja issue

PR #766, opened 2026-02-25
CORS middleware

viktike:
Use FastAPI's CORS middleware to serve HTTP OPTIONS requests.
In my case ChatBox.app initially requests the /models endpoint with OPTIONS instead of GET.


Blaizzy (approved):
LGTM, thanks!

Blaizzy (approved):
LGTM, thanks!

PR #776, opened 2026-02-26
Add KV cache (quantization) parameters to server.

viktike:
Memory constrained systems can benefit from quantization of the KVCache, not just the model.
This change adds the same KVCache command line arguments to the server as the CLI (generate).

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -19,6 +19,7 @@ from typing_extensions import Required, TypeAlias, TypedDict
 from .generate import (
     DEFAULT_MAX_TOKENS,
     DEFAULT_MODEL_PATH,
+    DEFAULT_QUANTIZED_KV_START,
     DEFAULT_SEED,
     DEFAULT_TEMPERATURE,
     DEFAULT_TOP_P,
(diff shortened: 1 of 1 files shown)


mlx_vlm/server.py
```diff
@@ -1095,9 +1145,38 @@ def main():
...
+    parser.add_argument(
+        "--quantized-kv-start",
+        type=int,
+        default=DEFAULT_QUANTIZED_KV_START,
+        help="Start index (of token) for the quantized KV cache.",
+    )
     args = parser.parse_args()
     if args.trust_remote_code:
         os.environ["MLX_TRUST_REMOTE_CODE"] = "true"
+    os.environ["KV_BITS"] = str(args.kv_bits)
```
Blaizzy, line 1175: Shouldn't we check if the values are set as env vars or not?

Because if the CLI args are always going to override the env vars (they have defaults that are not none), then why use env vars in the first place?

viktike: > Shouldn't we check if the values are set as env vars or not?

It's not intended to be set as env vars (just command line arguments) by the user. 
 
> Because if the CLI args are always going to override the env vars (they have defaults that are not none), then why use env vars in the first place?

They are only env vars because python vars are not propagated to the fastapi server workers. The default values are intentional.

Let me know if my explanation wasn't clear.

viktike: The MLX_TRUST_REMOTE_CODE is checked before because it's a boolean, meaning the absence of it is a valid false. These ones are ints.

Defaults are for unquantized KV cache, how it was always working, so if the user doesn't provide these command line arguments, it will work as before, no matter what env vars were present when the server started.

Blaizzy: Makes sense, thanks!

PR #784, opened 2026-03-03
Forward extra_body params (e.g. enable_thinking) to chat template

francip:
> [!NOTE]
> This patch was developed with the assistance of a coding agent.

## Problem

When using the OpenAI-compatible `/chat/completions` endpoint, extra body parameters like `enable_thinking` are silently dropped and never reach the Jinja chat template.

For models like Qwen3.5 whose chat template checks `enable_thinking` to toggle thinking/reasoning output, there is currently no way to control this behavior through the API — the model always produces its full thinking trace regardless of what the client sends.

```python
# This has no effect today — model always thinks
client.chat.completions.create(
    model="mlx-community/Qwen3.5-35B-A3B-4bit",
    messages=[{"role": "user", "content": "Hello"}],
    extra_body={"enable_thinking": False},
)
```

## Root cause

Two places where kwargs are dropped:

1. **`server.py`** — `chat_completions_endpoint` builds the `apply_chat_template()` call with only `num_images` and `num_audios`. Extra fields captured by `FlexibleBaseModel` (via `extra="allow"`) are never forwarded.

2. **`prompt_utils.py`** — `apply_chat_template()` accepts `**kwargs` and passes them to `get_message_json()`, but the final call to `get_chat_template()` on line 710 drops them. Since `get_chat_template()` already accepts and forwards `**kwargs` to the processor's `apply_chat_template()`, the fix is just adding `**kwargs` to that call.

## Fix

Minimal two-line patch:

- **`server.py`**: Extract `request.__pydantic_extra__` and pass as `**template_kwargs` to `apply_chat_template()`
- **`prompt_utils.py`**: Forward `**kwargs` from `apply_chat_template()` to `get_chat_template()`

## Testing

Tested with `mlx-community/Qwen3.5-35B-A3B-4bit` on macOS (Apple M4 Pro):

```bash
# With thinking disabled — clean response
curl localhost:2027/chat/completions -H "Content-Type: application/json" \
  -d '{"model":"mlx-community/Qwen3.5-35B-A3B-4bit",
       "messages":[{"role":"user","content":"Say hello"}],
       "max_tokens":256,"enable_thinking":false}'
# → {"choices":[{"message":{"content":"Hello!"}}], ...}

# With thinking enabled — full reasoning trace
curl localhost:2027/chat/completions -H "Content-Type: application/json" \
  -d '{"model":"mlx-community/Qwen3.5-35B-A3B-4bit",
       "messages":[{"role":"user","content":"What is 17*23?"}],
       "max_tokens":256,"enable_thinking":true}'
# → {"choices":[{"message":{"content":"Thinking Process:\n\n1. ..."}}], ...}
```

Verified that the Qwen3.5 chat template checks `enable_thinking` at render time and emits an empty `<think>\n\n</think>` block when `false`, suppressing the reasoning output.

### DIFF
--- a/mlx_vlm/lora.py
+++ b/mlx_vlm/lora.py
@@ -19,17 +19,21 @@ from .trainer.utils import (
     not_supported_for_training,
     unfreeze_modules,
 )
-from .utils import load, load_image_processor
+from .utils import load
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
+logging.getLogger("httpx").setLevel(logging.WARNING)
+logging.getLogger("httpcore").setLevel(logging.WARNING)
+logging.getLogger("huggingface_hub").setLevel(logging.WARNING)
 
 
 def transform_dataset_to_messages(dataset, model_type, custom_prompt_format=None):
     """
     Only transform dataset to messages format for VLMs with single-turn QA and image columns.
     If the dataset already has a 'messages' column, return as is.
-    Otherwise, require 'question', 'answer', and 'image' or 'images' columns.
+    Otherwise, require 'question' and 'answer' columns.
+    If present, 'image' or 'images' columns are included in the user content.
     No multi-turn or template logic. No audio support.
     """
     has_messages = (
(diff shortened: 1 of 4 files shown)


Blaizzy:
Hey @francip 

Thanks a lot for the contribution and the notice it does help to know! 

This is good start, there are just a couple changes I will make in the morning before merging :)

francip:
Sorry I didn't see this in time. Glad it's in.

PR #811, opened 2026-03-10
feat(server): add --model and --adapter-path flags for startup preloading

### LINKED ISSUE #810: allow `--model` option in server to load model in start up
While I understand the intention behind dynamic loading/unloading to change out models without spinning down/up the server, is there any point not to have `--model` option to give the user an option to load model on startup and not cold start?

Having model loading strictly to `v1/chat/completions` or `v1/responses` is not the best user experience for the first person to ping the endpoint (i.e timeout due to slow download speed if model has not even been downloaded, slow model loading)

I have a PR to implement this change.

auggie246:
## Overview

Adds optional `--model` and `--adapter-path` CLI flags to preload a model at server startup, eliminating the cold-start latency on the first request. Uses FastAPI's lifespan context manager, which also provides clean model unloading on shutdown.

## Key Changes

- Added `--model` and `--adapter-path` CLI arguments in `main()`, matching mlx-lm's style
- Added `lifespan` context manager that conditionally calls `get_cached_model()` on startup and `unload_model_sync()` on shutdown
- Uses environment variables (`PRELOAD_MODEL`, `PRELOAD_ADAPTER`) to pass CLI values across the uvicorn import boundary, consistent with existing patterns (`KV_BITS`, `PREFILL_STEP_SIZE`, etc.)
- No behavior change when `--model` is omitted — models still load lazily on first request

Closes #810

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -86,10 +86,13 @@ async def lifespan(app):
     model_path = os.environ.get("PRELOAD_MODEL")
     adapter_path = os.environ.get("PRELOAD_ADAPTER") or None
     if model_path:
-        print(f"Preloading model: {model_path}")
-        get_cached_model(model_path, adapter_path)
+        try:
+            print(f"Preloading model: {model_path}")
+            get_cached_model(model_path, adapter_path)
+        except Exception as e:
+            print(f"Failed to preload model: {e}")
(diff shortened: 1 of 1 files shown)


Blaizzy (approved):
LGTM, thanks!

PR #814, opened 2026-03-12
Add common sampling parameters to server endpoints

spicyneuron:
This PR wires up four additional sampling parameters through the server endpoints. Requests now also support:

- `top_k`
- `min_p`
- `repetition_penalty`
- `logit_bias`

Added tests for both endpoints and `generate_step` to verify proper forwarding.

I held off on `chat_template_kwargs` since actual usage in this project seemed a bit nuanced. Keys like `num_images` and `skip_image_token` aren't really meant for the template, but are currently included. And simply passing request-level kwargs to the template opens up questions about collisions and precedence.

This PR only handles the straightforward cases, but I'd be happy to submit a follow up if you weigh in on preferred future direction for `chat_template_kwargs`. Keep them mixed and filter out reserved keys? Refactor to separate the concerns? Something else?

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -395,7 +395,6 @@ def generate_step(
             return y, logprobs.squeeze(0)
 
     with mx.stream(generation_stream):
-
         # Get input embeddings (handles both multimodal and text-only)
         embedding_output = model.get_input_embeddings(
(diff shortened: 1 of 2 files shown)


Blaizzy (approved):
LGTM, thanks!

PR #818, opened 2026-03-13
Add an 'index' for each tool call as per OpenAI spec.

viktike:
Turns out the OpenAI spec mandates an index integer for each tool call.

Most clients don't mandate it's presence, but I found that OpenCode does.


Blaizzy (approved):
LGTM, thanks!

PR #823, opened 2026-03-15
fix: pass images to vision encoder during LoRA training for Qwen/Gemma/SmolVLM

kikoncuo:
## Summary

`VisionDataset.process()` passes `images=None` to `prepare_inputs` for models with `use_embedded_images=True` (Qwen, Gemma, SmolVLM). This means **LoRA training runs without any actual image data** — the model trains on `<|image_pad|>` tokens with meaningless embeddings.

**Impact**: After training, the adapter corrupts generation — the model outputs only vision tokens (`<|vision_start|><|image_pad|><|vision_end|>`) instead of text.

## Root cause

```python
# Line 109 — images are discarded for Qwen/Gemma/SmolVLM
inputs = prepare_inputs(
    processor=self.processor,
    images=None if use_embedded_images else (images if images else None),
    ...
)
```

The `use_embedded_images` flag was intended to indicate that images should be embedded inline via the processor, but the implementation discards them entirely.

## Before vs After

| Metric | Before (broken) | After (fixed) |
|--------|-----------------|---------------|
| Tokens per sample | 23 | **1,076** |
| Trained tokens (10 iters) | 230 | **10,760** |
| Initial training loss | 2.66 | **22.14** |
| Loss after 10 iters | 0.43 (stuck) | **3.76** (dropping) |
| Peak memory | 1.3 GB | 6.6 GB |
| Generation after training | Vision tokens only | Normal text |

## The fix

For embedded-image models:
- Tokenize the text prompt normally (keeps raw `<|image_pad|>` placeholder tokens)
- Process images separately via the image processor to get `pixel_values` + `image_grid_thw`
- Expand placeholder tokens to match vision encoder output count (`prod(thw) // merge_size^2`)
- Return 1D tensors (no batch dim) since `iterate_batches` expects this

Also fixes the same bug in `PreferenceVisionDataset`.

## Testing

Tested with `mlx-community/Qwen3.5-0.8B-8bit` on the [keysay-vlm-context-training](https://huggingface.co/datasets/Enriqueag26/keysay-vlm-context-training) dataset (206 screenshot images).

```bash
python3 -m mlx_vlm.lora \
  --model-path mlx-community/Qwen3.5-0.8B-8bit \
  --dataset Enriqueag26/keysay-vlm-context-training \
  --iters 10 --batch-size 1 --learning-rate 1e-5 \
  --lora-rank 8 --lora-alpha 16 --max-seq-length 4096 \
  --grad-checkpoint --steps-per-report 1
```

**Before fix:**
```
Iter 1: Train loss 2.665, Tokens/sec 112, Trained Tokens 23
Iter 10: Train loss 0.430, Trained Tokens 230  <- stuck, ~23 tok/iter
```

**After fix:**
```
Iter 1: Train loss 22.143, Tokens/sec 95, Trained Tokens 1076
Iter 10: Train loss 3.760, Trained Tokens 10760  <- learning, ~1076 tok/iter
```

### DIFF
--- a/mlx_vlm/trainer/datasets.py
+++ b/mlx_vlm/trainer/datasets.py
@@ -2,6 +2,7 @@ import json
 import warnings
 
 import mlx.core as mx
+import numpy as np
 
 
 def get_prompt(model_type, processor, conversation):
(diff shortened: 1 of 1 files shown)


kikoncuo:
No problem I was testing it and got more errors downstream, I think it won't work anyways because how the LoRaLayer in this repo interacts with Qwen3.5's during inference, I'll open the issue

PR #829, opened 2026-03-17
Clean up server + generate args, validation, and defaults

spicyneuron:
My original intent was to add a couple more server arguments to bring `mlx-vlm.server` closer to parity with `mlx-lm.server`. But once I got into it, it felt like some cleanup would make that work easier. Thus this PR.

The main goal here is to simplify how generation and template arguments are defined, validated, and forwarded through the CLI and server paths. The result is less duplication, more consistency, and a cleaner base for follow-up work.

A few behavior changes are worth calling out:

- single-value `resize_shape` shorthand is now official in the server API. Before, lower-level code already had the concept of expanding a single value to a square shape, but the server still effectively required a two-value field. Now passing a single value like `[512]` is explicitly supported and normalized to `(512, 512)`.
- `resize_shape` validation is stricter now. Before, odd or ambiguous inputs could make it farther down the stack or fail in less obvious ways. Now invalid values are rejected instead.
- request validation now happens through the request models. Before, this path returned a custom 400 for some invalid server inputs. Now invalid request bodies surface through the FastAPI-standard validation response.

I also normalized a few default values that were previously inconsistent:

- `temperature`: `0.5` -> `0.0`
- `generate_step(..., quantized_kv_start=...)`: `0` -> `5000`
- `BatchGenerator.max_tokens`: `128` -> `256`

Let me know if you prefer different values / changes. Would be happy to adjust.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 test.ipynb
 **.pyc
-
- Byte-compiled / optimized / DLL files
+.claude
+Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
 
--- a/mlx_vlm/models/mistral3/config.py
+++ b/mlx_vlm/models/mistral3/config.py
@@ -18,14 +18,32 @@ class TextConfig(BaseModelConfig):
     max_position_embeddings: Optional[int] = None
     num_key_value_heads: Optional[int] = None
     rope_theta: float = None
-    rope_parameters: Optional[Dict[str, Union[float, str]]] = None  # For Ministral3
+    rope_parameters: Optional[Dict[str, Union[float, str]]] = None
     rope_traditional: bool = False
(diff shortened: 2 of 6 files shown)


mlx_vlm/generate.py
```diff
@@ -196,13 +205,34 @@ def parse_arguments():
...
+    shape = []
+    for value in values:
+        if isinstance(value, bool) or not isinstance(value, Integral):
+            raise ValueError("resize_shape must contain 1 or 2 integers")
+        shape.append(int(value))
+
+    if len(shape) not in (1, 2):
+        raise ValueError("resize_shape must contain 1 or 2 integers")
+
+    return (shape[0], shape[0]) if len(shape) == 1 else tuple(shape)
```
Blaizzy: Could you simplify this?

spicyneuron: Sure thing. Done!

PR #838, opened 2026-03-19
Fix multimodal prompt templating and Qwen image forwarding in dataset prep

quist00:
Thank you all for having this project.  I was trying to replicate a training I did in PyTorch on NRP locally under mlx-vlm, but I ran into some issues that with AI support I have separated into two pull requests.  Here is the first one.  Thanks in advance for your consideration.

Goal
I was fine-tuning Qwen2.5-VL on book spine images and needed reliable multimodal supervision, where both chat template formatting and image tensor routing work consistently during training.

What was breaking

- Multimodal chat template formatting could fail in tokenizer-level paths when message content is structured as image/text - chunks.
- Qwen image forwarding behavior in dataset preparation was gated in a way that could bypass expected pixel-value flow during supervised training.

Why this patch was necessary

- Processor-first chat templating is the correct multimodal path for Qwen VL style message structures.
- Keeping tokenizer fallback preserves compatibility for non-multimodal or older setups.
- Enabling proper Qwen image forwarding ensures the model is actually trained with visual signal instead of text-only degradation.

### DIFF
--- a/mlx_vlm/trainer/datasets.py
+++ b/mlx_vlm/trainer/datasets.py
@@ -98,9 +98,7 @@ class VisionDataset:
                 "Config must contain 'image_token_index' or 'image_token_id'"
             )
 
-        use_embedded_images = (
-            model_type.startswith("gemma") or model_type == "smolvlm"
-        )
+        use_embedded_images = model_type.startswith("gemma") or model_type == "smolvlm"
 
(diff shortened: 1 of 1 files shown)


Blaizzy:
Thanks, please open a GH issue first with a reproducible example :)

PR #918, opened 2026-04-04
Centralize server config and add CLI flags

spicyneuron:
An incremental followup to https://github.com/Blaizzy/mlx-vlm/pull/829.

Main goal was to centralize how server-side generation config is resolved, and expose a few more server CLI flags so that they don't need to be carried by request arguments. Most of the change is internal cleanup / deduplication rather than new behavior. 

## Changes

At a high level, this PR:
- centralizes generation config/default resolution in shared helpers
- routes both `/responses` and `/chat/completions` through the same config merge path
- adds server CLI flags for common generation defaults
- keeps request-level values taking precedence over server defaults

There are also a couple of smaller behavioral fixes folded in:
- `trust_remote_code` is now part of the model cache key, so cached loads better reflect the actual load configuration
- `model` is now optional in both endpoints IF a default `--model` was defined (previously was inconsistent)

Also added coverage around the shared config resolution and the server merge path, including default handling and request overrides.

## Adjacent Question

One thing I'm unsure about is `--reload`. Is hot reloading Uvicorn actually useful enough to justify the extra complexity needed to pass configs to reloaded processes?

That accounts for the current environment variable pattern. If `reload` is not especially valuable, I think we could simplify this even further by keeping config in-process.


mlx_vlm/generate.py
```diff
@@ -46,6 +46,55 @@
...
+    resolved.update(filter_generation_config(generation))
+
+    resolved["resize_shape"] = normalize_resize_shape(resolved["resize_shape"])
+
+    if resolved["kv_bits"] == 0:
+        resolved["kv_bits"] = None
+    if resolved["max_kv_size"] == 0:
+        resolved["max_kv_size"] = None
+
+    if model_name and "qat" in model_name.lower() and resolved["kv_bits"] is not None:
```
Blaizzy, line 92: what's this "qat" condition for?

Blaizzy (approved):
LGTM, thanks!

I just have one question before we merge

mlx_vlm/generate.py
spicyneuron: @Blaizzy It was introduced to server.py in https://github.com/Blaizzy/mlx-vlm/pull/776/changes and I moved it to generate.py for consistency. Though admittedly, I haven't used this combination myself.

Just pushed another incremental commit so that generate.py also uses `resolve_generation_config` and logs the same warning.

Blaizzy:
Hey @spicyneuron 

Let's do this after todays release. There are a lot of breaking changes coming.

So let's close this PR and re-do it after the release, ok?

spicyneuron:
@Blaizzy No sweat. This entire thing has been a yak shave off of adding --max-tokens to the CLI. Don't care if it's my PR that lands it — just happy to have a more reliable, extensible server.

PR #920, opened 2026-04-04
fix: preserve LoRA adapter across requests in OpenAI responses endpoint

JasonOA888:
## Problem

When starting `mlx_vlm server` with `--adapter-path`, the adapter loads correctly on startup but gets dropped after the first request. The OpenAI-compatible `/v1/responses` endpoint calls `get_cached_model(model)` without passing `adapter_path`, so the cache key becomes `(model, None)` instead of `(model, adapter_path)`. This triggers a cache mismatch, unloads the adapted model, and reloads the base model.

## Fix

- Add `server_adapter_path` module-level variable to track the adapter path set at startup
- Pass `server_adapter_path` to `get_cached_model()` in both the lifespan preload and the responses endpoint
- Now the cache key matches and the adapter persists across requests

## Testing

1. Start server with `--adapter-path ./my-adapter`
2. Send multiple requests to `/v1/responses`
3. Verify adapter is still loaded via `/health` endpoint

Closes #907

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -87,13 +87,14 @@ def get_quantized_kv_start():
 
 @asynccontextmanager
 async def lifespan(app):
+    global server_adapter_path
     # Startup
     model_path = os.environ.get("PRELOAD_MODEL")
-    adapter_path = os.environ.get("PRELOAD_ADAPTER") or None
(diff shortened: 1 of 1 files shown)


lucasnewman:
Thanks! This is fixed for you original case, and we're covering the newer endpoints here: https://github.com/Blaizzy/mlx-vlm/pull/1251

PR #921, opened 2026-04-04
fix: prevent crash when args.adapter_file is None in sft_trainer

JasonOA888:
## Problem

`sft_trainer.py` crashes with `TypeError: argument should be a str or an os.PathLike object` when `args.adapter_file` is `None`. This happens when `TrainingArgs` is constructed with an explicit `None` value for `adapter_file` (e.g. from programmatic usage).

## Fix

- Add defensive fallback: `adapter_file = args.adapter_file or "adapters.safetensors"` at both save points (checkpoint and final)
- Create parent directory before saving to prevent `FileNotFoundError`

Both the checkpoint save (line ~471) and final save (line ~484) are covered.

Closes #908

### DIFF
--- a/mlx_vlm/trainer/sft_trainer.py
+++ b/mlx_vlm/trainer/sft_trainer.py
@@ -468,20 +468,24 @@ def train(
 
         # Save checkpoint
         if it % args.steps_per_save == 0 and rank == 0:
-            save_adapter(model, args.adapter_file)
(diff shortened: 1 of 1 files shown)


lucasnewman:
Resolved in https://github.com/Blaizzy/mlx-vlm/pull/1255

PR #935, opened 2026-04-05
Fix Gemma 4 quantized per-layer projection loading

spicyneuron:
While trying to run [unsloth/gemma-4-E2B-it-UD-MLX-4bit](https://huggingface.co/unsloth/gemma-4-E2B-it-UD-MLX-4bit), I hit a loading error:

```python
ValueError: Unable to quantize model of type <class 'mlx_lm.models.gemma4_text.ScaledLinear'>
```

After cross-checking this against the Gemma 4 implementation in Transformers ([constructor](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma4/modeling_gemma4.py#L1524-L1531), [projection path](https://github.com/huggingface/transformers/blob/main/src/transformers/models/gemma4/modeling_gemma4.py#L1667-L1678)), I replaced the custom `ScaledLinear` wrapper used for `per_layer_model_projection` with a standard bias-free `nn.Linear` and moved the `hidden_size**-0.5` scale into the projection path explicitly.

The math stays the same, but the layer now works with MLX's normal quantization and loading flow.

For reference, here's the same fix on mlx-lm: https://github.com/ml-explore/mlx-lm/pull/1112


Blaizzy (approved):
LGTM, thanks!

PR #936, opened 2026-04-05
Add --max-tokens CLI argument to server

nnorris7:
## Summary

- The server defaults to 256 max tokens when the client doesn't send `max_tokens` in the request, which is too low for conversational use (e.g. via Open WebUI)
- Adds a `--max-tokens` CLI flag so operators can set a higher default at launch time (e.g. `--max-tokens 8192`)
- Per-request `max_tokens` from clients still takes precedence

## Test plan

- [x] `python -m mlx_vlm.server --help` shows the new `--max-tokens` flag with default 256
- [ ] Server launched with `--max-tokens 8192` uses 8192 when client omits `max_tokens`
- [ ] Client-provided `max_tokens` still overrides the server default

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -45,6 +45,10 @@ DEFAULT_SERVER_HOST = "0.0.0.0"
 DEFAULT_SERVER_PORT = 8080
 
 
+def get_default_max_tokens():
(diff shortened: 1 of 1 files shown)


lucasnewman:
This is resolved in 0.6.0 and later.

PR #939, opened 2026-04-06
Flash-style tiled TBQ prefill: bound activation memory

Blaizzy:
## Summary

- Rewrite `prefill_attention` to tile along the KV sequence dimension with online softmax (Welford-style), mirroring MLX's flash attention strategy
- Score matrix per chunk is capped at ~512 MB instead of scaling linearly with context length
- No new Metal kernels — reuses existing `_multi_query_prod_score_kernel` + `_single_tile_value_weighted_sum_kernel` on chunks of T

### Before (old `prefill_attention`)

Materializes full `(B×H×R, L, T)` score matrix in float32:

| Context | BL Activ. | TBQ Activ. | Score matrix |
|---------|-----------|------------|--------------|
| 40K     | 0.45 GB   | **10.04 GB**  | 9.8 GB       |
| 80K     | 0.76 GB   | **19.80 GB**  | 19.5 GB      |
| 106K    | 0.95 GB   | **26.15 GB**  | 25.9 GB      |
| 128K    | 1.12 GB   | **31.52 GB**  | 31.2 GB      |

### After (tiled with online softmax)

Processes KV in chunks, score matrix bounded at ~512 MB per chunk:

| Context | BL Activ. | TBQ Activ. | chunk_T | Score / chunk |
|---------|-----------|------------|---------|---------------|
| 40K     | 0.45 GB   | **1.15 GB**   | 2048    | 512 MB        |
| 80K     | 0.76 GB   | **1.15 GB**   | 2048    | 512 MB        |
| 106K    | 0.95 GB   | **1.15 GB**   | 2048    | 512 MB        |
| 128K    | 1.12 GB   | **1.15 GB**   | 2048    | 512 MB        |

**TBQ activations reduced from 31 GB → 1.15 GB (constant) at 128K context.**

> **Why BL is still lower:** BL uses a single fused flash attention Metal kernel where scores stay in registers/threadgroup memory and never hit global memory. TBQ requires two separate Metal kernel dispatches (score → value sum), so scores must be written to global memory between them. A fused L>1 kernel (extending `_fused_integer_decode_single_tile_kernel` to multi-query) would close this gap.

## KV Cache Savings (unchanged)

| Context | BL KV    | TBQ KV   | Savings |
|---------|----------|----------|---------|
| 40K     | 3.83 GB  | 1.70 GB  | 56%     |
| 80K     | 6.88 GB  | 2.61 GB  | 62%     |
| 106K    | 8.87 GB  | 3.20 GB  | 64%     |
| 128K    | 10.55 GB | 3.71 GB  | 65%     |

## Test plan

- [x] Smoke test on `gemma-4-e4b-it` (10K context) — same output, TBQ prefill 1547 t/s
- [x] Smoke test on `gemma-4-31b-it` (10K context) — same output agreement
- [ ] Test with longer contexts (40K+) on 31B to verify memory bounded
- [ ] Verify output quality matches old path on LongBench-v2

### DIFF
--- a/mlx_vlm/turboquant.py
+++ b/mlx_vlm/turboquant.py
@@ -71,101 +71,6 @@ def _mse_score_kernel():
 
 
 @lru_cache(maxsize=None)
-def _mse_score_tiled_kernel(repeat_count: int):
-    """Tiled MSE score: preload R queries into registers, loop over tokens.
-    Processes TokTileSize tokens per threadgroup instead of 1.
-    Reduces threadgroup count by TokTileSize×, amortizing dispatch overhead."""
-    if not _metal_available() or repeat_count < 1:
-        return None
-
-    lines = [
-        "        auto lane = thread_position_in_grid.x;",
-        "        auto n = thread_position_in_grid.z;",
-        "",
-        "        auto token_count = norms_shape[2];",
-        "        auto kv_heads = norms_shape[1];",
-        "        auto num_tok_tiles = (token_count + TokTileSize - 1) / TokTileSize;",
-        "        auto bh = n / num_tok_tiles;",
-        "        auto tok_tile = n % num_tok_tiles;",
-        "        int t_start = tok_tile * TokTileSize;",
-        "        int t_end = min(t_start + TokTileSize, (int)token_count);",
-        "",
-        "        auto q_base = q_rot + (bh * RepeatCount) * Dim;",
-        "        auto pk = packed + bh * token_count * PackedWidth;",
-        "        auto nm = norms + bh * token_count;",
-        "",
-        "        // Preload query values into registers (constant across all tokens)",
-    ]
-
-    for r in range(repeat_count):
-        lines += [
-            f"        float qr_{r}[DimsPerLane];",
-            f"        for (int i = 0, d = lane; i < DimsPerLane; i++, d += 32)",
-            f"            qr_{r}[i] = (d < Dim) ? static_cast<float>(q_base[{r} * Dim + d]) : 0.0f;",
-        ]
-
-    lines += [
-        "",
-        "        // Precompute bit offsets for key unpacking",
-        "        int k_words[DimsPerLane], k_offs[DimsPerLane];",
-        "        bool k_spills[DimsPerLane];",
-        "        for (int i = 0, d = lane; i < DimsPerLane; i++, d += 32) {",
-        "            int bo = d * Bits;",
-        "            k_words[i] = bo / 32;",
-        "            k_offs[i] = bo % 32;",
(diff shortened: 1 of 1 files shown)


Landon-Molt:
## Benchmark: gemma-4-31B-it attention dimensions (32K & 64K context)

Tested PR #939's tiled prefill vs the current dequantize+SDPA fallback using a standalone simulation of all 60 gemma-4-31B layers (50 SWA x 16 KV heads x D=256 + 10 full-attn x 4 KV heads x D=512). TQ 4-bit, B=1, 128 query tokens.

### Memory: Constant overhead -- confirmed

| Context | TQ cache | Tiled activation | Dequantize activation | Reduction |
|---------|----------|------------------|-----------------------|-----------|
| 32K | 7.03 GB | **68.5 MB** | 4,624 MB | 67.5x |
| 64K | 14.06 GB | **68.5 MB** | 9,232 MB | 134.7x |

Tiled activation memory is **constant at 68.5 MB** regardless of context length. This matches the PR description -- bounded by O(L x chunk_T) per chunk instead of O(L x T).

### Speed: Regression from multi-dispatch overhead

| Context | Tiled (60 layers) | Dequantize+SDPA (60 layers) | Slowdown |
|---------|-------------------|----------------------------|----------|
| 32K | 53.99s | 3.11s | 17.4x |
| 64K | 100.56s | 6.04s | 16.6x |

The dequantize path benefits from `mx.fast.scaled_dot_product_attention` -- a single fused Metal kernel where scores stay in registers. The tiled path dispatches two separate Metal kernels (score -> value sum) per chunk in a Python loop, with scores written to global memory between them. This matches the PR's own note:

> BL uses a single fused flash attention Metal kernel where scores stay in registers/threadgroup memory and never hit global memory. TBQ requires two separate Metal kernel dispatches

### Analysis

The memory improvement is dramatic and confirmed at scale. The speed gap is the main blocker. A fused L>1 Metal kernel (extending `_fused_integer_decode_single_tile_kernel` to multi-query) would close this gap.

For context: we discovered this issue downstream in [omlx](https://github.com/jundot/omlx) while investigating why TurboQuant showed zero memory savings ([jundot/omlx#661](https://github.com/jundot/omlx/issues/661)). The root cause is that `prefill_attention()` always returns `None` after #909 dropped ProdCodec, causing the dequantize fallback to trigger on every prefill chunk. See #1016 for the full trace.

### Test setup
- Platform: MacBook Pro M5 Max 128GB (macOS Darwin 25.4.0)
- MLX, standalone layer simulation (not full model inference)
- Simulates all 60 layers with correct gemma-4-31B architecture dimensions

PR #944, opened 2026-04-06
Fix TurboQuant crash with PromptCacheState prefix reuse

damonvjanis:
`stream_generate` crashes with `AttributeError: 'TurboQuantMSEState' object has no attribute 'shape'` when using TurboQuant (`kv_bits=3.5`) together with `prompt_cache_state` for multi-turn prefix reuse.

The cache trimming code assumes `.keys` is a raw tensor and does `c.keys.shape[2]` / `c.keys[:, :, :n, :]`. TurboQuant caches store quantized NamedTuples instead. Fix: use `c.trim()` which `TurboQuantKVCache` already provides.

Repro: call `stream_generate` twice with `prompt_cache_state=PromptCacheState()` and `kv_bits=3.5, kv_quant_scheme="turboquant"`. Second call crashes at line 691.

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -688,12 +688,19 @@ def stream_generate(
             # Trim cache to prefix_len in case it includes generated tokens
             for c in kv_cache:
                 if hasattr(c, "keys") and c.keys is not None:
-                    cached_len = c.keys.shape[2]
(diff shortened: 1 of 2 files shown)


lucasnewman:
Resolved in https://github.com/Blaizzy/mlx-vlm/pull/1748

PR #946, opened 2026-04-06
Wire up PromptCacheState in server for cross-request KV cache reuse

trevorgordon981:
## Summary

The `generate.py` module already has `PromptCacheState` with prefix matching and KV cache persistence across turns, but the server never passes it through. Every request creates a fresh KV cache from scratch, so TTFT scales with full context length on every call — even when the system prompt is identical.

This patch:
- Creates a `PromptCacheState` per loaded model in `model_cache`
- Passes it to `stream_generate`/`generate` in both `/v1/chat/completions` and `/responses` endpoints
- Reports `cached_tokens` in `prompt_tokens_details` in the response (OpenAI-compatible)

**28 lines added, 0 removed. No changes to generate.py or any other file.**

## Results

Tested on Qwen3.5-397B-A17B-4bit (hybrid GDN + MoE architecture, Mac Studio M3 Ultra 512GB):

| Request | Cached Tokens | TTFT | Speedup |
|---------|--------------|------|---------|
| Cold | 1/1824 | 5.43s | — |
| Warm 1 | 1811/1824 | 1.66s | 3.3x |
| Warm 2 | 1811/1824 | 0.62s | 8.8x |

This is especially impactful for agent workloads where a large system prompt (tool definitions, personality config, memory) is repeated across every request in a conversation.

## Context

With mlx-lm PR #1072 (merged), hybrid models like Qwen3.5 now support KV cache trimming and reuse. `PromptCacheState` in mlx-vlm's `generate.py` already handles the prefix matching, cache trimming, and state persistence — it just wasn't wired into the server.

## Test plan

- [x] Non-streaming `/v1/chat/completions` with repeated system prompt shows `cached_tokens > 0` on 2nd+ request
- [x] TTFT drops proportionally to cached prefix length
- [x] Tested on hybrid GDN/MoE architecture (Qwen3.5-397B-A17B)
- [ ] Verify streaming path caches correctly
- [ ] Verify `/responses` endpoint caches correctly
- [ ] Verify vision models with image inputs still work (cache should skip when image tokens change)

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -685,9 +685,18 @@ def stream_generate(
                 kwargs.pop("cached_image_features", None)
             # Reuse the saved KV cache (trimmed to prefix length)
             kv_cache = prompt_cache_state.cache
-            # Trim cache to prefix_len in case it includes generated tokens
+            # Trim cache to prefix_len in case it includes generated tokens.
+            # Use the cache's own trim() method when available (handles KVCache,
+            # QuantizedKVCache, RotatingKVCache). Fall back to direct array
+            # slicing for plain array caches. Skip non-trimmable caches
+            # (ArraysCache for SSM/recurrent layers in hybrid models).
             for c in kv_cache:
-                if hasattr(c, "keys") and c.keys is not None:
+                if hasattr(c, "trim") and hasattr(c, "offset"):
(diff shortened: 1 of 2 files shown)


Blaizzy:
Closed in favour of #1103

PR #773, opened 2026-02-25
Tool calling in server

viktike:
Various models support tool calling, but the server doesn't implement it.
mlx_lm also supports tool calling.

This PR adds tool calling to the server's "chat/completions" endpoint with various tool parsers implemented and imported from mlx_lm.

This WiP initial version works _(I'm using osaurus.ai for testing)_, and any input, improvements or changes are welcome. I'm not happy with it yet, because this implementation leaves the model output in the assistant's turn, but I have no solution for it in the streaming case, because stream_generate returns multiple tokens at a time.

Fixes: #772

PR #964, opened 2026-04-07
Set finish_reason to "tool_calls" when the model emits tool calls

michaelstingl:
Returns `finish_reason: "tool_calls"` from the chat completions endpoint when the model emits tool calls, instead of always returning `"stop"`. Per the OpenAI Chat Completions spec, the `finish_reason` enum is "`tool_calls` if the model called a tool" ([openai-openapi `CreateChatCompletionResponse.choices[].finish_reason`](https://github.com/openai/openai-openapi/blob/498c71ddf6f1c45b983f972ccabca795da211a3e/openapi.yaml#L19676-L19696)).

Both `server.py:1182` (streaming, `ChatStreamChoice`) and `server.py:1264` (non-streaming, `ChatChoice`) hardcoded `"stop"`. The variable `tool_calls["calls"]` is already in scope at both spots, so the fix is a local conditional.

The bug is visible in [#773](https://github.com/Blaizzy/mlx-vlm/pull/773) (the PR that added tool calling) — the test response posted in the discussion shows `"finish_reason":"stop"` alongside a populated `tool_calls` array. The mapping was outside the scope of that PR.

Strict OpenAI-spec clients use `finish_reason == "tool_calls"` as the branch condition for entering the tool-execution loop and currently treat tool-emitting responses as plain stops.

## Tests

Two tests added to `mlx_vlm/tests/test_server.py` following the existing pattern (`FastAPI TestClient` + `patch.object` on `get_cached_model` / `apply_chat_template` / `generate`, plus `_infer_tool_parser` / `load_tool_module` / `process_tool_calls` for the tool path):

- `test_chat_completions_finish_reason_is_tool_calls_when_tools_emitted` — covers the non-streaming positive case. Verified to fail without the fix.
- `test_chat_completions_finish_reason_stays_stop_when_no_tool_calls` — regression guard for the unchanged path.

```
$ python -m pytest mlx_vlm/tests/test_server.py
======================== 11 passed, 2 warnings in 0.80s ========================
```

The streaming path (line 1182) is structurally identical to the non-streaming path for the `finish_reason` branch and not separately covered. Happy to add a streaming test if you'd like — it would need an SSE-parsing helper that doesn't exist yet in `test_server.py`. Let me know your preference.

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -1179,7 +1179,9 @@ async def chat_completions_endpoint(request: ChatRequest):
                     # Signal stream end
                     choices = [
                         ChatStreamChoice(
-                            finish_reason="stop",
+                            finish_reason=(
+                                "tool_calls" if tool_calls["calls"] else "stop"
+                            ),
                             delta=ChatMessage(
                                 role="assistant",
                                 content="",
@@ -1261,7 +1263,7 @@ async def chat_completions_endpoint(request: ChatRequest):
 
                 choices = [
                     ChatChoice(
-                        finish_reason="stop",
+                        finish_reason=("tool_calls" if tool_calls["calls"] else "stop"),
                         message=ChatMessage(
                             role="assistant",
                             content=tool_calls["remaining_text"],
--- a/mlx_vlm/tests/test_server.py
+++ b/mlx_vlm/tests/test_server.py
@@ -88,6 +88,113 @@ def test_responses_endpoint_forwards_new_sampling_args(client):
     assert mock_generate.call_args.kwargs["thinking_start_token"] == "<think>"
 
 
+def test_chat_completions_finish_reason_is_tool_calls_when_tools_emitted(client):
+    model = SimpleNamespace()
+    tokenizer = SimpleNamespace(chat_template="dummy")
+    processor = SimpleNamespace(tokenizer=tokenizer)
+    config = SimpleNamespace(model_type="gemma4")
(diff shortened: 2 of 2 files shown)


michaelstingl:
Verified end-to-end against a live mlx-vlm running this patch, driven by [OpenCode](https://github.com/anomalyco/opencode) (via [`@ai-sdk/openai-compatible`](https://github.com/vercel/ai/tree/main/packages/openai-compatible) 1.0.32) against `mlx-community/gemma-4-26b-a4b-it-4bit` on an M5 Air.

Captured SSE deltas show the expected branching on both code paths:

**Turn 1 — tool call emitted:**

```json
{
  "choices": [
    {
      "index": 0,
      "finish_reason": "tool_calls",
      "delta": {
        "role": "assistant",
        "content": "\n",
        "tool_calls": [
          {
            "type": "function",
            "index": 0,
            "id": "f918…",
            "function": {
              "name": "read",
              "arguments": "{\"filePath\": \"README.md\"}"
            }
          }
        ]
      }
    }
  ]
}
```

**Turn 2 — plain assistant reply after tool result:**

```json
{
  "choices": [
    {
      "index": 0,
      "finish_reason": "stop",
      "delta": {
        "role": "assistant",
        "content": "",
        "tool_calls": []
      }
    }
  ]
}
```

OpenAI Chat Completions spec compliance confirmed for both branches.

lucasnewman:
This has been resolved in v0.6.3 and later.

PR #967, opened 2026-04-07
Fix race condition in TurboQuant fused fast-quantize kernels

Chedrian07:
## Summary

The fast-path Metal kernels introduced in #909 pack each thread's quantization index into shared `packed_shared[w]` with a plain `|=`:

```cpp
packed_shared[w] |= idx_val << shift;
```

That is **not** atomic on Metal threadgroup memory. For `bits=4` and `Dim=128`, eight threads write to `packed_shared[0]`, eight to `packed_shared[1]`, and so on. Only one of the contributions per word survives the read-modify-write race, so the kernels silently emit indices like

```
[non-zero, 0, 0, 0, 0, 0, 0, 0,  non-zero, 0, 0, 0, 0, 0, 0, 0, ...]
```

i.e. one valid entry per 32-bit word and zeros everywhere else. The corruption is **worse at higher bit-widths** because each word holds more index slots, so a 4-bit cache is *more* corrupted than a 2-bit one.

The slow `_quantize_unit` path uses `_pack_lowbit_kernel` (thread-per-word, no race) and is unaffected, so the bug only fires on decode-time `T=1` calls — every generated token after prefill — making per-token decode quality drop sharply at 4-bit while leaving prefill round-trips clean. That's why all existing tests (which use `T>1`) pass even on the unfixed code.

## How I noticed it

Running KV-quant ablations (TurboQuant 2/3/3.5/4-bit) on Qwen3.5-0.8B for VLM benchmarks, 4-bit consistently produced **worse** results than 2-bit, the opposite of the `MSE ~ 1/4^b` prediction:

| condition | DocVQA ANLS | TextVQA acc | ChartQA acc | MMMU exact |
|---|---|---|---|---|
| baseline FP16 | 0.83 | 0.73 | 0.52 | 0.20 |
| tq 4-bit | **0.69** ⬇⬇ | 0.65 ⬇ | 0.43 ⬇ | 0.10 |
| tq 3.5-bit | 0.74 | 0.72 | 0.48 | 0.10 |
| tq 3-bit | 0.79 | 0.72 | 0.48 | 0.13 |
| tq 2-bit | **0.80** | 0.72 | 0.49 | 0.13 |

Inspecting decode-time predictions showed the model emitting the correct first tokens and then failing to stop — e.g. ChartQA `"Rep/Lean Rep"` (gold) became `"Rep/Lean Rep, the chart shows that 82% of Republicans..."` at 4-bit but stayed correct at 2-bit. That looked like KV cache corruption that grows with bit-width, which led me to compare `_TurboQuantMSECodec.quantize` between fast (`T=1`) and slow (`T>1`) paths on the same input.

## Root cause

`_fused_norot_quantize_kernel`, `_fused_mse_quantize_kernel`, and `_fused_kv_quantize_kernel` all share the same packing pattern:

```cpp
threadgroup uint packed_shared[PackedWidth];
if (d < PackedWidth) packed_shared[d] = 0;
threadgroup_barrier(mem_flags::mem_threadgroup);

if (d < Dim) {
    uint idx_val = shared_idx[d] & MASK;
    int bo = d * BITS;
    int w = bo >> 5;
    int shift = bo & 31;
    packed_shared[w] |= idx_val << shift;          // ← race
    if (shift + BITS > 32)
        packed_shared[w + 1] |= idx_val >> (32 - shift);  // ← race
}
```

Multiple threads write to the same `packed_shared[w]` concurrently with a non-atomic read-modify-write. The result is GPU-implementation-defined; in practice exactly one contribution per word is preserved.

## Fix

Switch `packed_shared` to `threadgroup atomic_uint` and use `atomic_store_explicit` / `atomic_fetch_or_explicit` / `atomic_load_explicit` (relaxed ordering — the existing `threadgroup_barrier`s already provide the visibility guarantees we need). Same fix applied to all three fused kernels. `_pack_lowbit_kernel` is untouched (its threads already write disjoint output slots).

Atomic ORs on threadgroup memory have a small per-op cost relative to a plain store, but `PackedWidth` is much smaller than `Dim` (e.g. 16 vs 128 for 4-bit) so the overhead lands well below the kernel's overall budget. I haven't measured a regression in per-token decode latency.

## Verification

Synthetic round-trip on Beta-distributed unit vectors (`dim=128`, the `use_rht` fast path):

| bits | slow MSE | fast MSE **before** fix | fast MSE **after** fix |
|---|---|---|---|
| 2 | 9.06e-04 | 2.40e-02 (26×) | **9.02e-04** |
| 3 | 2.65e-04 | 3.70e-02 (140×) | **2.61e-04** |
| 4 | 7.24e-05 | 5.58e-02 (770×) | **7.15e-05** |
| 5 | 2.12e-05 | 5.76e-02 (2716×) | **2.10e-05** |
| 6 | 7.66e-06 | 6.28e-02 (8194×) | **7.70e-06** |

After the fix, fast and slow paths produce **byte-identical** packed indices for every bit-width I tested (2/3/4/5/6, dim 64 and 128). MSE is monotone in bits as expected.

## Tests

Added a parametrised regression test `test_turboquant_mse_quantize_fast_path_matches_slow_path` that asserts:

1. fast-path (`T=1`) and slow-path (`T=N`) produce **identical** reconstructions for the same input vectors,
2. per-coordinate MSE stays well below the broken value (the unfixed kernel jumps to ~2-6e-2 for every bit-width).

The new test fails on `main` and passes after the fix. **All 15 existing turboquant tests still pass** — none of them exercised the `T=1` fast path, which is why this slipped through. The new test covers that gap (5 bits × 2 dims = 10 cases).

```
$ pytest mlx_vlm/tests/test_turboquant.py
............................. (25 passed in 2.5s)
```

## Related but out of scope

I also noticed that token-by-token `TurboQuantKVCache.update_and_fetch` produces NaN-laced state on `main` (independent of this fix — same NaN with the fix applied). That's a separate issue and I'll file it separately so this PR can stay focused on the race.

## Test plan

- [x] `pytest mlx_vlm/tests/test_turboquant.py` — 25 passed
- [x] New regression test fails on `main` (race) and passes here
- [x] Manual fast-vs-slow byte equality across `bits ∈ {2,3,4,5,6}` and `dim ∈ {64,128}`
- [x] End-to-end Qwen3.5-0.8B decode produces correct (non-runaway) outputs at 4-bit after the fix

### DIFF
--- a/mlx_vlm/turboquant.py
+++ b/mlx_vlm/turboquant.py
@@ -2748,27 +2748,31 @@ def _fused_kv_quantize_kernel(key_bits: int, val_bits: int):
         }}
         threadgroup_barrier(mem_flags::mem_threadgroup);
 
-        // Step 5: Pack indices
-        threadgroup uint packed_shared[Dim];  // oversized but safe
-        if (d < pw) packed_shared[d] = 0;
-        threadgroup_barrier(mem_flags::mem_threadgroup);
-
-        if (d < Dim) {{
-            uint idx_val = shared_idx[d] & idx_mask;
-            int bo = d * bits;
-            int w = bo >> 5;
-            int shift = bo & 31;
-            packed_shared[w] |= idx_val << shift;
-            if (shift + bits > 32)
-                packed_shared[w + 1] |= idx_val >> (32 - shift);
-        }}
-        threadgroup_barrier(mem_flags::mem_threadgroup);
-
+        // Step 5: Pack indices — thread-per-word, race-free, no atomics.
+        // Each thread d (d < pw) walks the dims whose
+        // [i*bits, (i+1)*bits) range intersects [32d, 32(d+1)) and
+        // accumulates them into a private register, then writes the
+        // word once. `bits` and `pw` are runtime-uniform here so the
+        // loop bound is well-defined per dispatch.
         if (d < pw) {{
+            uint w_val = 0u;
+            int word_start = (int)d * 32;
+            int i_min = word_start / bits;
+            int i_max = (word_start + 31) / bits;
+            if (i_max >= (int)Dim) i_max = (int)Dim - 1;
+            for (int i = i_min; i <= i_max; i++) {{
+                uint idx_val = shared_idx[i] & idx_mask;
+                int bit_off = i * bits - word_start;
+                if (bit_off >= 0) {{
+                    w_val |= idx_val << bit_off;
+                }} else {{
+                    w_val |= idx_val >> (-bit_off);
+                }}
+            }}
             if (is_val)
-                out_val_packed[bh * pw + d] = packed_shared[d];
+                out_val_packed[bh * pw + d] = w_val;
             else
-                out_key_packed[bh * pw + d] = packed_shared[d];
+                out_key_packed[bh * pw + d] = w_val;
         }}
     """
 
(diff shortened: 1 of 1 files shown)


Blaizzy:
I'm just finishing some tests around the your new thread per word apporach.

Blaizzy (approved):
LGTM, thanks!

PR #969, opened 2026-04-07
fix(gemma4): resolve NaN gradients in vision encoder training

ukint-vs:
## Summary

Gemma 4 vision fine-tuning produces NaN gradients due to three interacting issues in the forward pass. This PR fixes all three, unblocking Gemma 4 vision LoRA training on Apple Silicon.

**Root causes:**

1. **`-inf` attention mask → NaN softmax backward** (`vision.py`): `float("-inf")` causes `0 * -inf = NaN` in the gradient when all-padding rows hit softmax. Fixed with `-1e4` which is safe for both float16 (max ~65504) and float32.

2. **`.item()` breaks autograd and `mx.compile`** (`vision.py`): The pooling loop called `.item()` which detaches from the computation graph AND triggers `mx.eval()` inside `mx.compile` (illegal). Replaced with mask multiplication + `default_output_length` slice. No `.item()` calls remain.

3. **`@mx.compile` blocks gradient flow** (`base.py`): The decorator on `ensure_fused_sdpa` caches a forward-only trace that produces NaN during training. Removed.

## Test results

| Test | Result |
|------|--------|
| Unit tests (10 tests covering each patch) | All pass |
| Existing gemma4 model tests (4 tests) | All pass |
| Existing qwen3_vl tests (5 tests, also uses `ensure_fused_sdpa`) | All pass |
| Real training: Gemma 4 E2B-it 8bit, 5 steps, real images | **Loss: 15.99 → 7.89 → 4.67 → 1.98 → 0.67** |

## Compatibility

- Compatible with `mx.compile` on the SFT training step (no `.item()` or `mx.eval()` calls in the forward path)
- `-1e4` mask value is safe for both float16 and float32 (the original `-1e9` overflows to `-inf` in float16)

## Test plan

- [x] `python test_gemma4_nan_fixes.py` — 10 unit tests including float16 safety
- [x] `python test_gemma4_training_smoke.py` — real Gemma 4 model, real images, finite loss
- [x] `python -m unittest mlx_vlm.tests.test_models -k gemma4` — existing tests pass
- [x] `python -m unittest mlx_vlm.tests.test_models -k qwen3` — qwen3_vl tests pass

### DIFF
--- a/mlx_vlm/models/base.py
+++ b/mlx_vlm/models/base.py
@@ -383,7 +383,8 @@ def chunked_attention(
     return mx.concatenate(outputs, axis=2)  # (B, n_heads, L, head_dim)
 
 
-@mx.compile
+# Not compiled: @mx.compile caches a forward-only trace that produces
+# NaN gradients during vision encoder backward passes in training.
 def ensure_fused_sdpa(q, k, v, scale, mask=None):
     fused_dims = (64, 80, 128)  # supported by MLX's fused SDPA kernel
     d = q.shape[-1]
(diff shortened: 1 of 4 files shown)


Blaizzy:
Fixed in main

PR #970, opened 2026-04-07
fix(trainer): multi-turn completion masking with auto-detected turn tokens

ukint-vs:
## Summary

`--train-on-completions` previously only masked tokens before the first assistant marker. In multi-turn conversations it continued training on later user turns, and the default assistant token was Qwen-specific.

This PR moves completion masking into dataset/collator preparation so trainer and loss code consume a precomputed `completion_mask` instead of building masks inside the training step.

## Changes

- Adds `build_completion_mask()` in `mlx_vlm/trainer/datasets.py` to toggle assistant-token masking on assistant turns and off at an end-turn token or user token.
- Adds `resolve_completion_token_ids()` to auto-detect assistant, end-turn, and user token IDs from the processor chat template when possible.
- Extends `VisionDataset` and `PreferenceVisionDataset` to return `completion_mask`, `chosen_completion_mask`, and `rejected_completion_mask` when `--train-on-completions` is enabled.
- Updates SFT and ORPO collators to pad completion masks alongside `input_ids` and `attention_mask`.
- Keeps SFT and ORPO loss/logprob code free of mask construction; they now require the dataset batch to provide `completion_mask` when completion-only training is enabled.
- Adds CLI flags `--assistant-id`, `--end-turn-id`, and `--user-id` for explicit overrides when auto-detection is not enough.
- Adds regression coverage under `mlx_vlm/tests/test_completion_masking.py` for multi-turn masking, token-id resolution, dataset outputs, SFT/ORPO collation, and loss guards.

## Reviewer Note

This addresses the review request to keep masking in `mlx_vlm/trainer/datasets.py` and the collator/data path rather than constructing it in the trainer or loss function.

## Verification

- `python3 -m compileall mlx_vlm/lora.py mlx_vlm/trainer/datasets.py mlx_vlm/trainer/sft_trainer.py mlx_vlm/trainer/orpo_trainer.py mlx_vlm/tests/test_completion_masking.py`
- `git diff --check`
- `uv run python -m unittest mlx_vlm.tests.test_completion_masking mlx_vlm.tests.test_trainer mlx_vlm.tests.test_trainer_utils mlx_vlm.tests.test_prompt_utils`


Blaizzy:
Thanks @ukint-vs!

Looping in @Goekdeniz-Guelmez

Goekdeniz-Guelmez:
Hey @ukint-vs you can make it a lot more minimal and future safer, how about moving the masking into the dataset/collator creation part inside `mlx_vlm/trainer/datasets.py` and not in the trainer let alone loss function like the OG.

ukint-vs:
> Hey @ukint-vs you can make it a lot more minimal and future safer, how about moving the masking into the dataset/collator creation part inside `mlx_vlm/trainer/datasets.py` and not in the trainer let alone loss function like the OG.

hey. sure, I will take a look into this. Just moved forward from my mlx-vlm experiments already.

Goekdeniz-Guelmez:
@ukint-vs thanks!! you can for example move the helper to the `datasets.py` file and add it into the return structure.

```python
**({"completion_mask": completion_mask} if completion_mask is not None else {}),
```

ukint-vs:
@Goekdeniz-Guelmez updated.

lucasnewman:
Closing as obsolete.

PR #971, opened 2026-04-07
perf(trainer): add mx.compile to SFT step + wire up LR scheduling

ukint-vs:
## Summary

Two training optimizations that already exist in the codebase but aren't wired up:

1. **`mx.compile` on SFT training step** — One-line decorator addition. The ORPO trainer already has this (line 342), and mlx-lm uses the identical pattern with gradient accumulation (`prev_grad`/`do_update`). Reduces kernel dispatch overhead through graph fusion.

2. **Wire up learning rate scheduling** — `get_learning_rate()` in `utils.py:44` implements warmup + cosine decay, and `TrainingArgs` already has `warmup_steps` and `min_learning_rate` fields, but neither SFT nor ORPO trainer ever calls the function. Learning rate stays constant the entire run.

3. **Guard `warmup_steps >= iters`** — `get_learning_rate()` divides by `(iters - warmup_steps)`. With default `warmup_steps=100` and `--iters 100`, this is `0/0 = NaN`. Added fallback to linear ramp when warmup covers the entire run.

## Benchmark (Gemma 4 E2B-it 8bit, M4 Max, MLX 0.31.1)

| Config | Time/step | Peak Memory | Notes |
|--------|-----------|-------------|-------|
| Baseline | 0.259s | 7.78GB | |
| **+ mx.compile** | **0.137s** | 8.18GB | **1.9x faster** |
| + compile + LR | 0.103s | 8.18GB | Full stack |

Numerical correctness verified: 20 steps compiled vs uncompiled with same seed, loss values match within 1e-5.

## Changes

- `sft_trainer.py`: add `@partial(mx.compile, inputs=state, outputs=state)` to `step()`, add `get_learning_rate()` call before each step
- `orpo_trainer.py`: add `get_learning_rate()` call before each step (already has `mx.compile`)
- `utils.py`: guard division by zero when `warmup_steps >= iters`

Total: 3 files, +22/-2 lines.

## Test plan

- [x] Existing trainer tests pass (`python -m unittest mlx_vlm.tests.test_trainer`)
- [x] Correctness: compiled vs uncompiled loss matches within 1e-5 over 20 steps
- [x] Throughput: 1.9x speedup measured on Gemma 4 E2B
- [x] LR schedule: warmup ramps correctly, cosine decay reaches min_learning_rate
- [x] Edge case: `warmup_steps=100, iters=100` produces valid LR (no NaN)

### DIFF
--- a/mlx_vlm/trainer/orpo_trainer.py
+++ b/mlx_vlm/trainer/orpo_trainer.py
@@ -13,7 +13,7 @@ from mlx.utils import tree_map
 from tqdm import tqdm
 
 from .sft_trainer import TrainingArgs, _squeeze_leading_batch_dim
-from .utils import Colors, grad_checkpoint, save_adapter
+from .utils import Colors, get_learning_rate, grad_checkpoint, save_adapter
 
 
 @dataclass
@@ -421,6 +421,14 @@ def train_orpo(
                 )
 
             tic = time.perf_counter()
+
+        # Update learning rate schedule
+        optimizer.learning_rate = get_learning_rate(
+            iters=args.iters, step=it,
+            warmup_steps=args.warmup_steps,
+            learning_rate=args.learning_rate,
+            min_learning_rate=args.min_learning_rate,
+        )
 
         # Training step
         lvalue, toks = step(chosen_batch, rejected_batch)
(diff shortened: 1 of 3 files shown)


lucasnewman:
@ukint-vs Sorry for the delay. In general this looks reasonable. However, I don't understand the claim that changing the learning rate would affect per-step performance -- those should be completely orthogonal. Can you explain?

lucasnewman:
Closing as we have a wide range of trainer changes in progress.

PR #972, opened 2026-04-07
fix(generate): auto-apply chat template for vision/audio prompts

ukint-vs:
## Summary

`generate()` and `stream_generate()` return empty output for Gemma 4 vision models when called with raw text prompts and images via the Python API.

**Root cause**: The CLI path (`main()` in generate.py) calls `apply_chat_template()` before `stream_generate()`, but the Python API path does not. Without the template, the processor tokenizes raw text without inserting image tokens (`258880`), so the model never sees the image.

```python
# Before (empty output):
generate(model, processor, "What do you see?", image=["photo.jpg"])
# → GenerationResult(text="")

# After (works):
generate(model, processor, "What do you see?", image=["photo.jpg"])
# → GenerationResult(text="This image shows a traffic camera recording...")
```

## Fix

Detect raw prompts by checking if they contain any special tokens from `tokenizer.all_special_tokens`. If none found, auto-apply `apply_chat_template()`. Pre-templated prompts (from CLI, `chat_ui.py`, `server.py`) are detected and left unchanged.

## Test results

- Raw prompt + image → produces real text (was empty before)
- Pre-templated prompt + image → no double-templating, same output
- Tested on Gemma 4 E2B-it 8bit with real security camera frames

## Changes

1 file, +30 lines. `mlx_vlm/generate.py` only.

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -627,6 +627,36 @@ def stream_generate(
     image_token_index = getattr(model.config, "image_token_index", None)
     vision_cache = kwargs.pop("vision_cache", None)
 
+    # Apply chat template if the prompt is raw text (not already templated).
+    # The CLI path (main()) and callers like chat_ui.py/server.py apply
+    # apply_chat_template() before calling stream_generate(). The direct
+    # API path (generate()/stream_generate()) receives raw text from users.
+    # Without the template, models like Gemma 4 don't get image tokens in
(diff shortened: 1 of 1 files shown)


lucasnewman:
It's not appropriate to apply the chat template to all models. We're improved the docs for this specific case here: https://github.com/Blaizzy/mlx-vlm/pull/1300/changes

PR #973, opened 2026-04-07
Fix rotation-basis mismatch in TurboQuant L=1 fast quantize

nnorris7:
## Summary

`_fused_kv_quantize_kernel` (the L=1 fast path used by `TurboQuantKVCache.update_and_fetch` for per-token decode) rotates with the dense matrix `self.key_codec.rotation`, but `_TurboQuantMSECodec.quantize` (the L≥1 path used by prefill via `from_cache`) uses `mx.hadamard_transform` whenever `use_rht=True` (power-of-two head dims). The two are different orthogonal projections of the same input vector, so quantization indices end up in incompatible bases. The cache buffer becomes `[Hadamard prefix from prefill] + [dense suffix from decode]` and `decode_attention` reads it all assuming Hadamard.

**Affects every model with a power-of-two head dim (D=64, 128, 256) when used with `--kv-quant-scheme turboquant`** — i.e. most VLMs and LMs.

## Symptom

After #967 fixed the related race-condition in the same kernels, this rotation mismatch was the remaining cause of corrupted per-token decode state. Reproducer with D=128, bits=4 on the post-#967 main:

```python
import mlx.core as mx
from mlx_vlm.turboquant import TurboQuantKVCache

mx.random.seed(99)
B, H, D = 1, 8, 128
k17 = mx.random.normal((B, H, 17, D)).astype(mx.float16)
v17 = mx.random.normal((B, H, 17, D)).astype(mx.float16)

# Same 17 tokens, two routes through the public API:
a = TurboQuantKVCache(bits=4.0)
a.update_and_fetch(k17, v17)                                       # one L=17 call

b = TurboQuantKVCache(bits=4.0)
b.update_and_fetch(k17[:, :, :16, :], v17[:, :, :16, :])           # prefill L=16
b.update_and_fetch(k17[:, :, 16:17, :], v17[:, :, 16:17, :])       # decode L=1

print(a.state[0].indices[0,0,16,0].item())  # 847406903
print(b.state[0].indices[0,0,16,0].item())  # 2628291478  ← different basis
```

Same input, same codec, different indices because the L=1 path quantized in the dense rotation basis while the L=N path used Hadamard.

Decode-attention divergence on this case before the fix:

| dim | use_rht | rel err `‖a - b‖ / ‖a‖` | indices byte-eq |
|-----|---------|-------------------------|-----------------|
| 64  | True    | 0.3486                  | False           |
| 128 | True    | 0.4014                  | False           |
| 256 | True    | 0.2330                  | False           |
| 80  | False   | 0.0000                  | **True**        |
| 96  | False   | 0.0000                  | **True**        |

D=80 and D=96 are byte-identical because `use_rht=False` makes `_TurboQuantMSECodec.quantize` fall back to dense rotation, matching the kernel. The bug only fires on power-of-two dims.

## Why it slipped through

Same coverage gap as #967: every existing turboquant test uses `TurboQuantKVCache.from_cache(fp_cache)` to build the cache in one shot, then runs `decode_attention`. None of them exercise the prefill-then-per-token-update flow that's used in actual generation. So the L=1 fast path has never been compared against the L=N path on the same input.

It's also why the symptom is subtle in end-to-end use: the model still produces tokens, just with degraded attention scores on every decode-time position. You'd see it as quality drop, not as an outright crash. Likely contributed to the surprising "4-bit worse than 2-bit" pattern that #967's author observed on Qwen3.5-0.8B (which has D=128) — the race condition was the dominant effect, but the rotation mismatch was still adding noise after #967 landed.

## Fix

Gate `_try_fused_kv_quantize` on `not self.key_codec.use_rht and not self.value_codec.use_rht`. For power-of-two dims, fall through to `_TurboQuantMSECodec.quantize`'s own L=1 fast path, which already handles `use_rht` correctly via `mx.hadamard_transform` + `_fused_norot_quantize_kernel` (a no-rotation packing kernel that consumes pre-rotated input). For non-power-of-two dims the K+V fusion is unchanged.

12 lines, no kernel changes.

## Validation

After the fix, on top of `origin/main` (b2cffea, post-#967):

```
D= 64  rel err=0.000000  bytes_eq=True  PASS
D=128  rel err=0.000000  bytes_eq=True  PASS
D=256  rel err=0.000000  bytes_eq=True  PASS
D= 80  rel err=0.000000  bytes_eq=True  PASS  (non-pow2, unchanged)
D= 96  rel err=0.000000  bytes_eq=True  PASS  (non-pow2, unchanged)
```

L=N batch and L=N-1+1 split now produce **byte-identical** quantized state and **bit-identical** decode_attention output across every dim I tested.

All 28 existing tests still pass:
```
pytest mlx_vlm/tests/test_turboquant.py mlx_vlm/tests/test_kv_cache_quantization.py
============================ 28 passed in 0.66s ============================
```

## Out of scope (follow-ups)

- The proper long-term fix is to teach `_fused_kv_quantize_kernel` to apply Hadamard externally (the way `_TurboQuantMSECodec.quantize` already does for L=1) and use a no-rotation K+V packing kernel. That preserves the K+V fusion across both rotation modes. Probably 30–50 lines of kernel work and not blocking this fix.
- A regression test that exercises the prefill+decode mixing path on a `TurboQuantKVCache` (asserting L=N batch == L=N-1+1 split byte-equality) would have caught both this bug and #967. Worth adding alongside the fix that #967 already added for codec-level fast vs slow paths.
- I noticed this while validating an unrelated SWA TurboQuant cache that subclasses `TurboQuantKVCache`; that work is on a separate branch and will be a follow-up PR once this lands.

cc @Chedrian07 since this is in the same neighborhood as #967 and you mentioned a separate "NaN-laced state" finding in passing — wondering if this might have been part of what you saw.

## Test plan

- [x] `pytest mlx_vlm/tests/test_turboquant.py mlx_vlm/tests/test_kv_cache_quantization.py` — 28 passed
- [x] Reproducer above produces byte-identical state across D=64, 128, 256 after the fix
- [x] D=80, 96 (non-power-of-two) remain unchanged and identical
- [ ] End-to-end smoke test on a real Gemma 4 / Qwen run — would appreciate someone with a quantized model handy verifying perplexity / output quality before merge


Blaizzy (approved):
LGTM, thanks!

Issue #900, opened 2026-04-03
Streaming: raw tool call tokens leak into content chunks instead of tool_calls field


kht2199:
## Problem

In streaming mode (`stream: true`), raw tool call tokens (e.g. `<|tool_call>call:func{...}<tool_call|>`) are forwarded to the client as `content` in intermediate SSE chunks. Clients receive the raw tool call syntax as text instead of a structured `tool_calls` object.

This affects **all models** that use tool calling in streaming mode.

## Reproduction

```bash
curl http://localhost:8080/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "<any-model>",
    "messages": [{"role": "user", "content": "Search for latest AI news"}],
    "tools": [{"type": "function", "function": {"name": "web_search", ...}}],
    "stream": true
  }'
```

Intermediate SSE chunks contain:
```json
{"choices": [{"delta": {"content": "<|tool_call>call:web_search{query:...", "tool_calls": []}}]}
```

The final stop chunk correctly has `tool_calls` populated via `process_tool_calls`, but by then the client has already received the raw tokens as content.

## Root cause

The streaming path in `chat_completions_endpoint` yields every `chunk.text` directly as `content`. `process_tool_calls` only runs **after** the loop ends, so tool call tokens are never suppressed from intermediate chunks.

The non-streaming path works correctly because `process_tool_calls` runs on the complete `gen_result.text` before building the response.

## Fix

Once `tool_call_start` appears in the accumulated output, stop emitting content chunks — tool calls are fully delivered in the final stop chunk via `process_tool_calls`:

```python
tool_call_started = False
for chunk in token_iterator:
    output_text += chunk.text
    if tool_call_start_tok and tool_call_start_tok in output_text:
        tool_call_started = True
    visible = "" if tool_call_started else chunk.text
    if visible:
        yield SSE chunk with visible

# Final stop chunk — process_tool_calls runs here on full output_text
```

This ensures no raw tool call syntax is sent as content, while the structured `tool_calls` field in the stop chunk remains intact.

PR #974, opened 2026-04-07
Strip tool-call markup from streamed delta.content

michaelstingl:
## What

The streaming chat/completions loop forwarded `chunk.text` to the client verbatim, so Gemma 4's raw tool-call markup (`<|tool_call>call:NAME{ARGS}<tool_call|>`) landed in `delta.content` even though the final delta also carried a structured `tool_calls` array. Strict OpenAI Chat Completions clients treat `delta.content` as user-visible assistant output, so the raw parser syntax rendered on top of the tool execution frame in the client UI.

## Observed

Captured end-to-end against `mlx-community/gemma-4-26b-a4b-it-4bit` on an M5 Air, driven by [OpenCode](https://github.com/anomalyco/opencode) (Vercel AI SDK via `@ai-sdk/openai-compatible` 1.0.32):

```json
{
  "choices": [
    {
      "index": 0,
      "finish_reason": null,
      "delta": {
        "role": "assistant",
        "content": "<|tool_call>call:read{filePath:<|\"|>README.md<|\"|>}<tool_call|>"
      }
    }
  ]
}
```

The raw markup landed in `delta.content`. A separate, final delta carried the structured `tool_calls` array, so the call still executed — but the client TUI also rendered the parser syntax as assistant output.

The non-streaming path was already correct: `process_tool_calls` returns `remaining_text` with the tool-call blocks replaced and stripped, and `ChatChoice` uses that as the message content.

## Fix

Switch the streaming loop to progressive emission based on a new `_streamable_prefix` helper that:

1. Calls `process_tool_calls` on the accumulated `output_text` to drop any complete tool-call blocks.
2. Suppresses an unclosed block (start marker without its matching end) from the marker onwards.
3. Holds back a partial prefix of the start marker at the tail so a later chunk can complete it.

Each delta then contains only the newly-safe portion since the previous chunk, tracked via a monotonic `streamed_len` counter so that whitespace-stripping inside `remaining_text` cannot cause a double-emission.

This handles all three chunk-boundary cases — atomic block, split start/end across chunks, and partial start marker at the tail — without any per-parser state machine. Works for any tool parser with `tool_call_start` / `tool_call_end` attributes (both the existing `mlx_vlm.tool_parsers.gemma4` and the parsers loaded from `mlx_lm.tool_parsers.*` expose these).

## Tests

Four streaming tests added to `mlx_vlm/tests/test_server.py`, sharing a small `_parse_sse` helper and a `_run_stream_with_chunks` fixture:

| Test | Scenario |
|---|---|
| `test_chat_completions_stream_strips_tool_call_markup` | Atomic chunk (Gemma 4's real behaviour) |
| `test_chat_completions_stream_handles_markup_split_across_chunks` | Start marker in chunk 1, end marker in chunk 2 |
| `test_chat_completions_stream_holds_back_partial_start_marker` | Chunk 1 ends with `<|t`, chunk 2 completes the marker |
| `test_chat_completions_stream_passes_through_plain_content` | Regression guard: tool-capable chat template with plain-text output must flow through unchanged |

```
$ python -m pytest mlx_vlm/tests/test_server.py
13 passed in 1.09s
```

The three markup tests were verified to fail against the pre-fix streaming loop with the expected leak messages.

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -694,30 +694,70 @@ def _streamable_prefix(text, tool_module, tools):
     """Return the portion of ``text`` safe to include in a streamed
     ``delta.content``.
 
-    Complete ``tool_call_start`` / ``tool_call_end`` blocks are dropped
-    (the structured ``tool_calls`` array is still built from the full
-    accumulated output in the final delta). An unclosed tool call at
-    the tail — either a start marker without its matching end, or a
-    partial prefix of the start marker — is held back so a later chunk
-    can complete it. Called per chunk from the streaming loop.
+    Two kinds of markup can appear in a raw stream and must not reach
+    ``delta.content``:
+
+    - Tool-call blocks delimited by ``tool_module.tool_call_start`` /
+      ``tool_module.tool_call_end``. The authoritative ``tool_calls``
+      array is built from the full accumulated output in the final
+      delta, so the raw markup is pure noise for streaming clients.
+    - Thinking / reasoning blocks delimited by optional
+      ``tool_module.think_start`` / ``tool_module.think_end``
+      attributes. Gemma 4 emits these as internal reasoning and
+      Google's chat template removes them via a ``strip_thinking``
+      Jinja macro; mlx-lm's server routes them through a reasoning
+      state machine (ml-explore/mlx-lm#1114).
+
+    For each marker pair, three things can happen inside a chunk:
+
+    1. A complete block: drop it.
+    2. A start marker without its matching end (yet): suppress from
+       the start marker onwards.
+    3. A partial prefix of the start marker at the tail: hold it back
+       so a later chunk can complete it without leaking half of the
+       marker first.
+
+    Called per chunk from the streaming loop.
     """
     parsed = process_tool_calls(model_output=text, tool_module=tool_module, tools=tools)
     cleaned = parsed["remaining_text"] if parsed["calls"] else text
 
-    tc_start = tool_module.tool_call_start
+    # Strip complete thinking blocks if the parser declares them.
+    think_start = getattr(tool_module, "think_start", None)
+    think_end = getattr(tool_module, "think_end", None)
+    if think_start and think_end and think_start in cleaned and think_end in cleaned:
+        think_pattern = re.compile(
+            f"{re.escape(think_start)}.*?{re.escape(think_end)}", re.DOTALL
+        )
+        cleaned = re.sub(think_pattern, "", cleaned).strip()
 
-    # Unclosed tool call: suppress from the start marker onwards.
-    if tc_start in cleaned:
-        cleaned = cleaned[: cleaned.index(tc_start)]
+    # Collect the start markers we care about for the "unclosed" and
+    # "partial tail" cases. Tool-call start is always present; the
+    # thinking start marker is optional and only added when declared.
+    start_markers = [tool_module.tool_call_start]
+    if think_start and think_end:
+        start_markers.append(think_start)
 
-    # Partial prefix of the start marker at the tail (the marker is
-    # being generated across chunks): hold it back so the next chunk
-    # can complete it without leaking half of the marker first.
-    max_partial = min(len(cleaned), len(tc_start) - 1)
-    for n in range(max_partial, 0, -1):
-        if cleaned.endswith(tc_start[:n]):
-            cleaned = cleaned[:-n]
-            break
+    # Unclosed block: suppress from the EARLIEST start marker present.
+    earliest = min(
+        (cleaned.index(m) for m in start_markers if m in cleaned),
+        default=-1,
+    )
+    if earliest >= 0:
+        cleaned = cleaned[:earliest]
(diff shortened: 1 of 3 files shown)


michaelstingl:
While running this fix against a live server driven by OpenCode I hit a closely related leak in the same streaming path: Gemma 4 also emits `<|channel>…<channel|>` blocks as internal reasoning (token IDs 100 / 101 in the tokenizer — these are real Gemma 4 special tokens, not hallucinated). Google's chat template ships a `strip_thinking` Jinja macro that removes the content between these markers from rendered assistant output, which confirms they should not reach `delta.content` either.

mlx-lm addressed the same class of bug in [ml-explore/mlx-lm#1114](https://github.com/ml-explore/mlx-lm/pull/1114) (merged 2026-04-07) by adding multi-token thinking support via a new `_infer_thinking` helper and a `reasoning` state in the streaming state machine.

Marking this PR as draft while I extend `_streamable_prefix` to strip the `<|channel>`/`<channel|>` pair as well — it already owns "strip markup pairs from the accumulated output" and the second pair is a clean addition (same three-phase approach: complete block strip, unclosed suppression, partial prefix holdback). Will undraft once the expanded scope is covered by tests.

gofastercloud:
+1 on this — independent reproduction from a second production user.

Hit the same bug yesterday during a migration from Ollama to \`mlx_vlm.server\` on Gemma 4 26B-A4B (\`mlx-community/gemma-4-26b-a4b-it-4bit\`) with \`--kv-bits 3 --kv-quant-scheme turboquant\`. Driver is the OpenClaw agent gateway via its OpenAI-compatible streaming client. Observed exactly the symptom described in #900 and in your PR description: the raw \`<|tool_call>call:exec{command: <|"|>...<|"|>}<tool_call|>\` framing lands in accumulated \`delta.content\` before the terminal SSE chunk emits a structured \`tool_calls\` array, and the gateway's state machine commits to \"this was a plain-text reply\" before the parsed calls arrive. The agent then loops trying to re-issue the request until its turn timeout tripped.

We carried a much simpler local patch in the interim (flip a \`suppress_content\` flag once \`tool_module.tool_call_start\` appears in the accumulated \`output_text\`, and stop yielding content deltas from there) just to unblock the migration — it works for the common atomic-chunk case on our workload but doesn't handle the partial-start-marker-at-tail or split-marker-across-chunks edge cases that your \`_streamable_prefix\` approach does. Your version is strictly better; we'll retire our local patch once this lands.

Happy to test on our hardware if that's useful. For reference: M4 Pro 24 GB, Gemma 4 26B-A4B at ~21k-token prompts via the OpenClaw gateway, ~50–100 real agent turns/day. Lmk if you'd like logs from our repro.

abe238:
Confirming this fix works on Qwen3.6 (not just Gemma 4) — `mlx-community/Qwen3.6-35B-A3B-8bit` on an M5 Pro.

**On stock `mlx-vlm 0.4.4`** the streamed `delta.content` chunks contain the raw Qwen XML tool-call markup:

```
<tool_call>
<function=terminal>
<parameter=command>
ls /tmp
</parameter>
</function>
</tool_call>
```

…and the structured `tool_calls[]` only lands in the final chunk. Downstream agents (Hermes Agent, in our case) render the XML as assistant text and then *also* execute the tool — so the user sees both. That part is in the upstream bug, not introduced here.

**On this branch** (`michaelstingl/blaizzy-mlx-vlm@fix/stream-strip-tool-call-markup`, sha `ec7f918`):

```
CONTENT chunks (joined): ''
contains <tool_call> markup? False
TOOL_CALLS emitted: [
  {
    "type": "function",
    "id": "...",
    "function": { "name": "terminal", "arguments": "{\"command\": \"ls /tmp\"}" }
  }
]
```

Clean. The `_streamable_prefix` helper handles Qwen's `<tool_call><function=...><parameter=...>` XML format generically — no Qwen-specific changes needed. End-to-end test through Hermes Agent (which always streams to OpenAI-compatible endpoints) now shows structured tool calls instead of leaked XML.

Repro:

```bash
pip install --force-reinstall --no-deps \
  "git+https://github.com/michaelstingl/blaizzy-mlx-vlm@fix/stream-strip-tool-call-markup"
python -m mlx_vlm.server --model mlx-community/Qwen3.6-35B-A3B-8bit --port 8080
# then any streamed chat.completions request with tools=[...]
```

Happy to help finish off the Gemma thinking-marker piece if it would unblock landing this — it's a real pain point for any agent built on top of mlx-vlm right now.

Blaizzy:
Hey guys thanks for flagging, this is fixed in main :)

I encountered and addressed it prior to seeing this PR

michaelstingl:
For anyone landing here later: the thinking-marker strip that landed on `main` doesn't cover tool-call markers (`<|tool_call>` / `<tool_call|>` etc.) — those still leak into `delta.content` during streaming. I've opened #1035 with a minimal symmetric fix matching the existing thinking-filter pattern, plus regression tests.

abe238:
Amazing!

PR #826, opened 2026-03-15
Fix preprocessing for image input for trainer

Goekdeniz-Guelmez:
addresses #824

Issue #824, opened 2026-03-15
LoRA training broken for Qwen3.5 VLM: 3 bugs causing corrupted generation


kikoncuo:
## Summary

LoRA fine-tuning Qwen3.5 VLM models produces adapters that corrupt generation — the model outputs only vision tokens (`<|vision_start|><|image_pad|><|vision_end|>`) instead of text. We traced this to **3 independent bugs** that compound together.

## Reproducible Example

```bash
# Train for 30 iterations on any image+text dataset
python3 -m mlx_vlm.lora \
  --model-path mlx-community/Qwen3.5-0.8B-8bit \
  --dataset Enriqueag26/keysay-vlm-context-training \
  --iters 30 --batch-size 1 --learning-rate 1e-5 \
  --lora-rank 8 --lora-alpha 16 --max-seq-length 4096 \
  --grad-checkpoint --steps-per-report 10 \
  --output-path my-lora
```

After training, load the adapter and generate:

```python
from mlx_vlm import generate, load
from mlx_vlm.prompt_utils import apply_chat_template

model, processor = load("mlx-community/Qwen3.5-0.8B-8bit", adapter_path="my-lora")
prompt = apply_chat_template(processor, config=model.config, prompt="Describe this image.", num_images=1)
result = generate(model, processor, prompt=prompt, image="test.png", max_tokens=200)
print(result.text)
# Expected: "This image shows..."
# Actual:   "<|vision_start|><|image_pad|><|vision_end|>"
```

## Bug 1 (There are other problems I'm still trying to figure out): `VisionDataset` discards images for Qwen/Gemma/SmolVLM

**File:** `mlx_vlm/trainer/datasets.py`, line 109

```python
inputs = prepare_inputs(
    processor=self.processor,
    images=None if use_embedded_images else (images if images else None),  # Bug
    ...
)
```

For models with `use_embedded_images=True` (Qwen, Gemma, SmolVLM), `images=None` is passed to `prepare_inputs`. Training runs without any actual image data — the model trains on `<|image_pad|>` tokens with meaningless embeddings.

**Evidence:**
- `Trained Tokens` shows only ~23 tokens per sample (should be ~1000+ with image expansion)
- After fixing: tokens jumped to 1076/sample, loss dropped from 22.1 to 3.8 in 10 iters

## Environment

- mlx-vlm 0.4.0 (pip)
- Model: mlx-community/Qwen3.5-0.8B-8bit
[...]

PR #979, opened 2026-04-08
fix(trainer): pass images to prepare_inputs for Gemma, Qwen, and SmolVLM

ukint-vs:
## Summary

- `VisionDataset.process()` discards images for Gemma, Qwen, and SmolVLM models by passing `images=None` to `prepare_inputs()`
- The model trains on text-only with `<|image|>` placeholder tokens but **no actual image features**
- Fix: remove the `use_embedded_images` conditional, always pass images

## The bug

```python
# Before (datasets.py line 101-109):
use_embedded_images = (
    model_type.startswith("gemma")
    or model_type.startswith("qwen")
    or model_type == "smolvlm"
)
inputs = prepare_inputs(
    processor=self.processor,
    images=None if use_embedded_images else (images if images else None),  # <-- discards images
    ...
)
```

When `images=None`, `prepare_inputs` returns only `input_ids` + `attention_mask` with no `pixel_values`. The vision tower is never called during training.

## Impact

| Metric | Without images (bug) | With images (fix) |
|--------|---------------------|-------------------|
| Tokens/sample (Gemma4) | 37 | 304 (266 image + 38 text) |
| pixel_values | None | (1, 3, 672, 912) |
| Loss (20 iters) | 5.3 → 1.7 | 0.62 → 0.23 |

Same bug exists in `PreferenceVisionDataset` (ORPO trainer). Fixed in both.

## Verified on

- Gemma4 E2B (gemma4): pixel_values present, 266 image tokens, training loss drops correctly
- Gemma3 4B (gemma3): pixel_values present, 256 image tokens, dataset loads correctly
- All 8 existing unit tests pass

## Note on Qwen and SmolVLM

Qwen2.5-VL and SmolVLM processors fail after this fix because their prompt templates don't insert `<image>` placeholder tokens that the processors expect. This is a **pre-existing** bug in the message formatting pipeline (see #824, #826), not caused by this change. Previously these models "worked" by silently training without images. Now they properly fail, making the real issue visible.

Relates to #824 (Bug 1)

### DIFF
--- a/mlx_vlm/trainer/datasets.py
+++ b/mlx_vlm/trainer/datasets.py
@@ -98,15 +98,9 @@ class VisionDataset:
                 "Config must contain 'image_token_index' or 'image_token_id'"
             )
 
-        use_embedded_images = (
-            model_type.startswith("gemma")
-            or model_type.startswith("qwen")
-            or model_type == "smolvlm"
-        )
-
         inputs = prepare_inputs(
             processor=self.processor,
-            images=None if use_embedded_images else (images if images else None),
+            images=images if images else None,
             audio=audio if audio else None,
             prompts=prompts,
             image_token_index=image_token_index,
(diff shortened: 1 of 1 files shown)


Goekdeniz-Guelmez:
@ukint-vs don't need this pr anymore has been fixed already, you can go ahead and close this

ukint-vs:
fixed

PR #986, opened 2026-04-09
Fix LoRA scaling: divide alpha by rank (#845)

### LINKED ISSUE #845: LoRaLayer uses raw alpha instead of alpha/rank — 8x scaling error with default settings
## Bug

`LoRaLayer.__call__` in `mlx_vlm/trainer/lora.py` uses `self.alpha` as the scaling factor instead of `self.alpha / rank`. This makes the effective LoRA perturbation `rank` times larger than intended.

With the default `alpha=16, rank=8`:
- **mlx-vlm**: scale = 16 (raw alpha)
- **PEFT / original LoRA paper**: scale = 16/8 = 2

The LoRA contribution is **8x larger** than standard implementations.

## Proof

```python
import mlx.core as mx
import mlx.nn as nn
from mlx_vlm.trainer.lora import LoRaLayer

linear = nn.Linear(4, 4)
lora = LoRaLayer(linear, rank=8, alpha=16.0)

# Set known weights
lora.A = mx.ones((4, 8))
lora.B = mx.ones((8, 4))
x = mx.ones((1, 4))

base = linear(x)
actual = lora(x)
lora_contribution = (actual - base)[0, 0].item()

# mlx-vlm: alpha * (x @ A @ B) = 16 * 32 = 512
# PEFT:    (alpha/rank) * (x @ A @ B) = 2 * 32 = 64
print(f"mlx-vlm LoRA contribution: {lora_contribution}")  # 512.0
print(f"Expected (alpha/rank):     {2.0 * 32.0}")          # 64.0
print(f"Ratio: {lora_contribution / 64.0}x")               # 8.0x
```

## Impact

- Training with default settings applies 8x more perturbation than expected
- Users who set `alpha=16, rank=8` expecting PEFT-equivalent behavior get 8x amplified updates
- For sensitive architectures like Qwen3.5's GatedDeltaNet, this causes training instability
- The workaround `--lora-alpha 2` works but only because it sets the raw multiplier to what `alpha/rank` would compute

## Location

`mlx_vlm/trainer/lora.py`, line 38:
```python
return y + (self.alpha * lora_update).astype(x.dtype)
```

Should be:
```python
return y + (self.scale * lora_update).astype(x.dtype)
```

Where `self.scale = alpha / rank` is computed in `__init__`.

Same issue in `replace_lora_with_linear` at line 47:
```python
lora_update = layer.alpha * (layer.A @ layer.B)
```

## Reference

- [Original LoRA paper](https://arxiv.org/abs/2106.09685): "We then scale ΔWx by α/r"
- [PEFT implementation](https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/layer.py): `self.scaling = lora_alpha / r`
- [MLX-LM implementation](https://github.com/ml-explore/mlx-examples/blob/main/lora/lora.py): uses `alpha / rank`

H-A-Khan:
## What

Apply the standard LoRA scaling factor `alpha / rank` (Hu et al. 2021) in `LoRaLayer.__call__` and `replace_lora_with_linear`. Previously the layer multiplied the LoRA update by raw `alpha`, making the effective scaling rank-times too large for the documented defaults — for example `r=8`, `alpha=16` gave an effective scaling of 16 instead of the intended 2.

This matches every other PEFT implementation, including HuggingFace `peft` and the original Microsoft LoRA reference.

Closes #845.

## Changes

- `LoRaLayer.__init__` now stores `self.rank` and `self.scaling = alpha / rank`.
- `LoRaLayer.__call__` multiplies the update by `self.scaling`.
- `replace_lora_with_linear` uses `layer.scaling` so merged weights match what the trained adapter applies during inference.
- Two regression tests in `mlx_vlm/tests/test_trainer_utils.py`:
  - `test_lora_layer_uses_alpha_over_rank_scaling` — checks the stored attribute.
  - `test_lora_layer_forward_matches_alpha_over_rank` — checks the actual forward-pass output against the expected `base + (alpha/rank) * (x A B)` formula with a non-zero `B`.

## Backwards compatibility

⚠️ This is a **behavioural change**. Adapters trained against the previous (broken) scaling will behave 8× weaker after this fix when `r=8`, `alpha=16`. Users who want the old effective scaling can multiply their alpha by rank — e.g. set `--lora-alpha 128` to recover the old `r=8, alpha=16` behaviour. The recommended action is to retrain with the documented defaults now that they actually mean what the docs say.

If maintainers prefer to gate this behind a flag for one release I am happy to do that — let me know.

## Tests

```bash
python -m pytest mlx_vlm/tests/test_trainer_utils.py -v
```

```
test_find_all_linear_names PASSED
test_get_module_by_name PASSED
test_get_peft_model PASSED
test_lora_layer_forward_matches_alpha_over_rank PASSED
test_lora_layer_uses_alpha_over_rank_scaling PASSED
test_set_module_by_name PASSED
======================== 6 passed in 4.97s =========================
```

### DIFF
--- a/mlx_vlm/tests/test_trainer_utils.py
+++ b/mlx_vlm/tests/test_trainer_utils.py
@@ -1,8 +1,10 @@
 import unittest
 from unittest.mock import MagicMock, patch
 
+import mlx.core as mx
 import mlx.nn as nn
 
+from mlx_vlm.trainer.lora import LoRaLayer
 from mlx_vlm.trainer.utils import (
     find_all_linear_names,
     get_module_by_name,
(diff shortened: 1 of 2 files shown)


Blaizzy:
Thanks a lot @H-A-Khan, this has been fixed in main ✅

PR #989, opened 2026-04-09
feat: concurrency guard for Metal GPU serialization

eloe:
## Summary

MLX generation runs single-threaded on Metal. When multiple API requests hit the server concurrently, they corrupt shared GPU state and produce garbled output or crashes. This PR adds an `asyncio.Semaphore`-based concurrency guard that serializes access to the generation pipeline.

- Wraps both streaming and non-streaming paths in `/v1/chat/completions` and `/responses` with semaphore acquire/release
- Defaults to `MAX_CONCURRENT_REQUESTS=1` (safe for Metal); configurable via `--max-concurrent-requests` CLI flag or env var
- Uses `sem = None` + guard pattern to prevent semaphore over-release on early exit or client disconnect
- Integrates with `--request-timeout` to return HTTP 503 when the GPU is busy beyond the timeout window

### Tests

- `test_semaphore_exists_and_is_semaphore`
- `test_semaphore_default_value_is_one`
- `test_semaphore_respects_env_var`
- `test_semaphore_singleton`
- `test_concurrent_requests_both_succeed`

```
python -m pytest mlx_vlm/tests/test_server.py -k "concurrent" -v
```

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -126,11 +126,16 @@ model_cache = {}
 # Concurrency guard: MLX generation is single-threaded on Metal.
 # Concurrent requests would corrupt shared GPU state. The semaphore
 # serializes access to the generation pipeline.
+# NOTE: This guard assumes uvicorn runs with workers=1 (the default).
+# Multiple workers would each have their own semaphore, bypassing the guard.
 _generation_semaphore: Optional[asyncio.Semaphore] = None
 
 
 def get_max_concurrent_requests() -> int:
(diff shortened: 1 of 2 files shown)


eloe:
Closing this older general concurrency-guard PR for now. As of April 21, 2026, upstream has moved on with newer targeted concurrency work, including , so rebasing this broad semaphore-based change no longer looks like the right next step. If a general guard is still needed after the newer work settles, I’ll reopen that effort as a fresh, narrower PR against current .

PR #990, opened 2026-04-09
fix: return finish_reason=tool_calls when tool calls detected

eloe:
## Summary

Per the [OpenAI Chat Completions spec](https://platform.openai.com/docs/api-reference/chat/object), `finish_reason` should be `"tool_calls"` when the model emits tool calls, not `"stop"`. Strict OpenAI-compatible clients (including agent frameworks like LangChain, CrewAI, and others) use `finish_reason == "tool_calls"` as the branch condition for entering the tool-execution loop.

Currently both streaming and non-streaming paths hardcode `"stop"`. This PR fixes both:

- **Non-streaming** (`ChatChoice`): `finish_reason = "tool_calls" if tool_calls.get("calls") else "stop"`
- **Streaming** (`ChatStreamChoice`): Final chunk uses `"tool_calls"` finish_reason when tools detected

### Tests

4 tests covering all combinations:

- `test_chat_completions_finish_reason_stop_no_tools` — plain text, no tools defined
- `test_chat_completions_finish_reason_tool_calls` — non-streaming with tool calls
- `test_chat_completions_finish_reason_stop_tools_no_calls` — tools defined but model doesn't call any
- `test_chat_completions_streaming_finish_reason_tool_calls` — streaming with tool calls

```
python -m pytest mlx_vlm/tests/test_server.py -k "finish_reason" -v
```

### Related

This overlaps with #964 by @michaelstingl which addresses the non-streaming path. This PR additionally covers the streaming path and includes 4 tests (streaming + non-streaming, with and without tool calls).

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -1,5 +1,4 @@
 import argparse
-import asyncio
 import gc
 import json
 import os
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -1,4 +1,5 @@
 import argparse
+import asyncio
 import gc
 import json
 import os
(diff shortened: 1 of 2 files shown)


eloe:
Closing this because  already returns  for tool-call completions as of April 21, 2026, so this PR is no longer needed in its current form. We have aligned our fork  to match , and I’m cleaning up the obsolete PR set accordingly.

PR #992, opened 2026-04-09
feat: enforce tool_choice parameter in chat/completions

eloe:
## Summary

Adds `tool_choice` enforcement to `/v1/chat/completions` and `/responses`, per the [OpenAI API spec](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tool_choice).

### Supported values

| Value | Behavior |
|---|---|
| `"auto"` (default) | Model decides whether to call tools |
| `"none"` | Tools are stripped from the request; model cannot call tools |
| `"required"` | A system instruction is injected requiring the model to call at least one tool |
| `{"type": "function", "function": {"name": "X"}}` | Tools are filtered to only the named function; instruction injected to require it |

### Implementation

`resolve_tool_choice()` returns a filtered tool list and an optional system instruction. This approach works with any chat template without requiring model-specific tool_choice handling.

### Tests

- `test_tool_choice_none_strips_tools`
- `test_tool_choice_required_injects_instruction`
- `test_tool_choice_specific_function`
- `test_tool_choice_auto_passes_through`

```
python -m pytest mlx_vlm/tests/test_server.py -k "tool_choice" -v
```

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -637,6 +637,10 @@ class ChatStreamChunk(BaseModel):
     usage: Optional[UsageStats]
 
 
+class InvalidToolChoiceError(ValueError):
+    """Raised when tool_choice is invalid."""
+
+
 def resolve_tool_choice(
     tools: Optional[list],
     tool_choice: Optional[Any],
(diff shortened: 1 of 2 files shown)


eloe:
Superseding this older PR in favor of the current issue-first path. As of April 27, 2026, tool_choice behavior is tracked in #1045, and the active follow-up implementation is in #1070 against current main. Since this branch predates the recent server changes and the issue discussion refined the intended behavior, I do not think rebasing this older PR is the right review unit anymore. Closing this one in favor of issue #1045 and replacement PR #1070.

PR #994, opened 2026-04-09
feat: add logprobs support to /chat/completions

eloe:
## Summary

Adds `logprobs` support to `/v1/chat/completions`, per the [OpenAI API spec](https://platform.openai.com/docs/api-reference/chat/create#chat-create-logprobs). When `logprobs: true` is set, the response includes per-token log probabilities in the standard OpenAI format.

### Implementation

- **Streaming**: Each chunk includes `logprobs.content[]` with the chosen token's logprob
- **Non-streaming**: Uses `stream_generate()` internally to collect per-token logprobs, then assembles the full response. This is necessary because `generate()` does not expose per-token data.
- **Format**: `ChoiceLogprobs { content: [TokenLogprob { token, logprob, bytes }] }`
- `top_logprobs` is accepted on the schema but validated to reject unsupported values

### Example

```json
{
  "model": "...",
  "messages": [{"role": "user", "content": "Hi"}],
  "logprobs": true
}
```

Response includes:
```json
{
  "choices": [{
    "logprobs": {
      "content": [
        {"token": "Hello", "logprob": -0.5, "bytes": [72, 101, 108, 108, 111]}
      ]
    }
  }]
}
```

### Tests

- `test_chat_completions_logprobs_returned_when_requested`
- `test_chat_completions_logprobs_absent_by_default`
- `test_chat_completions_logprobs_format`
- `test_chat_completions_streaming_logprobs`

```
python -m pytest mlx_vlm/tests/test_server.py -k "logprobs" -v
```

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -602,11 +602,31 @@ class UsageStats(OpenAIUsage):
 
 class ChatRequest(GenerationRequest):
     messages: List[ChatMessage]
+    logprobs: Optional[bool] = Field(
+        None, description="Whether to return log probabilities."
+    )
+    top_logprobs: Optional[int] = Field(
+        None,
+        ge=0,
+        le=20,
+        description="Number of most likely tokens to return at each position.",
+    )
+
+
+class TokenLogprob(BaseModel):
+    token: str
+    logprob: float
+    bytes: Optional[List[int]] = None
+
+
+class ChoiceLogprobs(BaseModel):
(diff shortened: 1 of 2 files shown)


eloe:
Closing this because  already includes  logprobs support as of April 21, 2026, so this PR is obsolete in its current form. We have aligned our fork  to match , and I’m cleaning up the PR set to reflect only the still-missing features.

PR #995, opened 2026-04-09
feat: prompt prefix caching with TTL eviction and TurboQuant support

eloe:
## Summary

Adds cross-request KV cache reuse to the server, significantly reducing time-to-first-token for multi-turn conversations and agentic workflows where the system prompt is stable across turns.

### How it works

1. **PromptCacheState** tracks token IDs + KV cache between requests
2. On each request, `find_prefix_length()` computes the common prefix with the cached state
3. The KV cache is trimmed to the prefix length and reused, skipping re-computation of shared tokens
4. Cache entries are keyed per model (default) or per `prompt_cache_key` for multi-session routing

### Features

- **TurboQuant compatible**: Uses `cache.trim()` instead of raw array slicing, works with `QuantizedKVCache`, `RotatingKVCache`, and `ArraysCache` (hybrid models like Qwen3.5)
- **TTL eviction**: Background asyncio task evicts stale entries after `--prompt-cache-ttl` seconds (default 300s), freeing GPU memory
- **LRU cap**: Max 64 cache entries with least-recently-used eviction to prevent unbounded memory growth
- **Cache key routing**: Optional `prompt_cache_key` field enables per-session cache isolation (useful for multi-user gateways)
- **cached_tokens reporting**: `usage.input_tokens_details.cached_tokens` in responses
- **Graceful recovery**: Cache mismatches (shape errors, stale state) are caught and fall back to fresh generation

### Configuration

| Flag | Env var | Default | Description |
|---|---|---|---|
| `--prompt-cache-ttl` | `PROMPT_CACHE_TTL` | 300 | Seconds before idle cache eviction. 0 = no expiry |

### Tests

9 tests covering cache state lifecycle, persistence, isolation, prefix matching, and cleanup:

```
python -m pytest mlx_vlm/tests/ -k "PromptCache" -v
```

### Related

This builds on the same direction as #946 by @trevorgordon981 (basic PromptCacheState wiring) and #944 by @damonvjanis (TurboQuant trim fix). This PR includes both of those fixes plus TTL eviction, LRU capping, per-session cache key routing, cached_tokens reporting, graceful stale-cache recovery, and 9 tests.

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -671,6 +671,8 @@ def stream_generate(
 
     if prompt_cache_state is not None and prompt_cache_state.cache is not None:
         prefix_len = prompt_cache_state.find_prefix_length(full_input_ids_list)
+        cached_len = len(prompt_cache_state.token_ids) if prompt_cache_state.token_ids else 0
+        print(f"[prompt-cache] prefix_match={prefix_len}/{cached_len}, new_input={input_ids.shape[1]}, reuse={'yes' if prefix_len > 0 and prefix_len < input_ids.shape[1] else 'no'}")
         if prefix_len > 0 and prefix_len < input_ids.shape[1]:
             reused_prefix_len = prefix_len
             # Trim to only new tokens
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -671,7 +671,9 @@ def stream_generate(
 
     if prompt_cache_state is not None and prompt_cache_state.cache is not None:
         prefix_len = prompt_cache_state.find_prefix_length(full_input_ids_list)
-        cached_total = len(prompt_cache_state.token_ids) if prompt_cache_state.token_ids else 0
+        cached_total = (
+            len(prompt_cache_state.token_ids) if prompt_cache_state.token_ids else 0
+        )
         # Only reuse if a substantial prefix matches (>= 50% of cached tokens).
         # Short matches on quantized KV caches (TurboQuant) can produce
         # corrupted output because trim() only adjusts the offset without
@@ -796,7 +798,10 @@ def stream_generate(
         # polluting the cache with short probe/capability-check requests that
         # some agent frameworks send before the real request.
         _MIN_CACHE_TOKENS = 1024
-        if prompt_cache_state is not None and len(full_input_ids_list) >= _MIN_CACHE_TOKENS:
+        if (
+            prompt_cache_state is not None
+            and len(full_input_ids_list) >= _MIN_CACHE_TOKENS
+        ):
             all_ids = full_input_ids_list + [
                 t.item() if hasattr(t, "item") else t for t in generated_tokens
             ]
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -671,7 +671,13 @@ def stream_generate(
 
     if prompt_cache_state is not None and prompt_cache_state.cache is not None:
         prefix_len = prompt_cache_state.find_prefix_length(full_input_ids_list)
-        if prefix_len > 0 and prefix_len < input_ids.shape[1]:
+        cached_total = len(prompt_cache_state.token_ids) if prompt_cache_state.token_ids else 0
+        # Only reuse if a substantial prefix matches (>= 50% of cached tokens).
+        # Short matches on quantized KV caches (TurboQuant) can produce
+        # corrupted output because trim() only adjusts the offset without
+        # clearing stale quantized data.
+        min_reuse = max(512, cached_total // 2)
+        if prefix_len >= min_reuse and prefix_len < input_ids.shape[1]:
             reused_prefix_len = prefix_len
             # Trim to only new tokens
             input_ids = input_ids[:, prefix_len:]
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -3,6 +3,7 @@ import codecs
 import contextlib
 import functools
 import json
+import os
 import time
 from collections.abc import Sequence
 from dataclasses import dataclass
@@ -736,7 +737,8 @@ def stream_generate(
         except Exception as e:
             # Cache reuse failed (e.g., shape mismatch, stale KV state).
             # Invalidate and fall back to fresh generation.
-            print(f"[prompt_cache] Cache reuse failed, invalidating: {e}")
+            if os.environ.get("VERBOSE", "").lower() in ("1", "true", "yes"):
+                print(f"[prompt_cache] Cache reuse failed, invalidating: {e}")
             prompt_cache_state.invalidate()
             reused_prefix_len = 0
             input_ids = _original_input_ids
@@ -784,7 +786,8 @@ def stream_generate(
                             f"Cache offset mismatch: expected {reused_prefix_len}, got {c.offset}"
                         )
         except (ValueError, IndexError, AttributeError) as e:
-            print(f"[prompt_cache] Cache validation failed, rebuilding: {e}")
+            if os.environ.get("VERBOSE", "").lower() in ("1", "true", "yes"):
+                print(f"[prompt_cache] Cache validation failed, rebuilding: {e}")
             if prompt_cache_state is not None:
                 prompt_cache_state.invalidate()
             reused_prefix_len = 0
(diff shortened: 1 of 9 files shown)


gofastercloud:
+1 from production use — this PR is doing real work for us. A huge thank-you to @eloe for pushing it upstream, and one bug report attached below that I think is worth folding in before merge.

## Empirical result

We're running the `mlx_vlm.server` 0.4.4 + `--kv-bits 3 --kv-quant-scheme turboquant` configuration on an M4 Pro 24 GB Mac Mini, serving `mlx-community/gemma-4-26b-a4b-it-4bit` as the backend for a local AI agent stack with a ~21 k-token system prompt per turn. Before this PR, every agent turn re-prefilled the full system prompt at ~500 tok/s → ~42 s of prefill before any useful work. Five tool-call round-trips in a single task was 3.5 minutes of pure prefill, routinely blowing the agent's turn timeout.

With this PR applied (plus the TurboQuant shape-check fix described below), measured on two back-to-back identical agent turns through our gateway:

| run | prompt size | duration | prefill rate |
|---|---:|---:|---:|
| turn 1 (cold) | 25 906 tok | 144 708 ms (~2:24) | ~500 tok/s |
| turn 2 (warm) | 25 906 tok | **5 949 ms (~5.9 s)** | cache hit (no prefill) |

**~24× speedup on end-to-end agent turns.** For direct `/v1/chat/completions` (non-streaming) the numbers are even cleaner: 13.9 s cold → **0.36 s warm** on a 6 647-token prompt with the same `prompt_cache_key`. The prefill progress bars vanish from `mlx-vlm.log` on warm calls — the server skips straight to decode.

Without this PR, none of that is possible on our hardware. The agent-use case on a single-machine deployment becomes actually viable with prefix caching where before it was drowning in prefill time. This is exactly the kind of thing a server for local agent use needs, and I hope it lands.

## Bug report — TurboQuant shape-check incompatibility

`generate.py` around L780 (post-patch), inside the cache validation that guards against stale KV state:

```python
if reused_prefix_len > 0:
    try:
        for c in kwargs["prompt_cache"]:
            if hasattr(c, "keys") and c.keys is not None:
                expected_seq = reused_prefix_len
                actual_seq = c.keys.shape[2] if len(c.keys.shape) >= 3 else c.offset
                if actual_seq != expected_seq:
                    raise ValueError(
                        f"Cache shape mismatch: expected seq={expected_seq}, got {actual_seq}"
                    )
```

For **unquantized** KV caches this is correct: `c.keys` is `(heads, kv_heads, seq_len, head_dim)` and `shape[2]` equals the token count, so the comparison is meaningful.

For **TurboQuant**-quantized KV caches, however, `c.keys` has a grouped/chunked layout where `shape[2]` reflects the **group size** (typically 1024), **not** the actual token offset. The check therefore fails on every call for quantized caches — even when the cache is byte-identical to what was just stored — and falls into the `invalidate + rebuild` branch. The net effect is that prompt caching looks functionally correct on any prose test (the code paths all run) but delivers **zero measured speedup** on a TurboQuant-configured server, with logs like this on every call:

```
[prompt_cache] Cache validation failed, rebuilding: Cache shape mismatch: expected seq=6182, got 1024
[prompt_cache] Cache validation failed, rebuilding: Cache shape mismatch: expected seq=6200, got 1024
[prompt_cache] Cache validation failed, rebuilding: Cache shape mismatch: expected seq=6218, got 1024
```

The `got 1024` is the dead giveaway — that's the quant group size, not the offset.

## Suggested fix

Trust `c.offset` first when present (it's the authoritative counter for both unquantized and quantized caches under the existing MLX cache abstractions), and only fall back to shape-based comparison when `offset` is unavailable:

```python
if reused_prefix_len > 0:
    try:
        for c in kwargs["prompt_cache"]:
            if hasattr(c, "keys") and c.keys is not None:
                expected_seq = reused_prefix_len
                if hasattr(c, "offset") and c.offset is not None:
                    actual_seq = c.offset
                else:
                    actual_seq = c.keys.shape[2] if len(c.keys.shape) >= 3 else 0
                if actual_seq != expected_seq:
                    raise ValueError(
                        f"Cache shape mismatch: expected seq={expected_seq}, got {actual_seq}"
                    )
```

That's the ~6-line change we're carrying locally. We went from 0 × speedup (with the upstream check) to 24 × speedup (with `c.offset` first) on an otherwise-identical setup. The non-TurboQuant path is unchanged because `c.offset` is already set on those caches too and points to the same value that `shape[2]` would have.

Happy to open a follow-up PR against your branch with the one-line fix if that's useful, or you might prefer to fold it in here since it's a one-character conceptual change and the commit already advertises TurboQuant support. Whatever's easier for you to land.

## Context

We carry both this PR's full runtime diff and the TurboQuant shape-fix as local patches, vendored in-tree with a dedicated README and retirement plan. Happy to share the full `apply-mlx-patches.sh` harness if anyone else is running into the same TurboQuant issue — the patch file will be at `docs/upstream-prs/03-mlx-vlm-prompt-cache-turboquant/` once we push it public.

Again — huge thanks for this PR. This is going to unblock a lot of single-machine agent stacks.

Blaizzy:
Thanks @eloe for the awesome contribution!

I made some changes to the server in main that add support for Continous batching with TBQ as well, could you check if main solves the issue? if not, please resolved the conflicts and we will merge this!

Blaizzy:
@gofastercloud could you check main and see if the issues remain?

gofastercloud:
The specific \"Cache shape mismatch: expected seq=X, got 1024\" errors from the validation block are gone from main — thanks for that.

There's a related issue remaining in the trimming path in `generate.py` (lines 689–696). For `TurboQuantKVCache`, `c.keys.shape[2]` still returns the quantized capacity dimension rather than the token count, and the raw array slice `c.keys[:, :, :prefix_len, :]` is incorrect on the quantized storage layout. `TurboQuantKVCache` also doesn't expose a `trim()` method, so there's no clean way to truncate it in-place.

The common case (new turn where the full prefix matches cached history exactly) works fine because the trim branch doesn't fire. But when the cache holds generated tokens from a prior turn that aren't in the new prompt's prefix, it would silently corrupt rather than fall back.

Safe fix would be: use `c.offset` for the length comparison, and for types without `trim()`, invalidate instead of raw-slicing:

```python
for c in kv_cache:
    offset = getattr(c, "offset", None)
    cached_len = offset if offset is not None else (
        c.keys.shape[2] if hasattr(c, "keys") and c.keys is not None else 0
    )
    if cached_len > prefix_len:
        if hasattr(c, "trim"):
            c.trim(prefix_len)
        elif offset is None and hasattr(c, "keys") and c.keys is not None:
            # Unquantized cache without offset: raw slice is safe
            c.keys = c.keys[:, :, :prefix_len, :]
            c.values = c.values[:, :, :prefix_len, :]
        else:
            # Quantized cache (no trim() method): invalidate, rebuild on next turn
            c.offset = 0
            c.keys = None
            c.values = None
```

We don't have a TurboQuant-configured server to test against right now, but happy to provide a patch against #995 or main if useful.

Blaizzy:
Perfect! 

Can you also check #1030 and if needed target the changes to that branch which will be in main soon.

gofastercloud:
One correction to our suggested patch from yesterday: `trim(n)` removes `n` tokens from the end, so the call should be `c.trim(cached_len - prefix_len)`, not `c.trim(prefix_len)`. Here's the corrected version targeting main:

```python
for c in kv_cache:
    offset = getattr(c, "offset", None)
    cached_len = offset if offset is not None else (
        c.keys.shape[2] if hasattr(c, "keys") and c.keys is not None else 0
    )
    if cached_len > prefix_len:
        trim_amount = cached_len - prefix_len
        if hasattr(c, "trim"):
            c.trim(trim_amount)
        elif offset is None and hasattr(c, "keys") and c.keys is not None:
            # Unquantized cache without offset: raw slice is safe
            c.keys = c.keys[:, :, :prefix_len, :]
            c.values = c.values[:, :, :prefix_len, :]
        else:
            # Quantized cache (no trim() method): invalidate, rebuild on next turn
            c.offset = 0
            c.keys = None
            c.values = None
```

Also worth noting: the existing code's `c.keys.shape[2]` would actually crash on `TurboQuantKVCache` — `c.keys` is a `TurboQuantMSEState` NamedTuple which has no `.shape` attribute. It only doesn't crash in practice because for the common case (prefix matches cached history exactly) the trim branch never fires.

Happy to open a PR against `main` with this if that's easier.

Blaizzy:
Awesome! 

Could you open a GitHub issue first with reproducible examples.

Then we can discuss solutions and then move into a PR.

eloe:
I merged the latest `main` into this branch and rechecked after #1030 landed on April 19, 2026.

At this point, the runtime prompt-cache / TurboQuant handling appears to already be present on current `main`, and the remaining diff from this branch is effectively just additional TTL/scenario tests. So I don’t think this PR is still needed as a feature PR anymore.

Thanks again @gofastercloud for the detailed TurboQuant reports. If there’s still a reproducible issue on current `main`, I think the next best step is the separate issue requested above with a minimal repro. Otherwise, I’m happy to close this PR as superseded by `main`, or open a much smaller follow-up PR with just the extra tests if that’s useful.

PR #1009, opened 2026-04-10
fix: use OpenAI chat-completion field names in /chat/completions usage

chilang:
## Summary

`UsageStats` (used for `/v1/chat/completions` responses) inherits from `OpenAIUsage`, which models the **OpenAI Responses API** (`/v1/responses`) — that spec uses `input_tokens` / `output_tokens`. But `/v1/chat/completions` is a different spec: the `usage` object must contain `prompt_tokens`, `completion_tokens`, and `total_tokens`.

The net effect: any OpenAI-compatible client hitting mlx-vlm's `/v1/chat/completions` fails to read the `usage` payload because the field names don't match. I hit this reproducing Gemma 4 benchmarks with [llama-benchy](https://github.com/eugr/llama-benchy), which errors out during warmup with `Warmup failed: 'prompt_tokens'`.

## Fix

- Stop inheriting `UsageStats` from `OpenAIUsage`. Keep `OpenAIUsage` as-is for the Responses API (`/v1/responses`) where the spec is correct.
- Declare the chat-completion fields directly on `UsageStats`:
  - `prompt_tokens: int`
  - `completion_tokens: int`
  - `total_tokens: int`
  - (existing `prompt_tps` / `generation_tps` / `peak_memory` extras preserved)
- Update the two call sites in `chat_completions_endpoint` (streaming SSE chunk + non-streaming final response) to build `UsageStats` with the new field names.
- `/v1/responses` is untouched — it keeps using `OpenAIUsage` with `input_tokens` / `output_tokens` per OpenAI's Responses API spec.

## Test plan

- Added regression test `test_chat_completions_response_uses_openai_usage_field_names` that mocks `generate()` and asserts the JSON response body contains `usage.prompt_tokens`, `usage.completion_tokens`, `usage.total_tokens`, and does **not** contain the Responses-API field names.
- `python -m pytest mlx_vlm/tests/test_server.py` — 10 passed.
- Manually verified with `curl` against `/v1/chat/completions` on a locally running server — the `usage` object now matches the OpenAI Chat Completions spec.
- Verified with `llama-benchy` against `mlx-community/gemma-4-E4B-it-4bit` and `mlx-community/gemma-4-26b-a4b-it-4bit` — warmup now succeeds and benchmarks complete.
- `black --check` and `isort --profile=black --check` pass on changed files.

## References

- OpenAI Chat Completions: https://platform.openai.com/docs/api-reference/chat/object — `usage` object fields `prompt_tokens` / `completion_tokens` / `total_tokens`
- OpenAI Responses: https://platform.openai.com/docs/api-reference/responses/object — `usage` object fields `input_tokens` / `output_tokens`

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -586,11 +586,14 @@ class GenerationRequest(VLMRequest):
     )
 
 
-class UsageStats(OpenAIUsage):
+class UsageStats(BaseModel):
     """
-    Inherits from OpenAIUsage and adds additional fields for usage statistics.
+    OpenAI chat-completion usage with extra MLX stats.
     """
 
+    prompt_tokens: int
+    completion_tokens: int
+    total_tokens: int
     prompt_tps: float = Field(..., description="Tokens per second for the prompt.")
     generation_tps: float = Field(
         ..., description="Tokens per second for the generation."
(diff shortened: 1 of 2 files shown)


Blaizzy:
Please open a github issue

PR #1013, opened 2026-04-11
Resolve no images crash for qwen3_vl and qwen3_vl_moe generate call

urimem:
The code blindly slices visual_pos_masks during chunked prefill without checking if it's None.
The image parameter in the generate call is optional and with no images, visual_pos_masks is never set, but chunked prefill still passes n_to_process as a kwarg.

### DIFF
--- a/mlx_vlm/models/qwen3_vl/language.py
+++ b/mlx_vlm/models/qwen3_vl/language.py
@@ -503,7 +503,7 @@ class LanguageModel(nn.Module):
     ):
         # Slicing visual_pos_masks when prefilling
         n_to_process = kwargs.get("n_to_process", None)
-        if n_to_process is not None:
+        if n_to_process is not None and visual_pos_masks is not None:
             visual_pos_masks = visual_pos_masks[:, n_to_process:]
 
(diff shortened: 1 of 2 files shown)


conceptDawg:
Independent repro of this on a different model in the family — hit the exact crash with `mlx-community/Qwen3-VL-30B-A3B-Thinking-4bit` routed through `mlx_vlm.server` on any text-only prompt long enough to trigger chunk prefill:

```
TypeError: 'NoneType' object is not subscriptable
  File ".../qwen3_vl_moe/language.py", line 548, in __call__
    visual_pos_masks = visual_pos_masks[:, n_to_process:]
```

Applied this patch locally (both `qwen3_vl` and `qwen3_vl_moe`) against 0.4.4. Text-only chat completions now stream at ~80 tok/s on an M2 Ultra, and vision prompts still work correctly. The guard matches the pattern used elsewhere in the codebase for optional `mx.array` kwargs, so this looks like the right fix to me.

+1 from a downstream user — would be great to get this merged.

Blaizzy (approved):
LGTM, thanks!

PR #1014, opened 2026-04-12
server: indicate finish reason properly when model made a tool call.

viktike:
Finish reason should be "tool_calls" when the model called a tool, not "stop".


lucasnewman:
This has been resolved in v0.6.3 and later.

PR #1015, opened 2026-04-12
fix: propagate the verbose to the Prefill tqdm

PeterStaar-IBM:
1. generate_step signature (line 397): Added verbose: bool = False parameter before **kwargs.
  2. Prefill tqdm (line 533): Added disable=not verbose to suppress the bar when verbose=False.
  3. stream_generate (line 604): Added verbose = kwargs.pop(verbose, False) to extract verbose from kwargs before it reaches generate_step.
  4. generate_step call in stream_generate (line 733): Added verbose=verbose to forward it explicitly.
  5. generate → stream_generate call (line 858): Added verbose=verbose so the top-level generate() function's verbose flag also controls the Prefill bar.

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -394,6 +394,7 @@ def generate_step(
     sampler: Optional[Callable[[mx.array], mx.array]] = None,
     logits_processors: Optional[List[Callable[[mx.array, mx.array], mx.array]]] = None,
     prefill_step_size: Optional[int] = DEFAULT_PREFILL_STEP_SIZE,
+    verbose: bool = False,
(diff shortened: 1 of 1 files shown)


PeterStaar-IBM:
@Blaizzy Let me know if we can pull this in and release. We are a big fan of mlx-vlm in Docling (https://github.com/docling-project/docling), but we are currently seeing a lot of log pollution due to the Prefill tqdm.

lucasnewman:
@PeterStaar-IBM Thanks! I updated your PR to apply against main and we'll resolve it in https://github.com/Blaizzy/mlx-vlm/pull/1250

PR #1019, opened 2026-04-13
refactor: improve model loading and resource handling in utils.py

SyedaAnshrahGillani:
Key Changes:
   * **Fix Parameter Propagation:** Updated load_config and load_image_processor to correctly pass revision and force_download arguments to get_model_path. Previously, these were ignored when a string repository ID was provided, potentially leading to incorrect model versions being used.
   * **Robust Network Resource Management:** Improved load_image to use a context manager (with statement) for HTTP requests. This ensures that network connections are properly closed immediately after use, preventing potential connection leaks.
   * **Safer Image Handling:** Modified load_image to buffer URL response content into a BytesIO object before opening it with PIL. This is a safer pattern than passing the raw response stream directly to the image processor.
   * Enhanced Error Context: Preserved the original image source in load_image exception handling to provide clearer, more actionable error messages during failures.
   * **Documentation Cleanup:** Removed duplicate quantize_activations parameter descriptions in the docstrings for the load and load_model functions.
   * **Import Optimization:** Moved the safetensors import to the top-level to follow standard Python practices and avoid redundant local imports during model loading.

### DIFF
--- a/mlx_vlm/utils.py
+++ b/mlx_vlm/utils.py
@@ -12,6 +12,7 @@ import mlx.core as mx
 import mlx.nn as nn
 import numpy as np
 import requests
+import safetensors
 from huggingface_hub import snapshot_download
 from mlx.utils import tree_flatten, tree_map
 from PIL import Image, ImageOps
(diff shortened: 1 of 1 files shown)


Blaizzy (approved):
LGTM, thanks!

PR #1026, opened 2026-04-16
Security: Remote image fetch lacks timeout and payload size limits (DoS risk)

tomaioo:
## Summary

Security: Remote image fetch lacks timeout and payload size limits (DoS risk)

## Problem

**Severity**: `Medium` | **File**: `computer_use/utils.py:L19`

`requests.get(image_input)` is called without a timeout, and response content is loaded directly into memory before decoding with PIL. A malicious endpoint can hold connections open or return very large/decompression-bomb images, causing resource exhaustion.

## Solution

Use `requests.get(..., timeout=(connect_timeout, read_timeout), stream=True)`, enforce a maximum download size before reading body, and validate image dimensions/file size before decoding. Consider PIL safety settings (e.g., `Image.MAX_IMAGE_PIXELS`) to mitigate decompression bombs.

## Changes

- `computer_use/utils.py` (modified)

### DIFF
--- a/computer_use/utils.py
+++ b/computer_use/utils.py
@@ -6,6 +6,12 @@ import requests
 from PIL import Image, ImageDraw
 
 
+REQUEST_TIMEOUT = (5, 10)
+MAX_IMAGE_DOWNLOAD_SIZE = 10 * 1024 * 1024
+MAX_IMAGE_PIXELS = 25_000_000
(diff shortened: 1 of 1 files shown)


Blaizzy:
Hey @tomaioo 

Thanks for the contribution! 

Could we keep this PR focused on the timeout only?

PR #1030, opened 2026-04-17
Add KV cache quantization for continuous batching

Blaizzy:
## Summary
- Adds `BatchQuantizedKVCache` (uniform `mx.quantize`) and `BatchTurboQuantKVCache` with full `extend()`/`filter()` support for continuous batching
- Wires `--kv-bits`, `--kv-group-size`, `--kv-quant-scheme` through `BatchGenerator` → `ResponseGenerator` → server
- Skips quantizing the last layer (sensitive in deep models like gemma-4-31b)
- Adds attention dispatch in `base.py` for batch TurboQuant (dequantize + standard sdpa)

### Memory savings (B=4, H=32, S=512, D=128)
| Config | KV Memory | Reduction |
|--------|-----------|-----------|
| Unquantized | 67.1 MB | 1x |
| Uniform 8-bit | 18.9 MB | **3.6x** |
| Uniform 4-bit | 10.5 MB | **6.4x** |

### Short context (gemma-4-26b-a4b-it, ~80 prompt tokens)
| Config | Response | Gen tok/s | KV Cache | Peak Memory |
|--------|----------|-----------|----------|-------------|
| No quant | The capital of Japan is Tokyo. | 65.2 | 62.4 MB | 51.69 GB |
| Uniform 8-bit | The capital of Japan is Tokyo. | 63.8 | 60.4 MB | 51.69 GB |
| TurboQuant 3.5-bit | The capital of Japan is Tokyo. | 63.9 | 58.3 MB | 51.69 GB |

### Long context (gemma-4-26b-a4b-it, 20K prompt tokens)
| Config | Gen tok/s | KV Cache | KV Reduction | Peak Memory |
|--------|-----------|----------|--------------|-------------|
| No quant | 50.3 | 0.624 GB | 1x | 54.58 GB |
| Uniform 8-bit | 52.6 | 0.469 GB | **1.33x** | 54.69 GB |
| TurboQuant 3.5-bit | 25.6 | 0.365 GB | **1.71x** | 54.69 GB |

> Note: gemma-4 has 25 sliding-window layers (fixed-size `RotatingKVCache`, unquantized) and only 5 full-attention layers that get quantized. Models with all full-attention layers (e.g. Qwen, LLaMA) will see larger KV reductions. TurboQuant gen tok/s is lower due to dequantize-at-attention (custom Metal kernels not yet batch-aware).

## Test plan
[...]

PR #1032, opened 2026-04-17
Attempt to fix Qwen VL generation

pcuenca:
Chunked prefill for this family appears to have broken after CB was introduced.

Reproducer:

```bash
mlx_vlm.generate --model mlx-community/Qwen3-VL-30B-A3B-Instruct-bf16 --prompt "Describe this image" --image https://huggingface.co/datasets/huggingface/documentation-images/resolve/0052a70beed5bf71b92610a43a52df6d286cd5f3/diffusers/rabbit.jpg
```

This fails before this patch and works after.

`get_input_embeddings` already resets the cache positions, so Claude claims it's not necessary to do it again in `_process_prompts`.

I tested generation (example above) and CB against the server and both work.

**Note** it does not fail for all images, just when chunked prefill is engaged.

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -1106,13 +1106,6 @@ class BatchGenerator:
         left_padding = [max_length - l for l in lengths]
         inputs = _left_pad_prompts(inputs, max_length=max_length)
 
-        # Reset cached position state from previous prefills
-        # (e.g. Qwen3.5 caches _position_ids / _rope_deltas on the model)
-        if hasattr(self.model, "_position_ids"):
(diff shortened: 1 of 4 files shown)


Blaizzy:
Can you try running continous batching on the same model? https://github.com/Blaizzy/mlx-vlm?tab=readme-ov-file#continuous-batching

I have another PR ironing our these issues #1030

pcuenca:
> Can you try running continous batching

I did, mentioned above. Details: launched a server, then launched 5 background curl requests with a script.

pcuenca:
Addressed in #1030

PR #1035, opened 2026-04-18
Strip tool-call markup from streamed delta.content

michaelstingl:
## Summary

The streaming loop in `chat_completions_endpoint` already filters thinking markers (`<|channel>thought` / `<channel|>`, `<think>` / `</think>`) from `delta.content` via an inline state machine. Tool-call markers (`<|tool_call>` / `<tool_call|>` for Gemma 4, `[TOOL_CALLS]` for Mistral, etc.) follow the same contract but are not filtered today — tokens inside a tool-call block leak into `delta.content` before the final structured `tool_calls` chunk arrives, and downstream OpenAI-compatible clients render the raw markup.

This PR adds a symmetric `in_tool` branch to the existing inline state machine, driven by the active tool parser module's `tool_call_start` / `tool_call_end` strings. No new dependencies, no refactor of the surrounding code — just the missing second half of the existing pattern.

The leak mostly hurts agent frameworks (OpenCode, Hermes Agent, OpenClaw and similar) that stream to OpenAI-compatible endpoints and commit to "this is a text reply" as soon as non-empty `delta.content` arrives. See the two independent reproductions in #974 for concrete cases.

## What changes

- **`mlx_vlm/server.py`** (stream branch of `chat_completions_endpoint`)
  - New `in_tool` boolean alongside the existing `in_thinking`.
  - Reads `tc_start` / `tc_end` from `tool_module` once (both `None` when no parser is matched — the new branch is then inert).
  - Two new transitions: `normal → in_tool` when `tc_start in accumulated`, `in_tool → normal` when `tc_end in accumulated`.
  - While `in_tool`, tokens are still appended to `full_output` (so `process_tool_calls()` sees the full sequence at turn end) but are not streamed as `delta.content`.

- **`mlx_vlm/tests/test_server.py`**
  - New `TestStreamStateRouting` class with 4 tests over representative token sequences: tool-call markers stripped, thinking still works when a tool parser is active, Gemma 4 `<|channel>thought … <channel|><|tool_call> … <tool_call|>` combined pattern, and content passthrough without a parser.
  - The helper mirrors the inline loop's routing rules; a docstring note flags that the two must stay in sync.

## Scope I deliberately kept narrow

- No extraction of the state-update logic into a helper function — the diff stays surgical and symmetric to the existing thinking code.
- No partial-marker suppression. For Gemma 4 `<|tool_call>` is typically a single vocab token so the transition fires atomically. Multi-token markers that stream token-by-token could leak a few characters before the transition; that's the same trade-off the existing thinking filter makes and can be a follow-up.
- Mistral's empty `tool_call_end` means once `in_tool` is `True`, the turn stays in tool mode until EOS. That matches Mistral's convention that tool calls are terminal.

## Test plan

- [x] `python -m pytest mlx_vlm/tests/test_server.py` — 31 passed (27 existing + 4 new)
- [x] `pre-commit run --files mlx_vlm/server.py mlx_vlm/tests/test_server.py` — black + isort + autoflake clean
- [x] End-to-end verification against `mlx-community/gemma-4-26b-a4b-it-4bit` via mlx-lm's server + OpenCode (single tool call, multi-turn with `tool_response` feedback, parallel tool calls, nested arguments with `<|"|>` escape tokens) — no marker leakage, structured `tool_calls` emitted at the right boundary.

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -1607,10 +1607,26 @@ async def chat_completions_endpoint(request: ChatRequest):
 
                         output_tokens = 0
                         request_id = f"chatcmpl-{uuid.uuid4()}"
-                        # Track thinking state for reasoning/content split
+                        # Track thinking / tool-call state for content routing
                         in_thinking = False
+                        in_tool = False
                         accumulated = ""
                         full_output = ""  # raw output for tool call parsing
+
+                        # Marker strings for the active tool parser (if any).
+                        # Tokens emitted while in_tool is True are absorbed
+                        # into full_output for process_tool_calls() to parse,
+                        # and never streamed as delta.content.
+                        tc_start = (
+                            tool_module.tool_call_start
+                            if tool_module is not None
+                            else None
+                        )
+                        tc_end = (
+                            tool_module.tool_call_end
+                            if tool_module is not None
+                            else None
+                        )
 
                         def _next_token():
                             try:
@@ -1626,13 +1642,17 @@ async def chat_completions_endpoint(request: ChatRequest):
                             accumulated += token.text
                             full_output += token.text
 
-                            # Detect thinking boundaries
+                            # Detect thinking / tool-call boundaries
                             delta_reasoning = None
                             delta_content = None
 
-                            if not in_thinking and (
-                                "<|channel>thought" in accumulated
-                                or "<think>" in accumulated
+                            if (
+                                not in_thinking
+                                and not in_tool
+                                and (
+                                    "<|channel>thought" in accumulated
+                                    or "<think>" in accumulated
+                                )
                             ):
                                 in_thinking = True
                                 accumulated = ""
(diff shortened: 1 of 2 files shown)


michaelstingl:
For context on where this can go next: Gemma 4 already publishes its marker convention as structured fields in `tokenizer_config.json` — `stc_token` / `etc_token` for tool calls, `soc_token` / `eoc_token` for the thinking channel, etc. (see [Google's Gemma 4 prompt-formatting spec](https://ai.google.dev/gemma/docs/core/prompt-formatting-gemma4)). HuggingFace's `AutoTokenizer` surfaces these as attributes on the tokenizer object, so a streaming server can read them directly instead of relying on a per-model parser module matching the chat-template Jinja.

I've opened https://github.com/ml-explore/mlx-lm/pull/1163 on mlx-lm that reads `stc_token` / `etc_token` from the tokenizer and feeds them into mlx-lm's existing token-level state machine. If that convention takes hold, mlx-vlm could adopt the same auto-discovery and the tool-call filter here would work for any future Gemma-4-style model without a new entry in `_infer_tool_parser`.

This PR (#1035) stays surgical and matches the existing thinking-filter pattern — the structured-field angle is a separate follow-up.

Blaizzy:
Hey @michaelstingl 
Thanks for the awesome contribution!
I have added you as a co-author in #1037 which addresses the issue and couple other things.
So closing this one in favour of #1037

PR #1039, opened 2026-04-20
Fix Kimi VL concurrent Metal crash and mixed-batch text degradation

Blaizzy:
## Summary
- **Metal crash fix**: `mx.async_eval(inputs_embeds)` after vision encoding to prevent Metal command buffer conflicts between the default stream and the generation stream during concurrent requests
- **Mixed-batch text fix**: Route mask through `cache[0]` for BatchKVCache left-padding awareness in Kimi VL's language model

## Tested with
- `mlx-community/Kimi-VL-A3B-Thinking-4bit`
- Single ✅, Concurrent x3 ✅, Mixed (img+txt) ✅

## Test plan
- [x] Single image request returns coherent output
- [x] 3 concurrent same-image requests all return correct responses  
- [x] Mixed batch (1 image + 1 text) returns coherent results for both
- [x] CLI generate still works


lucasnewman (approved):
LGTM

PR #1051, opened 2026-04-22
fix: Use thread-local generation stream

### LINKED ISSUE #1049: Crash on mlx 0.31.2: 'There is no Stream(gpu, N) in current thread' when generate() runs in a worker thread
## Summary

`mlx_vlm.generate.stream_generate` crashes with `RuntimeError: There is no Stream(gpu, N) in current thread.` on **mlx 0.31.2** when invoked from any thread other than the one that imported the module. This breaks every async/threaded caller of `mlx_vlm` (e.g. MinerU, FastAPI servers wrapping `generate` in `asyncio.to_thread`, and any `BatchGenerator`-style worker).

This is the same root cause as ml-explore/mlx-lm#1179 / #1181: mlx 0.31.2 made streams **thread-local** (mlx PRs #3281, #3405). Streams created on one thread can no longer be used from another.

## Environment

- mlx: 0.31.2
- mlx-metal: 0.31.2
- mlx-vlm: 0.3.9
- Python: 3.10
- macOS: Darwin (Apple Silicon)

## Reproduction

Any caller that wraps `mlx_vlm.generate(...)` in `asyncio.to_thread` (or otherwise calls it from a non-import thread) crashes on the first generation. Concrete real-world repro via MinerU:

```bash
mineru -p some.pdf -o . -b vlm-mlx-engine
```

```
File ".../mlx_vlm/generate.py", line 425, in stream_generate
    with wired_limit(model, [generation_stream]):
File ".../contextlib.py", line 153, in __exit__
    self.gen.throw(typ, value, traceback)
File ".../mlx_vlm/generate.py", line 183, in wired_limit
    mx.synchronize(s)
RuntimeError: There is no Stream(gpu, 1) in current thread.
```

## Root cause

`mlx_vlm/generate.py` uses a stream that was created on the importing thread, but `wired_limit` / `stream_generate` are called from a worker thread. The fix mirrors what mlx-lm is doing in ml-explore/mlx-lm#1090: create the stream on the thread that uses it, instead of at import time.

## Workaround

Pin `mlx<0.31.2`:

```bash
pip install 'mlx<0.31.2' 'mlx-metal<0.31.2'
```

## Related

- ml-explore/mlx-lm#1179
- ml-explore/mlx-lm#1181
- ml-explore/mlx-lm#1090 (fix in mlx-lm)
- ml-explore/mlx#3281, ml-explore/mlx#3405 (the breaking change)

spicyneuron:
- Fixes https://github.com/Blaizzy/mlx-vlm/issues/1049

## Summary

Fixes the MLX 0.31.2+ worker-thread stream crash:

```
RuntimeError: There is no Stream(gpu, 0) in current thread.
```

Solution is to create the shared generation stream as a thread-local stream when the installed MLX version supports it. This mirrors the direction of ml-explore/mlx-lm#1090.

## Details

- Uses `mx.new_thread_local_stream` for `generation_stream` when available
- Falls back to `mx.new_stream` for compatibility with older MLX versions
- Keeps the change scoped to the existing generation stream abstraction

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -3,6 +3,7 @@ import codecs
 import contextlib
 import functools
 import json
+import threading
 import time
 from collections.abc import Sequence
 from dataclasses import dataclass
@@ -269,8 +270,26 @@ def normalize_resize_shape(
 
 
 # A stream on the default device just for generation.
-_new_generation_stream = getattr(mx, "new_thread_local_stream", mx.new_stream)
-generation_stream = _new_generation_stream(mx.default_device())
+_thread_local_generation_stream = (
+    mx.new_thread_local_stream(mx.default_device())
+    if hasattr(mx, "new_thread_local_stream")
+    else None
+)
+_generation_stream_local = threading.local()
+
+
+def get_generation_stream():
+    if _thread_local_generation_stream is not None:
+        return _thread_local_generation_stream
+
+    stream = getattr(_generation_stream_local, "stream", None)
+    if stream is None:
+        stream = mx.new_stream(mx.default_device())
+        _generation_stream_local.stream = stream
+    return stream
+
+
+generation_stream = get_generation_stream()
 
 
 def maybe_quantize_kv_cache(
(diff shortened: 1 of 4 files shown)


Blaizzy:
Thanks @spicyneuron!

However, closing in favour of existing and ongoing #1050

spicyneuron:
Yup, this was a bit more involved that it appeared at first pass. Will defer to you on this one.

PR #1065, opened 2026-04-24
fix(gemma4): handle JSON-formatted keys in tool call parser

afanty2021:
## Summary

Gemma 4 models may output standard JSON format for tool call arguments:
```
call:func_name{"key": "value"}  ← Standard JSON
```
Instead of the Gemma 4-specific format:
```
call:func_name{key<|"|>value<|"|>}  ← Gemma 4 format
```

The parser's `_parse_object` function was directly using the raw key text,
which included the JSON quotes, resulting in keys like `'"action"'` instead
of `'action'`.

## Problem

When the parser received JSON-formatted input:
1. `_parse_object` extracted the key as `"action"` (with quotes)
2. This created a dict: `{'"action"': 'add'}`
3. `json.dumps` escaped the existing quotes: `'{"\"action\"": "add"}'`
4. Downstream tools received malformed parameter names

## Solution

Modified `_parse_object` to detect and handle JSON-formatted keys:
```python
if key.startswith('"') and key.endswith('"'):
    try:
        key = json.loads(key)  # Properly unescape
    except (json.JSONDecodeError, ValueError):
        key = key[1:-1]  # Fallback: strip quotes
```

## Test Cases

Added comprehensive tests covering:
- ✅ Simple JSON-formatted keys
- ✅ Multiple JSON keys  
- ✅ Nested objects/arrays with JSON keys
- ✅ Unicode escape sequences in keys
- ✅ Mixed Gemma 4 and JSON formats

All 12 tests pass.

## Impact

- Fixes tool call parameter parsing for Gemma 4 models
- Maintains backward compatibility with existing Gemma 4 format
- No changes needed to model code or chat templates

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


Blaizzy:
Hey @afanty2021, thanks!

Please open a issue first

PR #1069, opened 2026-04-25
Support OpenAI stop sequences in server

### LINKED ISSUE #1044: OpenAI-compatible server accepts `stop` but silently ignores it on `/chat/completions` and `/responses`
## Summary

On current `main`, the OpenAI-compatible server accepts a `stop` parameter on both `/chat/completions` and `/responses`, returns `200`, and then silently ignores it.

This happens because the request models allow extra fields, but `stop` is not declared as a first-class field on the public request models and is never translated into generation-time stopping behavior.

## Why this matters

OpenAI-compatible clients reasonably expect `stop` to bound generation. When the server accepts the field and ignores it, clients cannot rely on stop sequences to prevent control-token leakage or over-generation.

## Minimal repro

This repro uses FastAPI `TestClient` with a mocked model so the behavior is independent of any particular model:

```python
from unittest.mock import patch
from fastapi.testclient import TestClient
import mlx_vlm.server as server

client = TestClient(server.app)
model = SimpleNamespace()
processor = SimpleNamespace(tokenizer=SimpleNamespace(chat_template=""))
config = SimpleNamespace(model_type="qwen2_vl")
result = SimpleNamespace(
    text="ok",
    prompt_tokens=1,
    generation_tokens=1,
    total_tokens=2,
    prompt_tps=1.0,
    generation_tps=1.0,
    peak_memory=0.1,
)

with (
    patch.object(server, "get_cached_model", return_value=(model, processor, config)),
    patch.object(server, "apply_chat_template", return_value="prompt"),
    patch.object(server, "generate", return_value=result) as mock_generate,
):
    for name, path, payload in [
        (
            "chat",
            "/chat/completions",
            {
                "model": "demo",
                "messages": [{"role": "user", "content": "hi"}],
                "stop": ["STOP"],
            },
        ),
        (
            "responses",
            "/responses",
            {
                "model": "demo",
                "input": "hi",
                "stop": ["STOP"],
            },
        ),
    ]:
        resp = client.post(path, json=payload)
        print(name, resp.status_code)
        print(sorted(mock_generate.call_args.kwargs.keys()))
        print(mock_generate.call_args.kwargs)
```

## Actual result

Both requests return `200`, but `generate()` receives no `eos_tokens` or any other `stop`-derived stopping configuration.

In my local repro on April 21, 2026, the relevant kwargs reaching `generate()` were:

```python
{
    "max_tokens": 2048,
    "top_p": 1.0,
    "top_k": 0,
    "min_p": 0.0,
    "enable_thinking": True,
}
```

No `eos_tokens`, no `stop`, and no server-side translation of the provided stop strings.

## Expected result

At minimum:

- `stop` should be a first-class field on the public OpenAI-compatible request models for both endpoints.
- The server should resolve up to 4 stop sequences and pass the corresponding stopping behavior into generation.
- If a given endpoint/model combination cannot support this safely, the server should reject the field explicitly instead of silently accepting and ignoring it.

## Notes

I previously had an open PR in this area, but per maintainer guidance on #1043 I’m opening an issue first with a concrete repro instead of leading with a PR.

eloe:
## Summary

Fixes Blaizzy/mlx-vlm#1044.

This adds OpenAI-compatible `stop` handling to the MLX-VLM server for `/chat/completions`, plus matching mlx-vlm compatibility support for `/responses` so both server endpoints honor the same caller intent.

The implementation normalizes `stop` as either a string or a list of one to four non-empty strings, then incrementally filters decoded text so stop sequences are trimmed even when they span token/chunk boundaries. It is wired through continuous batching, speculative decoding, and non-batching fallback paths. When a server-side stop sequence is matched, generation is cancelled/removed from the active batch and the response reports `finish_reason: "stop"`. When a requested stop sequence is not emitted before generation exhausts, chat completions report `finish_reason: "length"`.

I also fixed a related Responses streaming consistency issue discovered while testing: `response.completed.response.output_text` now matches the final trimmed `response.output_text.done` text.

## OpenAI spec validation

Validated against the current official OpenAI documented OpenAPI spec at `https://app.stainless.com/api/spec/documented/openai/openapi.documented.yml`.

- Chat Completions defines `stop` via `StopConfiguration`: nullable string or array with `minItems: 1` and `maxItems: 4`; returned text must not contain the stop sequence.
- Chat completion `finish_reason` includes `stop` for a natural stop or provided stop sequence, and `length` when the token limit is reached.
- Streaming chat examples finish with a final chunk whose `delta` is empty and whose `finish_reason` is `stop`.
- The current OpenAI `CreateResponse` schema does not define a top-level `stop` parameter, so `/responses` handling here is intentionally treated as mlx-vlm compatibility behavior rather than a claim that OpenAI Responses currently accepts top-level `stop`.

## Validation

Automated tests:

```bash
uv run --with pytest python -m pytest mlx_vlm/tests/test_server.py
# 59 passed, 3 warnings
```

Additional checks:

```bash
python3 -m compileall -q mlx_vlm/server.py mlx_vlm/tests/test_server.py
git diff --check
```

Live worktree integration validation against `mlx-community/Qwen3.6-35B-A3B-nvfp4` on the local MLX-VLM server:

- Chat non-streaming string stop: passed
- Chat streaming string stop: passed
- Responses non-streaming string stop: passed
- Responses streaming string stop: passed
- Chat newline stop: passed
- Responses newline stop: passed
- No matching stop reaches `length`: passed
- List stop chooses earliest emitted stop: passed
- Rejects more than four stop strings: passed
- Rejects empty stop string: passed

Integration sweep result: `10 passed, 0 failed`.

Broader repo note: `pytest -q` still fails during collection on an unrelated existing mismatch in `mlx_vlm/tests/test_utils.py`, which imports `get_class_predicate` from `mlx_vlm.utils` although that symbol is not present.

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -121,6 +121,7 @@ class GenerationArguments:
     enable_thinking: bool = True
     thinking_budget: Optional[int] = None
     thinking_start_token: Optional[str] = None
+    stop: Optional[List[str]] = None
 
     def to_generate_kwargs(self) -> dict:
         """Convert to kwargs dict for generate()/stream_generate()."""
@@ -170,6 +171,56 @@ class StreamingToken:
     finish_reason: Optional[str]
     peak_memory: float = 0.0
     top_logprobs: Optional[List[Tuple[int, float]]] = None
+
+
+class StopMatcher:
+    """Incrementally trims OpenAI stop strings from decoded token text."""
+
+    def __init__(self, stop: Optional[List[str]] = None):
+        self.stop = stop or []
+        self.buffer = ""
+        self.max_stop_len = max((len(s) for s in self.stop), default=0)
+        self.stopped = False
+
+    def append(self, text: str) -> Tuple[str, bool]:
+        if self.stopped or not text:
+            return "", self.stopped
+        if not self.stop:
+            return text, False
+
+        self.buffer += text
+        stop_at = None
+        for stop in self.stop:
+            idx = self.buffer.find(stop)
+            if idx != -1 and (stop_at is None or idx < stop_at):
+                stop_at = idx
+
+        if stop_at is not None:
+            out = self.buffer[:stop_at]
+            self.buffer = ""
+            self.stopped = True
+            return out, True
+
+        keep = max(self.max_stop_len - 1, 0)
+        if keep == 0:
+            out = self.buffer
+            self.buffer = ""
+            return out, False
+        if len(self.buffer) <= keep:
+            return "", False
+        out = self.buffer[:-keep]
+        self.buffer = self.buffer[-keep:]
+        return out, False
+
+    def feed(self, text: str) -> Tuple[str, bool]:
+        return self.append(text)
+
+    def flush(self) -> str:
+        if self.stopped:
+            return ""
+        out = self.buffer
+        self.buffer = ""
+        return out
 
 
 class ResponseGenerator:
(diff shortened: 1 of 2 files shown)


Blaizzy:
Thanks, lets think about this further, I believe there is a cleaner solution to this.

Issue #1045, opened 2026-04-21
OpenAI-compatible server accepts `tool_choice` but does not enforce it on `/chat/completions` or `/responses`


eloe:
## Summary

On current `main`, the OpenAI-compatible server accepts `tool_choice`, returns `200`, and then does not enforce the requested tool-calling policy on either `/chat/completions` or `/responses`.

The `tools` payload may survive far enough to reach template rendering in some paths, but `tool_choice` itself is not modeled or enforced, so values like `"required"` behave the same as omitting the field.

## Why this matters

OpenAI-compatible clients use `tool_choice` to control whether tool calls are optional, forbidden, or required. Silently accepting and ignoring the field makes function-calling integrations unreliable and hard to debug.

## Minimal repro

This uses a mocked model to avoid model-specific tool behavior and checks what the server actually does with the request:

```python
from types import SimpleNamespace
from unittest.mock import patch
from fastapi.testclient import TestClient
import mlx_vlm.server as server

client = TestClient(server.app)
model = SimpleNamespace()
processor = SimpleNamespace(tokenizer=SimpleNamespace(chat_template=""))
config = SimpleNamespace(model_type="qwen2_vl")
result = SimpleNamespace(
    text="ok",
    prompt_tokens=1,
    generation_tokens=1,
    total_tokens=2,
    prompt_tps=1.0,
    generation_tps=1.0,
    peak_memory=0.1,
)

tools = [
    {
        "type": "function",
        "function": {
            "name": "ping",
            "parameters": {"type": "object", "properties": {}},
        },
    }
]

with (
    patch.object(server, "get_cached_model", return_value=(model, processor, config)),
    patch.object(server, "apply_chat_template", return_value="prompt") as mock_template,
    patch.object(server, "generate", return_value=result) as mock_generate,
):
    for name, path, payload in [
        (
            "chat",
            "/chat/completions",
            {
                "model": "demo",
                "messages": [{"role": "user", "content": "hi"}],
                "tools": tools,
                "tool_choice": "required",
            },
        ),
        (
            "responses",
            "/responses",
            {
                "model": "demo",
                "input": "hi",
                "tools": tools,
                "tool_choice": "required",
            },
        ),
    ]:
        resp = client.post(path, json=payload)
        print(name, resp.status_code)
        print(resp.json())
        print(sorted(mock_generate.call_args.kwargs.keys()))
        print(mock_template.call_args.kwargs)
```

## Actual result

[...]

PR #1070, opened 2026-04-25
Handle OpenAI tool_choice requests

eloe:
## Summary

Addresses #1045 by adding explicit OpenAI-compatible `tool_choice` handling to the `/chat/completions` and `/responses` endpoints.

This follows the approach discussed in the issue thread:

- `tool_choice` omitted: preserve existing behavior and pass tools through when provided.
- `tool_choice: "none"`: omit tools from chat-template rendering and suppress tool-call parsing.
- `tool_choice: "auto"`: render tools normally and parse model-emitted tool calls.
- `tool_choice: "required"` and named function choices: return a clear 400 because current VLM chat templates do not provide a reliable way to force required or pinned tool use.

## Why

The larger tool-capable VLMs we checked, including Qwen3.5/Qwen3.6 and Gemma-4/Hermes-Gemma-4 variants, support tool schemas in their chat templates, but the templates do not consume `tool_choice` to change prompt rendering. Accepting `none` and `auto` improves compatibility with OpenAI-style clients while avoiding a false guarantee for unsupported enforcement modes.

The Responses API path also now emits parsed function-call output items when the loaded template has a supported parser, and streaming `response.completed` events preserve the final `output_text` aggregate.

## Validation

- `uv run --with pytest python -m pytest mlx_vlm/tests/test_server.py`
  - `51 passed, 3 warnings`
- `uv run --with pytest python -m pytest mlx_vlm/tests/test_server.py mlx_vlm/tests/test_gemma4_tool_parser.py`
  - `58 passed, 3 warnings`
- `python3 -m compileall -q mlx_vlm/server.py mlx_vlm/tests/test_server.py`
- `git diff --check`

Warnings were the existing MLX/Pydantic serializer warnings observed in the local test environment.


lucasnewman:
This is now supported in main.

PR #1071, opened 2026-04-25
Fix stale test_utils.py regressions + extract get_class_predicate

mdstaff:
## Summary

Fixes four stale regressions in `test_utils.py` that left the suite uncollectable. Because the first regression is an `ImportError` on `test_utils.py` itself, it cascades through the package import chain and blocks collection of 8 other test files. Also extracts `get_class_predicate` back to module level so it remains testable.

| | Before | After |
|--|--|--|
| `pytest mlx_vlm/tests/ --ignore=mlx_vlm/tests/test_smoke.py` | 8 collection errors, 0 tests run | 464 passed, 1 skipped |

> **CI note:** The pipeline excludes `test_utils.py` from its pytest invocation, so the green check does not validate the fixes in this PR. See the test plan below for local results.

### mlx_vlm/utils.py

- Extract `get_class_predicate` as a module-level factory `(skip_vision, weights, quantization_config) -> predicate`. The function was inlined as a closure in #821, breaking its existing unit test. Behavior at the call site in `load_model()` is identical; closure-captured variables become explicit parameters.
- Restore image/prompt mismatch validation lost when the text-only fast path was added in #401. When `images=None`, that path calls `processor.tokenizer(...)` directly rather than routing through `processor(...)`, silently bypassing per-model `<image>` token validation. Fixed by checking `getattr(processor, "image_token", None)` at the top of the fast path — covers ~90% of processors; models that use token IDs or multi-token constants handle their own validation downstream.

### mlx_vlm/tests/test_utils.py

1. **`MockProcessor.tokenizer` not callable** (#821 + #401): replace the non-callable `DummyTokenizer` with a `MagicMock`; add optional `tokenizer_return_value` parameter so only tests exercising the text-only fast path need to configure it; add `image_token = "<image>"` attribute consistent with real processors.
2. **`test_load_passes_revision` stale assertion** (#391): add missing `force_download=False` to `assert_called_with`.
3. **`test_invalid_url_raises` stale regex** (#803): update match string from `"Failed to load image from URL"` to `"Failed to load image from"` to match the current error format.
4. **`get_class_predicate` import error** (#821): resolved by the extraction above.

## Test plan

- [x] `pytest mlx_vlm/tests/ --ignore=mlx_vlm/tests/test_smoke.py` — 464 passed, 1 skipped
- [x] `pytest mlx_vlm/tests/test_utils.py` — 18 passed (local only, excluded from CI)


mlx_vlm/utils.py
```diff
@@ -1194,6 +1199,16 @@ def prepare_inputs(
...
+        image_token = getattr(processor, "image_token", None)
+        if isinstance(image_token, str) and prompts:
+            prompts_list = [prompts] if isinstance(prompts, str) else prompts
+            for prompt in prompts_list:
+                count = prompt.count(image_token)
+                if count > 0:
+                    raise ValueError(
+                        f"Number of image tokens in prompt_token_ids ({count}) "
+                        f"does not match number of images (0)"
+                    )
```
Blaizzy: please revert this

Blaizzy (commented):
Mostly, LGTM!

mlx_vlm/utils.py
mdstaff: Thanks for the feedback, I've pushed that change next up is resolving the merge conflict.

Blaizzy: My pleasure!

mdstaff: Should be all set now

lucasnewman (approved):

PR #1074, opened 2026-04-26
Fix silent MoE to dense fallback in qwen3_5, qwen3_vl, and glm4v

### LINKED ISSUE #1073: Qwen3.5 MoE models use dense MLP instead of SparseMoeBlock
## Bug

`mlx_vlm/models/qwen3_5/language.py` hardcodes `Qwen3_5MLP` (dense MLP) for every decoder layer, ignoring `num_experts` in the model config. Qwen3.5 MoE models (35B-A3B with 256 experts, 122B-A10B) require `SparseMoeBlock` when `num_experts > 0`.

The reference implementation in `mlx_lm/models/qwen3_5.py` handles this correctly:

```python
# mlx_lm (correct)
if args.num_experts > 0:
    self.mlp = SparseMoeBlock(args)
else:
    self.mlp = MLP(args.hidden_size, args.intermediate_size)
```

The current mlx_vlm code:

```python
# mlx_vlm (bug -- always dense)
self.mlp = Qwen3_5MLP(args.hidden_size, args.intermediate_size)
```

## Measured impact

I tested on Qwen3.6-35B-A3B-VLM-MTP-8bit. The MLLM path (which goes through `mlx_vlm`) produces garbage text ("Ber Ber ber 数 Ber") because expert routing is entirely skipped. The dense path runs at ~32 tok/s (fast precisely because it is computing the wrong function) while the correct MoE path (using `mlx_lm` TextModel directly) runs at ~5.5 tok/s and produces correct output.

Every Qwen3.5 MoE model served through mlx_vlm will produce incorrect outputs.

## Expected fix

The `DecoderLayer.__init__` in `mlx_vlm/models/qwen3_5/language.py` should check `args.num_experts` and use `SparseMoeBlock` when `num_experts > 0`, matching what `mlx_lm` does.

Thump604:
## Summary

Fixes #1073.

Three model families had the same bug: the base `DecoderLayer` unconditionally created a dense MLP, ignoring MoE fields from the model config. MoE variants loaded through these module paths compute the wrong function and produce garbage output.

The fix is the same pattern for each family: add MoE config fields with safe zero/None defaults to the base TextConfig, add the MoE block class to the base language module, make `DecoderLayer.__init__` select MoE vs dense based on the config, and simplify the `_moe` variant module to thin re-exports.

### qwen3_5 (original commit)

- Added `Qwen3_5SparseMoeBlock` to `qwen3_5/language.py`
- DecoderLayer checks `num_experts > 0` to select MoE vs dense MLP
- Simplified `qwen3_5_moe/language.py` to thin wrappers

### qwen3_vl

- Added MoE config fields (`num_experts`, `num_experts_per_tok`, `moe_intermediate_size`, `decoder_sparse_step`, `mlp_only_layers`) to `qwen3_vl/config.py` TextConfig
- Added `Qwen3VLSparseMoeBlock` to `qwen3_vl/language.py`
- DecoderLayer checks `num_experts > 0` with `decoder_sparse_step` and `mlp_only_layers` guards
- Simplified `qwen3_vl_moe/language.py` to thin wrappers

### glm4v

- Added MoE config fields (`n_routed_experts`, `n_shared_experts`, `moe_intermediate_size`, `num_experts_per_tok`, `first_k_dense_replace`, `n_group`, `topk_group`, `routed_scaling_factor`, `norm_topk_prob`, `scoring_func`, `topk_method`) to `glm4v/config.py` TextConfig
- Added `MoEGate`, `MoE`, `group_expert_select` to `glm4v/language.py`
- DecoderLayer checks `n_routed_experts is not None` and `layer_idx >= first_k_dense_replace`
- Simplified `glm4v_moe/language.py` to thin wrappers

## Measured impact

Without this fix, MoE models served through mlx_vlm produce garbage text because expert routing is entirely skipped. The dense path runs fast precisely because it is computing the wrong function.

## Test plan

- [ ] Load a Qwen3.5 MoE model (e.g. Qwen3.5-35B-A3B) and verify coherent output
- [ ] Load a Qwen3.5 dense model and verify no regression
- [ ] Load a Qwen3 VL MoE model and verify coherent output
- [ ] Load a Qwen3 VL dense model and verify no regression
- [ ] Load a GLM-4V MoE model and verify coherent output
- [ ] Load a GLM-4V dense model and verify no regression
- [ ] Verify `_moe` module paths still work (backwards compatibility of re-exports)


Blaizzy:
Hey @Thump604 

Thanks for the contribution!

But looking at the solution, I'm not sure I agree.

Could you improve your issue with detailed reproducible examples and we will take it from there.

Thump604:
Follow-up from my local runtime audit after the clarification on #1073:

I agree this PR is not the right upstream shape as written. The important correction is that `qwen3_5` / `qwen3_vl` should remain the dense modules, and MoE configs should route through `qwen3_5_moe` / `qwen3_vl_moe`.

My local Qwen 35B config reports `model_type=qwen3_5_moe` with the expected MoE fields, and static inspection of the installed package shows the loader imports by `config["model_type"]`. I am treating the old dense-module MoE patch as a local historical carry to remove or prove during the next controlled runtime window, not as an upstreamable patch.

I will open a fresh minimal issue/PR only if I can reproduce an upstream loader path that routes a MoE config through a dense module.

Blaizzy:
No, please patch your config because that’s not an issue on our side it’s a misconfiguration of the model config file.

PR #1076, opened 2026-04-26
Fix Unicode byte-fallback decoding in server streaming responses

spicyneuron:
After recent server changes, I noticed that emoji characters would frequently appear as � in responses.

Root cause was in the continuous batching server path. It decoded output by repeatedly calling `tokenizer.decode(tokens)` on the growing generated-token list, then slicing out the new text delta. That works for normal tokens, but it breaks when a character is represented as multiple byte-fallback tokens. For example, an emoji may arrive as several UTF-8 byte tokens; decoding before the full byte sequence is available can produce a replacement character, and the server would then treat that replacement as final output.

The fix changes the batched server response path to use a per-request streaming detokenizer, matching the safer generation path already used elsewhere in the repo. The detokenizer buffers byte tokens until they form valid UTF-8, so incomplete byte sequences are not emitted prematurely.

This PR also adds a test that simulates a byte-fallback emoji split across tokens.


Blaizzy:
Thanks, @spicyneuron!

Please open a GH issue :)

PR #1078, opened 2026-04-26
fix: remap per-layer quantization keys for mixed-bit models

ivanfioravanti:
## Summary

Models with per-layer mixed-precision quantization (e.g. Intel AutoRound with mixed 4-bit/6-bit per layer) fail to load because per-layer quantization overrides are never matched during `nn.quantize()`.

## Root Cause

The `quantization` config stores per-layer overrides using **weight-key format**:
```
model.language_model.layers.0.linear_attn.in_proj_qkv
```

But `nn.quantize()` walks the module tree using **module-path format** (after `sanitize()` remapping):
```
language_model.model.layers.0.linear_attn.in_proj_qkv
```

So `get_class_predicate()` never finds the per-layer config, falls through to the global `bits` value, and the model gets quantized with the wrong bit-width — causing a shape mismatch on `load_weights()`.

Example error:
```
ValueError: Expected shape (10240, 640) but received shape (10240, 960)
for parameter language_model.model.layers.0.linear_attn.in_proj_qkv.weight
```

(640 = 4-bit packed, 960 = 6-bit packed — the override was ignored)

## Fix

Remap `model.language_model.` → `language_model.model.` keys in the `quantization` dict before calling `nn.quantize()`, so `get_class_predicate()` can match per-layer overrides.

## Testing

Verified with `Intel/Qwen3.6-27B-4.5b-mlx-AutoRound` which uses mixed 4-bit/6-bit quantization — loads and runs successfully after this fix.

### DIFF
--- a/mlx_vlm/models/granite4_vision/config.py
+++ b/mlx_vlm/models/granite4_vision/config.py
@@ -11,7 +11,7 @@ class TextConfig(BaseModelConfig):
     hidden_size: int = 2560
     num_hidden_layers: int = 40
     intermediate_size: int = 8192
-    shared_intermediate_size: int = 8192
+    shared_intermediate_size: Optional[int] = None
     num_attention_heads: int = 40
     rms_norm_eps: float = 1e-5
     vocab_size: int = 100353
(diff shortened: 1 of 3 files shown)


Blaizzy (commented):
Thanks a lot Ivan, this seems to be an issue that can be solved in the model sanitize methods 👌🏽

I will fix it by EoD, there are few other issues related to qwen3.6

ivanfioravanti:
I was thinking the same while looking at mlx-lm, here in utils.py is centralized but it has hardcoded path. Not the most elegant solution. Sanitize better, but not centralized. Eager to see what solution will you choose. Thanks!

Blaizzy:
Thanks! 

If no one opens a PR by then I will override this one and merge with you as co-author 👌🏽

Issue #1072, opened 2026-04-25
feature: add missing metadata in server endpoints


goniz:
I want to add support for mlx_vlm.server in my OpenCode plugin: https://github.com/goniz/opencode-local-provider

My plugin allows detection of server type by url and enumeration of loaded models and their required metadata to pass on to OpenCode automatically.

I'm missing a simple and reliable way to detect it's a vlm server, probably in: https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/server.py#L2496
Example: https://github.com/goniz/opencode-local-provider/blob/main/src/providers/ollama.ts#L10

In addition I'm missing what model is loaded (beyond https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/server.py#L2502) either in a dedicated endpoint or in /v1/models.
I need it to include:
- context length (configured, not train_ctx)
- whether tool calling is supported by the model
- whether it's a vision enabled model or not

Reference: https://github.com/goniz/opencode-local-provider/blob/main/src/types.ts#L5

PR #1082, opened 2026-04-26
server: added 'server' header in responses

goniz:
Added a “Server” response header in the mlx_vlm.server cli to help identifying the running server.

This helps #1072 in detecting the server type from url only.


Blaizzy (approved):
LGTM, thanks!

PR #1088, opened 2026-04-28
Nemotron H Nano Omni: rename SoundConfig and add processor

Blaizzy:
## Summary
- Renamed `SoundConfig` → `AudioConfig` across the `nemotron_h_nano_omni` package, `__init__` exports, audio module type hints, and tests. The `sound_config` field name on `ModelConfig` is kept to match the upstream HF config key.
- Added `NemotronHNanoOmniProcessor` (`processing_nemotron_h_nano_omni.py`), ported from NVIDIA's reference processor at `nvidia/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-BF16`. Adapted to mlx-vlm conventions: drop torch, use flat `images_kwargs`/`videos_kwargs`/`audio_kwargs`/`text_kwargs`, return mlx tensors via `to_mlx`, and keep `sound_clips` as a list of raw waveforms so the model's `SoundFeatureExtractor` can consume them at runtime.
- `from_pretrained` loads `AutoTokenizer` + `AutoImageProcessor` (with `trust_remote_code=True`) and applies `load_chat_template`; optional overrides come from `processor_config.json`.
- Audio file loading goes directly through `mlx_audio` — no librosa/soundfile fallback.
- Registered with `install_auto_processor_patch` for both the upstream model_type (`nemotronh_nano_omni_reasoning_v3`) and the canonical mlx-vlm name.

### DIFF
--- a/mlx_vlm/models/nemotron_h_nano_omni/__init__.py
+++ b/mlx_vlm/models/nemotron_h_nano_omni/__init__.py
@@ -1,4 +1,5 @@
 from .config import AudioConfig, ModelConfig, TextConfig, VisionConfig
+from .image_processing_nemotron_h_nano_omni import NemotronHNanoOmniImageProcessor
 from .language import LanguageModel
 from .nemotron_h_nano_omni import Model
 from .processing_nemotron_h_nano_omni import NemotronHNanoOmniProcessor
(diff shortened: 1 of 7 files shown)


lucasnewman (approved):
LGTM

PR #1092, opened 2026-04-30
server: added loaded model's context size and tool call parser to /health endpoint

### LINKED ISSUE #1072: feature: add missing metadata in server endpoints

goniz:
Added "loaded_context_size" and "loaded_tool_parser" fields to /health endpoint.
Will be used by opencode-local-provider plugin to detect and probe the loaded models :) 

Went with the smallest, easy to implement method. point me to the right direction if you feel it should be implemented somewhere else! thanks!

Closes #1072


Blaizzy (approved):
LGTM, thanks!

PR #1093, opened 2026-04-30
fix: Prevent batched cache metadata lazy graph buildup

spicyneuron:
## Summary

This fixes lazy graph accumulation in long-running batched generation by explicitly evaluating lightweight cache metadata arrays during each decode step. 

The issue is similar to the upstream `mlx-lm` fix in ml-explore/mlx-lm#1232. Cache metadata such as `offset`, `left_padding`, `lengths`, and `_lengths` can be updated token-by-token without being materialized often enough, eventually leading to `[metal::malloc] Resource limit (...) exceeded` failures even when full KV tensors are not unexpectedly growing.

## Details

- Adds `_iter_cache_eval_arrays()` to discover lightweight cache metadata arrays
- Handles nested cache structures through `caches`
- Ignores non-array or missing metadata fields
- Adds those metadata arrays to the existing `mx.async_eval()` call in `GenerationBatch._step`
- Avoids evaluating full KV cache tensors


Blaizzy:
Thanks!

Could you please open an issue with a reproducible example?

spicyneuron:
@Blaizzy My deployed fork had a couple other changes merged, making it a bit harder to isolate. I've reset it to the latest stock `mlx-vlm` and see if batching causes the same issues.

PR #1094, opened 2026-04-30
fix: Batch generation breaks top-p sampling

spicyneuron:
## Summary

This PR fixes a batched top-p sampling bug that can cause concurrent generations to choose tokens using another sequence's sorted vocabulary mapping.

## Problem

A general scenario is:

1. A request starts generation with `temperature > 0` and `top_p < 1.0`.
2. Another request arrives while the first is still generating.
3. The server batches the active sequences and calls the sampler with logits shaped `[B, vocab]`.
4. The old implementation attempts to call `sorted_indices.squeeze(0)`, which only works when `B == 1`.

With `B > 1`, this can fail with:

```text
ValueError: [squeeze] Cannot squeeze axis 0 with size 2 which is not equal to 1.
```

The underlying issue is that top-p sampling needs to preserve the per-row sorted vocabulary mapping. Squeezing away the batch dimension assumes unbatched input and breaks once concurrent sequences are sampled together.

## Solution

This updates `top_p_sampling` to normalize unbatched `[vocab]` logits to `[1, vocab]` internally, use `mx.take_along_axis` for per-row gathers, and preserve the existing scalar return shape for unbatched inputs.


Blaizzy (approved):
LGTM, thanks!

PR #1095, opened 2026-04-30
Fix per-sequence MRoPE alignment in mixed VL+text batches

neilmehta24:
GenerationBatch currently carries per-sequence rope_deltas for Qwen-style MRoPE models. In mixed batches where some sequences came from image prompts and some are text-only, make sure rope_deltas are always aligned one row per active sequence. A text-only sequence must not accidentally inherit another sequence's image MRoPE delta. When merging batches, if one side has rope_deltas and the other does not, synthesize zero-delta rows for the text-only side.


lucasnewman (approved):
Thanks!

PR #1097, opened 2026-05-01
feat(server): support tools / function-calling on /v1/responses

jasonvassallo:
## Problem

Today `/v1/responses` accepts a `tools` array but **silently drops it**: a function-definition request comes back as plain text with no way for the model to invoke the tool. The same request via `/v1/chat/completions` correctly emits `tool_calls`.

### Repro (before)

```bash
curl http://localhost:8000/v1/responses -d '{
  \"model\": \"mlx-community/Qwen3.6-35B-A3B-8bit\",
  \"input\": [{\"role\":\"user\",\"content\":\"What is the weather in Paris? Use the get_weather tool.\"}],
  \"max_output_tokens\": 256,
  \"tools\": [{\"type\":\"function\",\"name\":\"get_weather\",
              \"parameters\":{\"type\":\"object\",
              \"properties\":{\"city\":{\"type\":\"string\"}}}}]
}'
```

→ Returns a polite refusal text. `output[]` is just a `message` item; no `function_call` item is ever produced.

### After this PR

Same request returns:

```json
\"output\": [
  {
    \"id\": \"fc_...\",
    \"type\": \"function_call\",
    \"status\": \"completed\",
    \"call_id\": \"...\",
    \"name\": \"get_weather\",
    \"arguments\": \"{\\\"city\\\": \\\"Paris\\\"}\"
  }
]
```

…and streaming yields `response.output_item.added` (function_call) → `response.function_call_arguments.delta` → `response.function_call_arguments.done` → `response.output_item.done`, per the [OpenAI Responses spec](https://platform.openai.com/docs/api-reference/responses-streaming/response/function_call_arguments/delta).

## What changed

* **`OpenAIRequest`** — add `tools` and `tool_choice`. Both the Responses-shape (`{type, name, description, parameters}`) and the Chat-Completions shape (`{type, function: {...}}`) are accepted.
* **`ResponsesInputItem`** — new model for `/v1/responses` input items so `function_call` and `function_call_output` items round-trip through multi-turn tool conversations. Kept separate from `ChatMessage` so the chat-completions request/response schema is unchanged.
* **`FunctionCallItem`** — new output-item model. `OutputItem = Union[MessageItem, FunctionCallItem]` is now used by `output_item.added` and `output_item.done`.
* Two new stream events: **`response.function_call_arguments.delta`** and **`response.function_call_arguments.done`** ([spec](https://platform.openai.com/docs/api-reference/responses-streaming/response/function_call_arguments/delta)).
* **`_responses_tools_to_chat_format()`** — small translator so `apply_chat_template(tools=...)` keeps accepting Chat-Completions-shaped tools internally while callers can send the Responses shape.
* **`responses_endpoint`** —
  * Translate `function_call` / `function_call_output` input items into the chat-template's tool-call / tool-result turns.
  * Pass `tools` through to `apply_chat_template`.
  * After generation, parse tool calls (**reusing the existing `process_tool_calls`** — no new generation logic) and emit them as `function_call` output items rather than dropping them as plain text.
  * **Streaming with tools buffers the generation** then emits per-output-item events at the end. Streaming raw text deltas would otherwise leak `<tool_call>...</tool_call>` markers into the message text instead of producing a clean `function_call` item.
  * **No-tools streaming path is byte-for-byte unchanged** — non-tool callers see zero behavioural change.

This extends the precedent set by #1047 (json_schema on `/v1/responses`): chat-completions-only logic gets reused on the Responses endpoint instead of being lifted into a parallel implementation. The diff is glue, not new generation logic.

## Tests

Added in `mlx_vlm/tests/test_server.py`:

* `TestResponsesToolsToChatFormat` — helper unit tests (Responses-shape conversion, Chat-shape passthrough, missing-parameters defaulting).
* `TestResponsesInputItemSchema` — Pydantic schema for input items.
* `TestResponsesEndpointTools` — end-to-end with mocks covering tools forwarded to the template, function_call output emission, and multi-turn input translation.

```
$ pytest mlx_vlm/tests/test_server.py mlx_vlm/tests/test_gemma4_tool_parser.py mlx_vlm/tests/test_prompt_utils.py
77 passed in 1.29s
```

## Empirical verification (Qwen3.6-35B-A3B-8bit)

| Test | /v1/responses (before) | /v1/responses (after this PR) |
|---|---|---|
| T1 — non-streaming function call | message refusal, `tool_calls: []` | `function_call` output item ✓ |
| T2 — streaming function call | text deltas containing `<tool_call>` markers | clean `response.function_call_arguments.delta/done` events ✓ |
| T3 — multi-turn (`function_call` + `function_call_output` input) | 422 (input model rejects non-message items) | model uses tool result to produce natural-language answer ✓ |
| T4 — `/v1/chat/completions` regression | `tool_calls` populated | unchanged ✓ |

cc @Blaizzy @avbiswas (#1047 is the precedent, and this is the natural follow-up for tool calling)

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -773,6 +773,48 @@ def suppress_tool_call_content(
     return in_tool_call, delta_content
 
 
+def _responses_tools_to_chat_format(tools):
+    """
+    Translate the Responses-shaped tools list into the Chat-Completions shape
+    that ``apply_chat_template`` expects.
+
+    Responses request shape::
+
+        [{"type": "function", "name": "...", "description": "...", "parameters": {...}}, ...]
+
+    Chat-Completions request shape (what ``apply_chat_template`` accepts)::
+
+        [{"type": "function", "function": {"name": "...", "description": "...", "parameters": {...}}}, ...]
+
+    Tools that already use the Chat-Completions nested shape are passed through
+    unchanged so callers that send Chat-Completions-style tools to /v1/responses
+    still work. Non-function tool types (if any) are also passed through.
+    """
+    if not tools:
+        return None
+    converted = []
+    for tool in tools:
+        if not isinstance(tool, dict):
+            continue
+        if tool.get("type") != "function":
+            converted.append(tool)
+            continue
+        if "function" in tool and isinstance(tool["function"], dict):
+            converted.append(tool)
+            continue
+        converted.append(
+            {
+                "type": "function",
+                "function": {
+                    "name": tool.get("name"),
+                    "description": tool.get("description"),
+                    "parameters": tool.get("parameters") or {},
+                },
+            }
+        )
+    return converted
+
+
 def process_tool_calls(model_output: str, tool_module, tools):
     """Parse tool calls from model output using the appropriate tool parser."""
     called_tools = []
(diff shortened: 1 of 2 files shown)


Blaizzy:
Thanks @jasonvassallo!

Please open a GH issue :)

PR #1098, opened 2026-05-01
minicpmo / fastvlm: fix pixel cast on quantized language models

### LINKED ISSUE #1091: MiniCPM-o-4_5 quantized: vision tower crashes with `Invalid input array with type uint32` from conv2d
## Environment

- **mlx-vlm**: 0.4.4 (pip install)
- **mlx**: 0.31.2
- **mlx-metal**: 0.31.2
- **mlx-lm**: 0.31.3
- **transformers**: 5.6.2
- **Python**: 3.11.5
- **Hardware**: Apple M1 Max, 64 GB Unified Memory, 32 GPU cores
- **OS**: macOS 26.4.1 (arm64)
- **Model**: `mlx-community/MiniCPM-o-4_5-5bit` (the 4-bit variant reproduces the same crash; see "Possibly related issues")

## Description

Vision generation crashes with a `ValueError` from `mx.conv2d` whenever an image is passed to a quantized MiniCPM-o-4_5 model. Text-only generation works. The unquantized bf16 variant of the same model — locally converted from `openbmb/MiniCPM-o-4_5` via `mlx_vlm convert --dtype bfloat16` — processes the same images correctly, so the bug is specific to the interaction between MiniCPM-o quantization and the vision pipeline.

## Steps to Reproduce

```python

model, processor = load(
    "mlx-community/MiniCPM-o-4_5-5bit",
    trust_remote_code=True,
)

# Text-only — works ✅
prompt = apply_chat_template(processor, config=model.config, prompt="Say hello")
print(generate(model, processor, prompt, max_tokens=20).text)

# With image — crashes ❌
prompt = apply_chat_template(
    processor, config=model.config,
    prompt="What is in this image?", num_images=1,
)
generate(model, processor, prompt, image="/path/to/any.png", max_tokens=128)
```

## What I observed

1. **Crash trace** (identical when reached via `mlx_vlm.server` `/v1/chat/completions` and via direct `generate()`):

   ```
   File ".../mlx_vlm/models/minicpmo/minicpmo.py", line 292, in get_vision_embedding
       hidden = self.vision_tower(...)
   File ".../mlx_vlm/models/minicpmo/vision.py", line 152, in __call__
       patch_embeds = self.patch_embedding(pixel_values)
   File ".../mlx/nn/layers/convolution.py", line 157, in __call__
       y = mx.conv2d(...)
   ValueError: [conv] Invalid input array with type uint32. Convolution currently only supports floating point types
   ```

2. **Root cause** (in `mlx_vlm/models/minicpmo/minicpmo.py`, `get_vision_embedding`, around line 251 of v0.4.4):

   ```python
   dtype = self.language_model.model.embed_tokens.weight.dtype
   ...
   cur_pixels = _to_mx_array(cur_pixels, dtype=dtype)
   ```

   When the language model is quantized, `embed_tokens.weight.dtype` is `mlx.core.uint32` (packed quantized weights with `scales`/`biases` siblings), not a float type. `_to_mx_array` then casts the float32 pixel values from the processor to `uint32`, which the vision tower's `patch_embedding` (`nn.Conv2d`) immediately rejects.

3. **Verified empirically**:

   ```python
   >>> from mlx_vlm import load
   >>> model, _ = load("mlx-community/MiniCPM-o-4_5-5bit", lazy=True)
   >>> w = model.language_model.model.embed_tokens.weight
   >>> w.dtype, hasattr(model.language_model.model.embed_tokens, "scales")
   (mlx.core.uint32, True)
   >>>
   >>> # locally converted via mlx_vlm convert --dtype bfloat16
   >>> model, _ = load("./minicpm-bf16", lazy=True)
   >>> model.language_model.model.embed_tokens.weight.dtype
   mlx.core.bfloat16
   >>> # vision works on the bf16 build with the exact same input image
   ```

4. **Fix that worked locally** (one-line guard in `get_vision_embedding`):

   ```python
   dtype = self.language_model.model.embed_tokens.weight.dtype
   if dtype not in (mx.float16, mx.bfloat16, mx.float32):
       # embed_tokens may be quant-packed (uint32 with scales) when
       # the language model is quantized — fall back to a float
       # compute dtype so the vision tower receives floats.
       dtype = mx.bfloat16
   ```

   With this patch both the 4-bit and 5-bit `MiniCPM-o-4_5` MLX variants accept image input and produce correct descriptions.

5. **Workaround for users today**: convert the original `openbmb/MiniCPM-o-4_5` to MLX bf16 without quantization:

   ```
   mlx_vlm convert --hf-path openbmb/MiniCPM-o-4_5 \
                   --mlx-path ./minicpm-bf16 \
                   --dtype bfloat16 \
                   --trust-remote-code
   ```

   ~17 GB on disk, but vision works out of the box because the embedding dtype is `bfloat16` rather than packed `uint32`.

## Possibly related issues

- #400 — Same `Invalid input array with type uint32` symptom for `gemma-3n-E4B-it-4bit`. Closed by #398, which patched `gemma3n.py`. The `minicpmo` model file was never touched.
- #398 — gemma3n pixel-cast fix; analogous logic could land in `models/minicpmo/minicpmo.py::get_vision_embedding`.

contrapuntal:
## What's broken

Quantized VLMs crash with `[conv] Invalid input array with type uint32` the moment you give them an image. Reported in #1091 against `mlx-community/MiniCPM-o-4_5-4bit`, but the same crash also fires on FastVLM as soon as you run the language model through `mlx_vlm convert --quantize`.

## Why

Two model files cast `pixel_values` to `self.language_model.model.embed_tokens.weight.dtype`. On unquantized models that happens to be bf16/fp16, so the cast is a no-op. On quantized models it's `uint32` — MLX stores the layer's actual numerical tensor lazily, and the raw `weight` attribute is the packed `uint32` storage container that lives next to the `scales` and `biases` arrays. The cast turns the float pixel array into `uint32`, then `mx.conv2d` rejects it.

## Fix

Source the dtype from the vision tower's first patch-embedding weight instead of the language model's embedding. Vision towers in mlx-vlm aren't quantized by convention, so this stays a float. ~13 other model files in the repo already follow this exact pattern: `qwen2_vl`, `qwen2_5_vl`, `qwen3_vl`, `qwen3_vl_moe`, `glm4v`, `glm4v_moe`, `glm_ocr`, `dots_ocr`, `ernie4_5_moe_vl`, `internvl_chat`, `youtu_vl`, `molmo`, `molmo2`. On unquantized models the new and old dtype source happen to be the same value, so this is a no-op there.

Same root cause as #398 (gemma3n). That PR removed the cast entirely; this PR keeps the explicit cast and changes its source. The reasoning for keeping the cast: dropping it can leave pixels in `float32` even when the conv weight is `bf16`, which silently inflates memory and breaks dtype consistency with the rest of the repo. An explicit cast to the patch-embed weight dtype guarantees the vision tower's input matches its compute dtype.

## Repro / verification

`mlx-community/MiniCPM-o-4_5-4bit` is the model from the issue. FastVLM doesn't have a published quantized MLX variant on the Hub yet, so the second row uses one I quantized locally with the standard `mlx_vlm convert` flow.

| | `mlx-vlm@main` | This PR |
|---|---|---|
| `mlx-community/MiniCPM-o-4_5-4bit` + image | ❌ uint32 conv error at `models/minicpmo/vision.py:152` | ✅ `"Got it, let me look at the image. The main subject is an airplane flying in"` |
| FastVLM-0.5B 4-bit, quantized via `mlx_vlm convert -q --q-bits 4 --q-group-size 64 --hf-path mlx-community/FastVLM-0.5B-bf16` | ❌ uint32 conv error at `models/fastvlm/vision.py:516` (`MobileOneBlock.reparam_conv`) | ✅ `"Airplane"` |
| `mlx-community/FastVLM-0.5B-bf16` (no quantization, non-regression) | ✅ works | ✅ still works |

## Test plan

- [x] Reproduce the crash on stock `main` for both quantized models, then confirm it's gone on this branch.
- [x] Confirm unquantized FastVLM still produces correct output on this branch (non-regression).
- [x] Pre-commit passes locally: Black `24.2.0`, isort `5.13.2` with `--profile=black`, autoflake `2.2.1`.

Fixes #1091.


lucasnewman (approved):
LGTM, thanks!

PR #1099, opened 2026-05-01
Fix Gemma4 chat templating for Python generate()

Girish011:
## Summary
- Apply the tokenizer's chat template for Gemma 3/4 when calling the Python `generate()`/`stream_generate()` APIs with a raw string prompt.
- This avoids pathological repetition/echoing for Gemma4 (e.g., repeated `Say hello.` loops) and matches the CLI behavior that already applies chat templates.
- Adds a unit test to ensure the templated prompt is passed into `prepare_inputs`, and allows opting out via `skip_chat_template=True`.

## Repro
```python
model, proc = load('mlx-community/gemma-4-e4b-it-4bit')
out = generate(model, proc, prompt='Say hello in one sentence.', max_tokens=40, temperature=0.7, top_p=0.9)
print(out.text)
```

Before: output tends to loop/repeat the prompt.
After: produces a normal short response.

## Test plan
- `python -m pytest mlx_vlm/tests/test_gemma4_chat_template.py -q`

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -1009,11 +1009,33 @@ def stream_generate(
         else []
     )
 
-    add_special_tokens = (
-        getattr(processor, "chat_template", None) is None
-        if model.config.model_type in ["gemma3", "gemma3n", "gemma4"]
-        else True
-    )
+    # Some models (e.g. Gemma 3/4) rely on a HF chat template to format turns.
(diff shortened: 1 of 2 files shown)


Blaizzy:
Hey @Girish011 

Please open an issue first :)

Girish011:
Thanks for the feedback — opened issue #1254 with a minimal repro and root-cause analysis. Happy to resubmit a PR once triaged.

PR #1104, opened 2026-05-03
granite4_vision: support standard granite MLP backbone for 4.1-4b

EliSchwartz:
The granite-4.0-3b-vision uses a GraniteMoeHybrid backbone with a fused SharedMLP (input_linear/output_linear), while granite-vision-4.1-4b uses a standard Granite backbone with separate gate_proj/up_proj/down_proj weights. Add an MLP class for the standard variant and select between the two in TransformerBlock based on text_config.model_type.


Blaizzy (approved):
LGTM, thanks @EliSchwartz! 🚀

PR #1108, opened 2026-05-05
Guard server generation loop after failures

### LINKED ISSUE #1107: Server generation exceptions can spin in a hot error loop
## Summary

When `mlx_vlm.server` hits an exception inside the continuous-batching generation thread, the server can get stuck in a tight repeat-fail loop instead of failing the affected request(s), resetting the poisoned batch state, and backing off.

I saw this with a local OpenAI-compatible client sending requests to `/v1/chat/completions` against a Qwen3.6/Qwen3.5-family model. The root model error was:

```text
Error in generation thread: list index out of range
  ...
  File ".../mlx_vlm/models/qwen3_5/language.py", line 767, in __call__
    cache[self.model.fa_idx]
IndexError: list index out of range
```

After that first exception, the generation thread kept logging the same error and pegged CPU. The HTTP client saw empty/no-content streaming responses and retries. On my Mac, the loop coincided with high memory pressure and eventually a macOS watchdog hard reboot.

## Why this seems server-specific

The underlying `qwen3_5/language.py` `IndexError` may be a separate model/cache bug, but the server loop makes any generation-step exception much worse.

Current `main` still has broad exception handlers in both generation paths:

- `ResponseGenerator._run`: logs `Error in generation thread` and continues
- `ResponseGenerator._run_speculative`: logs `Error in speculative generation thread` and continues

Those handlers do not appear to:

- notify active per-request queues with an exception
- terminate active streams
- clear the `active` map
- discard/reset the current `BatchGenerator`
- clear MLX cache / collect after a failed batch
- apply backoff or circuit-breaking for repeated failures

So if the exception comes from corrupted batch/KV/cache state, the next loop iteration can immediately hit the same bad state again.

## Related but not duplicate

This looks adjacent to:

- #988, where a failed vision request corrupts later server requests
- #1086, where DFlash concurrent requests crash inside the speculative path

But this issue is about server resilience after *any* generation-thread exception, not only the specific model bug.

## Expected behavior

On a generation-step exception, the server should fail the affected active request(s), close their streams, discard the poisoned batch generator state, and pause briefly before accepting more work. Ideally `/health` would expose enough state to show recent generation-loop failures.

## Actual behavior

The server logs the exception and continues with the same loop state, which can produce repeated empty streaming responses and a high-CPU error loop.

## Suggested fix

Add a small failure path around the generation loop:

- wrap the original exception in a request-visible `RuntimeError`
- push it plus a stream terminator to all active `rqueue`s
- clear `active`
- reset `batch_gen = None`
- clear MLX cache / run GC after the failed batch
- track consecutive generation errors and exponential backoff
- expose `consecutive_generation_errors` / `last_generation_error` in `/health`

I have a local dev patch in progress and can open a PR if this direction sounds acceptable.

eloe:
## Summary

- fail active and pending response queues when the server generation loop raises
- reset poisoned batch state, clear MLX cache, collect garbage, and back off after repeated generation failures
- expose recent generation error state in `/health`
- add focused `ResponseGenerator` tests for queue notification, backoff tracking, and health state

Fixes #1107.

## Tests

- `python3 -m compileall mlx_vlm/server.py mlx_vlm/tests/test_server.py`
- `uv run --with pytest python -m pytest mlx_vlm/tests/test_server.py::TestResponseGenerator -q`

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -102,6 +102,17 @@ def get_top_logprobs_k():
     return max(0, min(k, 20))
 
 
+GENERATION_ERROR_BACKOFF_SECONDS = float(
+    os.environ.get("MLX_VLM_GENERATION_ERROR_BACKOFF_SECONDS", 0.25)
+)
+GENERATION_ERROR_BACKOFF_MAX_SECONDS = float(
+    os.environ.get("MLX_VLM_GENERATION_ERROR_BACKOFF_MAX_SECONDS", 5.0)
+)
+GENERATION_ERROR_RESET_WINDOW_SECONDS = float(
+    os.environ.get("MLX_VLM_GENERATION_ERROR_RESET_WINDOW_SECONDS", 30.0)
+)
+
+
 # =============================================================================
 # ResponseGenerator - Concurrent Request Handling with Threaded Batching
 # =============================================================================
@@ -217,6 +228,9 @@ class ResponseGenerator:
         self._load_error: Optional[Exception] = None
         self._cancelled: set = set()
         self._cancel_lock = Lock()
+        self._consecutive_generation_errors = 0
+        self._last_generation_error: Optional[str] = None
+        self._last_generation_error_at = 0.0
         self._thread = Thread(target=self._run, daemon=True)
         self._thread.start()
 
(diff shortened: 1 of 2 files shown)


Blaizzy:
I love the intent, but not the best solution right now.

PR #1111, opened 2026-05-05
Fix token queue timeout during long prefills

### LINKED ISSUE #1068: Hardcoded 60s queue timeout in token_iterator silently truncates long-prompt prefills
## Summary

`mlx_vlm/server.py` line 307 (currently on `main` at `0f903f90`) hardcodes a 60-second timeout on the token-streaming queue:

```python
def token_iterator():
    ended = False
    try:
        while True:
            item = rqueue.get(timeout=60.0)   # <--- here
```

When a prompt is large enough that **prefill alone takes longer than 60s**, no tokens have been pushed to `rqueue` yet, so `queue.Empty` is raised, the iterator exits, and the server returns an HTTP 200 with an **empty body** — indistinguishable, from the client's perspective, from a successful zero-token completion.

This silently truncates legitimate requests on long contexts.

## Reproducing

Hardware: M4 Max 36 GB, MoE model with continuous batching enabled.

Setup:
```bash
python -m mlx_vlm server \
  --model unsloth/Qwen3.6-35B-A3B-UD-MLX-4bit \
  --kv-bits 3.5 --kv-quant-scheme turboquant \
  --prefill-step-size 8192 \
  --max-kv-size 131072
```

Send a chat completion with a ~90k+ token prompt. Observed prefill rate on this hardware is ~600–1000 prompt-tps for this model, so prefill of 90k tokens takes 90–150s — well past the 60s queue timeout.

Result:
```
HTTP/1.1 200 OK
Content-Type: text/event-stream

data: [DONE]
```
No content delta, no error message. Server logs the underlying traceback:

```
File ".../mlx_vlm/server.py", line 307, in token_iterator
    item = rqueue.get(timeout=60.0)
File ".../queue.py", line 179, in get
    raise Empty
_queue.Empty
```

## Why 60s isn't enough in practice

For long-context-capable models (Qwen3.6, Llama-3.1, etc., with native windows of 128k–262k), prefill at any realistic prefill-step-size will exceed 60s for prompts >50k tokens on consumer Apple Silicon. The empirical breakpoints I measured:

| Model | Hardware | Prompt size that exceeds 60s queue wait |
|---|---|---|
| unsloth/Qwen3.6-35B-A3B-UD-MLX-4bit | M4 Max 36 GB | ~70k tokens |
| mlx-community/Qwen3.6-35B-A3B-nvfp4 | M4 Max 36 GB | ~70k tokens |

The actual Metal/compute ceiling on this machine is ~100k actual tokens, but the queue timeout cuts off requests well before the compute ceiling.

## Proposed fix

Make the queue timeout configurable via env var with a generous default:

```python
def token_iterator():
    ended = False
    queue_timeout = float(os.environ.get("TOKEN_QUEUE_TIMEOUT", "600"))
    try:
        while True:
            item = rqueue.get(timeout=queue_timeout)
            ...
```

Rationale for 600s default:
- Covers prefills up to ~600 × prefill_tps, which on consumer Apple Silicon is roughly 200k–400k tokens (more than enough for native context windows of mainstream long-context models)
- A genuinely dead worker is detected within 10 minutes, which is acceptable for chat workloads
- Operators who need a tighter fail-fast behavior can set `TOKEN_QUEUE_TIMEOUT=60` (or anything else) explicitly

Alternative shapes I considered:

1. **CLI flag**: `--token-queue-timeout 600` — same effect, more discoverable. Happy to PR either form.
2. **Scale by `prefill-step-size`** — too clever; an env var is plenty.
3. **Return an error to the client on `queue.Empty`** instead of empty 200 — fixes the silent-truncation aspect even if the timeout itself remains low. Worth doing in addition.

## Patch

```diff
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -300,9 +300,10 @@
         def token_iterator():
             # Mark ended before yielding the final token so a consumer that
             # closes immediately after seeing finish_reason isn't treated
             # as a client abort.
             ended = False
+            queue_timeout = float(os.environ.get("TOKEN_QUEUE_TIMEOUT", "600"))
             try:
                 while True:
-                    item = rqueue.get(timeout=60.0)
+                    item = rqueue.get(timeout=queue_timeout)
                     if item is None:
                         ended = True
                         break
```

Happy to send a PR if this approach is acceptable. Would also be glad to add an explicit error-on-timeout (instead of silent empty response) as a separate fix.

## Environment

- macOS 25.4.0, Apple M4 Max, 36 GB unified memory
- Python 3.12 (Homebrew), mlx_vlm editable install from `main` (`0f903f90`)

eloe:
## Summary
- replace the hardcoded 60s continuous-batching token queue timeout with configurable `MLX_VLM_TOKEN_QUEUE_TIMEOUT` / `TOKEN_QUEUE_TIMEOUT` support
- default to 600s so long-context prefills do not silently terminate before first token
- raise a clear timeout error and cancel the active request when the queue really does time out

Fixes #1068.

## Tests
- `uv run --with pytest pytest mlx_vlm/tests/test_server.py::TestResponseGenerator -q`
- `uv run --with pytest pytest mlx_vlm/tests/test_server.py -q`
- `uv run --with pre-commit pre-commit run --files mlx_vlm/server.py mlx_vlm/tests/test_server.py`
- `python3 -m compileall mlx_vlm/server.py mlx_vlm/tests/test_server.py`


Blaizzy (approved):
LGTM, thanks!

PR #1113, opened 2026-05-05
Fix TPS in API response

Edsuns:
## Summary

Fix missing TPS (tokens per second) metrics in API responses. Previously prompt_tps and generation_tps were not properly calculated or returned, leaving clients unable to retrieve inference performance data.

## Changes

mlx_vlm/generate.py (BatchGenerator)

- Added _gen_time_counter member variable to independently track generation phase duration (excluding prefill)
- Wrapped generation_batch.next() with time.perf_counter() calls for precise timing
- Added generation_time and generation_tps fields to the stats() method
  - generation_tps = generation_tokens / generation_time
    
mlx_vlm/server.py

- StreamingToken dataclass: added prompt_tps and generation_tps fields
- ResponseGenerator: on finish_reason, fetches TPS from batch_gen.stats() and injects into StreamingToken
- responses endpoint and chat/completions endpoint: fully populated performance metrics:
  - Streaming responses: usage now includes prompt_tps, generation_tps, total_tokens, peak_memory
  - Non-streaming responses: same fields added
  - Blocking _blocking_generate: return value extended to include prompt_tps and generation_tps
- Minor fix: consolidated a split string literal in a logger format call
    
## Scope

- Affects 2 API endpoints (/v1/responses, /v1/chat/completions) for both streaming and non-streaming responses
- No changes to model inference logic — only affects statistics collection and reporting
- No breaking changes — clients can optionally consume the new fields

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -2004,6 +2004,7 @@ class BatchGenerator:
         self._prompt_tokens_counter = 0
         self._prompt_time_counter = 0
         self._gen_tokens_counter = 0
+        self._gen_time_counter = 0.0
         self._steps_counter = 0
 
         self._wire_stack = contextlib.ExitStack()
@@ -2109,6 +2110,12 @@ class BatchGenerator:
             else 0
         )
         stats.generation_tokens = self._gen_tokens_counter
+        stats.generation_time = self._gen_time_counter
+        stats.generation_tps = (
+            self._gen_tokens_counter / self._gen_time_counter
+            if self._gen_time_counter > 0
+            else 0
+        )
         stats.peak_memory = mx.get_peak_memory() / 1e9
         return stats
 
@@ -2118,7 +2125,9 @@ class BatchGenerator:
 
         # Decode-first: always emit a generation step before touching prefill.
         if len(self._generation_batch) > 0:
+            tic = time.perf_counter()
             generation_responses = self._generation_batch.next()
+            self._gen_time_counter += time.perf_counter() - tic
             self._gen_tokens_counter += len(generation_responses)
             self._steps_counter += 1
             if self._steps_counter % 512 == 0:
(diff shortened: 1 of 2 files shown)


Blaizzy:
Thanks @Edsuns! 

I believe I worked on this and the conclusion was, it's inaccurate to calculate these at the server level. If you compare both generate CLI and server they differ vastly. 

It's best done on the client side. 

But if you have a differing experience please do share and we can consider.

Edsuns:
@Blaizzy Thanks for the response. Let me address your concerns:

**1. TPS was already returned before continuous batching**

Before the BatchGenerator was introduced, `UsageStats` already had `prompt_tps`, `generation_tps` and `peak_memory` fields. The streaming path used `stream_generate` which computes per-request TPS identical to the CLI. When continuous batching replaced this path, TPS was silently dropped from the API response — this is a regression, not a new feature.

**2. The inaccuracy is in the current PR approach, not in server-side TPS itself**

You are right that `batch_gen.stats()` gives batch-aggregate TPS, not per-request TPS. But the fix is straightforward: `GenerationResult` (generate.py:374) already carries `prompt_tps` and `generation_tps` computed at the per-request level, matching the CLI behavior. The server should propagate these values instead of calling `batch_gen.stats()`. I can rework the PR to do this.

**3. Peak memory is impossible to measure client-side**

`mx.get_peak_memory()` is a runtime API that only the MLX process can call. A client has no way to know the actual GPU memory footprint. This is critical for capacity planning — users need to know if their model is close to OOM.

**4. Server-side TPS is also useful for diagnosing batch performance**

I observed that the server generation TPS is lower than the CLI for the same single-request workload. This is valuable diagnostic information — it reveals overhead from the batching loop (`_next()` checks prompt_batch, unprocessed_sequences every step, and the `next_steady()` fast path from commit ad05280 is unused). Without server-side TPS, users cannot distinguish "model is slow" from "server overhead is high".

**Summary**: The correct approach is to return per-request TPS from `GenerationResult` (not `batch_gen.stats()`), plus `peak_memory` which is server-only. I will update the PR accordingly.

Edsuns:
Measure TPS using [aiperf](https://github.com/ai-dynamo/aiperf).

PR #1116, opened 2026-05-05
Fix Gemma 4 MTP server dispatch

JakeATX:
## Summary

Fixes Gemma 4 MTP speculative decoding in `mlx_vlm.server`.

`mlx_vlm.generate` already dispatches `draft_kind="mtp"` to the Gemma 4 MTP round loops, but the server speculative worker still always imported/called `_dflash_rounds_batch`. That meant `mlx_vlm.server --draft-kind mtp` could load a Gemma 4 assistant drafter but route the decode loop through the wrong speculative implementation.

This PR:

- dispatches server speculative decoding to `_mtp_rounds_batch` when `--draft-kind mtp` is selected;
- keeps DFlash as the default path and leaves DFlash routing unchanged;
- uses Gemma MTP-specific prefill hooks (`return_hidden=True`, `return_shared_kv=True`);
- passes shared KV state and EOS IDs into the MTP batch loop;
- labels acceptance logs as `[MTP]` or `[DFLASH]`;
- adds small dispatch tests;
- adds Gemma 31B-safe mixed quantization predicates for memory-constrained local 31B deployments.

## Validation

Unit/static checks:

- `python -m py_compile mlx_vlm/server.py mlx_vlm/convert.py`
- `python -m pytest mlx_vlm/tests/test_server.py mlx_vlm/tests/test_convert_quant_predicates.py mlx_vlm/tests/test_generate.py::test_generate_cli_smoke -q`
- Result: `47 passed`

Local Apple Silicon smoke with Gemma 4 31B target + official 31B assistant:

- target: `/Users/jkooker/.lmstudio/models/lmstudio-community/gemma-4-31B-it-MLX-6bit`
- drafter: `mlx-community/gemma-4-31B-it-assistant-bf16`
- `generate` no-MTP: `13.34 tok/s` for 32 generated tokens
- `generate` MTP: `22.03 tok/s`, acceptance `1.385` tokens/round
- patched `server` MTP request completed with HTTP 200 and logged: `[MTP] batch=1 tokens=32 accept=1.46 rounds=13`

## Notes

The Gemma-safe quant predicates are intended as a fallback for memory-constrained 31B deployments. They follow the existing mixed quantization style while keeping embeddings/per-layer embedding paths and selected `v_proj` / `down_proj` tensors at higher precision.

PR #1117, opened 2026-05-05
Improve Gemma 4 MTP server batching

JakeATX:
## Summary

Follow-up to #1115, per @Blaizzy's request on #1116.

This PR targets `pc/fix-gemma-4-mtp-server` and keeps the diff limited to the remaining batching/correctness improvements from the local Gemma 4 31B MTP validation work.

## Changes

- Handle vector-valued `BatchKVCache.offset` in `_mtp_rounds_batch` by preserving per-row positions instead of collapsing every row to the max offset.
- Add `MLX_VLM_SPEC_BATCH_WAIT_MS` so speculative serving can briefly coalesce simultaneous requests into real MTP batches.
- Pass correct left-padding metadata to the speculative server prefill cache for mixed-length request batches.
- Make Gemma 4 assistant masks respect per-row offsets/valid lengths for batched rows.
- Add tests for the coalescing env var and batched assistant masks.

## Local validation

Target/runtime used locally:

- Hardware: Apple M4 Max, 64 GB RAM
- Target: `/Users/jkooker/.lmstudio/models/lmstudio-community/gemma-4-31B-it-MLX-6bit`
- Assistant: `mlx-community/gemma-4-31B-it-assistant-bf16`
- Endpoint: patched `mlx_vlm.server`, `/v1/chat/completions`
- Sampling: temperature 0, thinking disabled
- Coalescing: `MLX_VLM_SPEC_BATCH_WAIT_MS=1000`

Same-prompt/equal-length batch throughput after these changes:

| draft_block_size | user batch | aggregate tok/s | assistant accept/round |
| --- | ---: | ---: | ---: |
| 4 | 4 | 41.59 | 1.87 |
| 4 | 8 | 65.64 | 1.87 |
| 7 | 4 | 33.18 | 2.88 |
| 7 | 8 | 39.94 | 2.88 |

Mixed-length batches are now position/mask-correct, but acceptance is lower. For maximum throughput, production servers may want to batch by similar prompt length/template.

## Tests

```bash
/opt/miniconda3/bin/python -m pytest \
  mlx_vlm/tests/test_server.py \
  mlx_vlm/tests/test_gemma4_assistant_masks.py \
  mlx_vlm/tests/test_generate.py::test_generate_cli_smoke -q
```

Result: `50 passed`.

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -87,6 +87,11 @@ def _get_draft_block_size_from_env():
     return int(draft_block_size_str) if draft_block_size_str else None
 
 
+def _get_spec_batch_wait_from_env() -> float:
+    wait_ms = os.environ.get("MLX_VLM_SPEC_BATCH_WAIT_MS")
+    return max(0.0, float(wait_ms) / 1000.0) if wait_ms else 0.0
+
+
 def get_prefill_step_size():
     return int(os.environ.get("PREFILL_STEP_SIZE", DEFAULT_PREFILL_STEP_SIZE))
 
@@ -600,6 +605,7 @@ class ResponseGenerator:
         eos_set = set(self.stop_tokens) if is_mtp else None
         sampler = _make_sampler(temp=0)
         draft_block_size = _get_draft_block_size_from_env()
+        spec_batch_wait = _get_spec_batch_wait_from_env()
 
         while not self._stop:
             try:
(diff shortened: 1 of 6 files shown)


Blaizzy:
Awesome, LGTM!

Just one question, why do you need the wait? Because the generator already does that

mlx_vlm/server.py
```diff
@@ -87,6 +87,11 @@ def _get_draft_block_size_from_env():
     return int(draft_block_size_str) if draft_block_size_str else None
 
 
+def _get_spec_batch_wait_from_env() -> float:
+    wait_ms = os.environ.get("MLX_VLM_SPEC_BATCH_WAIT_MS")
+    return max(0.0, float(wait_ms) / 1000.0) if wait_ms else 0.0
```
Blaizzy, line 92: let's revert this

PR #1127, opened 2026-05-06
Fix mixed-length Gemma 4 batching

### LINKED ISSUE #1083: batch_generate produces token-level loops on Gemma 4 models at batch_size>=4 (works at B=1)
## Summary

`mlx_vlm.batch_generate` produces degenerate token-level loops on Gemma 4 models (`Gemma4ForConditionalGeneration` / `model_type="gemma4"`) at `batch_size >= 4`. At `batch_size = 1` the same models produce coherent output. The failure persists across both the `gemma-4-26b-a4b-it-4bit` MoE and `gemma-4-31b-it-8bit` dense variants. Same prompt, same sampler, same templates — only the batch size differs.

## Models tested

- `mlx-community/gemma-4-26b-a4b-it-4bit`
- `mlx-community/gemma-4-31b-it-8bit`

(Gemma 3 family — `mlx-community/gemma-3-4b-it-qat-4bit` — works correctly at all batch sizes in the same code path.)

## Symptom

Sample raw output for the same record at different batch sizes (max_tokens=80, temperature=0.3):

**`batch_size = 1` (works):**
```
{"action": "kept", "text": "Section Preface.\n\nEveryone will readily agree that it is of the highest importance to know whether we are not duped by morality."}
```

**`batch_size = 4` (loops):**
```
I-I-I-I-I-I-I-I-I-I-I-I-I-I-I-I-I-I-I-I-I-IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
```
```
TheTheTheTheTheTheTheTheTheTheTheTheTheTheTheTheTheTheTheTheTheTheTheTheTheTheTh
```
```
I's-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-s-
```

The same loops appear at `batch_size = 8`. Loops are model-output-level, not detokenizer-level — `prompt_tokens` and `generation_tokens` counts look normal, the model is just sampling degenerately.

## Reproduction

Minimal reproducer (~3 minutes, no external infrastructure beyond `mlx-vlm` itself):

```python
from mlx_vlm import load, batch_generate

model, processor = load("mlx-community/gemma-4-26b-a4b-it-4bit")

# Build any chat prompts via the processor's chat template
messages_list = [
    [{"role": "user", "content": "Briefly summarize: the cat sat on the mat."}],
    [{"role": "user", "content": "Briefly summarize: she opened the book."}],
    [{"role": "user", "content": "Briefly summarize: rain fell all night."}],
    [{"role": "user", "content": "Briefly summarize: the engine roared."}],
]
prompts = [
    processor.apply_chat_template(m, tokenize=False, add_generation_prompt=True)
    for m in messages_list
]

# B=1: works
for p in prompts:
    resp = batch_generate(model, processor, prompts=[p], max_tokens=80, verbose=False)
    print("B=1:", repr(resp.texts[0][:80]))

# B=4: every output is a token-level loop
resp = batch_generate(model, processor, prompts=prompts, max_tokens=80, verbose=False)
for t in resp.texts:
    print("B=4:", repr(t[:80]))
```

## Hypothesized root cause

The failure is fully isolated to the batched-cache code path. We have ruled out:

- **Chat template / kwargs** — failure persists with vanilla `apply_chat_template(tokenize=False, add_generation_prompt=True)` and no extra kwargs.
- **Repetition penalty** — failure persists with `logits_processors=make_logits_processors(repetition_penalty=2.0)` (also tested 1.5).
- **Sampler** — failure persists with `sampler=None` (default greedy) and with `make_sampler(temp=0.3, top_p=0.9)`.

The likely culprit is the **batched KV cache wrapper** in `mlx_vlm.generate.BatchGenerator._process_prompts`, which already special-cases the `B=1, no padding` path:

```python
elif len(uids) == 1 and max(left_padding) == 0:
    # Single prompt with no padding: use standard caches to avoid
    # numerical divergence from batch cache wrappers.
    prompt_cache = cache.make_prompt_cache(self.model)
else:
    prompt_cache = _make_cache(self.model, left_padding)
```

For Gemma 4 architecture specifically, this "numerical divergence" appears to be **catastrophic** rather than just imprecise — the cache state diverges enough that the model emits degenerate token loops. Plausible specific cause: Gemma 4's attention pattern (sliding window? mixed local/global?) or RoPE variant interacting badly with the left-padding-aware batched cache. We have not fully isolated it past the architecture-level finding.

## Environment

```
mlx-vlm 0.4.4
mlx 0.31.x  (also tested with mlx 0.30.6 + mlx-lm 0.30.7 pin — same result)
Python 3.13.6
Apple Silicon (M5 Pro 64GB unified memory)
```

## Workaround

Forcing `batch_size = 1` works but defeats the point of `batch_generate`. We've adopted this as a temporary production constraint while waiting on a fix.

---

I have a fuller diagnostic suite (per-batch-size output samples, additional rep-penalty / sampler ablations) — happy to share or fold into a smaller repro if helpful. Thanks for `mlx-vlm`.

### LINKED ISSUE #1124: Mixed-length B=4 still fails after #1123 — concat shape mismatch in generate.py _next path
## Summary

PR #1123 fixed the `PromptProcessingBatch` prefill path for E2B/E4B per_layer_inputs at mixed-length B≥2. Confirmed working: prefill no longer crashes.

But the **decode batching path** in `mlx_vlm/generate.py` `_next` (around line 3486) still raises the same error when 4 mixed-length text-only requests are batched together for non-speculative generation.

## Repro

mlx-vlm @ `8c9e560b` (latest main, includes #1123). Apple M3 Pro, 36 GB.

```
python -m mlx_vlm.server --model mlx-community/gemma-4-e4b-it-4bit --port 8083 --host 127.0.0.1
```

Then fire 4 concurrent POST /v1/chat/completions with mixed-length prompts (e.g. 21/25/27/28 tokens, max_tokens=200, no draft model). All four return 500.

Server stderr:
```
File "…/mlx_vlm/generate.py", line 3486, in _next
    merged_kwargs[k] = mx.concatenate(vs, axis=0)
ValueError: [concatenate] All the input array dimensions must match exactly except for the concatenation axis. However, the provided shapes are (1,25,42,256), (1,27,42,256), (1,28,42,256), and the concatenation axis is 0.
```

Same error message as #1123's repro, different stack frame. Reproduces on:
- `mlx-community/gemma-4-e4b-it-4bit`
- `mlx-community/gemma-4-e4b-it-8bit`
- `mlx-community/gemma-4-e2b-it-4bit`

Does NOT happen at B=2 (which works after #1123). Triggers at B≥4 mixed-length only.

## Notes

- Speculative-decoding path (`_run_speculative` with `--draft-model`) bypasses this code path entirely and **works** at B=4 mixed-length.
- The fix is presumably symmetric to #1123 — apply the same per-row length-aware slicing to length-aware kwargs in `_next`.
- Possibly the same root cause as #1083 (token loops at B≥4) — the loop bug may have been masking this 500.

Blaizzy:
## Summary

Fixes the mixed-length Gemma 4 continuous batching crash reported in #1124.

The cold prefill path already left-padded `inputs_embeds` to the prompt batch max length, but other sequence-aligned prompt kwargs such as Gemma 4 `per_layer_inputs` were concatenated without matching sequence lengths. Mixed-length B=4 requests could therefore fail with an MLX concatenate shape mismatch before generation.

This change adds scoped helpers for sequence-aligned prompt kwargs and applies the same padding rules when merging cold and APC mixed prompt batches. Non-sequence tensors such as image tensors and scalar-ish per-row metadata are left on their original axes.

## Validation

- `python -m pytest mlx_vlm/tests/test_generate.py -q` -> `59 passed`
- `python -m compileall -q mlx_vlm/generate.py mlx_vlm/tests/test_generate.py`
- Live server repro with `mlx-community/gemma-4-e4b-it-8bit`, 4 concurrent mixed-length `/v1/chat/completions`, `max_tokens=200` -> all 4 returned HTTP 200 with no server traceback

## Notes

This fixes the #1124 HTTP 500 / `[concatenate]` shape mismatch. The separate `batch_generate` B=4 Gemma 4 issue in #1083 also fixed.

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -2015,6 +2015,29 @@ def _prompt_kwarg_row(v: mx.array, row_idx: int, batch_size: int) -> mx.array:
     return v[:1]
 
 
+def _split_prompt_kwargs_per_row(prompt_kwargs: dict, batch_size: int) -> List[dict]:
+    """Normalize batched prompt kwargs into one dict per batch row.
+
+    ``model.get_input_embeddings()`` commonly returns batch-sized tensors
+    (notably ``inputs_embeds``). ``BatchGenerator.insert()`` stores prompt
+    kwargs per sequence, so passing the same batched dict for every row causes
+    the prompt builder to concatenate those batched tensors ``batch_size``
+    times, effectively squaring the batch dimension.
+    """
+    if batch_size <= 1:
+        return [prompt_kwargs or {}]
+
+    rows = [{} for _ in range(batch_size)]
+    for k, v in (prompt_kwargs or {}).items():
+        if isinstance(v, mx.array) and v.ndim > 0 and v.shape[0] >= 1:
+            for i in range(batch_size):
+                rows[i][k] = _prompt_kwarg_row(v, i, batch_size)
+        else:
+            for row in rows:
+                row[k] = v
+    return rows
+
+
 def _is_sequence_aligned_prompt_kwarg(
     key: str, v: mx.array, sequence_length: int
 ) -> bool:
@@ -3853,7 +3876,7 @@ def _generate_batch(
     uids = gen.insert(
         input_ids.tolist(),
         max_tokens,
-        prompt_kwargs=[gen_kwargs] * len(input_ids),
+        prompt_kwargs=_split_prompt_kwargs_per_row(gen_kwargs, batch_size),
         logits_processors=logits_processors,
     )
     results = {uid: [] for uid in uids}
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -1997,6 +1997,46 @@ def _right_pad_prompts(prompts, max_length=None):
     return mx.array([list(p) + [0] * (max_length - len(p)) for p in prompts])
 
 
+_SEQUENCE_ALIGNED_PROMPT_KWARGS = {
+    "attention_mask",
+    "decoder_inputs_embeds",
+    "deepstack_visual_embeds",
+    "visual_pos_masks",
+    "per_layer_inputs",
+    "full_text_row_masked_out_mask",
+    "position_ids",
+    "pos_hw",
+}
+
+
+def _prompt_kwarg_row(v: mx.array, row_idx: int, batch_size: int) -> mx.array:
+    if v.shape[0] == batch_size:
+        return v[row_idx : row_idx + 1]
+    return v[:1]
+
+
+def _is_sequence_aligned_prompt_kwarg(
+    key: str, v: mx.array, sequence_length: int
+) -> bool:
+    return (
+        key in _SEQUENCE_ALIGNED_PROMPT_KWARGS
+        and v.ndim >= 2
+        and v.shape[1] == sequence_length
+    )
+
+
+def _pad_sequence_aligned_prompt_kwarg(
+    v: mx.array, target_length: int, *, left: bool
+) -> mx.array:
+    pad = target_length - v.shape[1]
+    if pad <= 0:
+        return v
+    pad_shape = (v.shape[0], pad) + tuple(v.shape[2:])
+    pad_v = mx.zeros(pad_shape, dtype=v.dtype)
+    parts = [pad_v, v] if left else [v, pad_v]
+    return mx.concatenate(parts, axis=1)
+
+
 def _extend_cache(cache_a, cache_b):
     """Extend cache_a with cache_b along the batch dimension."""
     if not cache_a:
@@ -3142,6 +3182,7 @@ class BatchGenerator:
         # in _apc_extra_hash, never forwarded to the model.
         merged_kwargs: dict = {}
         per_row_keys: dict = {}
+        batch_size = len(prompt_kwargs_list)
         for i, kw in enumerate(prompt_kwargs_list):
             if not kw:
                 continue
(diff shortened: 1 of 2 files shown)


4north-8west:
Hi @Blaizzy — confirming PR #1127 closes both #1124 (mixed-length B≥2 HTTP 500s) and the #1083 surface on E4B same-length B=4. Validated on Apple M3 Pro / 36 GB / mlx-vlm 0.5.0.

**Validation — 0 generation failures in 60 streams**

Two passes, three quantizations, three batch configs each:

| Model | B=2 mixed | B=4 mixed | B=4 same |
|---|---|---|---|
| E4B 4-bit | 2/2 ok | 4/4 ok | 4/4 ok |
| E4B 8-bit | 2/2 ok | 4/4 ok | 4/4 ok |
| E2B 4-bit | 2/2 ok | 4/4 ok | 4/4 ok |

Pass A: thinking on (Gemma 4 default), `max_tokens=600`. Pass B: `chat_template_kwargs.enable_thinking:false`, `max_tokens=200`. Both passes: every row had non-empty `content` or `reasoning`; no row had both empty. Prompts were a mix of short factual and longer conceptual — the conceptual ones spend most of the budget inside `<|channel>thought` (handled cleanly by `_split_thinking`), so a strict `len(content) > 0` assertion will false-alarm on those rows. `len(content) + len(reasoning) > 0` is the right invariant for any future regression test on this path.

Happy to share per-stream JSONs or the harness if useful. LGTM.

Blaizzy:
Thanks @4north-8west 

Yes that's would be useful!

PR #1129, opened 2026-05-06
Fix streamed detokenization for byte-fallback tokens

### LINKED ISSUE #1126: Server: ResponseGenerator token decoding inconsistent with rest of codebase
Last week, I encountered a case of multibyte characters (emoji) being corrupted in streamed server output with Gemma 4. This bug proved difficult to reproduce cleanly — it only affected some models, some emoji, and then disappeared completely after a `uv cache prune`. So potentially even a side effect of changes in my own fork.

But in the process, I confirmed that `ResponseGenerator` decodes tokens differently from every other path in the codebase:

```python
# ResponseGenerator (server.py)
curr = self.tokenizer.decode(info["tokens"])
text = curr[len(info["prev_text"]):]
```

vs. `generate()`, `chat()`, and `stream_generate()`, which all use `processor.detokenizer.add_token()` / `last_segment` / `finalize()` — the `SPMStreamingDetokenizer` API designed for exactly this use case.

The naive incremental decode works as it is, but is out of step with how the rest of the codebase handles streaming output, and relies on `tokenizer.decode()` behavior that isn't explicitly guaranteed for incomplete byte sequences.

**PR [#1076](https://github.com/Blaizzy/mlx-vlm/pull/1076)** isn't fixing a confirmed bug, but still brings `ResponseGenerator` in line by introducing a per-request `SPMStreamingDetokenizer`, and includes a unit test that exercises the byte-fallback path directly.

Blaizzy:
## Summary

Fix server-side streamed detokenization for byte-fallback tokenizers by giving each active `ResponseGenerator` request its own streaming detokenizer instead of reconstructing deltas with cumulative `tokenizer.decode(...)` slices.

This aligns the server batching path with `stream_generate`, flushes pending byte sequences on stop/length, and covers the speculative server path as well.

Fixes https://github.com/Blaizzy/mlx-vlm/issues/1126

## Validation

- `pre-commit run --all-files`
- `pytest -q mlx_vlm/tests/test_server.py::TestResponseGenerator`
- `pytest -q mlx_vlm/tests/test_generate.py`
- `pytest -q mlx_vlm/tests/test_server.py mlx_vlm/tests/test_tokenizer_utils.py`
- `git diff --check`

I also benchmarked the changed detokenization path with `Qwen/Qwen3.5-4B`, `mlx-community/gemma-4-26b-a4b-it-4bit`, and `mlx-community/gemma-4-e4b-it-8bit` tokenizers; the new streaming detokenizer path is faster than cumulative decode/slice while preserving final text output.

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -22,6 +22,7 @@ from transformers import PreTrainedTokenizer
 from . import apc as _apc
 from .models import cache
 from .prompt_utils import apply_chat_template
+from .tokenizer_utils import make_streaming_detokenizer
 from .turboquant import BatchTurboQuantKVCache, TurboQuantKVCache, turboquant_enabled
 from .utils import (
     StoppingCriteria,
@@ -1706,8 +1707,7 @@ def stream_generate(
     total_prompt_tokens = reused_prefix_len + input_ids.size
 
     with wired_limit(model, [generation_stream]):
-        detokenizer = processor.detokenizer
-        detokenizer.reset()
+        detokenizer = make_streaming_detokenizer(processor)
         thinking_criteria = getattr(tokenizer, "thinking_budget_criteria", None)
         exact_checkpoint_len = None
         exact_checkpoint = None
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -703,25 +703,35 @@ class ResponseGenerator:
                 first_bonus = sampler(out.logits[:, -1:]).squeeze(-1)
                 mx.eval(first_bonus, hidden, out.logits)
 
+                finished_uids = set()
+
                 # Send first bonus tokens to clients
                 fb_list = first_bonus.tolist()
                 for j, uid in enumerate(uids):
                     tok = int(fb_list[j])
                     token_lists[uid].append(tok)
-                    text = self._stream_text(stream_infos[uid], tok, None)
+                    is_stop = tok in self.stop_tokens
+                    is_max = len(token_lists[uid]) >= max_tokens_map[uid]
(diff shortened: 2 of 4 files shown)


Edsuns:
This PR introduced a streaming regression: tokens no longer arrive one-by-one. Instead, the client sees a sequence of empty `text` deltas followed by a sudden burst of accumulated text when a word boundary is hit.

**Root cause**: `StreamingDetokenizer` is designed for the `stream_generate` accumulation pattern (process all tokens, detokenizer gradually builds up text). It only flushes `last_segment` on word boundaries (space, newline, etc.). Between boundaries, `last_segment` returns `""`.

In the server streaming loop, this means:
1. Token 1: `detokenizer.add_token(tok1)` → `last_segment` = `""` (subword prefix)
2. Token 2: `detokenizer.add_token(tok2)` → `last_segment` = `""` (still accumulating)
3. Token 3 (word boundary): `detokenizer.add_token(tok3)` → `last_segment` = `"hello world"` (flush)

The client sees: `""`, `""`, `"hello world"` — which is the burst behavior.

**Why `tokenizer.decode([tok])` is correct**:
- It handles byte fallback tokens correctly: `decode([0xE4])` produces the raw byte, which is valid UTF-8 when concatenated with adjacent tokens
- It produces text for every token, giving true per-token streaming
- This is the same approach used by vLLM and TGI for per-token SSE output

The byte fallback issue this PR tried to fix should be handled at the tokenizer level (ensuring `decode([tok])` returns valid partial UTF-8), not by switching to a word-boundary detokenizer that breaks streaming granularity.

Blaizzy:
Thanks @Edsuns 

I will revert it and document it somewhere

PR #1137, opened 2026-05-07
Fix server chat completions text content array handling

spicyneuron:
This fixes `/v1/chat/completions` handling for messages where `content` is an array of text parts. Some harnesses (Claude Code in my case), frequently send user text as multiple content parts and expect the server to flatten them before applying the chat template. Previously, this endpoint kept only the last text part, so earlier parts were silently dropped.

Ex:

```json
[
    {
      "content": "You are Claude Code...",
      "role": "system"
    },
    {
      "content": [
        {
          "text": "<system-reminder>\nAs you answer the user's questions, you can use the following context:\n# claudeMd\nCodebase and user instructions are shown below. ...\n</system-reminder>\n\n",
          "type": "text"
        },
        {
          "text": "Test!",
          "type": "text"
        }
      ],
      "role": "user"
    }
  ]
```

Changes:
- Reuse the existing `extract_text_from_content(...)` helper for chat completion message arrays.
- Preserve existing user image/audio extraction behavior.
- Add a regression test covering multiple `type: "text"` parts.

One note: `mlx-lm` flattens text fragments with an empty-string join. This repo’s existing helper joins text parts with a space, so this PR keeps the local behavior rather than changing broader prompt formatting.

That said, should `mlx-vlm` standardize on the upstream `mlx-lm` empty-string join behavior? Both directions potentially introduce whitespace issues, though these are probably negligible in most cases.

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -47,7 +47,7 @@ from .generate import (
     normalize_resize_shape,
     stream_generate,
 )
-from .prompt_utils import apply_chat_template
+from .prompt_utils import apply_chat_template, extract_text_from_content
 from .sample_utils import top_p_sampling
 from .structured import build_json_schema_logits_processor
 from .tokenizer_utils import make_streaming_detokenizer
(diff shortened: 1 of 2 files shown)


Blaizzy (approved):
LGTM, thanks!

Issue #1138, opened 2026-05-07
Gemma assistant MTP sliding masks should use local rotating-cache offsets


Thump604:
# Gemma assistant MTP sliding masks should use local rotating-cache offsets

## Summary

Gemma assistant MTP drafter masks should compare query positions against the
local K/V positions present in the target rotating sliding-window cache. Using
absolute decode positions directly against a local rotating cache can mask out
the valid window on long prompts.

## Local repro

Local Runtime evidence:

- `/opt/ai-runtime/run/non-live-jobs-ab/20260507T020113Z-gemma4_31b_mtp_assistant_no_think`
- `/opt/ai-runtime/run/non-live-jobs-ab/20260507T015456Z-gemma4_26b_a4b_mtp_assistant_no_think`

Observed before the local fix:

- assistant-MTP on long-prompt workloads could produce repetitive or corrupt
  output patterns
- bounded shorter probes could engage and accept drafts
- the differential narrowed the issue to long-prompt sliding-window behavior
  rather than basic drafter loading

## Exact upstream code path compared

Repository: `Blaizzy/mlx-vlm`

Commit compared: `aec6998d329b7c02e1e51c57cf085c9437e8d97f`

Path/function:

- `mlx_vlm/speculative/drafters/gemma4_assistant/masks.py::make_drafter_masks`

The upstream function sets `kv_valid_len = query_offset` and passes the
absolute `query_offset` into `bidirectional_swa_mask` for sliding-window
layers. The K/V tensors being shared from the target model are already in a
local cache coordinate system.

## Expected behavior

For sliding-window layers, drafter query positions should be mapped onto the
local rotating-cache coordinate system before mask construction. Full-attention
layers can use the valid K/V length directly.

## Minimal patch/design ask

- Map absolute decode positions to `min(query_offset, kv_len)` before building
  sliding-window masks.
- Pass `kv_len` as the valid K/V length for shared K/V states.
- Add focused tests for long-prompt local-window mask construction.

## What this does not claim

- This is not a Jobs workload-quality claim.
- This does not claim assistant-MTP should be enabled by default.
- This does not claim every Gemma assistant-MTP issue is caused by masks.

lucasnewman:
Resolved in https://github.com/Blaizzy/mlx-vlm/pull/1139

PR #1139, opened 2026-05-07
Use local rotating-cache offsets for Gemma assistant MTP masks

Thump604:
# Use local rotating-cache offsets for Gemma assistant MTP masks

Related issue: #1138

## Summary

Updates Gemma assistant MTP mask construction so sliding-window layers compare
queries against the local rotating-cache window shared by the target model.
Absolute decode positions are clamped into the local cache coordinate system
before `bidirectional_swa_mask` is built.

## Local repro

Local Runtime evidence before this fix:

- `/opt/ai-runtime/run/non-live-jobs-ab/20260507T020113Z-gemma4_31b_mtp_assistant_no_think`
- `/opt/ai-runtime/run/non-live-jobs-ab/20260507T015456Z-gemma4_26b_a4b_mtp_assistant_no_think`

Observed behavior:

- long-prompt assistant-MTP requests could produce repetitive or corrupt output
  patterns
- shorter bounded probes could still engage MTP and accept drafts
- the failure pattern was tied to long-prompt sliding-window mask behavior, not
  basic drafter loading

## Exact upstream code path compared

Base commit: `aec6998d329b7c02e1e51c57cf085c9437e8d97f`

Compared path/function:

- `mlx_vlm/speculative/drafters/gemma4_assistant/masks.py::make_drafter_masks`

The previous implementation used absolute `query_offset` as the sliding-window
mask offset while the shared target K/V state is local to the rotating cache.

## What changed

- Added `_local_window_offset`.
- Used the local offset for sliding-window drafter masks.
- Used `kv_len` as the valid K/V length for shared K/V masks.
- Added focused tests for long-prompt local-window behavior.

## Validation

- `git diff --check`
- `python -m py_compile mlx_vlm/speculative/drafters/gemma4_assistant/masks.py mlx_vlm/tests/test_gemma4_assistant_masks_static.py mlx_vlm/tests/test_speculative.py`
- direct focused execution of:
  - `test_local_window_offset_clamps_absolute_decode_position`
  - `test_local_window_offset_clamps_batched_decode_positions`

I did not bypass the local Runtime safety gate to run the full pytest command
while a live LaunchAgent was loaded.

## What this does not claim

- This does not enable assistant-MTP by default.
- This is not a Jobs workload-quality claim.
- This does not claim every Gemma assistant-MTP issue is caused by masks.

### DIFF
--- a/mlx_vlm/speculative/drafters/gemma4_assistant/masks.py
+++ b/mlx_vlm/speculative/drafters/gemma4_assistant/masks.py
@@ -129,14 +129,19 @@ def make_drafter_masks(
     masks = {}
     for layer_type, kv in shared_kv_states.items():
         kv_len = _kv_len(kv)
-        kv_valid_len = query_offset
+        local_valid_len = _local_window_offset(query_offset, kv_len)
         if layer_type == "sliding_attention":
             masks[layer_type] = bidirectional_swa_mask(
-                query_len, query_offset, kv_len, sliding_window, kv_valid_len, dtype
+                query_len,
+                local_valid_len,
+                kv_len,
+                sliding_window,
+                local_valid_len,
+                dtype,
             )
         else:
             masks[layer_type] = bidirectional_full_mask(
-                query_len, kv_len, kv_valid_len, dtype
+                query_len, kv_len, local_valid_len, dtype
             )
     return masks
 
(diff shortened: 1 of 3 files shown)


Blaizzy (approved):
LGTM, thanks!

PR #1141, opened 2026-05-07
Fix top_p_sampling crash on 3D logits in speculative decoding

ivanfioravanti:
## Summary

  Speculative generation crashes with `ValueError: [squeeze] Cannot squeeze axis -1 with size N which is
  not equal to 1` whenever `top_p` sampling is enabled (i.e. `top_p ∈ (0, 1)` and `temperature > 0`).

  ```
  File "/.../mlx_vlm/server.py", line 804, in _run_speculative
      for tok_list, _ in rounds_iter:
  File "/.../mlx_vlm/generate.py", line 725, in _mtp_rounds_batch
      target_tokens = sampler(verify_out.logits)
  File "/.../mlx_vlm/server.py", line 456, in sampler
      return top_p_sampling(logprobs, args.top_p, args.temperature)
  File "/.../mlx_vlm/sample_utils.py", line 41, in top_p_sampling
      token = mx.take_along_axis(sorted_indices, sampled_pos[:, None], axis=-1).squeeze(-1)
  ValueError: [squeeze] Cannot squeeze axis -1 with size 4 which is not equal to 1.
  ```

  ## Root cause

  `top_p_sampling` was written assuming logits are 1D `[V]` or 2D `[B, V]` and used `sampled_pos[:, None]`
   to add a trailing axis for `take_along_axis`. In MTP speculative decoding (`mlx_vlm/generate.py:725`),
  the verify step runs the LM over the initial token concatenated with `bs - 1` draft tokens, so
  `verify_out.logits` is 3D `[B_active, bs, V]`.

  With 3D input:
  - `sampled_pos` ends up as `[B, bs]`,
  - `sampled_pos[:, None]` becomes `[B, 1, bs]`,
  - `take_along_axis(sorted_indices=[B, bs, V], indices=[B, 1, bs], axis=-1)` broadcasts to `[B, bs, bs]`,
  - `.squeeze(-1)` rejects the size-`bs` last axis.

  The `size 4` in the error matches the speculative block size (`bs=4`), confirming the path.

  The greedy / pure-categorical path in `Server._make_sampler` doesn't hit this because it skips
  `top_p_sampling` entirely.

  ## Fix

  One-line change: use `sampled_pos[..., None]` instead of `sampled_pos[:, None]` so the function works
  for any leading rank (`[V]`, `[B, V]`, `[B, T, V]`, …). Updated the docstring to reflect the broader
  contract.

  ## Tests

  Added two regression tests in `mlx_vlm/tests/test_sample_utils.py`:

[...]

PR #1142, opened 2026-05-07
gemma4 mtp: demote to AR on drafter mismatch + propagate speculative errors

ivanfioravanti:
## Summary

  Three connected improvements to Gemma 4 MTP speculative decoding in `mlx_vlm.server`:

  1. **Drafter ↔ target compat probe at startup** with graceful demotion to autoregressive (AR) generation on mismatch. Previously, an incompatible drafter
  (e.g. the `26B-A4B-it-assistant` against a `gemma-4-31B-it` target) crashed the GPU thread on the first request with a deep `pre_projection` matmul error.
   The operator now gets a clear startup warning and the server keeps serving requests via AR.
  2. **Bind-time compat check** in `Gemma4AssistantDraftModel.bind()` — the source of truth used by the startup probe. Reusable for the planned per-request
  drafter API (where a client explicitly opts into a drafter and *should* see the error rather than have it silently swapped).
  3. **Propagate `_run_speculative` exceptions to client queues** so any speculative-thread failure (the new compat error, OOM, kernel crash, …) reaches the
   client as an HTTP/SSE error instead of leaving the FastAPI handler awaiting forever.

  ## The original failures

  ```
  [matmul] Last dimension of first input with shape (1,1,10752) must match
  second to last dimension of second input with shape (5632,1024).
    ...pre_projection(inputs_embeds)
  ```

  `10752 = 2 × 5376` (31B-it target hidden), `5632 = 2 × 2816` (26B-A4B-it-assistant `backbone_hidden_size`). Drafter weights are physically wrong-shape —
  runtime resizing is impossible.

  The exception then hung the client forever. `_run_speculative` caught the error, logged it via `print()`, and silently swallowed it. The per-request token
   queue was never signalled, so the FastAPI streaming handler awaited tokens that would never arrive.

  ## Changes

  ### Commit 1 — `bind()` compat check (source of truth)

  `mlx_vlm/speculative/drafters/gemma4_assistant/gemma4_assistant.py`

  `Gemma4AssistantDraftModel.bind()` reads the target's `hidden_size` (via the same `target.language_model.config` / `target.config` path it already uses
  for `layer_types`) and raises `ValueError` with the full pairing table when it doesn't match `self.config.backbone_hidden_size`. Cheap, idempotent — used
  by the next two commits.

  ### Commit 2 — error propagation

  `mlx_vlm/server.py`

  - New `_broadcast_exception_to_active(rqueues, finished_uids, exc)` helper pushes `(exc, None)` to each unfinished per-request queue (mirrors the
  non-speculative path at `server.py:625-636`).
  - Per-iteration state in `_run_speculative` is initialized **outside** the `try` so the except handler can see partial state when the exception fires
  during phase-1 request collection.
  - The except handler folds in `pending` items whose rqueue wasn't yet registered (covers the gap between `requests.get` and `rqueues[uid] = rqueue`).
  - `print(...) + traceback.print_exc()` replaced with `logger.exception(...)`.
  - After broadcasting, GPU state is reset (`mx.clear_cache` + `gc.collect`).

  The streaming iterator at `server.py:417-419` already handles `isinstance(item, Exception): raise item` — it just never received one in the speculative
  path before this change.

  ### Commit 3 — startup probe + AR demotion

  `mlx_vlm/server.py`

  - New `_probe_drafter_compat(drafter, target_model) -> Optional[Exception]` helper — calls `drafter.bind(target_model)` and returns the `ValueError` (or
  `None` on success). Idempotent: `bind()` is re-called per-request anyway, so the probe doesn't disturb runtime state.
  - `_initialize_model` runs the probe right after `load_drafter` succeeds. On mismatch, logs a structured `logger.warning` (with the full reason) plus a
  one-line `print` for the operator watching stdout, then sets `draft_model = None` so the existing thread router (`server.py:570`) takes the AR path.

  `bind()` still raises on mismatch — the right behavior for the future per-request drafter API, where a client explicitly opts into a drafter and should
  see the error rather than have it silently swapped.

  ## Out of scope (separate follow-up PR)

  **Per-request drafter selection.** Add `draft_model` / `draft_kind` fields to `VLMRequest`, an in-memory drafter cache, and per-request routing between AR
   and speculative threads. Real architectural change; gets its own PR.

  ## Tests

  `mlx_vlm/tests/test_speculative.py` — 7 new tests, 22 total (15 existing pass unchanged):

  **Bind guard** (3): mismatch raises with both numbers + pairing table; matched pair doesn't raise; unwrapped `LanguageModel` target works.

  **Error broadcast** (2): `(exc, None)` delivered per active queue with finished uids skipped; broken queue's `.put()` exception doesn't block other
  clients.

  **Probe helper** (2): returns the `ValueError` on mismatch; returns `None` on a matched pair.

  Tests stub the drafter via `Gemma4AssistantDraftModel.__new__(...)` — fast, hermetic, no GPU.

  ## End-to-end verification (against a live Gemma 4 server)

  | Scenario | Server log | Client |
  |---|---|---|
  | Matched pair (`gemma-4-26B-A4B-it` ↔ `gemma-4-26B-A4B-it-assistant`) | `Drafter ready — speculative decoding enabled.` then `[MTP] batch=1 tokens=…
  accept=… rounds=…` | streams normally with speculative speedup |
  | Mismatched pair (`gemma-4-31B-it` ↔ `gemma-4-26B-A4B-it-assistant`) | `WARNING: drafter ↔ target incompatibility — speculative decoding DISABLED.` +
  full pairing table in log | streams normally via AR; `lbench` runs to completion |
  | Speculative-thread exception | `ValueError: [squeeze] Cannot squeeze axis -1 with size 4 …` | receives SSE error event in **0.76 s** (vs. indefinite
  hang on `main`) |

  The third row was verified before this branch was rebased onto `main` by leveraging the latent top_p `[squeeze]` bug from #1141. That bug is now fixed, so
   the same e2e isn't reproducible on the rebased branch organically — the underlying error-propagation code is unchanged and is covered by the
  broadcast-helper unit tests.

  ## Test plan

  - [x] `pytest mlx_vlm/tests/test_speculative.py` — 22 passed
  - [x] `pytest mlx_vlm/tests/test_server.py` — 58 passed
  - [x] `pre-commit run --files <changed>` — clean
  - [x] **E2E:** matched pair streams via speculative
  - [x] **E2E:** mismatched pair demotes to AR with WARNING; server keeps serving
  - [x] **E2E:** speculative-thread exception reaches the client (verified pre-rebase against the now-fixed top_p bug)

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -91,6 +91,26 @@ def _get_draft_block_size_from_env():
     return int(draft_block_size_str) if draft_block_size_str else None
 
 
+def _broadcast_exception_to_active(
+    rqueues: dict, finished_uids: set, exc: BaseException
+) -> None:
+    """Push (exc, None) onto each unfinished per-request queue.
+
+    The streaming iterator at ``ResponseGenerator.token_iterator`` checks
+    ``isinstance(item, Exception)`` and re-raises, which surfaces the error
+    to the client as an HTTP/SSE error instead of an indefinite hang.
+    Mirrors the non-speculative generation thread's recovery pattern.
+    """
+    for uid, rqueue in list(rqueues.items()):
+        if uid in finished_uids:
+            continue
+        try:
+            rqueue.put(exc)
+            rqueue.put(None)
+        except Exception:
+            pass
+
+
 def get_prefill_step_size():
     return int(os.environ.get("PREFILL_STEP_SIZE", DEFAULT_PREFILL_STEP_SIZE))
 
@@ -659,9 +679,15 @@ class ResponseGenerator:
         draft_block_size = _get_draft_block_size_from_env()
 
         while not self._stop:
+            # Per-iteration state — initialized outside the try so the
+            # except handler can broadcast errors back to any client whose
+            # request was already pulled off the request queue.
+            pending: list = []
+            uids: list = []
+            rqueues: dict = {}
+            finished_uids: set = set()
             try:
                 # --- Phase 1: collect pending requests ---
-                pending = []
                 timeout = 0.1
                 try:
                     item = self.requests.get(timeout=timeout)
@@ -683,8 +709,8 @@ class ResponseGenerator:
                     continue
 
                 # --- Phase 2: prefill new batch ---
-                uids = []
-                rqueues = {}
+                # uids, rqueues, finished_uids are pre-initialized above the
+                # try so error broadcast can see partial state.
                 token_lists = {}
                 stream_infos = {}
                 max_tokens_map = {}
@@ -735,7 +761,7 @@ class ResponseGenerator:
                 first_bonus = sampler(out.logits[:, -1:]).squeeze(-1)
                 mx.eval(first_bonus, hidden, out.logits)
 
-                finished_uids = set()
+                # finished_uids pre-initialized at top of loop iteration.
 
                 # Send first bonus tokens to clients
                 fb_list = first_bonus.tolist()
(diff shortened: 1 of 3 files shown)


ivanfioravanti:
@Blaizzy 

- check for drafter compatibility is pretty basic for now: drafter.config.backbone_hidden_size == target.text_config.hidden_size  
- check Out of Scope that is the other idea we were talking about.

PR #1143, opened 2026-05-07
Fix stale Qwen MRoPE rope_deltas in decode

neilmehta24:
In continuous batching usage, Qwen decode may read stale `rope_deltas` for `qwen2_5_vl`, `qwen3_5`, and `qwen2_vl`. 

`GenerationBatch._step()` forwards each row's correct `rope_deltas` as a kwarg, but the model impl ignores the kwarg and reads `self._rope_deltas`, which a newer in-flight prefill could have just overwritten.

For example:
1. Start request `A` with MRoPE `delta A`, and then decode for many tokens
2. While `A` is decoding, send request `B` with a different image prompt
3. `B`'s prefill mutates `language_model._rope_deltas` to `delta B`
4. `A`'s next decode step uses MRoPE `delta B`

To fix, consistently prefer the caller-supplied `rope_deltas`

### DIFF
--- a/mlx_vlm/models/qwen2_5_vl/language.py
+++ b/mlx_vlm/models/qwen2_5_vl/language.py
@@ -504,19 +504,18 @@ class LanguageModel(nn.Module):
                     self._position_ids = position_ids
             else:
                 batch_size, seq_length = inputs.shape
+                rope_deltas_src = (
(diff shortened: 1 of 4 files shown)


Blaizzy (approved):
LGTM, thanks!

PR #1147, opened 2026-05-07
Fix streamed byte-fallback token handling in server batching

Blaizzy:
## Summary

Fix server-side streaming so batched responses keep token-level text deltas while still buffering incomplete UTF-8 byte-fallback sequences until they form valid characters.

Addresses https://github.com/Blaizzy/mlx-vlm/pull/1129#issuecomment-4397556225

This change:
- introduces a server token streamer in `tokenizer_utils`
- updates the continuous batching and speculative server paths to use it
- preserves immediate streaming for normal SPM/BPE text tokens
- flushes trailing incomplete UTF-8 correctly at request finalization
- adds regression coverage for subword streaming, repeated emoji/text interleaving, finalize behavior, and 8-request batched streaming

## Before / After

For a streamed response like `hi😀`:

- before: `""`, `""`, `""`, `""`, `"hi😀"`
- after: `"hi"`, `""`, `""`, `""`, `"😀"`

This keeps normal text streaming immediately while still waiting for the final UTF-8 byte before emitting the emoji.

## Validation

- `pytest -q mlx_vlm/tests/test_server.py::TestResponseGenerator::test_run_batches_eight_streaming_requests mlx_vlm/tests/test_server.py::TestResponseGenerator::test_step_streams_spm_subword_tokens_immediately mlx_vlm/tests/test_server.py::TestResponseGenerator::test_server_token_streamer_flushes_incomplete_utf8_on_finalize mlx_vlm/tests/test_server.py::TestResponseGenerator::test_step_streams_multiple_utf8_emojis_with_text_between_them`
- manual validation with `Qwen/Qwen3.5-4B` on the local server using 8 concurrent streaming requests containing repeated emoji-heavy output

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -51,10 +51,7 @@ from .generate import (
 from .prompt_utils import apply_chat_template, extract_text_from_content
 from .sample_utils import top_p_sampling
 from .structured import build_json_schema_logits_processor
-from .tokenizer_utils import (
-    _ServerTokenStreamer,
-    make_streaming_detokenizer,
-)
+from .tokenizer_utils import _ServerTokenStreamer, make_streaming_detokenizer
 from .tool_parsers import _infer_tool_parser_from_processor, load_tool_module
 from .utils import load, prepare_inputs
 from .version import __version__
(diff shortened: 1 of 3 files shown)


lucasnewman (approved):
LGTM

PR #1148, opened 2026-05-07
Add TTS generation support to MiniCPM-o

lucasnewman:
This adds general audio generation support to `mlx_vlm.generate` and wires through speech generation in the existing model.

(This is currently a draft because it needs a minor release of `mlx-audio` for the speech codec.)


lucasnewman:
@Blaizzy This one is ready. This is the first omni model with audio output, so it uses the same `--output-modality` switch as image generation to enable it.

mlx_vlm/generate/dispatch.py
```diff
@@ -1404,13 +1418,205 @@ def generate(
...
+            f"Prompt: {last_response.prompt_tokens} tokens, "
+            f"{last_response.prompt_tps:.3f} tokens-per-sec"
+        )
+        print(
+            f"Generation: {last_response.generation_tokens} tokens, "
+            f"{last_response.generation_tps:.3f} tokens-per-sec"
+        )
+        print(f"Peak memory: {last_response.peak_memory:.3f} GB")
+
+    audio_generation_kwargs = dict(kwargs)
```
Blaizzy: Everything before here seems like normal AR generate loop 

Could we reuse/import the existing AR generate code?

mlx_vlm/generate/dispatch.py
```diff
@@ -1404,13 +1418,205 @@ def generate(
...
+    should_generate_audio = output_modality == "audio"
+    output_audio_path = getattr(args, "output", None) if should_generate_audio else None
+    ref_audio_path = getattr(args, "ref_audio", None)
+    if should_generate_audio and output_audio_path is None:
+        raise ValueError(
+            "--output is required when --output-modality audio is selected"
+        )
+    if should_generate_audio and getattr(args, "chat", False):
+        raise ValueError("--output-modality audio does not support --chat")
+
```
Blaizzy, line 1564: Let's refactor into a cascade conditional (if image, else if audio and else) so it's clear what this section does.

mlx_vlm/generate/dispatch.py
```diff
@@ -1665,14 +1873,28 @@ def main():
...
+            )
+            if args.verbose and output_audio_path:
+                print(f"\nAudio written to {output_audio_path}")
+        else:
+            result = generate(
+                model,
+                processor,
+                prompt,
+                **gen_kwargs,
+            )
```
Blaizzy: Image gen used generate()? Or it had a different path entirely?

Blaizzy (commented):
LGTM, just a couple nits 👌🏽

mlx_vlm/generate/dispatch.py
lucasnewman: It uses `run_image_generation_cli` in the branch higher up, since it needs to handle task selection and other details. Let me see if I can unify the modalities a bit better.

lucasnewman:
@Blaizzy Take a look now -- I was able to share a lot between audio and standard VL generation, and I make the output modality routing more clear.

PR #1149, opened 2026-05-07
Fix APC disk cache directory recovery

### LINKED ISSUE #1146: Critical Bug: APC Cache Fails on Second Run If Cache Folder Deleted (Only Triggered Without Code Changes)
Problem Description
APC (Activation Cache) has a critical, low-level robustness bug. It works on the first run but permanently fails on the second run if the cache directory is deleted.
Steps to Reproduce (100% reproducible)
First run: cache directory is created normally.
Delete the entire APC cache directory (~/.cache/mlx-vlm/apc/...).
Run the server a second time WITHOUT modifying any .py source file.
The system logs: APC disk shard header read failed
APC does NOT recreate the directory and becomes permanently disabled.
Root Cause
The cache directory (os.makedirs) is only created in _init_cache() (first run).
The _load_cache() method does NOT check or create the directory before reading files.
This is a basic missing defensive programming check.
Special Key Discovery
If you modify the .py file (even adding ONE SPACE), Python recompiles the bytecode, refreshes __pycache__, and the bug disappears.
This means developers will almost never catch this bug during development.
Only end users who run without changing code will trigger it.
Expected Behavior
On every run, the system must check if the cache directory exists.
If missing, it should recreate it automatically.
APC must recover without manual intervention.
Fix (One Line)
Add this line at the beginning of _load_cache():
python
运行
os.makedirs(self.cache_dir, exist_ok=True)
This is an elementary bug that breaks cache reliability for all end users.

Blaizzy:
## Summary
- Recreate the APC disk cache namespace directory when it is deleted externally during runtime.
- Clear stale disk indexes and shard caches when the on-disk cache disappears.
- Drop stale shard index entries for missing/unreadable shards so APC can recover cleanly.
- Add a regression test covering cache directory deletion and disk restore.

Fixes #1146.

## Validation
- `python -m pytest mlx_vlm/tests/test_apc.py -q`
- Server verification with `Qwen/Qwen3.5-4B`, APC disk enabled, separate `pi-session-*` tenants, live deletion of `/tmp/mlx-vlm-apc-server-pi/Qwen_Qwen3.5-4B`, and fresh-process disk restore (`disk_hits=1`, `exact_hits=1`, `matched_tokens=420`).

### DIFF
--- a/mlx_vlm/apc.py
+++ b/mlx_vlm/apc.py
@@ -806,6 +806,42 @@ class DiskBlockStore:
     def _shard_path(self, shard_id: str) -> Path:
         return self.dir / f"{shard_id}{self.SUFFIX}"
 
+    def _drop_index_for_path(self, path: Path) -> None:
+        with self._index_lock:
+            stale = [h for h, (sp, _) in self._index.items() if sp == path]
+            for h in stale:
+                del self._index[h]
+            stale_exact = [h for h, sp in self._exact_index.items() if sp == path]
+            for h in stale_exact:
+                del self._exact_index[h]
+        with self._mmap_cache_lock:
+            self._mmap_cache.pop(path, None)
+        with self._header_cache_lock:
+            self._header_cache.pop(path, None)
+
+    def _clear_disk_state_after_external_delete(self) -> None:
+        with self._index_lock:
+            self._index.clear()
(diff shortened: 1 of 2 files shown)


lucasnewman (approved):
LGTM

PR #1151, opened 2026-05-07
server: populate cached_tokens and tps in usage envelope

ZeroErrors:
Two `usage` fields were broken or missing. Fixing the parts that have an obvious source.

- `cached_tokens`: 9719c63 added `UsageStats.prompt_tokens_details.cached_tokens` to match OpenAI's response shape but never populated it. APC already records prefix hits per row in `apc_meta[i]["prefix_len"]`. Surface that on `PromptProgress`, carry through `StreamingToken`, write it onto the final `UsageStats` / `OpenAIUsage.input_tokens_details`.

- Streaming chat chunks were building `usage` from a partial dict (`prompt_tokens` / `completion_tokens` / `total_tokens` only). Pydantic coerced it into `UsageStats` with `prompt_tps`, `peak_memory`, and `prompt_tokens_details` filled from defaults — clients saw the fields but they were always 0. Pass a full `UsageStats` so chunks report the real values that 23ea54d already wired at the server level.

`generation_tps` is intentionally not addressed here: see #1113. The continuous-batching path doesn't have a per-uid generation timing source (only batch-aggregate via `batch_gen.stats()`, which the maintainer correctly flagged as inaccurate), and the proper fix is best landed alongside that PR's per-request approach. Calling it out so this PR isn't claiming a fix it doesn't deliver.

## Test plan

- 8 unit tests in `mlx_vlm/tests/test_usage_envelope.py` for the dataclass plumbing and envelope serialisation.
- Existing suite green (`test_models.py` / `test_smoke.py` excluded they need a real model / extra deps and are unrelated).
- Wants a real-model smoke before merge:
  - `/chat/completions` stream + non-stream with `APC_ENABLED=1` and a repeated prefix → `cached_tokens` non-zero on the second hit.
  - `/responses` stream + non-stream → `input_tokens_details.cached_tokens` populated.
  - Streaming chat chunks → `prompt_tps` and `peak_memory` non-zero (previously always 0 due to dict coercion).

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -2262,6 +2262,9 @@ class PromptProgress:
     prompt_tokens: int
     prompt_tps: float = 0.0
     prompt_time: float = 0.0
+    # Number of leading prompt tokens served from a prefix cache (APC).
+    # Reported back as ``usage.prompt_tokens_details.cached_tokens``.
+    cached_tokens: int = 0
 
 
 class GenerationBatch:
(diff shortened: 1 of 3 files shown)


lucasnewman:
@ZeroErrors Can you please run the formatter with `pre-commit run --all`?

mlx_vlm/generate.py
Blaizzy, line 1: Please remove the comments :)

mlx_vlm/server.py
```diff
@@ -1754,12 +1768,22 @@ class OpenAIRequest(FlexibleBaseModel):
     )
 
 
+class InputTokensDetails(BaseModel):
```
Blaizzy, line 1771: This is not necessary

mlx_vlm/server.py
```diff
@@ -3137,6 +3186,9 @@ def _blocking_generate():
                             gt += 1
                             tt.append(time.perf_counter())
                             ptps = getattr(token, "prompt_tps", ptps)
+                            tok_cached = int(getattr(token, "cached_tokens", 0) or 0)
+                            if tok_cached:
```
Blaizzy, line 3190: We don't need this condition if we default to 0

mlx_vlm/tests/test_usage_envelope.py
Blaizzy, line 1: We have existing server tests, please update the test cases there to reflect instead of creating. A new test file.

Blaizzy (commented):
Thanks @ZeroErrors! 

I left a few comments ;)

lucasnewman:
Resolved in https://github.com/Blaizzy/mlx-vlm/pull/1209

PR #1152, opened 2026-05-08
Fix inverted --verbose flag on chat / generate / video_generate CLIs

SuperMarioYL:
## Summary

The three CLI entry points (`mlx_vlm.chat`, `mlx_vlm.generate`, `mlx_vlm.video_generate`) declare `--verbose` with `action="store_false"`, which silently inverts the flag:

```python
parser.add_argument("--verbose", action="store_false", help="Enable verbose output")
```

With `store_false` and no explicit default, argparse sets the default to `True`. Passing `--verbose` then flips it to `False`. So the flag literally **disables** what its help text claims to enable. The newer eval scripts (`evals/mmmu.py`, `evals/mmstar.py`, `evals/ocrbench.py`, `evals/math_vista.py`) all use `store_true`, which is the convention this PR aligns the main entry points with.

## Repro

```console
$ python -m mlx_vlm.chat --verbose
# expected: streaming token-by-token output
# actual:   no streaming, only the final Markdown panel (verbose=False inside)

$ python -m mlx_vlm.generate --verbose --model ... --prompt ...
# expected: detailed/streaming output + summary
# actual:   only the final result text is printed (the `if not args.verbose: print(result.text)` branch)
```

## Fix

Switch to `argparse.BooleanOptionalAction` with `default=True`. This:

* preserves the **existing default behavior** (`verbose=True` when the flag is absent — no UX regression for users who don't pass the flag),
* makes `--verbose` idempotently keep verbose on, matching the help text,
* exposes a clean opt-out via `--no-verbose`.

`BooleanOptionalAction` is available since Python 3.9 and the project already requires `python>=3.10` (`pyproject.toml`), so this is safe.

## Files touched

| File | Change |
|---|---|
| `mlx_vlm/chat.py:194` | flip the inverted flag |
| `mlx_vlm/generate.py:135` | flip the inverted flag |
| `mlx_vlm/video_generate.py:462` | flip the inverted flag |
| `mlx_vlm/tests/test_cli_verbose_flag.py` | new — regression tests |

## Tests

`mlx_vlm/tests/test_cli_verbose_flag.py` has two layers:

1. **Parser-semantic tests** — build a tiny parser with the same `--verbose` definition and assert the resulting `Namespace` for `[]`, `["--verbose"]`, `["--no-verbose"]` on each entry point. Runs in any environment.
2. **Module-inspection tests** — import the real `chat.main` / `generate.parse_arguments` / `video_generate.main` and verify `argparse.BooleanOptionalAction` is in the source and `'"--verbose", action="store_false"'` is not. These skip gracefully when `mlx` isn't importable.

```console
$ python mlx_vlm/tests/test_cli_verbose_flag.py
sss...
Ran 6 tests in 0.001s
OK (skipped=3)   # 3 skipped under no-mlx env; all 6 will run in CI
```

`black --check` and `isort --check --profile=black` pass on the changed files.


Blaizzy (approved):
LGTM, thanks!

Blaizzy (approved):

PR #1155, opened 2026-05-08
Fix Gemma 4 MTP per-row K/V corruption in batched speculative decoding

### LINKED ISSUE #1144: Gemma 4 MTP drafter: corrupt output + ~0.30 acceptance at B≥2 via mlx_vlm.server (gemma-4-26B-A4B-it-bf16)
On `gemma-4-26B-A4B-it-bf16` ↔ `gemma-4-26B-A4B-it-assistant-bf16` via `mlx_vlm.server` with `--draft-kind mtp`, drafted decoding at `B≥2` produces:

1. **Catastrophic acceptance rate** — `~0.30` accepted/round at `block_size=3` (max possible ~2.0), while the same drafter gets `1.26–1.67` at `B=1` on the same server.
2. **Corrupt output** — short-range token loops, e.g. `"...evaporation, which is the sun's the sun shines down on oceans..."` — the README's "byte-identical greedy output" promise is broken whenever `B≥2`.

The README benchmark for this exact pair claims **3.94×** at B=4, accept ~2.6/round (`mlx_vlm/speculative/drafters/gemma4_assistant/README.md`). Observed in our repro: **0.58× (1.7× slower) than baseline** and unusable text.

This is a different failure from #1083 (which was baseline-only B≥4 looping, fixed in #1127). Here it's drafter-engaged at B≥2 producing both numerical and quality regressions.

## Environment

- `mlx-vlm` 0.5.0 at git main, commit `b22280d` ("Flatten multipart chat content", #1137 — latest)
- `mlx` 0.31.2, `mlx-metal` 0.31.2
- Python 3.13.13, macOS Darwin 25.4.0 (Tahoe), M5 Max 128 GB
- All weights local, downloaded fresh

## Repro

**Server (drafter):**

```bash
uv run python -m mlx_vlm.server \
  --host 127.0.0.1 --port 8765 \
  --model mlx-community/gemma-4-26b-a4b-it-bf16 \
  --draft-model mlx-community/gemma-4-26B-A4B-it-assistant-bf16 \
  --draft-kind mtp --draft-block-size 3
```

**Client** (4 concurrent `/v1/chat/completions`, `temperature=0`, `max_tokens=256`, four distinct prompts about water cycle / atom structure / WWII / combustion engine; full script attached below):

```python
from concurrent.futures import ThreadPoolExecutor
import httpx
prompts = [
  "Explain the water cycle in 4 paragraphs...",
  "Describe the structure of an atom...",
  "Walk through the major events of World War II...",
  "Explain how a basic combustion engine works...",
]
def send(p):
    return httpx.post("http://127.0.0.1:8765/v1/chat/completions", json={
        "model": "mlx-community/gemma-4-26b-a4b-it-bf16",
        "messages": [{"role": "user", "content": p}],
        "max_tokens": 256, "temperature": 0.0, "stream": False,
    }, timeout=600).json()
with ThreadPoolExecutor(max_workers=4) as ex:
    results = list(ex.map(send, prompts))
```

### Run 1 — server log

```
[MTP] batch=1 tokens=8   accept=1.67 rounds=3      <- warmup, clean
[MTP] batch=2 tokens=512 accept=0.29 rounds=394    <- corrupt
[MTP] batch=2 tokens=512 accept=0.33 rounds=383    <- corrupt
```

Wall: 25.5s, aggregate **40.1 tok/s**.

### Run 2 — server log

```
[MTP] batch=1 tokens=8   accept=1.67 rounds=3      <- warmup, clean
[MTP] batch=1 tokens=256 accept=1.26 rounds=113    <- one solo request, clean
[MTP] batch=3 tokens=768 accept=0.30 rounds=589    <- corrupt
```

Wall: 24.2s, aggregate **42.3 tok/s**.

The B=1 solo request inside Run 2 (`accept=1.26`) is consistent with the maintainer's expected MTP behavior — output is fine, modest accept rate. The B=3 batched request right after, with the same drafter, the same target, the same server process, collapses to `accept=0.30` and produces token loops.

### Baseline (no drafter), same server, same client

```bash
uv run python -m mlx_vlm.server --host 127.0.0.1 --port 8765 \
  --model mlx-community/gemma-4-26b-a4b-it-bf16
```

Wall: 14.9s, aggregate **68.6 tok/s**, all 4 in one B=4 batch, output clean.

## Output samples (first 120 chars, prompt #0 = water cycle)

- **Baseline B=4** (clean): `The water cycle begins with **evaporation**, which is the process of water turning from a liquid into an invisible gas c...`
- **Drafter B=2 (Run 1)** (corrupt): `The water cycle begins with **evaporation**, which is the sun's the sun shines on oceans, lakes, and rivers, it warms up...`
- **Drafter B=3 (Run 2)** (corrupt): `The water cycle begins with **evaporation**, which is the sun's the sun shines down on oceans, lakes, and rivers, it war...`

Note `"the sun's the sun shines"` — short-range token loop, identical pattern across both runs even though the Run 1 batch was 2 prompts vs Run 2 batch was 3 prompts. So the loop pattern is reproducible at the drafter level, not a sampling artifact.

## Aside

`_run_speculative` collects pending requests with a 100ms `Queue.get` timeout, then `get_nowait()`-drains the rest. Four concurrent client threads with their own per-thread connect overhead don't all land in the same 100ms window, so the server tends to form `B=2`-ish batches even when 4 requests are nominally concurrent. That's a separate concern from the correctness bug above (which reproduces at any `B≥2`), but worth flagging because it makes it harder for users to actually engage the B=4 path the README's 3.94× number is measured at.

## Suspected area

PR #1139 (#1138) fixed sliding-window mask offsets for the drafter, but the fix path was tested on long-prompt single-request and explicitly noted in its body: *"This does not claim every Gemma assistant-MTP issue is caused by masks."* The B≥2 corruption + acceptance collapse pattern looks like the shared K/V state from `out.shared_kv_states` in `server.py:_run_speculative` is being snapshotted in a way that's correct for B=1 but wrong (per-row mismatch?) when the prefill batch contains multiple rows with different prompt lengths.

Happy to bisect against earlier commits or run additional probes if useful.

jeremi:
Fixes #1144.

## Summary

In batched MTP speculative decoding for Gemma 4 (`_mtp_rounds_batch` with `B≥2`, mixed-length prompts), the target's `BatchKVCache.update_and_fetch` advances `offset` uniformly across all rows by the full draft block size. When rows accept different numbers of draft tokens, under-accepting rows are left with `(max_a − a_i)` slots of garbage K/V (rejected drafts) in their tail, and the next verify forward consumes those wrong slots. Acceptance collapses (~1.50 → ~0.47) and outputs corrupt with short-range token loops, exactly matching #1144.

## Root cause

`_mtp_rounds_batch` (in `mlx_vlm/generate.py`) calls `lm.rollback_speculative_cache` after each verify round to rewind the target cache. The pre-fix gemma4 hook only blanked the per-row tail K/V, but did **not** adjust per-row `offset` / `left_padding`, so:

- the cache's `_idx` and uniform `offset += block_size` left a per-row mismatch — under-accepting rows still saw their next forward as if they had advanced by the full block;
- the next attention pass ran with wrong RoPE positions (`+ block_size` instead of `+ a_i + 1`);
- the drafter's `set_shared_kv` snapshot, normalized by `dynamic_roll(-left_padding)`, had no per-row alignment information to undo the mismatch.

The result is the symptom report in #1144: at `B=3`, mean accepted/round drops to ~0.30 with the same drafter that gets ~1.5 at `B=1`, and outputs degenerate to "the sun's the sun shines down…"-style loops.

This is _not_ a sliding-mask issue (#1138/#1139 already addressed those); the corruption is in the post-verify cache layout, not the prefill snapshot.

## Fix

**`mlx_vlm/models/gemma4/language.py::rollback_speculative_cache`** — replace tail-zero with a per-row cyclic shift:

After the uniform `c.trim(bs − max_a − 1)`, when any row accepted fewer than `max_a` (per-row variance):

1. If `BatchRotatingKVCache`, call `c._temporal_order()` first so the buffer is unrolled.
2. `dynamic_roll(c.keys, row_padding[:, None], axis=2)` and same for `c.values`, where `row_padding[i] = max_a − a_i`.
3. `c.offset = c.offset − row_padding`
4. `c.left_padding = c.left_padding + row_padding`

Depending on the cache shape, the rejected K/V either lands past `_idx` in unused storage (`BatchKVCache`, step-padded — overwritten on the next update) or wraps modulo `shape[2]` into the head slots (`BatchRotatingKVCache` pre-rotation). Either way, bumping `left_padding[i]` by the same shift masks the new head slots from attention, and dropping `offset[i]` leaves the per-row logical length at `old_offset[i] + a + 1`.

We do this directly rather than via `c.prepare(lengths=…, right_padding=…)` + `c.finalize()` because `BatchRotatingKVCache.prepare` derives `_lengths = mx.array(lengths) + self.offset`, then `finalize` does `roll = max(0, self.offset − self._lengths)` which is `0` for non-negative `lengths_user`. The `prepare`/`finalize` API is designed for prefill-with-right-padding (called _before_ forward, completed _after_ offset advances), not for after-the-fact rollback. A single post-update `prepare`/`finalize` is a no-op for sliding-attention layers, which is half the Gemma 4 layers.

**`mlx_vlm/generate.py::_mtp_rounds_batch`** — three companion edits:

1. Snapshot `pre_rollback_left_padding = _batch_cache_left_padding(prompt_cache)` **before** calling `rollback_speculative_cache`. The drafter's `set_shared_kv(...)` normalizes the verify snapshot via `dynamic_roll(-left_padding)`, and the snapshot was captured pre-rollback, so it must be normalized with the pre-rollback `left_padding`. Using post-rollback `left_padding` would over-roll the snapshot for under-accepting rows.
2. Initialize `snap_left_padding = pre_rollback_left_padding` after the next-shared-kv block; in the continuous-batching `c.filter(keep_mx)` step, also filter `snap_left_padding`.
3. Widen the rollback gate from `if max_a < bs − 1:` to `if max_a < bs − 1 or any(a < max_a for a in accepted_list):` so per-row variance with `max_a == bs − 1` (e.g. accepts `[3, 1, 2, 3]` at `bs=4`) still triggers per-row alignment instead of being skipped.

## Tests

Three new unit tests in `mlx_vlm/tests/test_speculative.py`:

- `test_gemma4_rollback_per_row_variance_rolls_kv_and_adjusts_offsets` — `BatchKVCache`, `bs=4`, accepts `[3, 0, 1]` (no uniform trim, only the per-row path). Asserts `offset == [9, 6, 7]`, `left_padding == [0, 3, 2]`, and that each row's live K-slice contains its prefill prefix plus exactly `a_i + 1` verify tokens, with zeros from the step-padded buffer tail rotated into the (masked) left-padding head.
- `test_gemma4_rollback_per_row_variance_on_rotating_cache` — same scenario on `BatchRotatingKVCache` (used for sliding-attention layers, half of Gemma 4). Asserts the rejected K/V wraps modulo `shape[2]` into the head slots and is correctly masked.
- `test_gemma4_rollback_uniform_no_variance_only_trims` — `bs=4`, accepts `[1, 1, 1]`. Asserts that with no per-row variance the rollback degenerates to a uniform trim and leaves `left_padding` untouched.

```
PYTHONPATH=. pytest mlx_vlm/tests/test_speculative.py -q
# 19 passed
```

Adjacent suites also pass:

```
PYTHONPATH=. pytest mlx_vlm/tests/test_speculative.py mlx_vlm/tests/test_server.py \
                     mlx_vlm/tests/test_gemma4_assistant_masks_static.py \
                     mlx_vlm/tests/test_generate.py -q
# 148 passed
```

`pre-commit run --all-files` clean (black, isort, autoflake).

## Validation against #1144's repro

Same 4-prompt concurrent client, `gemma-4-26b-a4b-it-bf16` ↔ `gemma-4-26B-A4B-it-assistant-bf16`, `--draft-kind mtp --draft-block-size 4` on M5 Max:

| | mean accepted/round | output |
|---|---|---|
| pre-fix `B=3` (#1144) | ~0.30 | corrupt (short-range loops) |
| this PR `B=3` | ~1.40 | coherent, grammatical |
| this PR `B=1` (reference) | ~1.51 | coherent |

That's ~92% of the B=1 acceptance rate at B=3, vs the pre-fix collapse to ~20%. The remaining gap to perfect bit-exactness with B=1 is **not** a logic bug: batched matmul has a different floating-point summation order than per-row matmul, so close logits can flip on argmax. Outputs diverge sentence-by-sentence rather than degenerate.

## Test plan

- [x] `pytest mlx_vlm/tests/test_speculative.py` — 19 passed, including 3 new gemma4 rollback tests
- [x] `pytest mlx_vlm/tests/test_server.py mlx_vlm/tests/test_generate.py mlx_vlm/tests/test_gemma4_assistant_masks_static.py` — 129 passed
- [x] `pre-commit run --all-files` — clean
- [x] End-to-end #1144 repro shows acceptance and output recovery

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -854,10 +854,7 @@ def _mtp_rounds_batch(
                     K_next, V_next = next_shared_kv[k]
                     next_shared_kv[k] = (K_next[keep_mx], V_next[keep_mx])
                 if snap_left_padding is not None:
-                    if isinstance(snap_left_padding, mx.array):
-                        snap_left_padding = snap_left_padding[keep_mx]
-                    else:
-                        snap_left_padding = [snap_left_padding[s] for s in keep_slots]
+                    snap_left_padding = snap_left_padding[keep_mx]
                 active_idx = [active_idx[j] for j in keep_slots]
 
         # Re-bind drafter with new shared_kv and per-row positions.
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -772,9 +772,21 @@ def _mtp_rounds_batch(
                 b[orig] = new_tokens_list[j][-1]
             positions[orig] = positions[orig] + accepted_list[j] + 1
 
-        # Rollback target cache (uniform trim by ``bs - max_a - 1`` plus
-        # per-row tail-zero on rows that accepted less).
-        if max_a < bs - 1:
+        # Rollback target cache. Two cases:
+        #   (a) max_a < bs - 1: at least one row accepted fewer than the
+        #       full draft block, so the uniform tail of the cache must be
+        #       trimmed.
+        #   (b) per-row variance (some a_i < max_a): rows accepted
+        #       different counts, so each under-accepting row must be
+        #       cyclically shifted to move its garbage K/V slots out of
+        #       the live region (handled in ``rollback_speculative_cache``
+        #       via per-row ``dynamic_roll`` + per-row ``offset`` /
+        #       ``left_padding`` adjustment).
+        # Snapshot left_padding BEFORE rollback so the drafter's
+        # ``set_shared_kv`` normalizes the (pre-rollback) verify snapshot
+        # with the matching layout.
+        pre_rollback_left_padding = _batch_cache_left_padding(prompt_cache)
+        if max_a < bs - 1 or any(a < max_a for a in accepted_list):
             with mx.stream(generation_stream):
                 lm.rollback_speculative_cache(prompt_cache, None, accepted_arr, bs)
 
@@ -816,6 +828,13 @@ def _mtp_rounds_batch(
                 V_next = V_next * keep_f
             next_shared_kv[k] = (K_next, V_next)
 
+        # ``snap_left_padding`` is the left_padding to use when normalizing
+        # ``next_shared_kv`` inside ``set_shared_kv``. It must match the
+        # layout of the (pre-rollback) verify snapshot, not the
+        # post-rollback target cache (rollback bumps left_padding for
+        # under-accepting rows, which would over-roll the snapshot).
+        snap_left_padding = pre_rollback_left_padding
+
         # Continuous batching: filter finished sequences. Only safe when
         # the caches expose a .filter() method (e.g. BatchKVCache); the
         # plain KVCache / RotatingKVCache do not, so we keep all rows
@@ -834,6 +853,11 @@ def _mtp_rounds_batch(
                 for k in next_shared_kv:
                     K_next, V_next = next_shared_kv[k]
                     next_shared_kv[k] = (K_next[keep_mx], V_next[keep_mx])
+                if snap_left_padding is not None:
+                    if isinstance(snap_left_padding, mx.array):
+                        snap_left_padding = snap_left_padding[keep_mx]
+                    else:
+                        snap_left_padding = [snap_left_padding[s] for s in keep_slots]
                 active_idx = [active_idx[j] for j in keep_slots]
 
         # Re-bind drafter with new shared_kv and per-row positions.
@@ -846,7 +870,7 @@ def _mtp_rounds_batch(
             next_shared_kv,
             kv_offset=new_kv_offset,
             position=mx.array(positions_active),
-            left_padding=_batch_cache_left_padding(prompt_cache),
+            left_padding=snap_left_padding,
         )
 
         if sum(emitted) % 256 == 0:
--- a/mlx_vlm/models/gemma4/language.py
+++ b/mlx_vlm/models/gemma4/language.py
@@ -4,6 +4,7 @@ from typing import Any, List, Optional
 import mlx.core as mx
 import mlx.nn as nn
 from mlx.nn import RMSNorm
+from mlx_lm.models.cache import dynamic_roll
 
 from ..base import (
     LanguageModelOutput,
@@ -621,8 +622,6 @@ class LanguageModel(nn.Module):
         advance. ``gdn_states`` is accepted (and ignored) for API parity
         with qwen3_5's hook.
         """
-        from mlx_lm.models.cache import dynamic_roll
-
         del gdn_states  # API-parity placeholder; Gemma 4 has no SSM/GDN state.
         if isinstance(accepted, int):
             accepted = mx.array([accepted])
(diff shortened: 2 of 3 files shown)


Blaizzy:
Closing in favour of #1166

PR #1157, opened 2026-05-09
Fix Qwen3 VL vision config deserialization (patch_size ignored)

lyonsno:
## Summary

- `ModelConfig.from_dict()` in the Qwen3 VL-family config modules did not deserialize nested `vision_config` dicts into their dataclass types before constructing `ModelConfig`
- This caused `VisionConfig` defaults to silently override values from `config.json` on direct `ModelConfig.from_dict()` construction paths
- Critical for `patch_size`: `Qwen3VLVisionConfig` defaults `patch_size=14`, but Qwen3.6-VL models use `patch_size=16` (confirmed from actual weight shapes: `(1152, 2, 16, 16, 3)`)
- Result: the vision encoder can crash on image inputs with a reshape error when the processor prepares 16x16 patch input but the model config still uses 14x14 patches

## Root cause

The image processor can correctly use `patch_size=16` to compute pixel values while direct model config construction can leave `vision_config` as a raw dict or fall back to the dataclass default. In the failing path, `PatchEmbed` receives `patch_size=14`, so its reshape expects 14x14 patches while receiving 16x16-sized input.

## Fix

Explicitly deserialize nested `vision_config` dicts in `from_dict()` before passing them to the `ModelConfig` constructor. Full `text_config` dicts are also deserialized while partial text-config dicts remain supported for existing callers/tests. Applied to `qwen3_vl`, `qwen3_vl_moe`, `qwen3_5`, and `qwen3_5_moe`.

## Test plan

- [x] Verified `ModelConfig.from_dict(config)` now produces `vision_config.patch_size=16` for Qwen3.6-35B-A3B configs
- [x] Added regression coverage for `vision_config.patch_size=16` on `qwen3_5`, `qwen3_5_moe`, and `qwen3_vl_moe` nested config deserialization
- [x] Preserved existing partial `text_config` behavior for qwen3.5 configs
- [x] End-to-end image inference with `stream_generate` on Qwen3.6-35B-A3B-JANGTQ — previously crashed, now works
- [x] Tested with multiple image sizes (864x558 PNG, 840x560 JPEG, 448x448 PNG)

Generated with Claude Code and reviewed/patched with Codex.

### DIFF
--- a/mlx_vlm/models/qwen3_5/config.py
+++ b/mlx_vlm/models/qwen3_5/config.py
@@ -143,6 +143,23 @@ class ModelConfig(BaseModelConfig):
 
     @classmethod
     def from_dict(cls, params):
+        params = dict(params)
+        if isinstance(params.get("vision_config"), dict):
+            params["vision_config"] = VisionConfig(
+                **{
+                    k: v
+                    for k, v in params["vision_config"].items()
+                    if k in inspect.signature(VisionConfig).parameters
+                }
+            )
+        if isinstance(params.get("text_config"), dict):
+            params["text_config"] = TextConfig(
+                **{
+                    k: v
+                    for k, v in params["text_config"].items()
+                    if k in inspect.signature(TextConfig).parameters
+                }
+            )
         return cls(
             **{
                 k: v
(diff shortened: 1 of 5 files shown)


lucasnewman (approved):
Good catch, thanks!

PR #1160, opened 2026-05-09
Fix max kv size enforcement

### LINKED ISSUE #1154: Server --max-kv-size is not enforced as a request budget with quantized KV
## Summary

When the server is launched with both `--max-kv-size` and KV cache quantization (`--kv-bits`, for example TurboQuant 3.5-bit), the configured max KV size is not enforced as a request-level context budget. Direct API callers can submit prompts that exceed the intended runtime budget and only fail later during prefill or Metal allocation pressure.

## What I observed

On an M4 Max / 36 GB machine serving `mlx-community/Qwen3.6-35B-A3B-nvfp4` with:

```bash
--kv-bits 3.5 \
--prefill-step-size 8192 \
--max-kv-size 65536
```

`/health` reported the model native context (`262144`), and direct `/v1/chat/completions` requests could exceed the intended 65k budget. In local testing, adding a request-budget check made an oversized request return a clear HTTP 400 before any GPU work:

```text
Request needs 70076 context tokens (70012 prompt + 64 max generation), but MAX_KV_SIZE is 65536.
```

## Likely cause

`get_max_kv_size(model)` returns `None` when quantized KV is enabled, which is appropriate for cache construction because QuantizedKVCache/TurboQuant do not use the rotating-cache `max_kv_size` path. But that also means the server has no separate request admission guard for the configured budget.

## Expected behavior

`--max-kv-size` should continue to bound the server runtime budget even when it cannot be passed into the quantized cache constructor. A request whose prompt tokens plus requested max generation exceeds the configured limit should fail early with a 400-level error.

## Related PR

I opened a draft fix in #1153.

lucasnewman:
Based on https://github.com/Blaizzy/mlx-vlm/pull/1153, but expanded to cover both streaming an non-streaming responses so we consistently serve an HTTP 400 in all cases.

Resolves https://github.com/Blaizzy/mlx-vlm/issues/1154

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -65,6 +65,10 @@ METRICS_HISTORY_LIMIT = 100
 METRICS_RECENT_LIMIT = 32
 
 
+class PromptTooLongError(ValueError):
+    """Raised when a request exceeds the configured server context budget."""
+
+
 def _get_speculative_rounds_batch(draft_kind: str):
     if draft_kind == "mtp":
         return _mtp_rounds_batch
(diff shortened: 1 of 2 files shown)


Blaizzy (approved):
LGTM 🚀

PR #1161, opened 2026-05-09
Fix streaming thinking mode and tool-call content truncation

jankyjames:
## Summary

Fixes bugs in the streaming chat completions endpoint that caused thinking content to bleed into `content`, and text immediately before tool calls to be silently dropped.

### Bug 1 — `in_thinking` initialised to `False`

When `enable_thinking` is set, the `<think>` tag is injected as a **prompt prefix**, not a generated token. Generation begins already inside the think block, but `in_thinking` started `False` so all thinking content was emitted as regular `content`.

**Fix:** initialise `in_thinking = gen_args.enable_thinking`.

### Bug 2 — Boundary detected by string match on `accumulated`

The original check fired whenever the string `</think>` appeared anywhere in `accumulated` — including when the model writes **about** the tag mid-thinking (e.g. in a code example or explanation). This prematurely exited thinking mode and corrupted the output.

**Fix:** resolve the `</think>` special-token ID once at startup via `convert_tokens_to_ids("</think>")` and detect the boundary by **token ID** instead.

### Bug 3 — BPE decoder embeds `</think>` text in the next token

The BPE incremental decoder suppresses the `</think>` special token (`text=''`) but releases preceding-context bytes on the **next** regular token. That token's `.text` therefore contains the `</think>` string concatenated with the first content text (e.g. `'clearly.\n</think>\n\nThe answer is…'`).

**Fix:** a **multi-token lookahead window** accumulates tokens after the `</think>` token ID is seen. The boundary is confirmed only when the text after the tag starts with `\n\n` (Qwen3 always emits `\n\n` before the response — a single `\n` can appear inside prose or code and is not sufficient). On confirmation, the text before `</think>` is emitted as `delta_reasoning` (thinking tail) and the text after is emitted as `delta_content`. A `</think>` with no following `\n\n` is spurious and the accumulated text is re-emitted as `delta_reasoning`.

An IndexError guard prevents a crash when the window closes with an empty `_combined`.

### Bug 4 — Text before tool calls silently dropped

When the model produced text immediately before a `<tool_call>` marker (e.g. `"Here's my answer: <tool_call>…"`), the BPE decoder flushed both in the same token. `suppress_tool_call_content` detected `tc_start` in `full_output` and returned `None` for the entire delta, discarding the prefix text.

**Fix:** when `tc_start` first appears in `full_output`, compute how much of `delta_content` precedes the marker and emit only that prefix. Subsequent chunks (fully inside the tool call) are suppressed as before.

### Minor: suppress empty SSE chunks

`has_payload` now uses `bool()` rather than `is not None` so empty-string deltas don't produce SSE chunks that clients may interpret as a stalled stream.

## Test plan

- [ ] Basic thinking request with `stream: true` — `delta.reasoning` populated, `delta.content` has clean response with no `</think>` leakage
- [ ] Request where model discusses `</think>` tokens mid-thinking — spurious tag does not exit thinking mode; full reasoning and full response both delivered
- [ ] Request that ends in a tool call preceded by text — prefix text is preserved in `delta.content`
- [ ] Request with `enable_thinking: false` — no change to existing behaviour
- [ ] Non-thinking model (no `</think>` token in vocabulary) — falls back to string match, existing behaviour preserved

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -2810,10 +2810,44 @@ async def chat_completions_endpoint(request: ChatRequest, http_request: Request)
 
                         output_tokens = 0
                         request_id = f"chatcmpl-{uuid.uuid4()}"
-                        # Track thinking state for reasoning/content split
-                        in_thinking = False
+                        # Track thinking state for reasoning/content split.
+                        # Start in thinking mode if enable_thinking is set, since the
+                        # <think> tag is injected as a prompt prefix — generation begins
+                        # already inside the think block, so in_thinking must be True.
+                        in_thinking = gen_args.enable_thinking
                         accumulated = ""
                         full_output = ""  # raw output for tool call parsing
+                        # Resolve the </think> special-token ID once. Detecting the
+                        # boundary by token ID (not string) prevents the model writing
+                        # about </think> mid-thinking from prematurely exiting thinking
+                        # mode (e.g. in code examples or explanations).
+                        _think_end_id = None
+                        try:
+                            _rtok = response_generator.tokenizer
+                            if hasattr(_rtok, "think_end_id"):
+                                _think_end_id = _rtok.think_end_id
+                            elif hasattr(_rtok, "convert_tokens_to_ids"):
+                                _unk = getattr(_rtok, "unk_token_id", None)
+                                _id = _rtok.convert_tokens_to_ids("</think>")
+                                if _id != _unk:
+                                    _think_end_id = _id
+                        except Exception:
+                            pass
+                        # Once the thinking block has closed we never re-enter it.
+                        _thinking_done = not in_thinking
+                        # Multi-token lookahead window for </think> boundary detection.
+                        # The BPE incremental decoder suppresses the </think> special
+                        # token (text='') but releases preceding-context bytes on the
+                        # *next* regular token. That next token's .text therefore
+                        # contains the </think> string followed by the first content
+                        # text. We accumulate tokens until we see \n in the combined
+                        # text: Qwen3 always emits \n\n before the response, so \n
+                        # confirms a real boundary. A </think> written inside a code
+                        # example has no following \n and is treated as spurious.
+                        _THINK_PENDING_MAX = 5   # cap on non-empty tokens before deciding
+                        _THINK_PENDING_CAP = 25  # hard cap including all-empty tokens
+                        _think_end_pending = 0   # 0 = not pending, >0 = window size
+                        _pending_buffer = []
                         # Track tool-call state to suppress markup from content
                         in_tool_call = False
                         tc_start = tool_module.tool_call_start if tool_module else None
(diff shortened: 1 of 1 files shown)


scottslee-orig:
There is a similar streaming bug with Gemma 4 26B-A4B-IT (`mlx-community/gemma-4-26b-a4b-it-4bit`) that the fix in this PR doesn't fully cover. Can file as separate PR if devs prefer.

### Reproducer

Conversation: user asks `"What's 7*8? Use exec_python."`, model issues a tool call `exec_python({"code":"print(7*8)"})`, tool returns `{"stdout":"56\n","exit_code":0}`, model generates a 13-token follow-up.

Token-by-token dump via `mlx_vlm.generate.stream_generate`:

```
T0..T4:  token_ids=[100, 45518, 107, 101, 236832]   text=''
T5:      token_id=808                                text='<|channel>thought\n<channel|>7'
T6:      token_id=236743                             text=' *'
T7:      token_id=236828                             text=''
T8:      token_id=578                                text=' 8'
T9:      token_id=236743                             text=' ='
T10,T11: token_ids=[236810, 236825]                 text=''
T12:     token_id=106                                text=' 56'
```

Note T5: the opener `<|channel>thought`, the closer `<channel|>`, and the first content character `7` all appear in **one** token's `.text`. The BPE decoder bundles the empty-text special tokens with their following regular tokens — similar in spirit to Bug 3, but for Gemma 4 the entire thinking-channel block collapses into a single emission rather than spanning the closer.

Self-contained script (the tool "result" is pre-baked into the messages; no actual `exec_python` runtime needed):

```python
from mlx_vlm.generate import stream_generate

model, processor = load("mlx-community/gemma-4-26b-a4b-it-4bit")
messages = [
    {"role":"user","content":"What's 7*8? Use exec_python."},
    {"role":"assistant","content":"","tool_calls":[
        {"id":"x","type":"function","function":{
            "name":"exec_python",
            "arguments":'{"code":"print(7*8)"}'}}]},
    {"role":"tool","tool_call_id":"x","name":"exec_python",
     "content":'{"stdout":"56\n","exit_code":0}'},
]
tools = [{"type":"function","function":{
    "name":"exec_python","description":"Run code.",
    "parameters":{"type":"object",
                  "properties":{"code":{"type":"string"}},
                  "required":["code"]}}}]
prompt = apply_chat_template(processor, model.config, messages,
                             tools=tools, add_generation_prompt=True)
for i, t in enumerate(stream_generate(model, processor, prompt,
                                      max_tokens=20, temperature=0.0)):
    print(f"T{i}: token_id={t.token} text={t.text!r}")
```

### Gap in PR #1161

In the updated opener branch (paraphrased):

```python
elif not in_thinking and not _thinking_done and (
    "<|channel>thought" in accumulated or "<think>" in accumulated
):
    in_thinking = True
    accumulated = ""  # ← T5's trailing "<channel|>7" is dropped
```

When T5 lands, `accumulated == "<|channel>thought\n<channel|>7"`. The branch resets `accumulated` and flips `in_thinking=True`. So:
- The trailing `7` is silently lost
- T6–T12 (` *`, ` 8`, ` =`, ` 56`) subsequently emit as `delta_reasoning` instead of `delta_content`

The new lookahead window doesn't help here — the closer was inside T5's `.text`, not its own token, so the `token.token == _think_end_id` check doesn't trigger. And the `\n\n` confirmation wouldn't fire either: Gemma 4 doesn't reliably emit `\n\n` between closer and response (here it's just `7`).

### Suggested patch

Extend the opener branch to look for the matching closer in the same accumulated buffer. If present, treat the token as a fully-resolved single-token thinking block — split into reasoning (between opener and closer) and content (after closer), and set `_thinking_done=True` so subsequent tokens go to `content`:

```python
elif not in_thinking and not _thinking_done and (
    "<|channel>thought" in accumulated or "<think>" in accumulated
):
    if "<|channel>thought" in accumulated:
        _open, _close = "<|channel>thought", "<channel|>"
    else:
        _open, _close = "<think>", "</think>"
    _after_open = accumulated[accumulated.index(_open) + len(_open):]
    if _close in _after_open:
        # Opener and closer in the same token — split into reasoning + content
        # and stay out of thinking mode for subsequent tokens.
        _idx = _after_open.index(_close)
        _r = _after_open[:_idx].lstrip("\n") or None
        _c = _after_open[_idx + len(_close):].lstrip("\n") or None
        if _r: delta_reasoning = _r
        if _c: delta_content = _c
        _thinking_done = True
    else:
        in_thinking = True
        if _after_open:
            delta_reasoning = _after_open
    accumulated = ""
```

With this change, T5 emits `delta_content="7"` (and no `delta_reasoning`, since the inner `\n` is stripped) and `_thinking_done=True` keeps T6–T12 routed to `content` correctly. Multi-token opener/closer flows (the existing common case where `<|channel>thought` and `<channel|>` arrive as separate tokens) are preserved.

Blaizzy:
Fixed on main, thanks guys for flagging this ❤️

Issue #1156, opened 2026-05-08
20260508 -- Diagnostics Report — 7 failure(s), 2 harness issue(s) (mlx-vlm 0.5.0)


jrp2014:
# Diagnostics Report — 7 failure(s), 2 harness issue(s) (mlx-vlm 0.5.0)

**Run summary:** 55 locally-cached VLM model(s) checked; 7 hard failure(s), 2 harness/integration issue(s), 0 preflight warning(s), 48 successful run(s).

Test image: `20260502-173345_DSC09912_DxO.jpg` (27.3 MB).

---

## Issue Queue

Root-cause issue drafts are generated in
[issues/index.md](https://github.com/jrp2014/check_models/blob/main/src/output/issues/index.md).
Each row is intended to become one focused upstream GitHub issue.

| Target                                         | Problem                                                                                              | Evidence Snapshot                                                                                                                                               | Affected Models                                            | Issue Draft                                                                                                                                                              | Evidence Bundle                                                                                                                                                                                          | Fixed When                                                |
|------------------------------------------------|------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------|
| `mlx`                                          | Weight/config mismatch during model load                                                             | Model Error \| phase model_load \| ValueError \| 2 model cluster                                                                                                | 2: `LiquidAI/LFM2.5-VL-450M-MLX-bf16` (+1)                 | [issue draft](https://github.com/jrp2014/check_models/blob/main/src/output/issues/issue_001_mlx_mlx-model-load-model_001.md)                                             | [2 repro JSONs](https://github.com/jrp2014/check_models/blob/main/src/output/repro_bundles/20260509T112302Z_001_LiquidAI_LFM2.5-VL-450M-MLX-bf16_MLX_MODEL_LOAD_MODEL_853049863f38.json)                 | Load/generation completes or fails with a narrower owner. |
[...]

PR #1162, opened 2026-05-09
Fix LFM2-VL model

lucasnewman:
This resolves the first two issues reported in https://github.com/Blaizzy/mlx-vlm/issues/1156

1) The weights for the` LiquidAI/LFM2-VL-*` models include the layernorm weights even when the config specifies the project doesn't use it, so just load them at skip it at eval time instead. (Note the config has format = mlx so there's no sanitize_weights() hook.)
2) Implement a safe copy for NaiveStreamingDetokenizer, which also affected the quantized models, e.g. `mlx-community/LFM2-VL-1.6B-8bit`


Blaizzy (approved):
LGTM!

PR #1164, opened 2026-05-10
Fix an error when loading the MolmoPoint-8B model

lucasnewman:
Resolves item 6 in https://github.com/Blaizzy/mlx-vlm/issues/1156 which gives an error in main: `Loaded processor has no image_processor; expected multimodal processor`


Blaizzy (approved):
🚀

PR #1168, opened 2026-05-10
Fix crash with FastAPI workers when using Qwen3.5 family models

### LINKED ISSUE #1134: RuntimeError: There is no Stream(gpu, N) in current thread when calling generate from FastAPI worker threads
**Environment:**
- macOS Apple Silicon M1 Max 32GB
- mlx-vlm 0.5.0
- mlx 0.31.2
- FastAPI + Uvicorn (custom server, not mlx_vlm.server)

**Description:**
After upgrading from `0.4.4` to `0.5.0`, all calls to `mlx_vlm.generate` fail with:
`RuntimeError: There is no Stream(gpu, 0) in current thread.`

The error occurs inside `wired_limit` -> `stream_generate` -> `generate_step` -> `mx.async_eval(y)`.

**Root cause:**
In ` 0.4.4`, generation_stream was a global stream, accessible from any thread:
`generation_stream = mx.new_stream(mx.default_device())`

In `0.5.0` it was changed to thread-local:
`generation_stream = mx.new_thread_local_stream(mx.default_device())`

FastAPI/Uvicorn dispatches requests to worker threads that don't have this stream initialized. In `0.4.4` the global stream was visible from all threads and everything worked. In `0.5.0` wired_limit calls `mx.synchronize(generation_stream)` which fails in any thread other than the one that imported the module.

**Workaround:**
Downgrade to` mlx-vlm 0.4.4` +` mlx 0.31.1` + `mlx-lm 0.31.1`.

lucasnewman:
https://github.com/Blaizzy/mlx-vlm/issues/1134 is a little tricky -- because of the use of custom kernels in the Qwen3.5 model, we need to do an initial single-step eval after model load to properly materialize the thread-local stream state before it gets eval'd on the workers. This is the least invasive fix I could come up with.

Resolves https://github.com/Blaizzy/mlx-vlm/issues/1134

### DIFF
--- a/mlx_vlm/models/qwen3_5/language.py
+++ b/mlx_vlm/models/qwen3_5/language.py
@@ -27,6 +27,7 @@ class Qwen3_5RotaryEmbedding:
         inv_freq = 1.0 / (
             self.base ** (mx.arange(0, self.dim, 2).astype(mx.float32) / self.dim)
         )
+        mx.eval(inv_freq)
         self.inv_freq = inv_freq
 
         self.mrope_section = mrope_section
--- a/mlx_vlm/models/qwen3_5/qwen3_5.py
+++ b/mlx_vlm/models/qwen3_5/qwen3_5.py
@@ -3,7 +3,6 @@ from typing import Optional
 import mlx.core as mx
 import mlx.nn as nn
 
-from .. import cache
 from ..base import InputEmbeddingsFeatures
 from ..qwen3_vl import Model as Qwen3VLModel
 from ..qwen3_vl import processing_qwen3_vl  # noqa: F401
@@ -32,18 +31,6 @@ class Model(Qwen3VLModel):
         self.config = config
         self.vision_tower = VisionModel(config.vision_config)
         self.language_model = LanguageModel(config.text_config, config)
-
-    def post_load(self):
-        """Materialize state needed by MLX thread-local streams."""
-        input_ids = mx.array([[0]], dtype=mx.int32)
-        embeddings = self.get_input_embeddings(input_ids, None).inputs_embeds
-        outputs = self.language_model(
-            input_ids,
-            inputs_embeds=embeddings,
-            cache=cache.make_prompt_cache(self.language_model),
-        )
-        mx.eval(outputs.logits)
-        mx.clear_cache()
 
     def get_input_embeddings(
         self,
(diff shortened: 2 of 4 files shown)


Blaizzy (commented):
What's the issue and how do I replicate it?

Blaizzy:
I don't think this is the solution. I suspect they are using the wrong mlx version in the enviroment running the server somehow. Because that error only happens when you are not on the latest mlx core.

lucasnewman:
> I don't think this is the solution. I suspect they are using the wrong mlx version in the enviroment running the server somehow. Because that error only happens when you are not on the latest mlx core.

Yeah, I've been reducing it, it seems to be lazy initialization across streams. It might need an upstream fix or the contract of lazy eval with thread-local streams needs more specific care:

```
import argparse
import threading
import mlx.core as mx


class TinyNet:
    def __init__(self, eval_buffer: bool = False):
        # Lazy non-parameter buffer, e.g a rotary inv_freq table
        self.scale = 1.0 / (10000 ** (mx.arange(0, 8, 2).astype(mx.float32) / 8))
        if eval_buffer:
            mx.eval(self.scale)

    def __call__(self, x):
        return x * self.scale


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument(
        "--eval-buffer",
        action="store_true",
        help="Materialize the lazy buffer on the creating thread. This avoids the failure.",
    )
    args = parser.parse_args()

    net = TinyNet(eval_buffer=args.eval_buffer)
    errors = []

    def worker():
        try:
            x = mx.ones((4,), dtype=mx.float32)
            y = net(x)
            mx.eval(y)
            print("worker result:", y.tolist())
        except Exception as e:
            errors.append(e)
            raise

    thread = threading.Thread(target=worker)
    thread.start()
    thread.join()

    if errors:
        raise SystemExit(1)


if __name__ == "__main__":
    main()
```

lucasnewman:
> I don't think this is the solution. I suspect they are using the wrong mlx version in the enviroment running the server somehow. Because that error only happens when you are not on the latest mlx core.

FWIW it is a real bug, the reproduction script (the last comment) hits it 100% of the time.

lucasnewman:
@Blaizzy Take a look now, it's a lot simpler.

I also filed an issue with MLX to get some clarity on what's expected for this type of thing: https://github.com/ml-explore/mlx/issues/3529

Blaizzy (approved):
LGTM, this solution is much better!

PR #1170, opened 2026-05-11
fix: replace deprecated use_fast=True with backend=torchvision in load_processor

Jonathangadeaharder:
## Problem

`mlx_vlm/utils.py:load_processor()` hardcodes `use_fast=True`:

```python
processor = AutoProcessor.from_pretrained(model_path, use_fast=True, **kwargs)
```

Transformers 5.8+ deprecated `use_fast` parameter. This triggers a warning on every model load:

```
The `use_fast` parameter is deprecated and will be removed in a future version.
Use `backend="torchvision"` instead of `use_fast=True`.
```

## Fix

Replace with the equivalent non-deprecated API:

```python
processor = AutoProcessor.from_pretrained(model_path, backend="torchvision", **kwargs)
```

`backend="torchvision"` is the direct equivalent of `use_fast=True`.

## Related

PRs #757, #848, #865 touched `use_fast`-related paths but none changed the root cause in `load_processor()`.


Blaizzy:
Thanks for catching it!

Just remove the argument 

We don't use torch or torchvision as dependencies :)

Blaizzy (approved):
LGTM, thanks 🚀

PR #993, opened 2026-04-09
feat: add JSON mode via response_format parameter

eloe:
## Summary

Adds `response_format` support to `/v1/chat/completions` and `/responses`, per the [OpenAI API spec](https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format).

When `response_format.type` is `"json_object"`, a system instruction is prepended to the conversation requiring the model to respond with valid JSON only. This is a lightweight, model-agnostic approach that works with any model that follows system instructions.

### Example

```json
{
  "model": "...",
  "messages": [{"role": "user", "content": "List 3 colors"}],
  "response_format": {"type": "json_object"}
}
```

### Supported values

| `type` | Behavior |
|---|---|
| `"text"` (default) | No change |
| `"json_object"` | Injects JSON-only system instruction |

### Tests

- `test_resolve_response_format_json_adds_instruction`
- `test_resolve_response_format_text_no_change`
- `test_resolve_response_format_none_no_change`
- `test_responses_json_mode_accepted`

```
python -m pytest mlx_vlm/tests/ -k "json_mode or response_format" -v
```

Issue #561, opened 2025-10-30
Support for Structured Output for OpenAI API


4davelo:
Hi,

First: Amazing work getting MLX-VLM out of the ground, a pleasure to play with VLMs on the Apple ecosystem.

Feature request: would be great to be able to force the VLM to respond in structured way. 
I tried using the text_format on the client.responses.parse OpenAI endpoint only to discover that the parameter was ignored and the VLM did not comply to the pydantic structure making it difficult to consistently parse the output.

Keep up the great work!

David

eloe:
I reproduced this on current `main` on April 21, 2026 and this looks like the right tracker for the structured-output / `response_format` gap.

## Observed behavior
Requests that include `response_format={"type":"json_object"}` return `200`, but the field is silently ignored on both `/chat/completions` and `/responses`.

The current code path accepts the field because `FlexibleBaseModel` uses `extra="allow"`, but neither public request model declares `response_format`, and neither endpoint branches on it before calling `generate()`.

## Minimal repro
This avoids any model-specific behavior by patching the server internals and checking what reaches `generate()`:

```python
from types import SimpleNamespace
from unittest.mock import patch
from fastapi.testclient import TestClient
import mlx_vlm.server as server

client = TestClient(server.app)
model = SimpleNamespace()
processor = SimpleNamespace(tokenizer=SimpleNamespace(chat_template=""))
config = SimpleNamespace(model_type="qwen2_vl")
result = SimpleNamespace(
    text="ok",
    prompt_tokens=1,
    generation_tokens=1,
    total_tokens=2,
    prompt_tps=1.0,
    generation_tps=1.0,
    peak_memory=0.1,
)

with (
    patch.object(server, "get_cached_model", return_value=(model, processor, config)),
    patch.object(server, "apply_chat_template", return_value="prompt") as mock_template,
    patch.object(server, "generate", return_value=result) as mock_generate,
):
    for path, payload in [
        (
            "/chat/completions",
            {
                "model": "demo",
                "messages": [{"role": "user", "content": "hi"}],
                "response_format": {"type": "json_object"},
            },
        ),
        (
            "/responses",
            {
                "model": "demo",
                "input": "hi",
                "response_format": {"type": "json_object"},
            },
        ),
    ]:
        resp = client.post(path, json=payload)
        print(path, resp.status_code)
        print(sorted(mock_generate.call_args.kwargs.keys()))
        print(mock_template.call_args.kwargs)
```

## Actual result
[...]

PR #1173, opened 2026-05-12
Support response_format type=json_object

### LINKED ISSUE #1171: Support response_format type=json_object
## Summary

`response_format: {"type": "json_object"}` is still not accepted by the OpenAI-compatible server endpoints. The server currently supports `json_schema` structured outputs from #1047, but requests using the older OpenAI JSON mode value fail with `Unsupported response_format type: 'json_object'`.

## Background

Related prior work:

- #1047 added `json_schema` support and explicitly left `json_object` for future work.
- #993 and #959 attempted JSON mode support, but both were closed unmerged.
- #561 requested structured output support and was closed after #1047.

## Proposed behavior

Accept `json_object` in both OpenAI-compatible input shapes:

- Chat Completions: `response_format: {"type": "json_object"}`
- Responses API: `text: {"format": {"type": "json_object"}}`

Because the server now has llguidance-backed structured generation, `json_object` can be implemented as a permissive JSON object schema:

```json
{"type": "object"}
```

This preserves the OpenAI compatibility surface while avoiding the prompt-injection/whitespace failure mode of instruction-only JSON mode. It also keeps `json_schema` as the recommended path when callers need an explicit schema.

## Scope

- Accept `json_object` as a valid response format.
- Route it through the existing structured logits processor path.
- Cover Chat Completions and Responses API request shapes.
- Keep the existing limitation that structured outputs are not supported with speculative decoding.

## References

OpenAI documents `json_object` as older JSON mode and recommends `json_schema` for schema-constrained outputs: https://platform.openai.com/docs/guides/structured-outputs?api-mode=chat

eloe:
## Summary

Adds OpenAI-compatible `response_format: {"type": "json_object"}` support to the server.

`json_object` is routed through the existing llguidance-backed structured output path as a permissive JSON object schema:

```json
{"type": "object"}
```

This preserves compatibility with older OpenAI JSON mode while using the server's existing constrained generation machinery instead of prompt injection.

Fixes #1171.

## Details

Supported request shapes:

- Chat Completions: `response_format: {"type": "json_object"}`
- Responses API: `text: {"format": {"type": "json_object"}}`

The existing `json_schema` path remains unchanged and should still be preferred when callers need a specific schema. Existing speculative decoding behavior is unchanged: structured response formats are rejected for speculative decoding.

## Validation

```text
uv run --with pytest pytest mlx_vlm/tests/test_server.py -q
83 passed, 2 warnings
```

## Related Work

- #1047 added `json_schema` support and left `json_object` out of scope.
- #993 and #959 attempted JSON mode support but were closed unmerged.
- #561 requested structured output support and was closed after #1047.

## Note

Replaces closed draft #1172 after renaming the fork branch to remove the old prefix.

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -1417,6 +1417,8 @@ def _extract_response_format_schema(request) -> Optional[Union[str, dict]]:
     format_type = response_format.get("type")
     if format_type in (None, "text"):
         return None
+    if format_type == "json_object":
+        return {"type": "object"}
     if format_type != "json_schema":
         raise ValueError(f"Unsupported response_format type: {format_type!r}")
 
--- a/mlx_vlm/tests/test_server.py
+++ b/mlx_vlm/tests/test_server.py
@@ -545,6 +545,84 @@ def test_chat_completions_endpoint_forwards_explicit_sampling_args(client):
     assert mock_generate.call_args.kwargs["resize_shape"] == (512, 512)
 
 
+def test_chat_completions_endpoint_forwards_json_object_processor(client, monkeypatch):
+    model = SimpleNamespace()
+    processor = SimpleNamespace()
+    config = SimpleNamespace(model_type="qwen2_vl")
+    result = SimpleNamespace(
+        text='{"ok": true}',
+        prompt_tokens=8,
+        generation_tokens=4,
+        total_tokens=12,
+    )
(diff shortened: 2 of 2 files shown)


lucasnewman:
@eloe Thank you! We're fixing this here since this one has drifted a bit: https://github.com/Blaizzy/mlx-vlm/pull/1299

PR #1181, opened 2026-05-13
Compatibility bridge for non-VL models

lucasnewman:
This adds a compatibility shim to load non-VL models by way of `mlx-lm`, so we can load text-only models and properly handle input embeddings.


mlx_vlm/generate.py
Blaizzy, line 1: We can revert changes here

mlx_vlm/prompt_utils.py
```diff
@@ -686,6 +686,32 @@ def apply_chat_template(
...
+                    msg = dict(item) if isinstance(item, dict) else {"role": role}
+                    if role != "tool":
+                        msg["content"] = extract_text_from_content(content)
+                    messages.append(msg)
+        else:
+            messages = [{"role": "user", "content": str(prompt)}]
+
+        if return_messages:
+            return messages
+        return get_chat_template(processor, messages, add_generation_prompt, **kwargs)
```
Blaizzy, line 755: Not sure but should work for now

mlx_vlm/utils.py
Blaizzy, line 1: Loading should be pretty straight forward with our existing loaded. Perhaps the onyl change would be to check if `vision_config` exists, and if not we just retrieve just routing to `test_only.py` in `get_mode_and_args`

mlx_vlm/models/text_only.py
```diff
@@ -0,0 +1,199 @@
...
+    @property
+    def layers(self):
+        if hasattr(self._model, "layers"):
+            return self._model.layers
+        inner = self.model
+        if hasattr(inner, "layers"):
+            return inner.layers
+        if hasattr(inner, "h"):
+            return inner.h
+        return []
```
Blaizzy, line 77: mlx-lm always exposes this

mlx_vlm/models/text_only.py
```diff
@@ -0,0 +1,199 @@
...
+            if hasattr(module, attr):
+                return getattr(module, attr)
+        return None
+
+    def input_embeddings(self, input_ids: mx.array) -> mx.array:
+        embedding = self._token_embedding()
+        if embedding is None:
+            raise ValueError(
+                "The wrapped mlx-lm model does not expose token embeddings."
+            )
```
Blaizzy: Can use `inputs_embeds` as the main?

Blaizzy (commented):
Almost there 🚀

mlx_vlm/prompt_utils.py
lucasnewman: Yeah, I think we could route through a unified path at some point in the near future.

Blaizzy (approved):
LGTM 🚀

mlx_vlm/utils.py
```diff
@@ -421,7 +442,14 @@ def load(
...
     model = load_model(model_path, lazy, **kwargs)
     if adapter_path is not None:
-        model = apply_lora_layers(model, adapter_path)
+        if getattr(model, "_is_text_model", False):
+            from mlx_lm.utils import load_adapters
+
+            model.language_model._model = load_adapters(
+                model.language_model._model, adapter_path
+            )
+        else:
```
Blaizzy, line 451: Related to #1184

PR #1186, opened 2026-05-15
Unify LoRA loading for VLM/LM models

### LINKED ISSUE #1184: Unify LoRA adapter loading for text-only MLX-LM models
## Context

The text-only MLX-LM bridge currently needs a special adapter path in `mlx_vlm.utils.load()`:

```python
if getattr(model, "_is_text_model", False):
    from mlx_lm.utils import load_adapters

    model.language_model._model = load_adapters(
        model.language_model._model, adapter_path
    )
else:
    model = apply_lora_layers(model, adapter_path)
```

This is required because MLX-LM adapters and MLX-VLM adapters use different config schemas and target different model roots.

## Problem

`mlx_vlm.trainer.utils.apply_lora_layers()` only understands the MLX-VLM LoRA adapter format:

- expects `adapter_config.json` to contain `rank`, `alpha`, `dropout`, etc.
- discovers modules through `model.language_model`
- loads weights through the VLM-shaped wrapper

MLX-LM adapters instead use the MLX-LM tuner schema:

- `num_layers`
- `lora_parameters`
- `fine_tune_type`
- raw MLX-LM model weight keys

So text-only models cannot just use the existing `apply_lora_layers()` implementation unless it delegates to `mlx_lm.utils.load_adapters()` for `_is_text_model` wrappers.

## Proposed Change

Make `apply_lora_layers()` the single public adapter-loading entry point for `mlx_vlm.load()`, but teach it to dispatch internally:

```python
if getattr(model, "_is_text_model", False):
    from mlx_lm.utils import load_adapters

    model.language_model._model = load_adapters(
        model.language_model._model, adapter_path
    )
    return model
```

Then `mlx_vlm.utils.load()` can always call:

```python
model = apply_lora_layers(model, adapter_path)
```

and no longer needs to know about MLX-LM adapter internals.

## Acceptance Criteria

- `mlx_vlm.load(..., adapter_path=...)` still works for existing VLM LoRA adapters.
- `mlx_vlm.load(..., adapter_path=...)` works for text-only MLX-LM models wrapped by `mlx_vlm.models.text_only`.
- Add tests that cover both adapter config schemas or at least assert the dispatch behavior for `_is_text_model`.
- Keep the raw MLX-LM adapter loading targeted at `model.language_model._model`, not the VLM wrapper.

lucasnewman:
Resolves https://github.com/Blaizzy/mlx-vlm/issues/1184

### DIFF
--- a/mlx_vlm/tests/test_trainer.py
+++ b/mlx_vlm/tests/test_trainer.py
@@ -6,7 +6,7 @@ import mlx.nn as nn
 
 from mlx_vlm.trainer.datasets import VisionDataset
 from mlx_vlm.trainer.lora import LoRaLayer
-from mlx_vlm.trainer.sft_trainer import TrainingArgs, train
+from mlx_vlm.trainer.sft_trainer import TrainingArgs, iterate_batches, train
 
 
 class TestDataset(unittest.TestCase):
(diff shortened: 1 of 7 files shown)


lucasnewman:
@Blaizzy I updated the LoRA training to use the `mlx-lm`-type schema, which is a lot more robust, and kept in compatibility for loading the earlier schema. I also fixed an issue with batching images in VL models which I noticed during testing.

I trained `mlx-community/Qwen3-VL-2B-Instruct-4bit` on `sssqs/vqa-rad` (medical image QA dataset) on my M5 Max  and verified the new schema is used and the training / loading works as expected.

Blaizzy (approved):
LGTM🚀

PR #1190, opened 2026-05-16
Fix LFM2-VL image preprocessing for variants shipping input_data_format=channels_last

contrapuntal:
## Problem

Image input to `mlx-community/LFM2-VL-450M-bf16` and `mlx-community/LFM2-VL-1.6B-bf16` raises `Failed to process inputs with error: The size of tensor a (N) must match the size of tensor b (3) at non-singleton dimension 0`, where `N` is the image height in pixels. `mlx-community/LFM2-VL-3B-bf16` works on the same input.

Traceback:

```
File "mlx_vlm/models/lfm2_vl/processing_lfm2_vl.py", in _patched_call
    vision_inputs = self.image_processor(batched_images, **images_kwargs)
File "transformers/models/siglip2/image_processing_siglip2.py", in _preprocess
    image = self.rescale_and_normalize(image, ...)
File "transformers/image_processing_backends.py", in normalize
    return tvF.normalize(image, mean, std)
File "torchvision/transforms/v2/functional/_misc.py", in normalize_image
    image = image.sub(mean)
RuntimeError: The size of tensor a (224) must match the size of tensor b (3) at non-singleton dimension 0
```

The 450M and 1.6B variants ship `input_data_format: "channels_last"` in `preprocessor_config.json`. `_patched_from_pretrained` in `mlx_vlm/models/lfm2_vl/processing_lfm2_vl.py` already pops a list of incompatible keys (`image_processor_type`, `return_tensors`, etc.) before constructing `Siglip2ImageProcessor`, but leaves `input_data_format` in place. The hint survives into the constructed processor, so Siglip2's `rescale_and_normalize` passes a channels-last (HWC) tensor to `torchvision.transforms.v2.functional.normalize`. Because torchvision expects channels-first (CHW), the broadcast against the per-channel mean of shape `(3,)` fails at dim 0. The 3B variant ships no `input_data_format` key in its preprocessor_config, so its processor uses the safe default and works.

## Fix

Add `"input_data_format"` to the list of keys popped from the loaded preprocessor_config in `_patched_from_pretrained`:

```python
for key in (
    "image_processor_type",
    "processor_class",
    "image_seq_length",
    "return_row_col_info",
    "device",
    "disable_grouping",
    "return_tensors",
    "input_data_format",   # NEW
):
    image_processor_config.pop(key, None)
```

This drops a model-shipped hint that fits the upstream `Lfm2VlImageProcessorFast` but breaks the `Siglip2ImageProcessor` substitution. With the hint gone, Siglip2 detects the layout from the actual input and passes correctly shaped tensors to torchvision.

## How to test

Tested with `mlx-vlm==0.5.0`, `transformers==5.5.4`, `mlx==0.31.2` on Apple Silicon (Python 3.12). This fix needs no pytest changes; run the snippet below to verify:

```python
from PIL import Image
from mlx_vlm.utils import load as vlm_load, prepare_inputs

model, processor = vlm_load("mlx-community/LFM2-VL-450M-bf16", trust_remote_code=True)
print("input_data_format:", processor.image_processor.input_data_format)
# Before fix: "channels_last" → fails downstream
# After fix:  None             → works

img = Image.new("RGB", (224, 224), color=(73, 109, 137))
out = prepare_inputs(processor, images=[img], prompts=["<image>describe"])
print({k: getattr(v, "shape", type(v).__name__) for k, v in out.items()})
```

Before the fix the `prepare_inputs` call raises with the dim-0 broadcast error above; after the fix it returns the prepared input dict. Verified locally against `mlx-community/LFM2-VL-450M-bf16` and `mlx-community/LFM2-VL-1.6B-bf16`. `LFM2-VL-3B-bf16` behaves identically either way.

## Related

- #872 introduced the slow-Siglip2 substitution shim. The pop list this PR extends lives in the same function.
- #725 corrected a different image-shape mismatch in the same LFM2-VL processor patch (image-token-count divergence). Different surface, adjacent symptom.

## Scope and risk

- Files: `mlx_vlm/models/lfm2_vl/processing_lfm2_vl.py` (+1 line, inside the existing pop loop).
- Risk: extremely low. Dropping this config hint stops it from overriding Siglip2's default layout detection, so the processor infers the input layout correctly without breaking the substitution.
- No public API change.


Blaizzy (approved):
LGTM, thanks @contrapuntal!

PR #1194, opened 2026-05-17
Freeze MaskedEmbedder.token_ordering so optimizer can't corrupt it

DirectriX01:
## Summary
`MaskedEmbedder.token_ordering` is a static int32 cluster→vocab-id index, loaded from checkpoint and used purely as a gather index. `nn.Module` treats every `mx.array` attribute as a trainable parameter by default, so any optimizer step (AdamW, SGD, ...) will:

1. Walk `token_ordering` via `tree_map` in `apply_gradients`
2. Apply `param - lr * m / (sqrt(v) + eps)` — promoting the dtype to float32
3. Cause the next `ordering[topk_idx]` gather in `_selected_logits` to throw `ValueError: [gather] Got indices with invalid dtype. Indices must be integral.`

This silently breaks fine-tuning of the Gemma 4 assistant drafter the first time the optimizer updates.

## Fix
`self.freeze(keys="token_ordering", recurse=False)` after the buffer is created in `__init__`. Excludes the field from `trainable_parameters()` so the optimizer skips it entirely; `parameters()` still includes it, so saving/loading checkpoints is unchanged.

## Test
Added `test_masked_embedder_token_ordering_is_buffer_not_parameter` in `mlx_vlm/tests/test_speculative.py`. Asserts that:

- `token_ordering` is absent from `trainable_parameters()`
- An `AdamW` step (zero gradients) preserves its int32 dtype and contents

The test fails without the freeze (`{'centroids.weight', 'token_ordering'}` ⊃ `{'token_ordering'}`) and passes with it.


Blaizzy (approved):
LGTM, thanks!

DirectriX01:
thanks!!

Blaizzy:
Welcome :)

PR #1196, opened 2026-05-17
Add server support for Anthropic-style /v1/messages API

lucasnewman:
Based on the docs here: https://platform.claude.com/docs/en/build-with-claude/working-with-messages

I tested streaming and tool calling from Qwen3.5 locally and it works as expected.

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -15,7 +15,7 @@ from datetime import datetime
 from queue import Empty as QueueEmpty
 from queue import Queue
 from threading import Event, Lock, Thread
-from typing import Any, Callable, Iterator, List, Literal, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Iterator, List, Literal, Optional, Tuple, Union
 
 logger = logging.getLogger("mlx_vlm.server")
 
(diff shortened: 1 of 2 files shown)


Blaizzy (approved):
LGTM! 🚀

Blaizzy:
Follow up 

`Tool-call arguments shape inconsistency` — server.py:227-235
`_anthropic_tool_use_to_openai` passes arguments as a dict:
"arguments": block.get("input") or {},
But the existing `process_tool_calls `(line 1297-1302) emits arguments as a JSON string `(json.dumps(args, ...))`. Most chat templates downstream expect the string form. Even though the test in test_server.py:1191-1207 asserts the dict form, this will likely break with real chat templates that call json.loads or string-format arguments. Recommend `json.dumps(block.get("input") or {}, ensure_ascii=False)` for parity with the existing OpenAI tool-call path.

Blaizzy:
`tool_result` with image content silently dropped — server.py:144-169
`_anthropic_tool_result_content_to_text` only walks text / document blocks. Anthropic's spec allows image blocks in `tool_result.content`. They're silently discarded here.

PR #1197, opened 2026-05-19
Decode OpenAI input_audio base64 in chat completions

Quiet-Node-io:
## Summary

Fix `/v1/chat/completions` handling for OpenAI-style `input_audio` content blocks by decoding base64 audio data before it is passed to the audio loader.

The current server extracts `input_audio["data"]` and forwards that string as the audio input. For OpenAI-compatible requests this field is commonly base64-encoded audio bytes, so the loader treats the base64 string as a path-like audio reference and fails with an unsupported file format error. This PR decodes bare base64 and `data:audio/...;base64,...` values into `BytesIO` objects while preserving existing URL/path references.

## Background

Atlas RAN-474 Phase 1 captured a decoder reproduction in `docs/dev/current/ran-474/evidence/decoder-reproduction.md`: the same WAV bytes load successfully when passed as a filesystem path, but fail when the base64 string is passed through the server path as though it were a file path/URL.

## Tests

```bash
uv run --with pytest --with fastapi --with httpx --with mlx --with huggingface_hub --with pydantic --with typing_extensions --with numpy --with uvicorn pytest mlx_vlm/tests/test_server.py::test_chat_completions_decodes_input_audio_base64 mlx_vlm/tests/test_server.py::test_chat_completions_preserves_input_audio_references -q
```

Result: `3 passed, 2 warnings`.

```bash
uv run python -m py_compile mlx_vlm/server.py mlx_vlm/tests/test_server.py
```

Result: passed.

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -1,5 +1,7 @@
 import argparse
 import asyncio
+import base64
+import binascii
 import gc
 import json
 import logging
@@ -12,6 +14,7 @@ from collections import deque
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
 from datetime import datetime
+from io import BytesIO
 from queue import Empty as QueueEmpty
 from queue import Queue
 from threading import Event, Lock, Thread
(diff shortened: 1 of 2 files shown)


lucasnewman:
Thanks! Through no fault of your own, the server code is significantly restructured since you put this up. I went ahead and ported your changes over in https://github.com/Blaizzy/mlx-vlm/pull/1211 so we can get the fix in.

PR #1198, opened 2026-05-19
[mlx_vlm] Expose 'strict' parameter in load() function

zyguy:
**Motivation:** Currently, the high-level `mlx_vlm.load()` function does not expose the `strict` parameter. This makes it impossible to pass `strict=False` when loading certain converted models or "abliterated" models that contain harmless orphaned weights.

The underlying `load_model()` function already supports the `strict` argument. This PR simply surfaces it to the main `load()` entrypoint, maintaining `True` as the default to ensure full backward compatibility. This will allow dependent projects like `oMLX` to load a wider variety of fine-tuned or abliterated Vision-Language Models without crashing.

### DIFF
--- a/mlx_vlm/utils.py
+++ b/mlx_vlm/utils.py
@@ -409,6 +409,7 @@ def load(
     adapter_path: Optional[str] = None,
     lazy: bool = False,
     revision: Optional[str] = None,
+    strict: bool = True,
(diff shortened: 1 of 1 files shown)


lucasnewman (approved):
LGTM

PR #1199, opened 2026-05-19
Support /v1/responses stateful responses API in the server

lucasnewman:
This provides Codex-style harness support in the server. Responses are stored in a lightweight in-memory cache for the time being.

### DIFF
--- a/mlx_vlm/server.py
+++ b/mlx_vlm/server.py
@@ -1837,9 +1837,12 @@ class OpenAIRequest(FlexibleBaseModel):
         None, description="System/developer instructions for this response."
     )
     previous_response_id: Optional[str] = Field(
-        None, description="ID of a previous response whose input/output items should be included."
+        None,
+        description="ID of a previous response whose input/output items should be included.",
     )
-    tools: Optional[List[Any]] = Field(None, description="Responses API tool definitions.")
+    tools: Optional[List[Any]] = Field(
+        None, description="Responses API tool definitions."
+    )
     tool_choice: Optional[Any] = Field(None, description="Tool choice policy.")
     store: Optional[bool] = Field(
         True, description="Whether to store this response for later retrieval."
@@ -2102,7 +2105,9 @@ def _append_response_item_to_prompt(
                 elif part_type == "image_url":
                     image_url = part.get("image_url")
                     images.append(
-                        image_url.get("url") if isinstance(image_url, dict) else image_url
+                        image_url.get("url")
+                        if isinstance(image_url, dict)
+                        else image_url
                     )
             content = "\n".join(p for p in text_parts if p)
         chat_messages.append({"role": role, "content": content or ""})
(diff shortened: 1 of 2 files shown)


mlx_vlm/server.py
```diff
@@ -1989,6 +2009,334 @@ class ResponseCompletedEvent(BaseStreamEvent):
     ResponseCompletedEvent,
 ]
 
+
+RESPONSE_STORE_LIMIT = int(os.environ.get("MLX_VLM_RESPONSE_STORE_LIMIT", "1024"))
```
Blaizzy, line 2013: I have a couple questions:
1. What is this env var for? 
2. How does it affect different models? 

And can we document in the readme/docs if we keep it.

mlx_vlm/server.py
```diff
@@ -1989,6 +2009,334 @@ class ResponseCompletedEvent(BaseStreamEvent):
...
+        if not isinstance(item, dict):
+            raise HTTPException(status_code=400, detail="Invalid input format.")
+        item_type = item.get("type")
+        if item_type is None and item.get("role") is not None:
+            item = {**item, "type": "message"}
+        items.append(item)
+    return items
+
+
+def _response_call_to_chat_tool_call(item: Dict[str, Any]) -> Dict[str, Any]:
```
Blaizzy, line 2063: Is this praising the tool call? 

If so, can we reuse the existing parser?

mlx_vlm/server.py
```diff
@@ -1989,6 +2009,334 @@ class ResponseCompletedEvent(BaseStreamEvent):
...
+    if not isinstance(arguments, str):
+        arguments = json.dumps(arguments or {}, ensure_ascii=False)
+    return {
+        "type": "function",
+        "id": call_id,
+        "function": {"name": name or "tool", "arguments": arguments},
+    }
+
+
+def _append_response_item_to_prompt(
```
Blaizzy, line 2083: We should probably move such functions or prompt utils

Blaizzy (commented):
It mostly LGTM! 

Just a few nits.

A general idea I'm thinking about is to make server into a module and separate Anthropic and OpenAI changes into scripts we import into server so we can have clarity about provenances. Because I lost track of what does what.

We could also have a common.py if some functions are used in both.

mlx_vlm/server.py
lucasnewman: This is the max number of responses we'd store in the in-memory cache before aging them out. The responses themselves aren't large, but it seemed reasonable to put a (high) limit so it's not just growing memory unboundedly.

A more robust solution would be to have a datastore for them, but that adds a lot of code and complexity. In practice it seems more likely at the moment to be using this in a harness and only referencing recent-ish responses across a handful of contexts/threads.

Blaizzy (approved):
LGTM 🚀

mlx_vlm/server.py
Blaizzy: That makes sense! 

We can for disk offload after a threshold 

Our APC already does this automatically.

I would remove this and use APC as default, the only change missing is to add a knob for max warm-memory GB before offloading

lucasnewman: I think that makes sense if we default APC to on, I can adjust that.

PR #1200, opened 2026-05-19
Fix /v1/messages tool call image handling

lucasnewman:
Properly handle images in tool calls. Addresses the review feedback in https://github.com/Blaizzy/mlx-vlm/pull/1196


Blaizzy (approved):
🚀

PR #1203, opened 2026-05-19
Modularize server

lucasnewman:
This breaks the server down into a small number of files that handle specific areas to make it easier to read and maintain.

I tried to maintain exact backwards compatibility with anyone currently using imports from `mlx_vlm.server`, so from a functionality perspective this should be exactly the same as what we have now, and the `python -m mlx_vlm server` invocation should work the same way.

The new implementation is in `mlx_vlm/server` with the following files:

**__init__.py**: compatibility shim that re-exports the old mlx_vlm.server public surface.
**__main__.py**: lets python -m mlx_vlm.server run the server CLI.
**app.py**: FastAPI app setup, lifespan, model load/unload, health/models/metrics/cache routes, and route registration.
**cli.py**: argparse options and uvicorn launch setup.
**runtime.py**: shared mutable server state for model cache, response generator, APC manager, and metrics.
**generation.py**: generation arguments, model loading helpers, continuous batching, KV/APC config, and metrics helpers.
**schemas.py**: request/response models and typed payload definitions for server APIs.
**openai.py**: OpenAI-compatible /responses and /chat/completions route handlers.
**anthropic.py**: Anthropic-compatible /messages route handlers and message/tool conversion helpers.
**responses_state.py**: in-memory Responses API state, previous-response replay, item normalization, and tool-call parsing helpers.


Blaizzy (approved):
LGTM, thanks!

Issue #882, opened 2026-03-30
fix: Handle array-valued offset from BatchKVCache in Llama4 attention


inureyes:
## Summary

mlx-vlm added guards for `offset` being an `mx.array` (from `BatchKVCache`) in Llama4's attention and model forward pass. Without these guards, `mx.arange(offset + 1, ...)` fails when `offset` is array-valued, and `maybe_trim_front()` may be called on caches that don't support it.

## Priority

Low — only affects batch inference scenarios with Llama4.

PR #1205, opened 2026-05-20
fix: handle array-valued offset in BatchQuantizedKVCache.make_mask

lele872:
## Problem

`BatchQuantizedKVCache` stores `self.offset` as an `mx.array` (initialized in `__init__` as `mx.array([-lp for lp in left_padding])`). When `make_mask()` passes this array to `create_attention_mask()`, it reaches `create_causal_mask()` which calls `mx.arange(offset + N)`. Since `mx.arange` requires an `int`, this raises:

```
TypeError: arange(): incompatible function arguments
```

This causes a crash when using `--kv-bits` with the OpenAI-compatible server (`mlx_vlm.server`).

Related: #882 (same issue identified but closed without fix).

## Fix

Convert the array-valued offset to `int` in `make_mask()` before passing it to `create_attention_mask()`:
- For scalar arrays (single batch, `ndim == 0`): use `.item()`
- For 1D arrays (multi-batch): take the first element

## Testing

- Added 3 new tests in `TestMakeMask` class covering scalar array offset, 1d array offset, and single-token mask generation
- All 19 tests in `test_batch_quantized_cache.py` pass
- All 664 tests in the test suite pass
- `pre-commit` (black, isort, autoflake) passes

## Reproduction

```bash
python -m mlx_vlm.server --model mlx-community/gemma-4-26b-a4b-it-5bit \
  --port 9090 --max-kv-size 131072 --kv-bits 8
# Then send any chat completion request → TypeError
```

### DIFF
--- a/mlx_vlm/models/cache.py
+++ b/mlx_vlm/models/cache.py
@@ -387,7 +387,10 @@ class BatchQuantizedKVCache(_BaseCache):
     def make_mask(self, *args, **kwargs):
         from mlx_lm.models.cache import create_attention_mask
 
-        return create_attention_mask(*args, offset=self.offset, **kwargs)
+        offset = self.offset
+        if isinstance(offset, mx.array):
+            offset = int(offset.item()) if offset.ndim == 0 else int(offset[0])
+        return create_attention_mask(*args, offset=offset, **kwargs)
 
 
 def make_prompt_cache(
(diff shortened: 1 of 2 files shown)


Blaizzy:
Thanks! 

Please open an issue with a reproducible example.

This is a important component and can break then workflow to thousands of models. 

I suspect the issue is in the model itself

PR #1209, opened 2026-05-21
Track APC cached tokens for generation metrics

spicyneuron:
## Summary

This is groundwork for a follow-up PR to add generation metrics to the server.

Currently, APC cached-token count is not carried through generation results. Server code can see prompt token totals and throughput, but it cannot reliably tell how many prompt tokens came from APC.

## Changes

This PR threads the APC cached-token count through generation internals:

- Adds `cached_tokens` to `GenerationResult`, `PromptProgress`, `StreamingToken`
- Captures APC `prefix_len` per request row during prompt batch setup

The important decision here is to use APC’s per-row `prefix_len` directly. That keeps the value scoped to the request the client made, instead of deriving it later from batched tensor shapes, padding, or suffix lengths. It also avoids treating cache accounting as a batch-level metric.

## Related

This PR takes a different approach than https://github.com/Blaizzy/mlx-vlm/pull/1151, which reads cached tokens back from `apc_meta` when building prompt progress.

That works, but `apc_meta` is operational state that may be consumed or cleared as the prompt batch advances. This PR snapshots the per-request cached-token count when the prompt batch is built, then carries it through the existing generation metric types. That keeps server metrics tied to request-level generation state instead of APC’s internal lifecycle.

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -385,6 +385,7 @@ class GenerationResult:
     prompt_tps: float = 0.0
     generation_tps: float = 0.0
     peak_memory: float = 0.0
+    cached_tokens: int = 0
 
 
 class PromptCacheState:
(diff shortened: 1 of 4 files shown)


lucasnewman (approved):
LGTM

PR #1211, opened 2026-05-21
Handle base64-encoded input audio in chat completions

lucasnewman:
Based on https://github.com/Blaizzy/mlx-vlm/pull/1197, which was the correct fix but has conflicts from the modular server refactoring.


Blaizzy (approved):
LGTM!

PR #1215, opened 2026-05-23
Fix server `finish_reason` inconsistencies

spicyneuron:
Follow-up to #1209.

While working on generation metrics in server responses, I ran into edge cases in how the server reports `finish_reason`, as well as minor inconsistencies across batched and non-batched.

This PR contains standalone fixes that improve consistency across the board. This should also make the full metrics PR easier to review:
- https://github.com/Blaizzy/mlx-vlm/pull/1216

## Fixes

**Non-streaming endpoints reported the wrong `finish_reason`.** All three endpoints (chat completions, responses, messages) hardcoded `"stop"` on the non-streaming path. A request that hit `max_tokens` was indistinguishable from one the model decided to end. Now propagates `"length"` correctly.

**Streaming chat completions reported inconsistently across backends.** Whether `finish_reason` arrived, and on which chunk, depended on whether the request was served by continuous batching or `stream_generate`. Now: one terminal chunk carries `finish_reason`, regardless of backend.

**Tenant-scoped Responses requests crashed on the non-streaming path.** `apc_tenant` was passed twice — once explicitly, once via `**gen_args.to_generate_kwargs()`. Raised `TypeError` for any request with `tenant_id` set. Removed the duplicate.

**`stream_generate` didn't expose why it stopped.** Callers had to infer with heuristics. Added `finish_reason` to `GenerationResult` (`"stop"` for eos/stop-sequence, `"length"` for max_tokens).

**`stream_generate(max_tokens=0)` raised `UnboundLocalError`.** The terminal yield after the for loop referenced `token` and `n`, which were never bound when the loop iterated zero times. Now yields a zero-token terminal `GenerationResult` so consumers always see stream termination.

## Notes

- Anthropic streaming now honors `top_k`, `min_p`, `repetition_penalty`, `logit_bias`, `enable_thinking`, `thinking_budget`, and `thinking_start_token`. These were silently dropped before; requests that appeared to work may produce different output.
- Empty SSE deltas are no longer emitted on `/v1/responses` and the chat-completions `stream_generate` fallback.
- /v1/chat/completions wire shape changes to be more OpenAI spec compliant. `finish_reason` is now consistently sent on the final chunk instead of content chunks.
- The server's `generate()` / `stream_generate()` call sites are now routed through `**gen_args.to_generate_kwargs()` consistently.
- Tests for the new wire shapes will be included in the metrics PR.

### DIFF
--- a/mlx_vlm/generate.py
+++ b/mlx_vlm/generate.py
@@ -391,6 +391,9 @@ class GenerationResult:
     generation_tps: float = 0.0
     peak_memory: float = 0.0
     cached_tokens: int = 0
+    # Populated only on the terminal chunk yielded by ``stream_generate``:
+    # ``"stop"`` for eos/stop-sequence, ``"length"`` for max_tokens.
+    finish_reason: Optional[str] = None
 
 
 class PromptCacheState:
@@ -1055,6 +1058,7 @@ def stream_generate(
         tic = time.perf_counter()
 
         generated_tokens = []
+        finish_reason: Optional[str] = None
         for n, (token, logprobs) in enumerate(gen):
             if n == 0:
                 prompt_time = time.perf_counter() - tic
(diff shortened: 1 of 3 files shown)


lucasnewman (approved):
LGTM

PR #1216, opened 2026-05-23
Add spec-compliant `usage` and `timings` for server endpoints

spicyneuron:
Follow-up to #1209 and #1215 ([focused diff against #1215](https://github.com/spicyneuron/mlx-vlm/compare/server-metrics-cleanup...spicyneuron:mlx-vlm:server-metrics?diff=split&w#diff-a23c7cc391035f08973a9ad3d63c346d8491a54ba1f934be53a47eae1f7276f5)).

Currently, server generation reporting is incomplete: cache hits aren't surfaced, throughput stats sit inside `usage` instead of a sibling `timings`, and streaming `usage` is inconsistent.

This PR is a best-effort alignment with each respective spec ([OpenAI Chat](https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create#(resource)%20chat.completions%20%3E%20(model)%20chat_completion%20%3E%20(schema)%20%3E%20(property)%20usage) / [llama.cpp](https://github.com/ggml-org/llama.cpp/blob/master/tools/server/README.md#post-v1chatcompletions-openai-compatible-chat-completions-api), [OpenAI Responses](https://developers.openai.com/api/reference/resources/responses/methods/create), [Anthropic Messages](https://platform.claude.com/docs/en/build-with-claude/working-with-messages)).

## What's added

**`/v1/chat/completions`.** OpenAI `usage` + a sibling `timings` (llama.cpp-shaped). Streaming: both appear on the final chunk, gated by `stream_options.include_usage`.

**`/v1/responses`.** OpenAI `usage`. Streaming: on the terminal `response.completed`.

**`/v1/messages`.** Anthropic `usage` with the `cache_read_input_tokens` / `cache_creation_input_tokens` split. Streaming: initial values on `message_start`, updates on `message_delta`.

## Notes

- Anthropic's `message_start` is deferred until first-token so `cache_read_input_tokens` can be populated. This matches what their docs imply (non-zero tokens in `message_start`).
- `GenerationMetrics` (new dataclass) duck-types `StreamingToken` and `GenerationResult` via `getattr` — overlapping but non-identical fields. Unifying could be a follow-up, but I deferred to avoid more churn.
- `peak_memory` is mlx-vlm specific. Lives inside `timings`, not `usage`, to keep the spec-defined object clean.
- Tests cover the new wire shapes per endpoint, including the `include_usage` gate and the Anthropic cache split.


Blaizzy (commented):
This is awesome! 👌🏽

Blaizzy (approved):
LGTM, thanks!

PR #1217, opened 2026-05-23
Fix inference cancellation on stream abort

spicyneuron:
Currently, when a streaming client aborts the connection, the server does not reliably cancel the underlying inference request.

The endpoints already call `token_iter.close()` from their stream cleanup paths. However, `ResponseGenerator.generate()` returns a Python generator, and in the continuous batching path one thread may be blocked in `next(token_iter)` while another thread calls `close()` during disconnect cleanup.

CPython rejects that with `ValueError: generator already executing`. The endpoint cleanup catches and ignores the exception, so cancellation is never signaled and inference may continue until `max_tokens`.

## Reproduction

1. Monitor GPU usage (`macmon`, `btop`, etc.) in one terminal.
2. Start `mlx-vlm.server` in another.
3. Send a long streaming request:
   ```sh
   curl -N http://localhost:8000/v1/chat/completions \
     -H 'Content-Type: application/json' \
     -d '{
       "model": "...",
       "stream": true,
       "messages": [
         {
           "role": "user",
           "content": "Write a very long essay about penguins."
         }
       ],
       "max_tokens": 4096
     }'
   ```
4. Press `Ctrl+C` while tokens are streaming.

Before this fix, GPU activity continues long after disconnect. After this fix, stream close cancels the active request and GPU activity drops shortly after.

## Fix

This PR replaces the generator returned by `ResponseGenerator.generate()` with a small closeable iterator. Callers keep the same contract.

The iterator preserves the previous behavior for queued tokens, terminal sentinels, queued exceptions, and token queue timeouts. The key change is that `close()` can now safely cancel an unfinished request even while another thread is blocked in `__next__()`.

## Note

An alternative would be to patch each streaming endpoint to call `ResponseGenerator._cancel(ctx.uid)` directly on disconnect.

This PR keeps cancellation inside the token iterator lifecycle instead. Endpoints already express the intended behavior by closing the iterator in `finally`; the bug was that the old generator implementation could not reliably honor that contract under cross-thread cleanup.

That keeps private cancellation details out of the OpenAI/Anthropic handlers and makes future streaming callers less likely to miss cancellation.

### DIFF
--- a/mlx_vlm/server/generation.py
+++ b/mlx_vlm/server/generation.py
@@ -8,7 +8,7 @@ from dataclasses import dataclass
 from queue import Empty as QueueEmpty
 from queue import Queue
 from threading import Event, Lock, Thread
-from typing import Callable, Iterator, List, Optional, Tuple
+from typing import Callable, List, Optional, Tuple
 
 import mlx.core as mx
 from fastapi import HTTPException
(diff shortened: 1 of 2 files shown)


lucasnewman (approved):
LGTM

PR #1214, opened 2026-05-22
Improve GLM-OCR generation and repetition controls

Blaizzy:
## Summary
- Build server logits processors from request-level `repetition_penalty`, `presence_penalty`, `frequency_penalty`, and `logit_bias` before inserting requests into `BatchGenerator`.
- Port GLM-OCR image preprocessing and placeholder expansion closer to the Transformers implementation, with processor coverage for the new behavior.
- Fix GLM-OCR M-RoPE position indexing and chunked prefill position reuse so long OCR prompts do not drift into repeated table rows.

## Issue
Fixes #1021

## Validation
- `uv run mlx_vlm.generate --model mlx-community/GLM-OCR-bf16 --prompt "Text Recognition:" --image /Users/prince_canuma/Downloads/table_2.png --temperature 0.0 --no-verbose`
- `uv run python -m compileall -q mlx_vlm/models/glm_ocr/language.py mlx_vlm/tests/test_models.py`
- Focused GLM-OCR RoPE smoke check
- `uv run --with pytest python -m pytest mlx_vlm/tests/test_server.py::TestResponseGenerator::test_server_generation_builds_repetition_logits_processors -q` -> 1 passed
- `uv run --with pytest python -m pytest mlx_vlm/tests/test_server.py -q` -> 94 passed
- `uv run --with pytest --with psutil python -m pytest -q --ignore=mlx_vlm/tests/test_smoke.py` -> 712 passed, 1 skipped

Full bare `pytest -q` still collects `mlx_vlm/tests/test_smoke.py`, whose `test_model_loading` and `test_generation` expect external fixtures (`model_path`, `model`) and error under normal collection.

PR #1221, opened 2026-05-26
fix: fallback to defaults

GeneCodeSavvy:
## Summary

Fixes an issue where `mlx-vlm/server/app.py` unintentionally overrides the default `repetition_context_size`.

`mlx-vlm` already defines a default context size of `20` in:

`mlx_vlm/server/generation.py`

However, when a request does not include `repetition_context_size`, the app layer was passing `None`, which overwrote the intended default.

## What changed

- Avoid passing `None` for `repetition_context_size` when the request omits it.
- Preserve the default value of `20` defined in the generation layer.
- Ensure omitted request parameters do not override existing defaults.

## Why

introduced a bug : #1214 

f210f6c30f44 changed _build_gen_args() to always pass request values into GenerationArguments:

```python
  repetition_context_size=getattr(request, "repetition_context_size", None)
```

That is the bug. For Pydantic request models, omitted optional fields still exist as attributes with value None. So this does not mean **use the dataclass default if absent** --- it means **explicitly override the dataclass default with None.**

### DIFF
--- a/mlx_vlm/server/app.py
+++ b/mlx_vlm/server/app.py
@@ -13,7 +13,11 @@ from fastapi.middleware.cors import CORSMiddleware
 from huggingface_hub import scan_cache_dir
 
 from .. import apc as _apc
-from ..generate import DEFAULT_TEMPERATURE, DEFAULT_TOP_P
+from ..generate import (
+    DEFAULT_REPETITION_CONTEXT_SIZE,
+    DEFAULT_TEMPERATURE,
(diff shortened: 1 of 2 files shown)


GeneCodeSavvy:
@Blaizzy this is a simple fix. Can we get this merged?

thank you

Blaizzy (approved):
LGTM, thanks @GeneCodeSavvy!

PR #1229, opened 2026-05-27
Restructure generation into AR and diffusion engines

Blaizzy:
## Summary
- Add a generation package with shared generation state, an autoregressive wrapper, and diffusion generation support.
- Route diffusion models through the diffusion generator while keeping generate.py as the top-level orchestrator.
- Add static prefix KV cache support and preserve diffusion generation config loading.

## Validation
- python3 -m compileall -q mlx_vlm/generate.py mlx_vlm/generation mlx_vlm/models/cache.py mlx_vlm/utils.py

## Notes
- The local pre-commit hook could not run because pre-commit is not installed in this environment.


mlx_vlm/models/cache.py
```diff
@@ -1162,3 +1162,98 @@ def trim(self, n):
         n = min(self.offset, n)
         self.offset -= n
         return n
+
+
+class StaticPrefixKVCache(_BaseCache):
```
lucasnewman, line 1167: Where is this used? In an upcoming PR?

lucasnewman (approved):
LGTM

mlx_vlm/models/cache.py
Blaizzy: Its upcoming. I can remove it but I'm gonna use it soon

PR #1232, opened 2026-05-28
Add support for FLUX.2 base & klein models

lucasnewman:
This also unifies the image model loading path to use the same primitives as the VL loading path, which is the extent possible given that there's no top-level model type available in the image model repos right now.

```sh
python -m mlx_vlm generate_image \
--model black-forest-labs/FLUX.2-klein-9B \
--prompt "a glass orb etched with MLX VLM in a quiet studio, product photograph, crisp details, soft natural light" \
--size 1024x1024 --steps 4 --guidance 3.5 --seed 1234 --output test.png
```

test


Blaizzy (approved):
LGTM, thanks!

PR #1235, opened 2026-05-29
Improve masked diffusion text generation controls

Blaizzy:
## Summary

- Adds exposed masked diffusion text-generation controls for block length, transfer target, thresholds, and post-fill refinement.
- Routes masked diffusion text generation by `mask_token_id` instead of hardcoding a concrete model type.
- Uses model-card quality defaults for LLaDA-style masked diffusion generation and avoids forcing extra low-confidence tokens, reducing comma/repetition artifacts when `--num-to-transfer` is greater than 1.

## Root Cause

The previous transfer fallback forced `num_to_transfer` tokens even when fewer tokens cleared the confidence threshold. With aggressive transfer settings this could commit low-confidence punctuation or repeated filler tokens, and the stricter edit threshold made those artifacts hard to repair.

## Validation

- `uv run pytest mlx_vlm/tests/test_diffusion_models.py -q`
- `uv run python -m mlx_vlm.generate --model inclusionAI/LLaDA2.1-mini --prompt "Write a long story" --temperature 0.0 --enable-thinking --max-tokens 500 --num-to-transfer 2`

The long-story smoke test generated clean text without the previous comma/repetition collapse; measured generation speed was about 16 tok/s for 500 tokens.


lucasnewman (approved):
LGTM

PR #1236, opened 2026-05-29
Add support for image editing with FLUX.2 models

lucasnewman:
This extends the image generation support to image editing from reference images, along with CLI image editing and `/v1/images/edits` OpenAI-compatible API support. I also added support for the kv caching variant of the flux models, e.g. `black-forest-labs/FLUX.2-klein-9b-kv`.


mlx_vlm/models/bonsai/README.md
```diff
@@ -22,12 +22,6 @@ snapshot path as `--model`.
 Supported image sizes are 256 to 2048 pixels per side, and both dimensions must
 be multiples of 16.
 
-## Install
```
Blaizzy, line 25: Was this intentional

mlx_vlm/edit_image.py
Blaizzy, line 1: I don't have a strong opinion on this but

Do we need this file here? since it's under generate? What about tasks maybe?

Blaizzy (approved):
LGTM, I just have a couple nits

mlx_vlm/edit_image.py
lucasnewman: Yeah, I think that makes more sense and opens the door for interesting other tasks later. Let me make a change and we can see how it feels.

mlx_vlm/models/bonsai/README.md
lucasnewman: It seemed redundant on the model README itself but I'm fine with keeping it. I'll put it back in.

mlx_vlm/edit_image.py
lucasnewman: @Blaizzy Take a look at the latest and see if you like that approach better.

Blaizzy (approved):
LGTM!

mlx_vlm/edit_image.py
Blaizzy: It's great! 🔥

Issue #1241, opened 2026-05-30
bicubic_interpolate uses a=-0.5 (Keys') instead of PyTorch's a=-0.75 — degrades MoonViT pos-emb parity


beshkenadze:
## Summary

`mlx_vlm/models/kernels.py::bicubic_interpolate` uses the **Keys' cubic** coefficient `a = -0.5`, while **PyTorch's `F.interpolate(mode="bicubic")` uses `a = -0.75`**. This causes a measurable numerical divergence wherever a learnable position embedding is bicubic-interpolated to a non-native grid — i.e. the **MoonViT** vision tower (`kimi_vl`, and the LocateAnything-3B port in #PR).

## Evidence (numerical parity vs the HF/PyTorch reference)

MLX modules fed the **identical** `pixel_values` as the HF model (RTX 4090, `transformers==4.51`, fp32):

| scenario | `vision_model` cos | `mlp1` cos | mean\|Δ\| |
|---|---|---|---|
| grid 64×64 (== pos-emb native size → **no interpolation**) | **0.999937** | **0.999898** | ~1e-3 |
| grid 36×46 (pos-emb interpolated 64×64 → 36×46) | 0.9894 | 0.9971 | ~1.6e-2 |

When interpolation is bypassed the port is bit-faithful (cos ≈ 0.99994). The entire residual on non-native grids comes from this coefficient mismatch. (Connector cos > vision cos because its leading `LayerNorm` partly cancels the additive pos-emb error.)

## Root cause

`mlx_vlm/models/kernels.py`:
- `_cubic_weight` — `a = -0.5` (docstring: "Keys' cubic")
- Metal `cubic_kernel` — `const float a = -0.5f;`

PyTorch uses `A = -0.75` for `mode="bicubic"` (`align_corners=False`, `antialias=False` defaults; ATen `upsample_bicubic2d`, `cubic_convolution1/2`).

## Proposed fix

Switch the cubic coefficient to `-0.75` in **both** the pure-MLX path (`_cubic_weight`) and the Metal kernel (`cubic_kernel`).

## ⚠️ Needs cross-model verification before merge

`bicubic_interpolate` is shared by ~8 consumers; changing `a` affects all of them:
`kimi_vl`, `kimi_k25`, `lfm2_vl`, `gemma3n`, `phi4_siglip`, `deepseekocr/sam`, `rfdetr`, and the incoming `locateanything`.

[...]

PR #1239, opened 2026-05-29
Add Nemotron Labs Diffusion model

Blaizzy:
## Summary
- Add the Nemotron Labs Diffusion text model implementation with AR, diffusion, and linear self-speculative generation support.
- Add prompt formatting support and generic --gen-kwargs plumbing for model-specific generation options.
- Add focused diffusion model tests and a model README with usage examples.

## Validation
- uv run python -m compileall -q mlx_vlm/models/nemotron_labs_diffusion mlx_vlm/generate/dispatch.py mlx_vlm/utils.py
- uv run python -m pytest mlx_vlm/tests/test_diffusion_models.py mlx_vlm/tests/test_prompt_utils.py -q

PR #1242, opened 2026-05-30
Add NVIDIA LocateAnything-3B (MoonViT + Qwen2.5, AR + Parallel Box Decoding)

beshkenadze:
## Summary

Ports [`nvidia/LocateAnything-3B`](https://huggingface.co/nvidia/LocateAnything-3B) — a visual-grounding VLM (object detection / referring-expression grounding / pointing / GUI & text localization) — into mlx-vlm so it runs on Apple Silicon via `mlx_vlm.generate`.

**Architecture:** MoonViT-SO-400M vision tower (shared with Kimi-VL) + Qwen2.5-3B text backbone + a 2-layer MLP connector (`mlp1`). Output is structured coordinate tokens, e.g. `<ref>remote</ref><box><64><152><273><244></box>`, with coordinates quantized to `<0>`…`<1000>` (normalized).

## What's implemented

- **Model package** `mlx_vlm/models/locateanything/`:
  - `config.py` — `VisionConfig` (moonvit) / `TextConfig` (qwen2) / `ModelConfig` with grounding token ids.
  - `vision.py` — MoonViT tower (2D RoPE, per-image block attention, 2×2 patch merge), ported from `kimi_vl` and reconciled to LocateAnything's weight names, `PytorchGELUTanh` activation, and LayerNorm eps.
  - `language.py` — standard Qwen2.5-3B causal LM (1D RoPE, GQA 16/2, tied embeddings) + the non-causal "magi" block mask for PBD.
  - `locateanything.py` — `mlp1` connector + `Model` (vision → projector → scatter at `<IMG_CONTEXT>`) + `sanitize()`.
  - `pbd.py` — Parallel Box Decoding (MTP block decoder + bbox decode utils).
  - `image_processing_locateanything.py` / `processing_locateanything.py` + `chat_template.json` (preprocessing matched bit-for-bit to the HF reference — bicubic ceil-resize, not center-crop).
- `prompt_utils.py` — register `locateanything` → `LIST_WITH_IMAGE_FIRST`.
- `generate/dispatch.py` — a small **additive, opt-in** hook routing `fast`/`hybrid` to PBD (gated on `model_type == "locateanything"`); `slow` and every other model are unchanged.
- `tests/test_locateanything.py` — config, vision/language/full-forward shapes, sanitize coverage, magi-mask, PBD decode utils, max-tokens, image-input handling (**22 tests, all green**).

## Decoding modes

| mode | description | throughput* |
|------|-------------|-------------|
| `slow` (default) | pure autoregressive | 1× |
| `fast` | Parallel Box Decoding (MTP, parallel blocks) | ~2× |
| `hybrid` | PBD with AR fallback on format irregularity | ~2× |

\* 16-token COCO run. All three modes produce **byte-identical** grounding output.

## Parity vs the PyTorch (CUDA) reference

Verified the MLX port numerically against the original HF/PyTorch model on an **RTX 4090** (WSL, `transformers==4.51`, fp32). The HF `vision_model` + `mlp1` were dumped on identical `pixel_values`, then fed to the MLX modules in fp32 and compared:

| stage | grid 64×64 (no pos-emb interp) | grid 36×46 (with interp) |
|-------|-------------------------------|--------------------------|
| `vision_model` | **cos 0.999937**, mean\|Δ\| ≈ 1e-3 | cos 0.9894 |
| `mlp1` (connector) | **cos 0.999898** | cos 0.9971 |

- **The port's math is numerically faithful** (cos ≈ 0.99994 on identical inputs).
- The entire residual on non-square grids comes from one op: the learnable 2D pos-emb **bicubic interpolation** — MLX's shared `bicubic_interpolate` kernel uses `a = -0.5` vs PyTorch's `a = -0.75`. It is localized to the additive pos-emb (hence connector cos > vision cos: the connector's `LayerNorm` partly cancels it), affects every MLX MoonViT port (incl. `kimi_vl`), and does **not** change the grounding output. Tracked in #1241.
- **PBD parity:** by the model's design invariant (hybrid falls back to AR for consistency), `fast`/`hybrid` are verified **byte-identical** to the `slow` (AR) path on the same input — so the verified AR path is itself the oracle for the parallel path.

## Verification

- **Unit tests:** `python -m unittest mlx_vlm.tests.test_locateanything` → 22 passed.
- **End-to-end** (real weights, Apple Silicon):
  ```
  python -m mlx_vlm.generate --model nvidia/LocateAnything-3B \
    --image http://images.cocodataset.org/val2017/000000039769.jpg \
    --prompt "Detect all objects in the image." --max-tokens 128 --temperature 0.0
  ```
  → `<ref>remote</ref><box><64><152><273><244></box><box><522><160><578><390></box>` (boxes match the two remotes); prompt token count matches the HF processor exactly.
- **Large images:** fixed a vision-attention OOM — a single image's all-True block mask was passed explicitly, forcing `mx.fast.scaled_dot_product_attention` off the flash kernel and materializing a dense `[1,heads,S,S]` score tensor (OOM on big frames). Now `mask=None` for one image (flash, O(N) memory); multi-image keeps the block-diagonal mask.
- **Codex CLI review** (`codex exec review`): no P0/P1; two P2 edge-cases (PBD `max_tokens`, `mx.array` image input) fixed.

## Blast radius

Everything PBD/model-specific lives in the `locateanything` package; the only shared edits are one additive line in `prompt_utils.py` and one additive, gated hook in `dispatch.py`. No other model, the vision tower, the processor, or shared SDPA is affected; `slow` AR remains the default.

## Quantized weights

MLX builds published to mlx-community: [bf16](https://huggingface.co/mlx-community/LocateAnything-3B-bf16), [8bit](https://huggingface.co/mlx-community/LocateAnything-3B-8bit), [4bit](https://huggingface.co/mlx-community/LocateAnything-3B-4bit) (mixed 4/8-bit — pure 4-bit degrades the tied coordinate-token embedding).


mlx_vlm/models/locateanything/chat_template.json
```diff
@@ -0,0 +1,3 @@
+{
```
lucasnewman: Shouldn't this be in the model repo?

mlx_vlm/generate/dispatch.py
```diff
@@ -820,6 +820,50 @@ def stream_generate(
         )
         return
 
+    # LocateAnything Parallel Box Decoding (opt-in). Only fast/hybrid route here;
```
lucasnewman: The dispatching here is really for the AR / diffusion / image gen paths, not for model specific entrypoints. Can this just be an option on generate on the model itself?

mlx_vlm/models/locateanything/chat_template.json
beshkenadze: Good call — removed the bundled copy. It's loaded from the model repo at runtime (`base.load_chat_template` / the processor's `from_pretrained`), and the HF repo + the mlx-community quants all ship it, so the package copy was unused.

mlx_vlm/generate/dispatch.py
beshkenadze: Agreed the `model_type` hardcode didn't belong here. Reworked it into a generic, capability-based route: a `--generation-mode {slow,fast,hybrid}` flag plus a `hasattr(model, 'pbd_generate')` check in `stream_generate` — consistent with how the dispatcher already routes by model capability (`output_modality`, `hasattr(model, 'encode_image')`). `slow`/AR is unchanged for every other model, and `generation_mode` is consumed so it never propagates to `generate_step`. Happy to move it elsewhere if you'd prefer a different mechanism.

mlx_vlm/models/locateanything/processing_locateanything.py
Blaizzy, line 1: Missing save_pretrained() method

beshkenadze: I'll check it out, since it ran without any errors on a separate Mac mini.

mlx_vlm/generate/dispatch.py
Blaizzy, line 1: I would revert this this, because its model specific, only one model has this model so it's not meant to be here.

In #1239 I'm adding gen-kwargs which you will be able to use here.

beshkenadze: Ok, np.

Blaizzy:
@lucasnewman it's now ready to review and merge

Thanks @beshkenadze!

PR #1243, opened 2026-05-30
fix(kernels): match PyTorch bicubic coefficient (-0.75 non-AA / -0.5 AA) — closes #1241

### LINKED ISSUE #1241: bicubic_interpolate uses a=-0.5 (Keys') instead of PyTorch's a=-0.75 — degrades MoonViT pos-emb parity

beshkenadze:
## Summary

Fixes #1241. `mlx_vlm/models/kernels.py::bicubic_interpolate` used Keys' cubic coefficient `a = -0.5` for the plain (non-antialiased) path, which diverges from PyTorch's `F.interpolate(mode="bicubic")` (`a = -0.75`). This caused a measurable parity gap wherever a learnable position embedding is bicubic-interpolated to a non-native grid — i.e. the **MoonViT** vision tower (`kimi_vl`, and the incoming LocateAnything-3B port).

## Why two coefficients

PyTorch uses **different** cubic coefficients depending on `antialias`:

| path | coefficient | matches |
|------|-------------|---------|
| `F.interpolate(mode="bicubic")` (default) | **a = -0.75** | ATen `upsample_bicubic2d` |
| `F.interpolate(..., antialias=True)` | **a = -0.5** | Pillow's bicubic |

The antialiased path uses `a = -0.5` on **every** resized axis whenever `antialias=True` — including pure upscales and mixed up/down resizes (verified empirically below). Support is only *widened* on the axes that are actually downsampled.

The fix selects the coefficient on the `antialias` flag, in both the Metal kernel (`cubic_kernel` / `cubic_kernel_antialias`, plus the `antialias_flag` selector so AA upscales aren't routed to the non-AA kernel) and the pure-MLX `_cubic_weight` fallback.

> The kernel previously used `-0.5` everywhere, so a naïve flip of the constant to `-0.75` would have **regressed** the `antialias=True` consumers (`aya_vision`, `phi4mm`, `rfdetr`, `multi_modality`). Both that regression and a follow-up AA-upscale regression were caught by Codex review and fixed before this PR was pushed.

## Verification — parity vs the PyTorch reference

Measured `bicubic_interpolate` against `torch.nn.functional.interpolate(mode="bicubic", align_corners=False)` on identical fp64 inputs (cosine similarity of the outputs):

| regime | before (a=-0.5) | after |
|--------|-----------------|-------|
| non-AA, downsample 64→36×46 | 0.997701 | **0.999872** |
| AA, downsample 64→36×46 | — | **1.000000** |
| AA, upsample 16→32×40 | — | **1.000000** |
| AA, mixed 16→10×40 | — | **1.000000** |

- The non-AA coefficient now matches PyTorch (`a = -0.75`); the small remaining non-AA residual (cos ~0.9996) is pre-existing edge/border handling in MLX's bicubic, unrelated to this coefficient change.
- The AA path matches PyTorch **bit-for-bit** (mean|Δ| ≈ 3e-7).
- `kimi_vl` vision forward (the primary MoonViT consumer) is unaffected.

## Affected consumers

`bicubic_interpolate` is shared by the MoonViT/SigLIP pos-emb resizes in `kimi_vl`, `kimi_k25`, `lfm2_vl`, `gemma3n`, `phi4_siglip`, `phi4mm`, `aya_vision`, `rfdetr`, `multi_modality`, `deepseekocr/sam`. Non-AA consumers gain the corrected `-0.75`; AA consumers keep `-0.5` (now applied on all axes, including upscales).

Surfaced while porting `nvidia/LocateAnything-3B` (MoonViT + Qwen2.5); the model port verified the MoonViT math as faithful *except* for this bicubic divergence.

### DIFF
--- a/mlx_vlm/models/kernels.py
+++ b/mlx_vlm/models/kernels.py
@@ -110,12 +110,17 @@ def nearest_interpolate(x, size=None, scale_factor=None):
     return result
 
 
-def _cubic_weight(t):
-    """Bicubic kernel with a=-0.5 (Keys' cubic), pure MLX."""
+def _cubic_weight(t, a=-0.75):
+    """Cubic convolution kernel.
+
+    PyTorch's ``F.interpolate(mode="bicubic")`` uses ``a=-0.75`` (ATen
+    ``upsample_bicubic2d``) for the non-antialiased path, but its antialiased
+    path matches Pillow, which uses ``a=-0.5`` (Keys' cubic). The caller selects
+    the coefficient per axis based on whether antialiasing is active.
+    """
     at = mx.abs(t)
     at2 = at * at
     at3 = at2 * at
-    a = -0.5
     w1 = (a + 2.0) * at3 - (a + 3.0) * at2 + 1.0
     w2 = a * at3 - 5.0 * a * at2 + 8.0 * a * at - 4.0 * a
     return mx.where(at <= 1.0, w1, mx.where(at < 2.0, w2, mx.zeros_like(t)))
@@ -151,7 +156,9 @@ def _bicubic_interpolate_mlx(x, out_h, out_w, align_corners=False, antialias=Fal
         offsets = mx.arange(n_taps, dtype=mx.int32)  # (taps,)
         pix = start[:, None] + offsets[None, :]  # (out, taps)
         dist = coords[:, None] - pix.astype(mx.float32)  # (out, taps)
-        w = _cubic_weight(dist / fs)
+        # Antialiased resize matches Pillow (a=-0.5) on every axis; plain
+        # bicubic uses PyTorch's a=-0.75.
+        w = _cubic_weight(dist / fs, -0.5 if antialias else -0.75)
         # Zero out-of-bounds
         mask = (pix >= 0) & (pix < in_size)
         w = w * mask
@@ -220,7 +227,11 @@ def bicubic_interpolate(
     # Calculate antialiasing parameters
     # PyTorch uses support = 2.0 for bicubic when antialiasing
     support = 2.0
-    antialias_flag = 1.0 if (antialias and (scale_h < 1.0 or scale_w < 1.0)) else 0.0
+    # PyTorch's antialiased bicubic uses the Pillow coefficient (a=-0.5) for
+    # every resized axis whenever antialias is requested — including pure
+    # upscales and mixed up/down resizes. Support is only widened for the axes
+    # that are actually downsampled (filter_scale_* below).
+    antialias_flag = 1.0 if antialias else 0.0
 
     # When downsampling with antialias, PyTorch expands the filter support
     if antialias and scale_h < 1.0:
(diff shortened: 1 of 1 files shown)


lucasnewman:
@beshkenadze Thanks for the analysis! Nice find.

lucasnewman (approved):

PR #1247, opened 2026-05-30
fix(convert): don't abort conversion when a processor lacks save_pretrained

beshkenadze:
## Summary

`mlx_vlm.convert` aborts when the model's processor doesn't implement `save_pretrained`. Many custom processors registered via `install_auto_processor_patch` are plain classes (e.g. moondream3's `Moondream3Processor`), and `convert()` calls `processor.save_pretrained(mlx_path)` immediately **before** `save_config()`. So the missing method raises `AttributeError`, `save_config()` never runs, and the output directory is left with quantized weights but a `config.json` that has **no `quantization` block** — the resulting checkpoint then fails to load (the loader builds full-precision layers against quantized weights).

## Repro

```
mlx_vlm.convert --hf-path <moondream3> --mlx-path out -q --q-bits 4
# ... [INFO] Quantizing -> save_weights OK ...
# AttributeError: 'Moondream3Processor' object has no attribute 'save_pretrained'
# -> out/config.json has no "quantization"; load() then fails.
```

## Fix

Route the processor save through a small `_save_processor()` helper:

- If the processor has **no** `save_pretrained`, skip it with a warning (its files are non-critical — e.g. moondream3 loads its tokenizer from a separate repo) and let conversion continue to `save_config()`.
- Otherwise save normally and **let genuine save failures propagate** — don't mask serialization/permission errors behind a broken artifact.

`save_config()` ordering is unchanged (still last), so the quantization-bearing `config.json` stays authoritative even for processors whose `save_pretrained` writes their own config files.

## Tests

Adds three unit tests in `mlx_vlm/tests/test_utils.py` for the helper: missing method (skipped), present method (called once with the path), and a failing save (propagates).

### DIFF
--- a/mlx_vlm/convert.py
+++ b/mlx_vlm/convert.py
@@ -71,6 +71,29 @@ def _preserve_existing_deepseek_v4_quantization(
     config["quantization_config"] = quantization
 
 
+def _save_processor(processor, mlx_path: Path) -> None:
+    """Save processor files, tolerating processors without ``save_pretrained``.
+
+    Some custom mlx-vlm processors (e.g. moondream3's ``Moondream3Processor``)
+    are plain classes registered via ``install_auto_processor_patch`` and don't
+    implement ``save_pretrained``. Skipping the save in that case lets conversion
+    proceed to ``save_config``, which writes the model ``config.json`` (including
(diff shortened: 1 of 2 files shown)


Blaizzy:
Thanks Aleksandr, but this is not a convert issue, it's a missing method on the processor.

PR #1248, opened 2026-05-30
Fix cache merging for batched quantized caches

### LINKED ISSUE #1231: KV-cache concatenate ValueError across consecutive requests with differently-shaped prompts (v0.5.0)
## Environment

- **mlx-vlm:** 0.5.0
- **Model:** `mlx-community/Qwen3.5-122B-A10B-4bit`
- **Host:** Mac Studio M2 Ultra, 256 GB unified memory
- **Server invocation:** `python -m mlx_vlm.server --model <path> --host 0.0.0.0 --port 18080 --max-tokens 8192 --prefill-step-size 1024 --kv-bits 4 --kv-group-size 64 --vision-cache-size 20`
- **Client pattern:** OpenAI-compatible POSTs to `/v1/chat/completions` from a small fleet of background agents. Single-concurrency reverse-proxy in front, so requests serialize. Each request carries the same large system+rules prefix (~3-5 k tokens) followed by a per-request candidate tail; **no two prompts are identical** but they share a long common prefix.

## Symptom

Periodically, a chat-completion request returns HTTP 500 with:

```
Generation failed: [concatenate] All the input array dimensions must
match exactly except for the concatenation axis. However, the provided
shapes are (1,2,13257,32), (1,2,13312,32), and the concatenation axis is 0.
```

The shape pair varies between failures. From a 24 h window we see four distinct shape classes:

| Shape A | Shape B | Axis |
| --- | --- | --- |
| `(1,2,13257,32)` | `(1,2,13312,32)` | 0 |
| `(1,2,13312,32)` | `(1,2,13208,32)` | 0 |
| `(1,3,12288)` | `(2,10,12288)` | 1 |
| `(1,3,12288)` | `(2,15,12288)` | 1 |
| `(1,3,12288)` | `(2,1024,12288)` | 1 |
| `(2,3,12288)` | `(3,1024,12288)` | 1 |

The traceback ends at:

```python
File "<mlx-vlm>/.../cache.py", line ?, in update_and_fetch
    self.keys = tuple(mx.concatenate([s, o]) for s, o in zip(sk, ok))
```

(Happy to paste the full traceback in the comment thread; redacted here for brevity.)

## Reproduction

Send two POSTs back-to-back to `mlx_vlm.server` with prompts that share a long common prefix but differ in length. The failure isn't 100% deterministic in our workload (~22 hits over ~24 h across thousands of requests), but it's frequent enough to materially affect a multi-agent setup where each agent issues 10s-of-1000s of requests per day. We see it most often when:

- One request finishes generation, KV cache state is retained.
- The next request shares the same system+rules prefix (tokenizes identically for the first N tokens) but the user-message tail has a different token count.

The shapes suggest the cache from request N (with its sequence dim) is being concatenated with new tokens from request N+1, and the batch / sequence dims don't line up.

## What we've ruled out

- **Not concurrent requests:** our proxy enforces concurrency=1, so requests strictly serialize.
- **Not a tokenizer-config issue:** `tokenizer_config.json` is the one shipped with the model (we set `tool_parser_type=qwen3_coder` for tool-call routing but the bug fires equally on requests with no `tools` field).
- **Not specific to the tool-parser patch we apply:** the same bug surface exists on a vanilla install — only the call rate drops because our patched build sees more diverse tool-call response shapes.
- **Not OOM:** unified memory has ~25 GB headroom at peak; no swap activity.
- **Not vision-related:** the failing requests are text-only chat completions.

## What we've tried

- Bouncing the server clears cache state and resolves the immediate batch of failures; the bug returns after the cache fills again (minutes to hours, depending on request diversity).
- Lowering `--prefill-step-size` from 2048 → 1024 didn't change frequency.
- No `--max-kv-size` cap is set (we want the full 262 k native context); setting a small cap would force more frequent cache eviction but at a real throughput cost.

## Asks

1. **A `--no-prompt-cache-reuse` (or similar) CLI flag** to disable inter-request cache reuse, even at the cost of full prefill on every request. We'd accept the prefill-latency hit until a real fix lands.
2. **Or a fix in cache.py** where the shape mismatch happens — likely tightening the prefix-match check so a length-divergent request invalidates the cache rather than concatenating into it.

Happy to test patches against our workload — we have a steady 1 k+ req/day stream that reliably reproduces the failure within a few hours.

## Related

- mlx-vlm-server cache management code at `<mlx-vlm>/cache.py` — happy to link to the upstream commit in the comment thread.
- This codepath is hit on `mx.concatenate` in `update_and_fetch` — first stack trace we've found in 0.5.0; happy to bisect against 0.4.x if useful.

CC anyone you'd like in for triage.

lucasnewman:
Makes quantized batch cache merging normalize both caches to a shared sequence capacity before concatenating them, so filtered and rounded cache lengths don't hit a ValueError from mismatched shapes. Resolves https://github.com/Blaizzy/mlx-vlm/issues/1231

### DIFF
--- a/mlx_vlm/models/cache.py
+++ b/mlx_vlm/models/cache.py
@@ -281,30 +281,42 @@ class BatchQuantizedKVCache(_BaseCache):
             return
 
         max_idx = max(self._idx, other._idx)
+        self_len = 0 if self.keys is None else self.keys[0].shape[-2]
+        other_len = 0 if other.keys is None else other.keys[0].shape[-2]
+        max_size = max(max_idx, self_len, other_len)
+        ref_keys = self.keys if self.keys is not None else other.keys
+        ref_values = self.values if self.values is not None else other.values
+
+        def _empty_like(parts, batch_size: int):
+            out = []
+            for part in parts:
+                shape = list(part.shape)
+                shape[0] = batch_size
+                shape[-2] = 0
+                out.append(mx.zeros(tuple(shape), dtype=part.dtype))
+            return tuple(out)
 
         def _pad_quant(cache_obj):
             if cache_obj.keys is None:
-                return None
+                batch_size = int(cache_obj.offset.shape[0])
+                keys = _empty_like(ref_keys, batch_size)
+                values = _empty_like(ref_values, batch_size)
+            else:
+                keys = cache_obj.keys
+                values = cache_obj.values
             left = max_idx - cache_obj._idx
-            right_total = max(
-                cache_obj.keys[0].shape[-2],
-                max_idx,
-            )
-            right = right_total - cache_obj.keys[0].shape[-2] - left
+            right = max_size - keys[0].shape[-2] - left
             if right < 0:
-                trimmed_keys = tuple(
-                    k[..., : right_total - left, :] for k in cache_obj.keys
-                )
-                trimmed_values = tuple(
-                    v[..., : right_total - left, :] for v in cache_obj.values
-                )
+                trimmed_keys = tuple(k[..., : max_size - left, :] for k in keys)
+                trimmed_values = tuple(v[..., : max_size - left, :] for v in values)
                 right = 0
             else:
-                trimmed_keys = cache_obj.keys
-                trimmed_values = cache_obj.values
+                trimmed_keys = keys
+                trimmed_values = values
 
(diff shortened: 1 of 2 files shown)


Blaizzy (approved):
LGTM, thanks Lucas!

PR #1249, opened 2026-05-30
security: require patched Starlette

Thump604:
## Summary
- Add an explicit `starlette>=1.0.1` dependency floor.
- Refresh `uv.lock` so the locked server/UI stack resolves to patched Starlette.

## Why
Starlette versions below the patched BadHost floor have a Host-header URL parsing vulnerability. `mlx-vlm` already depends on FastAPI, but the project dependency file did not directly express a patched Starlette floor, and the lock file still resolved `starlette==0.50.0`.

## Validation
- Checked current server source for `request.url` / `request.url.path` usage; none found in `mlx_vlm` server code.
- `uv lock --check` passed.
- Resolver dry-run for the UI extra on Python 3.12 completed successfully:
  `uv pip install --python /tmp/mlx-vlm-badhost-resolve/bin/python --dry-run '.[ui]'`
- The resolved set selected `fastapi==0.136.3`, `gradio==6.15.2`, and `starlette==1.2.0`.
- `git diff --check` passed.

## Explicit non-claims
- This PR does not claim an application-level BadHost exploit in `mlx-vlm`.
- This PR does not change server behavior.
- This PR only hardens dependency resolution and the lock file.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,5 +10,6 @@ mlx-lm>=0.31.3
 mlx-audio>=0.4.3
 opencv-python>=4.12.0.88
 fastapi>=0.95.1
+starlette>=1.0.1
 uvicorn
 numpy
--- a/uv.lock
+++ b/uv.lock
@@ -18,15 +18,6 @@ resolution-markers = [
 ]
 
 [[package]]
-name = "aiofiles"
-version = "24.1.0"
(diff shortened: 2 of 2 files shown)


Blaizzy (approved):
LGTM, thanks!

PR #1250, opened 2026-05-30
Don't show tqdm progress bars for prefill by default

lucasnewman:
Same fix as https://github.com/Blaizzy/mlx-vlm/pull/1015 but updated for the new generation file split.


Blaizzy:
You sure we don't want to show this by default?

Blaizzy (approved):
LGTM

PR #1269, opened 2026-06-03
Fix system role normalization for /v1/messages API

### LINKED ISSUE #1264: 400 Validation Error with Claude and /v1/messages endpoint: `messages.x.role` input should be 'user' or 'assistant', got 'system'
**Describe the bug**
When using [Claude Code](https://docs.anthropic.com/en/docs/claude-code) (v2.1.160) to connect to the `mlx-vlm` local server via the `/v1/messages` endpoint, the request fails with a 400 Bad Request error. The error indicates a Pydantic validation failure where a message in the `messages` array has the role `'system'`.

**To Reproduce**
1. Start the `mlx-vlm` server with a compatible model (e.g., Qwen3.6-35B-A3B-nvfp4).
2. Run Claude Code and configure it to use the local `mlx-vlm` endpoint (`/v1/messages`).
3. Send any prompt (e.g., "Introduce the project in the current directory").
4. Observe the 400 API error.

**Expected behavior**
The API should accept the request and return a valid response. According to the [Anthropic Messages API specification](https://docs.anthropic.com/en/api/messages), the `system` prompt should be passed as a top-level `system` field, and the `messages` array should strictly contain only `user` and `assistant` roles.

**Error Logs**
```text
❯ Introduce the project in the current directory                                                                                                                                                      
  ⎿  API Error: 400 Invalid request body: 1 validation error for AnthropicRequest
     messages.1.role
       Input should be 'user' or 'assistant' [type=literal_error, input_value='system', input_type=str]
         For further information visit https://errors.pydantic.dev/2.12/v/literal_error
```

**Environment**
- `mlx-vlm` version: 549a75a
- Model used: `mlx-community/Qwen3.6-35B-A3B-nvfp4`
- Client: Claude Code v2.1.160
- OS: Mac OS 26.5

**Additional Context / Possible Cause**
The error trace shows that `AnthropicRequest` is validating a `messages` array where `messages[1].role == 'system'`. 

Since the official Anthropic API strictly forbids `role: 'system'` inside the `messages` array (it must be in the top-level `system` field), it is highly likely that `mlx-vlm`'s `/v1/messages` router is incorrectly processing the request. 

Specifically, it might be:
1. Extracting the top-level `system` field and incorrectly appending it to the `messages` list *before* passing it to the Pydantic model for validation.
2. Or, the Pydantic model `AnthropicRequest` itself has an incorrect definition that allows/forces `system` into the `messages` list during internal parsing.

Could you please check the request handling logic and Pydantic models for the `/v1/messages` endpoint?

Thank you!

lucasnewman:
Resolves https://github.com/Blaizzy/mlx-vlm/issues/1264

### DIFF
--- a/mlx_vlm/server/anthropic.py
+++ b/mlx_vlm/server/anthropic.py
@@ -100,6 +100,38 @@ def _anthropic_system_text(system: Optional[Union[str, List[Any]]]) -> Optional[
     return "\n".join(parts).strip() or None
 
 
+def _normalize_anthropic_system_messages(body: Any) -> Any:
+    if not isinstance(body, dict):
+        return body
+    messages = body.get("messages")
+    if not isinstance(messages, list):
+        return body
+
+    normalized_messages = []
+    system_parts = []
+    saw_system_message = False
+    for message in messages:
+        if isinstance(message, dict) and message.get("role") == "system":
+            saw_system_message = True
+            text = _anthropic_system_text(message.get("content"))
+            if text:
+                system_parts.append(text)
+            continue
+        normalized_messages.append(message)
+
+    if not saw_system_message:
+        return body
+
+    normalized_body = dict(body)
+    normalized_body["messages"] = normalized_messages
+    existing_system = _anthropic_system_text(normalized_body.get("system"))
+    if existing_system:
+        system_parts.insert(0, existing_system)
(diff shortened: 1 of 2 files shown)


Blaizzy (approved):
LGTM!🚀

PR #1272, opened 2026-06-03
Make sure to evaluate cached MRoPE parameters on model load

lucasnewman:
I found this issue while working on a separate bug -- the current MRoPE implementation can leave mx.arrays un-eval'ed which will crash when moving the model across threads/thread-local streams. This change does eager eval for the cache parameters so we can avoid this without failing strict weight loading.

I verified 9 MRoPE-based models still run correctly with the change, and moving the model across streams no longer crashes.


Blaizzy:
Thanks for catching it Lucas, LGTM! 

Just have one suggestion that you can test if it still fixes the issue.
How about this? https://gist.github.com/Blaizzy/33709336102d728fc13b0c2c90c121ef

lucasnewman:
Yep, that works fine, I updated the implementation.

Blaizzy (approved):
LGTM 🚀

PR #1279, opened 2026-06-03
Fix Nemotron-H Nano Omni audio crash under nvfp4 quantization

### LINKED ISSUE #1278: Nemotron-H Nano Omni audio path crashes on nvfp4 quantized models
Loading `mlx-community/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-nvfp4` (or any other nvfp4 quant of the same architecture) and sending a request that exercises the audio modality crashes inside the Parakeet encoder:

```
File ".../mlx_vlm/models/nemotron_h_nano_omni/audio.py", line 304, in __call__
    hidden_states = layer(hidden_states)
File ".../mlx/nn/layers/convolution.py", line 157, in __call__
    y = mx.conv2d(...)
ValueError: [conv] Invalid input array with type uint8. Convolution currently only supports floating point types
```

### Root cause

`Model._extract_sound_features` in `nemotron_h_nano_omni.py` derives a compute dtype for the audio mel features from the language-model head's quantization scales:

```python
lm_head = self.language_model.lm_head
compute_dtype = (
    lm_head.scales.dtype if hasattr(lm_head, "scales") else lm_head.weight.dtype
)
input_features = input_features.astype(compute_dtype)
```

This assumes `scales.dtype` is a floating-point type. That holds for the `affine`, `mxfp4`, and `mxfp8` quant modes — but **nvfp4 packs the per-group scales as bytes**, so for nvfp4 `scales.dtype == mx.uint8`. The mel features get cast to `uint8`, and the next `mx.conv2d` rejects the input.

### Scope

- Affected: any nvfp4 quant of Nemotron-H Nano Omni, audio modality only.
- Unaffected: text and vision paths (they don't go through this dtype derivation).
- Other quant modes (affine / mxfp4 / mxfp8) are fine because their scales are float.

### Minimal repro

The regression test in the linked PR reproduces it without needing model weights — it attaches a `uint8` `.scales` attribute to a synthetic `lm_head` and runs the audio path. Pre-fix it raises the same `ValueError`.

### Proposed fix

Fall back to `mx.bfloat16` (the model's declared `torch_dtype`) when `scales.dtype` is an integer packing type. PR with the two-line guard and a regression test is opened in the next message.

txdadlab:
Resolves #1278.

## Bug

`Model._extract_sound_features` casts the audio mel features to `lm_head.scales.dtype`. For nvfp4-quantized models that is `uint8`, and the subsampling `Conv2d` then rejects the input:

```
ValueError: [conv] Invalid input array with type uint8.
Convolution currently only supports floating point types.
```

Crash trace ends at `ParakeetEncoderSubsamplingConv2D.__call__` (audio.py:304).

## Fix

When `lm_head.scales` is an integer packing type (`uint8` for nvfp4, `uint32` also covered for other packed modes), fall back to `mx.bfloat16` — which matches the model's declared `torch_dtype`. Float-scaled quant modes (`affine`, `mxfp4`, `mxfp8`) keep their previous behavior.

```python
lm_head = self.language_model.lm_head
if hasattr(lm_head, "scales"):
    s_dtype = lm_head.scales.dtype
    compute_dtype = (
        mx.bfloat16 if s_dtype in (mx.uint8, mx.uint32) else s_dtype
    )
else:
    compute_dtype = lm_head.weight.dtype
input_features = input_features.astype(compute_dtype)
```

## Test

New regression test `test_audio_path_handles_nvfp4_uint8_lm_head_scales` attaches a `uint8` `.scales` attribute to a synthetic `lm_head` and exercises `Model.get_input_embeddings` with audio features. The test fails on the original code with the exact `Invalid input array with type uint8` error, and passes with the fix.

```
$ uv run --with pytest --with-editable . python -m pytest \
    mlx_vlm/tests/test_nemotron_h_nano_omni.py -v
...
test_sound_feature_extractor_shapes_and_masks PASSED
test_sound_encoder_and_projection_tiny_smoke PASSED
test_model_merges_sound_features_into_input_embeddings PASSED
test_audio_path_handles_nvfp4_uint8_lm_head_scales PASSED   ← new
test_model_rejects_sound_token_feature_count_mismatch PASSED
test_sanitize_audio_and_projection_weights PASSED
6 passed
```

## Manual repro

Loaded `mlx-community/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-nvfp4` through `waybarrios/vllm-mlx` 0.31.x and sent a 30-second WAV clip via `/v1/chat/completions`. Pre-fix: 500 error with the crash above. Post-fix: clean transcription end-to-end (a 16 kHz mono WAV of `say "The quick brown fox..."` came back verbatim).

## Scope

Two files, +55 / -3 lines. The fix is localized to the dtype derivation in `_extract_sound_features`. No public API changes.

### DIFF
--- a/mlx_vlm/models/nemotron_h_nano_omni/nemotron_h_nano_omni.py
+++ b/mlx_vlm/models/nemotron_h_nano_omni/nemotron_h_nano_omni.py
@@ -122,9 +122,17 @@ class Model(nn.Module):
         if not isinstance(input_features, mx.array):
             input_features = mx.array(input_features)
         lm_head = self.language_model.lm_head
-        compute_dtype = (
-            lm_head.scales.dtype if hasattr(lm_head, "scales") else lm_head.weight.dtype
-        )
+        if hasattr(lm_head, "scales"):
+            s_dtype = lm_head.scales.dtype
+            # nvfp4 packs the per-group scales as uint8 (and other packed
+            # quant modes can use uint32), so the scales' dtype is not a
+            # usable *compute* dtype for the audio features. Fall back to
+            # bfloat16, which matches the model's declared torch_dtype.
+            compute_dtype = (
+                mx.bfloat16 if s_dtype in (mx.uint8, mx.uint32) else s_dtype
+            )
+        else:
+            compute_dtype = lm_head.weight.dtype
         input_features = input_features.astype(compute_dtype)
 
         if feature_attention_mask is not None and not isinstance(
--- a/mlx_vlm/models/nemotron_h_nano_omni/nemotron_h_nano_omni.py
+++ b/mlx_vlm/models/nemotron_h_nano_omni/nemotron_h_nano_omni.py
@@ -128,9 +128,7 @@ class Model(nn.Module):
             # quant modes can use uint32), so the scales' dtype is not a
             # usable *compute* dtype for the audio features. Fall back to
             # bfloat16, which matches the model's declared torch_dtype.
-            compute_dtype = (
-                mx.bfloat16 if s_dtype in (mx.uint8, mx.uint32) else s_dtype
-            )
+            compute_dtype = mx.bfloat16 if s_dtype in (mx.uint8, mx.uint32) else s_dtype
         else:
             compute_dtype = lm_head.weight.dtype
         input_features = input_features.astype(compute_dtype)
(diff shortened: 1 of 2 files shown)


lucasnewman:
@txdadlab Can you run `pre-commit run --all` to format and clear the tests?

lucasnewman (approved):
Thanks!

PR #1282, opened 2026-06-04
Fix multi-image prompts in get_rope_index for remaining MRoPE families

scyyh11:
## Summary

The qwen2_vl-derived `get_rope_index` implementations count vision blocks with:

```python
vision_start_indices = mx.sum(
    mx.where(
        input_ids == vision_start_token_id,
        mx.arange(input_ids.shape[0]),
        mx.zeros_like(input_ids),
    )
)
vision_tokens = input_ids[vision_start_indices + 1]
image_nums = (vision_tokens == image_token_id).sum().item()
```

`mx.sum` collapses **all** vision-start positions into a single scalar which is a port of `torch.argwhere` that only works when the prompt contains exactly one vision block. With two or more images, `input_ids[sum + 1]` reads an arbitrary token, `image_nums` collapses to 0 or 1, and every image after the first silently receives flat text positions, corrupting M-RoPE `position_ids` and `mrope_position_deltas` (no exception is raised).

#1135 already fixed this for `qwen3_vl` and `qwen3_5` (mask-filtered Python token scan + compact-position scatter). This PR ports that exact fix to the seven families still carrying the original pattern:

`glm4v`, `glm4v_moe`, `paddleocr_vl`, `qwen2_5_vl`, `qwen2_vl`, `qwen3_omni_moe`, `qwen3_vl_moe`

### Concrete example

For a two-image prompt `[1, 2, VS, IMG×4, 5, VS, IMG×4, 7]` with grids `[[1,4,4], [1,4,4]]` (merge size 2), the two vision-start indices 2 and 8 are summed to 10; `input_ids[11]` happens to be an image token, so only one image is counted — the second image's tokens get sequential text positions instead of the (t, h, w) grid pattern, and the rope delta is wrong.

## Changes

- Replace the `mx.sum` vision-start collapse with the #1135 token scan (mask-filtered `input_tokens`, list-comprehension `vision_tokens`, Python counts) — identical code to qwen3_vl/qwen3_5.
- Port the matching compact-position scatter and delta semantics (`compact_max_position + 1 - len(input_tokens)`).
- Add `TestMultiImageMRoPE`: a two-image prompt (plus a left-padded variant) is checked against hand-derived HF-reference positions across **all nine** MRoPE families — the seven fixed here plus qwen3_vl/qwen3_5 as references — pinning cross-family consistency.

## Testing

- `pytest mlx_vlm/tests/test_models.py::TestMultiImageMRoPE mlx_vlm/tests/test_models.py::TestChunkedPrefillRoPE` → **6 passed, 18 subtests passed** (the chunked-prefill tests exercise the patched glm4v/glm4v_moe single-image path, confirming no regression).
- Single-image positions/deltas are unchanged from the previous behavior (which was correct for exactly one image).
- `black` 24.2.0 / `isort` 5.13.2 (pre-commit pins): no changes.

### DIFF
--- a/mlx_vlm/models/glm4v/language.py
+++ b/mlx_vlm/models/glm4v/language.py
@@ -306,21 +306,20 @@ class LanguageModel(nn.Module):
             )
             image_index, video_index = 0, 0
             for i, input_ids in enumerate(total_input_ids):
-                input_ids = mx.where(
-                    attention_mask[i] == 1, input_ids, mx.zeros_like(input_ids)
-                )
+                row_mask = attention_mask[i].tolist()
+                input_tokens = [
+                    token
+                    for token, keep in zip(input_ids.tolist(), row_mask)
+                    if keep == 1
+                ]
                 image_nums, video_nums = 0, 0
-                vision_start_indices = mx.sum(
-                    mx.where(
-                        input_ids == vision_start_token_id,
-                        mx.arange(input_ids.shape[0]),
-                        mx.zeros_like(input_ids),
-                    )
-                )
-                vision_tokens = input_ids[vision_start_indices + 1]
-                image_nums = (vision_tokens == image_token_id).sum().item()
-                video_nums = (vision_tokens == video_token_id).sum().item()
-                input_tokens = input_ids.tolist()
+                vision_tokens = [
+                    input_tokens[idx + 1]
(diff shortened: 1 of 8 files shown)


lucasnewman (approved):
LGTM, thanks!

PR #1284, opened 2026-06-04
feat(mla): FP8 KV cache for MLA models with absorbed-decode Metal kernel

popfido:
## FP8 KV cache for MLA models (vLLM-style) + absorbed-decode Metal kernel

Adds an **opt-in, off-by-default** FP8 (e4m3, per-128-element-group) KV cache for
MLA-attention models (DeepSeek-V2/V3 family), served through the **absorbed-MLA** decode form.
Wired into `deepseek_vl_v2`; the default path is byte-identical when the flag is off.

### Motivation

mlx-vlm currently **expands** MLA back into a full per-head MHA cache before storing it, which
throws away MLA's entire memory advantage. Caching the `kv_lora_rank` latent in FP8 + the bf16
RoPE key reproduces vLLM/SGLang's DeepSeek-V3 layout (`512` fp8 + `4×4` fp32 scales + `64×2`
bf16 = **656 B/token**).

| DeepSeek-V2 (60 layers, 128 heads) | KV bytes/token |
|---|--:|
| current mlx-vlm — expanded bf16 | ~4.8 MB |
| absorbed bf16 latent | 67.5 KB |
| **absorbed FP8 latent (this PR)** | **38.4 KB** |

→ **~43% smaller than the bf16 latent**, **~125× smaller than the current expanded cache**.
At 128K context (batch 1) the latent cache drops **9.06 GB → 5.16 GB**; vs the expanded cache,
long context was not reachable at all.

### How it works

- The `kv_lora_rank` latent is FP8 (e4m3) quantized with **per-128-element group scales**
  (absmax/448); the RoPE key stays in the model dtype.
- Decode runs in the **absorbed** form (`kv_b_proj` folded into `embed_q`/`unembed_out`, latent
  serves as both K and V — MQA), derived lazily from `kv_b_proj` including quantized weights —
  same construction as mlx-lm's `deepseek_v3`.
- Three decode strategies (`MLX_VLM_MLA_FP8_DECODE`):
  - **`mlx`** *(default)* — fused einsum; fastest on MLX 0.31.2 (MLX fuses the FP8 dequant into
    the matmul → matrix units, no bf16 materialisation).
  - **`kernel`** — an all-in-one scalar **Metal** absorbed-decode kernel (split-K flash, online
    softmax). No bf16 materialisation → lowest extra memory.
  - **`sdpa`** — fused FP8-dequant Metal kernel + MLX SDPA.

### Performance (M-series, MLX 0.31.2, B=1, H=128, kv_lora_rank=512) — decode latency/token

`bf16 absorbed SDPA` is the *hypothetical* unquantised-latent baseline (2× the memory); the
current expanded cache OOMs before these contexts.

| context | bf16 absorbed SDPA (2× mem) | FP8 `mlx` (default) | FP8 `kernel` | FP8 `sdpa` |
|--:|--:|--:|--:|--:|
| 1K  | 0.53 ms | **0.33 ms** | 0.83 ms | 0.33 ms |
| 4K  | 0.37 ms | **0.43 ms** | 1.81 ms | 0.53 ms |
| 16K | 0.76 ms | **1.11 ms** | 5.65 ms | 1.34 ms |
| 64K | 2.18 ms | **3.86 ms** | 20.7 ms | 4.54 ms |

**Honest note on the Metal kernel:** the hand-written absorbed kernel is correct and uses the
least extra memory, but it is *slower* than MLX's fused einsum — a scalar `metal_kernel` can't
beat MLX's matrix-unit GEMM here. So the default is the `mlx` path; the custom kernel is kept
as the lowest-memory option. Beating bf16 SDPA on FP8 would need an MMA (simdgroup-matrix)
kernel that reads FP8 directly — left as future work. The headline win of this PR is **memory**
(plus accuracy parity), at decode speed that is at parity-to-faster at short context and ~1.8×
at 64K vs the (hypothetical, 2×-memory) bf16 latent.

### Accuracy

- Grouped FP8 round-trip RMS on the latent: **~2.5%** (`N(0,1)`; stable under outliers/heavy tails).
- Fused kernel / SDPA decode vs fp32 reference: **cosine 1.000000**.
- Absorbed FP8 vs default expanded bf16 on a real `DeepseekV2Attention` layer: **cosine 0.9997**.
- Full **4-bit quantized** 6-layer `deepseek_vl_v2` model, prefill + 24 greedy decode steps,
  FP8 vs bf16 cache: next-token **logit cosine 0.9993** (real `QuantizedLinear` `kv_b_proj`).

### Usage

```bash
export MLX_VLM_MLA_FP8=1            # enable FP8 MLA KV cache (default: off)
export MLX_VLM_MLA_FP8_GROUP=128    # latent quant group size
export MLX_VLM_MLA_FP8_DECODE=mlx   # mlx (default) | kernel | sdpa
```

### Changes

- `mlx_vlm/models/deepseek_vl_v2/mla_fp8.py` *(new)* — quant/dequant, `Fp8MLAKVCache`, absorbed attention, decode paths, Metal kernels.
- `mlx_vlm/models/deepseek_vl_v2/{language,deepseek_vl_v2}.py` — opt-in dispatch + `make_cache` (default unchanged).
- `mlx_vlm/tests/test_mla_fp8.py` *(new)* — 35 tests (quant, cache, kernel-vs-reference, layer parity, guards).
- `mlx_vlm/tests/bench_mla_fp8.py` *(new)* — reproduces the tables above.

### Testing

`python -m pytest mlx_vlm/tests/test_mla_fp8.py` → **35 passed**. A code review pass was applied
(fixed cache regrow over-allocation, added a kernel `kv_lora_rank % 32` guard, padding-mask
pass-through, dequant divisibility check, and trim-then-grow coverage).

### Scope / limitations

- Wired into `deepseek_vl_v2` only (other MLA models can adopt it by routing through
  `mla_fp8.mla_fp8_attention` + returning `Fp8MLAKVCache` from `make_cache`).
- Real pretrained MLA `deepseek_vl_v2` checkpoints are 9 GB+; the end-to-end check uses a full
  quantized model with random weights (network-bound download on the test host).
- The `kernel` decode path requires `kv_lora_rank % 32 == 0` (guarded; falls to `mlx` otherwise).


mlx_vlm/models/deepseek_vl_v2/mla_fp8.py
Blaizzy, line 1: Could you move this to deepseek_vl_v2?

Blaizzy (commented):
LGTM, just one nit

popfido:
## Update — rebased onto current `main`; MMA decode reference added (default unchanged)

Rebased this branch onto the latest upstream `main`, and added an **MMA-tiled (8×8 simdgroup), FP8-fused** absorbed-decode kernel as a new **non-default** `kernel_mma` method, plus a batched decode benchmark. The default decode stays **`mlx`** (the fused einsum).

**Honest negative result for the custom kernel:** the MMA `metal_kernel` is numerically correct (cosine ≈ 1.0, 51 tests pass) but a **perf regression at every batch size**. Single-token MLA decode is overhead/latency-bound (not FLOP- or bandwidth-bound), so the matrix units stay starved and a `mx.fast.metal_kernel` can't out-pipeline MLX's compiled SDPA/einsum. (Isolation check: removing the FP8 dequant entirely saved only ~0.5 ms of 7.7 ms — the cost is per-tile `simdgroup_load`/barriers, not the FP8 math.)

**Where FP8-MLA actually wins — batched decode + memory**, via the existing `mlx` path:

| context (B=8) | bf16 absorbed SDPA | FP8 `mlx` | speedup |
|--:|--:|--:|--:|
| 4K | 4.31 ms | 2.09 ms | 2.06× |
| 16K | 18.86 ms | 7.92 ms | 2.38× |
| 64K | 113.94 ms | 32.72 ms | 3.48× |

At **B=1** it's a memory win (≈parity-to-slower on speed); at **B≥4** it's **1.5–3.5× faster** while halving KV memory — the real serving regime. Caveat: the bf16 baseline uses SDPA's 128-way MQA broadcast, so part of the gap is SDPA-vs-einsum; the pure FP8-vs-bf16 same-path effect is ~neutral. FP8's value is making the absorbed-latent einsum form viable at batch.

Net: `mlx` stays default; `kernel_mma` is kept as a documented reference. A single-token decode speedup would need a non-JIT steel-GEMM kernel or an upstream MLX SDPA FP8-latent fast path.

mlx_vlm/models/deepseek_vl_v2/mla_fp8.py
popfido: Done in 5ae7a0f — moved to `mlx_vlm/models/deepseek_vl_v2/mla_fp8.py` (pure move + import-path updates; 51/51 tests pass).

lucasnewman:
Closing as obsolete. Please reopen if this is still needed and valid.

Issue #1270, opened 2026-06-03
APC not storing any cache entries


Edsuns:
### Environment

- **OS**: macOS (Apple Silicon) 26.5.1
- **Model**: `mlx-community/Qwen3.6-35B-A3B-4bit`
- **mlx-vlm version**: 549a75a
- **Client**: [aiperf](https://github.com/ai-dynamo/aiperf)

### Server Startup Command

```bash
APC_ENABLED=1 \
APC_NUM_BLOCKS=1024 \
APC_DISK_PATH=~/.cache/mlx-vlm/caching \
APC_DISK_MAX_GB=8 \
APC_DISK_SHARD_MAX_BLOCKS=256 \
mlx_vlm.server --model /path/to/models/mlx-community/Qwen3.6-35B-A3B-4bit \
  --log-level DEBUG \
  --max-kv-size 131072 \
  --port 12341
```

### Client Command

```bash
aiperf profile \
  --model "/path/to/models/mlx-community/Qwen3.6-35B-A3B-4bit" \
  --streaming \
  --endpoint-type chat \
  --tokenizer Qwen/Qwen3.6-35B-A3B \
  --url http://localhost:12341 \
  --request-count 3
```

### Observed Behavior

After running the benchmark with 3 requests, querying the cache stats endpoint shows **all values at 0**:

```bash
curl http://localhost:12341/v1/cache/stats
```

```json
{
  "block_size": 16,
  "num_blocks": 1024,
  "pool_used": 0,
  "lookups_hit": 0,
  "lookups_miss": 0,
  "matched_tokens": 0,
  "served_tokens": 0,
  "token_hit_rate": 0.0,
  "evictions": 0,
  "stores": 0,
  "disk_hits": 0,
  "disk_writes": 0,
  "exact_hits": 0,
  "exact_stores": 0,
  "disk_bytes": 0,
  "disk_max_bytes": 8589934592,
  "disk_evictions": 0,
  "disk_files": 0,
  "disk_blocks_indexed": 0,
  "disk_exact_indexed": 0,
  "enabled": true
}
```

Key observations:
- `"enabled": true` — APC is recognized as enabled
- `"stores": 0` and `"exact_stores": 0` — **No cache entries are being stored at all**
- `"disk_writes": 0` — Nothing is being written to disk
- `"lookups_miss": 0` — No lookups are even being attempted

### Expected Behavior

With `APC_ENABLED=1` and multiple requests being sent:
[...]

PR #1287, opened 2026-06-04
Fix automatic prefix caching for single-request chats

spicyneuron:
### Problem

APC silently does nothing for models that use exact-mode caching (Gemma 4, Qwen 3.5, etc.) during single-request chat. Every turn reprocesses the prompt and APC stats show zero hits. No errors or warnings either.

Plain attention models and block APC aren't affected.

### Repro

1. Load the server with Gemma 4 26B (or any other model where APC uses "exact" mode) and `APC_ENABLED=1`
2. Open /v1/cache/stats in the browser
3. Send 2 or more messages
4. Refresh stats

Expected: Non-zero `exact_stores`, `exact_hits`, `matched_tokens`

Actual:
```json
{"block_size":16,"num_blocks":2048,"pool_used":0,"lookups_hit":0,"lookups_miss":0,"matched_tokens":0,"served_tokens":0,"token_hit_rate":0.0,"evictions":0,"stores":0,"disk_hits":0,"disk_writes":0,"exact_hits":0,"exact_stores":0,"enabled":true}
```

### Fix

After prefill, APC snapshots the prompt cache so the next turn can reuse it. The snapshot step assumed a batch-aware cache and bailed quietly when it didn't find one. Single-stream caches expose a 1-row tensor directly and don't implement the batch-aware `extract()` interface so the snapshot never ran.

The simplest fix was to give single-row caches a direct deep clone instead. The batch-aware path is unchanged. Multi-row caches without a batch interface still bail — there didn't seem to be a safe way to pick a row.

### DIFF
--- a/mlx_vlm/apc.py
+++ b/mlx_vlm/apc.py
@@ -311,9 +311,9 @@ def _cache_entry_is_single_row_for_apc(c: Any) -> bool:
             lm_cache.ChunkedKVCache,
         ),
     ):
-        return _array_has_single_row(getattr(c, "keys", None)) and _array_has_single_row(
-            getattr(c, "values", None)
-        )
+        return _array_has_single_row(
+            getattr(c, "keys", None)
+        ) and _array_has_single_row(getattr(c, "values", None))
     if isinstance(c, lm_cache.ArraysCache):
         states = getattr(c, "cache", [])
         if any(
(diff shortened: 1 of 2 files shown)


spicyneuron:
Potentially fixes: https://github.com/Blaizzy/mlx-vlm/issues/1270

lucasnewman:
Thanks! I did a slightly more surgical fix here: https://github.com/Blaizzy/mlx-vlm/pull/1291

PR #1288, opened 2026-06-04
Adjust Gemma4 quantization predicate

Blaizzy:
## Summary
- Let Gemma4 MLP projections use the requested global quantization bits instead of forcing them to 8-bit
- Keep router modules explicitly quantized at 8-bit
- Add a regression test for the predicate behavior

## Tests
- .venv/bin/python -m pytest mlx_vlm/tests/test_models.py::TestModels::test_gemma4_quant_predicate_keeps_mlp_at_requested_bits
- .venv/bin/python -m pytest mlx_vlm/tests/test_models.py::TestModels::test_gemma4


mlx_vlm/models/hunyuan_vl/language.py
```diff
@@ -125,14 +125,19 @@ def apply_rotary_pos_emb(
         cos: Cosine values with shape (seq_len, head_dim)
         sin: Sine values with shape (seq_len, head_dim)
     """
+    origin_dtype = q.dtype
+
     # Expand cos/sin to (1, 1, seq_len, head_dim) for broadcasting
     cos = cos[None, None, :, :]
     sin = sin[None, None, :, :]
 
+    q, k = q.astype(mx.float32), k.astype(mx.float32)
```
lucasnewman, line 134: Did you intend to include the changes in this file?

lucasnewman (approved):
See comment but otherwise looks good.

mlx_vlm/models/hunyuan_vl/language.py
Blaizzy: Yes, it was a bug I found experimenting with prefil chunking testing

PR #1289, opened 2026-06-04
Add RealtimeSession: incremental audio for realtime interaction

fum0passiv0:
## What

Adds `RealtimeSession` — incremental, persistent-cache multimodal generation for
omni models. Instead of waiting for the whole utterance and prefilling it
(`examples/omni.py`'s VAD-batch loop), it appends audio into **one persistent KV
cache in ~200 ms micro-turns *during* speech**, so the reply starts the moment
the user stops — no end-of-utterance re-prefill.

This leans on Gemma 4's **encoder-free audio**: a 200 ms chunk is just N×640 raw
samples → one linear projection → N tokens appended to the cache (no encoder, no
lookback). Optional image-first turns (`IMG {text} AUD`) match the intended
layout.

## Why

`omni.py` already does live mic+camera, but the latency that matters for
*interaction* — end-of-speech → first token — is dominated by re-prefilling the
utterance. Ingesting perception as it arrives removes that.

## Result (M4 Max, `mlx-community/gemma-4-12B-it-4bit`)

- audio ingest **~4× realtime** (~50 ms / 200 ms chunk, during speech)
- **~110 ms** time-to-first-token after end-of-audio (vs full-utterance prefill)
- clean replies; stops correctly via `config.eos_token_id`

## Changes

- `mlx_vlm/realtime.py` — `RealtimeSession`: `start_turn(image=None)` /
  `append_audio(pcm)` / `respond()`. Built on `make_prompt_cache` +
  `make_sampler`; derives turn frames from the chat template. ~150 lines, no new
  deps. Generalises to omni models exposing `get_audio_features` + an audio
  token; validated on `gemma4_unified`.
- `examples/gemma4_realtime.py` — live-mic voice chat (`--camera` for a frame per
  turn, `--audio file` for one-shot).

## Open questions / notes

This is a **proposal** — happy to reshape the API, move it, or close it if you'd
rather build it your way. Two things I'd love your steer on:

1. **Placement** — `mlx_vlm/realtime.py` vs `generate/`, and naming.
2. **Tests** — I validated end-to-end locally but didn't add a unit test yet:
   `RealtimeSession` needs a real processor (chat template + feature extractor),
   whereas the model tests feed tensors directly. I can add a test in whatever
   fixture style you prefer (or refactor a lower-level tensor-append primitive
   that fits `test_models.py`). Just say the word.

### DIFF
--- a/examples/gemma4_realtime.py
+++ b/examples/gemma4_realtime.py
@@ -1,0 +1,126 @@
+"""Realtime voice chat with Gemma 4 (local, MLX) — talk and it replies.
+
+Audio is ingested into a PERSISTENT KV cache in 200 ms micro-turns *while you
+speak*, so the reply starts the moment you stop — no end-of-utterance re-prefill.
+Exploits Gemma 4's encoder-free audio (a 200 ms chunk = 5×640 raw samples → one
+linear projection → 5 tokens appended).
+
+    python examples/gemma4_realtime.py                 # live mic (default)
+    python examples/gemma4_realtime.py --audio q.wav   # one-shot from a file
+"""
+
+import argparse
+import time
+
+import numpy as np
+from mlx_vlm import load
+from mlx_vlm.realtime import RealtimeSession
+
+
+def _snap(cap):
+    """Grab one RGB frame from an open cv2 capture, or None."""
+    import cv2
+    from PIL import Image
+
+    ok, frame = cap.read()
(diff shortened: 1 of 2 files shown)


lucasnewman:
@fum0passiv0 This looks cool but I don't see a strong reason for it to exist in `mlx-vlm` -- you could just construct it in whatever project you need that uses it. If you wanted to wire it through the `/v1/realtime` endpoint that could be interesting as a potential avenue for more general usage.

PR #1297, opened 2026-06-05
Add regression tests for ThinkingBudgetCriteria.apply_forced_token

glenncameron2:
## Summary

`ThinkingBudgetCriteria` has unit tests (`TestThinkingBudgetCriteria`), but none of them call `apply_forced_token` — the method where an uninitialised `forced_token_id` previously surfaced as an `AttributeError`.

The crash path: when the first generated token is the thinking **start** or **end** delimiter, both branches in `__call__` return early without setting `forced_token_id`. On older builds (pre-`ec0f235`, e.g. the v0.4.4 release) `__init__` didn't initialise it either, so the very next `apply_forced_token` raised:

```
AttributeError: 'ThinkingBudgetCriteria' object has no attribute 'forced_token_id'
```

I hit this on Gemma-style models (`thinking_start_token="<|think|>"`, `thinking_end_token="<channel|>"`) during a ~5.8k-sample batch inference run — the exception was swallowed by the caller and silently produced empty outputs for the affected samples.

`forced_token_id` is now initialised in `__init__` on `main`, so these tests **pass as-is**. This PR just locks that behaviour in and guards against a regression, since the existing suite would not have caught it.

## What's added

Four cases in `TestThinkingBudgetCriteria`, all exercising `apply_forced_token`:
- before the first `__call__` (the bare regression);
- immediately after the **start** delimiter early-return;
- immediately after the **end** delimiter early-return;
- the budget-exceeded path, asserting the token `__call__` returns is the one `apply_forced_token` injects.

## Test plan

- [x] New tests pass against current `main` (verified by exec'ing the current `ThinkingBudgetCriteria` source directly).
- [x] Confirmed the tests **fail** (`AttributeError`) when the `__init__` initialisation is removed, i.e. they genuinely reproduce the pre-`ec0f235` crash.
- [ ] CI: `pytest mlx_vlm/tests/test_generate.py::TestThinkingBudgetCriteria`

Test-only change; no library code touched.

### DIFF
--- a/mlx_vlm/tests/test_generate.py
+++ b/mlx_vlm/tests/test_generate.py
@@ -1442,6 +1442,55 @@ class TestThinkingBudgetCriteria:
         assert criteria.thinking_token_count == 0
         assert criteria.budget_exceeded is False
 
+    def _make_criteria(self, enable_thinking=True):
+        return ThinkingBudgetCriteria(
+            tokenizer=FakeTokenizer(),
+            thinking_budget=5,
+            thinking_end_token="</think>",
+            thinking_start_token="<think>",
+            enable_thinking=enable_thinking,
+        )
+
+    def test_apply_forced_token_safe_before_first_call(self):
+        """Regression: apply_forced_token must be safe before __call__ ever runs.
+
+        forced_token_id has to be initialised in __init__; otherwise the first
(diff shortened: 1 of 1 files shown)


lucasnewman (approved):
LGTM

PR #1298, opened 2026-06-05
[Cohere] Add Cohere tool-calling and thinking parsing support

Terrencezzj:
## Fix cohere generation parsing

- Add a new cohere tool parser that handles Cohere's <|START_ACTION|> / <|END_ACTION|> delimited tool calls, supporting both single and batched (array) calls

- Add Cohere thinking markers (<|START_THINKING|> / <|END_THINKING|>) to the streaming thinking state machine

- Strip Cohere content markers (<|START_TEXT|> / <|END_TEXT|>) from output in both streaming and non-streaming paths, including correct handling when markers span chunk boundaries
 
- Allow parse_tool_call to return a list of calls from a single action block, with correct index tracking

## Tests
- test_cohere_thinking_markers_strip_text_markers — non-streaming thinking + content marker parsing
- test_cohere_text_markers_are_suppressed_across_chunks — streaming with markers split across chunks
- test_parser_can_return_multiple_tool_calls — multi-call list support in process_tool_calls
- test_cohere_action_array_parses_to_openai_tool_calls — array action block parsing
- test_cohere_single_action_object_parses_to_openai_tool_call — single object action block parsing


lucasnewman (approved):
LGTM

Issue #1254, opened 2026-05-31
Python generate()/stream_generate() does not apply chat template for Gemma 3/4 (causes repetition/echo loops)


Girish011:
## Bug

When calling the **Python API** (`mlx_vlm.generate()` / `stream_generate()`) with Gemma 3/3n/4 models and a raw string prompt, the tokenizer chat template is **not applied**. This causes pathological repetition/echo loops (the model repeats the user prompt instead of answering).

The **CLI path already applies the chat template** before generation, so CLI and Python API behavior are inconsistent.

## Environment

- macOS (Apple Silicon)
- `mlx-vlm==0.4.4` (also reproduced on current `main` as of 2026-05-31)
- Model: `mlx-community/gemma-4-e4b-it-4bit`

## Steps to reproduce

```python
from mlx_vlm import load, generate

model, proc = load("mlx-community/gemma-4-e4b-it-4bit")
out = generate(
    model,
    proc,
    prompt="Say hello in one sentence.",
    max_tokens=40,
    temperature=0.7,
    top_p=0.9,
)
print(out.text)
```

**Actual output (example):**
```
Say hello in one sentence. Say hello in one sentence. Say hello in one sentence. ...
```

**Expected:** a normal short response (e.g. `Hello!`).

## Root cause

In `mlx_vlm/generate/dispatch.py` `stream_generate()`, prompts are passed directly to `prepare_inputs()` without applying `apply_chat_template()`.

By contrast, the CLI entrypoint applies the chat template explicitly:

```python
prompt = apply_chat_template(processor, config, prompt, ...)
```

For Gemma 3/3n/4, raw string prompts without turn markers (`<|turn>user`, `<turn|>`, etc.) lead to degenerate generation.

## Workaround

Manually apply the chat template before calling `generate()`:

```python
from mlx_vlm.prompt_utils import apply_chat_template
from mlx_vlm.utils import load_config

config = load_config("mlx-community/gemma-4-e4b-it-4bit")
prompt = apply_chat_template(
    proc, config, "Say hello in one sentence.",
    num_images=0, num_audios=0, add_generation_prompt=True,
)
out = generate(model, proc, prompt=prompt, max_tokens=40, temperature=0.7, top_p=0.9)
```

This produces correct output (`Hello!`).

## Proposed fix

[...]

PR #1300, opened 2026-06-05
docs(gemma4): clarify base vs instruct and thinking modes

Girish011:
## Summary

Adds a **Base vs Instruct models** section to `mlx_vlm/models/gemma4/README.md` per maintainer feedback on #1254:

- **Base** checkpoints: no chat template; raw prompts work with the same Python API
- **Instruct (`-it`)** checkpoints: callers should call `apply_chat_template()` before `generate()` (CLI still applies it automatically)
- **Thinking vs non-thinking** instruct usage: default is non-thinking; how to enable thinking via CLI / Python

Includes short Python examples for base (raw prompt) vs instruct (templated prompt).

## Context

Follow-up to #1254 - thanks for clarifying that explicit chat templating is expected API behavior for instruct models.

## Test plan

- [x] Markdown renders correctly
- [x] Examples match existing README patterns (`apply_chat_template`, `--enable-thinking`)

### DIFF
--- a/mlx_vlm/models/gemma4/README.md
+++ b/mlx_vlm/models/gemma4/README.md
@@ -20,6 +20,55 @@ Capabilities:
 
 > **K-eq-V**: Full attention layers reuse key projections as values (no separate `v_proj`), reducing parameters and memory while using `num_global_key_value_heads` for the KV dimension.
 
+## Base vs Instruct models
+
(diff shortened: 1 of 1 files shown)


lucasnewman (approved):
@Girish011 Thank you!

PR #1301, opened 2026-06-05
Fix gemma4 load for HF checkpoints with num_kv_shared_layers > 0

Blaizzy:
## Summary

Loading `google/gemma-4-*` HF checkpoints (e.g. `gemma-4-E2B-it-qat-q4_0-unquantized`) fails with `ValueError: Missing 60 parameters`, listing `k_proj` / `v_proj` / `k_norm` weights for the upper layers.

These layers are KV-shared: when `num_kv_shared_layers > 0`, the last `num_kv_shared_layers` layers reuse KV from an earlier layer of the same `layer_type` and don't ship their own K/V tensors in the checkpoint.

The `Attention` class already handles this via `kv_shared_only=True` (skips creating `k_proj` / `v_proj` / `k_norm` / `v_norm`). But `Gemma4TextModel.__init__` only forwarded the module-level `kv_shared_only` flag uniformly to every `DecoderLayer`, so during conversion (with default `kv_shared_only=False`) every layer still allocated its own KV modules → `load_weights` raised.

This PR sets `kv_shared_only=True` for layers in the shared range (`i >= num_hidden_layers - num_kv_shared_layers`), matching what the `Attention` class expects.

For `gemma-4-E2B-it-qat-q4_0-unquantized`: `num_hidden_layers=35`, `num_kv_shared_layers=20` → layers 15..34 are shared → 20 × {k_proj, v_proj, k_norm} = 60 missing tensors, which matches the error exactly.

## Verification

Patched in-place against `mlx-vlm 0.6.1` on a Linux/RTX-PRO-6000 box, then ran `python -m mlx_vlm convert --hf-path google/gemma-4-E2B-it-qat-q4_0-unquantized --mlx-path /tmp/out -q --q-bits 4`. Model loads cleanly, quantization runs, output safetensors written (4.33 GB), upload to `mlx-community/gemma-4-E2B-it-qat-4bit` succeeds.

## Test plan

- [ ] CI passes
- [ ] `python -m mlx_vlm convert --hf-path google/gemma-4-E2B-it-qat-q4_0-unquantized --mlx-path /tmp/out -q --q-bits 4` succeeds
- [ ] Inference on a non-KV-shared model (e.g. an earlier gemma model) still works — `kv_shared_only` path is additive, default config (`num_kv_shared_layers=0`) is unchanged

### DIFF
--- a/mlx_vlm/models/gemma4/language.py
+++ b/mlx_vlm/models/gemma4/language.py
@@ -759,6 +759,8 @@ class LanguageModel(nn.Module):
         for k, v in weights.items():
             if "self_attn.rotary_emb" in k:
                 continue
+            if self._is_unused_shared_kv_weight(k):
+                continue
             if any(
                 s in k for s in ["input_max", "input_min", "output_max", "output_min"]
             ):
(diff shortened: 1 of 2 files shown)


lucasnewman (approved):
LGTM

PR #1313, opened 2026-06-05
Fix Qwen quantized KV cache prompt state

### LINKED ISSUE #1310: Quantized KV cache (--kv-bits) silently corrupts the prefilled prompt context on 0.6.1 (in-context recall fails; f16 and 0.5.0 are correct)
### Summary

With KV cache quantization enabled (`--kv-bits`), mlx_vlm.server **silently loses the prefilled prompt context**: the model can no longer recall information stated in the prompt, while its *parametric* knowledge stays correct. There is **no error** — the output is just wrong. The identical command is correct on **0.5.0**, so this is a 0.6.x regression. It reproduces with every scheme/precision I tried (`uniform` and `turboquant`, `--kv-bits` 4 and 8). Plain f16 (no `--kv-bits`) is correct.

This looks like it lives in the same area as #1206 (`BatchQuantizedKVCache` `offset` / `make_mask`): that issue fixed the `TypeError` *crash* with `--kv-bits`, but the quantized-KV path now appears to produce a **wrong attention mask / offset** rather than crashing — i.e. the prompt KV is effectively not attended.

### Environment
- mlx-vlm **0.6.1** (0.5.0 = known-good), mlx 0.31.2, mlx-lm 0.31.3, transformers 5.9.0
- Python 3.14, macOS (Apple Silicon, M5 Max), Metal
- Models: `mlx-community/Qwen3.6-27B-mxfp8` (also reproduced on `mlx-community/Qwen3.6-35B-A3B-8bit`)

### Reproduction
Start the server **with** KV quantization:
```bash
python -m mlx_vlm.server --model mlx-community/Qwen3.6-27B-mxfp8 \
  --kv-bits 8 --kv-quant-scheme uniform --kv-group-size 64 \
  --host 127.0.0.1 --port 8295
```
Ask it to recall a value that exists **only in the prompt** (the `model` field is set to the loaded model, and I verified the response's echoed `model` matches — to rule out the default-model routing footgun; `temperature: 0` for determinism):
```bash
curl -s http://127.0.0.1:8295/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model": "mlx-community/Qwen3.6-27B-mxfp8",
  "messages": [{"role":"user","content":"Retiens ce nombre : 73914. Quel nombre dois-tu retenir ? Réponds uniquement le nombre."}],
  "max_tokens": 40, "temperature": 0 }'
```

### Expected vs actual
- **Expected:** `73914`
- **Actual (0.6.1 + `--kv-bits 8`):** `7`  — the prompt context is lost.

### Controls (single variable isolated; all send `model`, verified via the echoed response `model`; `temperature: 0`)

| Config | in-context recall (`73914`) | parametric (`capital of France`) |
|---|---|---|
| 0.6.1 + `--kv-bits 8 --kv-quant-scheme uniform --kv-group-size 64` | ❌ `7` | ✅ `Paris` |
| 0.6.1, **no** `--kv-bits` (f16) | ✅ `73914` | ✅ `Paris` |
| 0.5.0 + `--kv-bits 8 --kv-quant-scheme uniform --kv-group-size 64` | ✅ `73914` | ✅ `Paris` |

### Notes
- **Parametric answers stay correct** even with the bug (e.g. "capital of France" → Paris), which masks it — the failure is specific to the **prefilled prompt KV**.
- Reproduced with `--kv-quant-scheme uniform` **and** `turboquant`, at `--kv-bits` **4** and **8** — all fail on 0.6.1; only f16 is correct.
- Same observable behavior as the 0.6.0 reports of "answers as if the prompt were empty".
- Possibly the silent-correctness counterpart of the (now-closed) crash in #1206 — same `--kv-bits` / `BatchQuantizedKVCache` `make_mask`/`offset` path.

Blaizzy:
## Summary
- keep Qwen3.5/3.6 quantized batch KV caches on their in-place update path instead of the single-row BatchKVCache shortcut
- handle batched cache offsets when attention builds fallback position ids
- add regressions for quantized single-row Qwen batch cache state retention and shortcut routing

Fixes #1310.

## Verification
- `.venv/bin/python -m pytest mlx_vlm/tests/test_speculative.py -q`
- `.venv/bin/python -m pytest mlx_vlm/tests/test_batch_quantized_cache.py -q`
- `.venv/bin/python -m pytest mlx_vlm/tests/test_turboquant.py -q`
- `.venv/bin/python -m pytest mlx_vlm/tests/test_server.py -q`
- real server smoke: `mlx-community/Qwen3.6-27B-mxfp8 --kv-bits 8 --kv-quant-scheme uniform --kv-group-size 64` returned `73914`
- real server smoke: `mlx-community/Qwen3.6-27B-mxfp8 --kv-bits 4 --kv-quant-scheme turboquant` returned `73914`

### DIFF
--- a/mlx_vlm/models/qwen3_5/language.py
+++ b/mlx_vlm/models/qwen3_5/language.py
@@ -687,10 +687,13 @@ def _extract_row_cache(cache_entry, row: int):
 
 def _is_single_row_batch_cache(cache_entry) -> bool:
     left_padding = getattr(cache_entry, "left_padding", None)
+    # Quantized batch caches must update in place; the singleton shortcut
+    # rebuilds only plain BatchKVCache state after the row-wise forward.
     return (
         isinstance(left_padding, mx.array)
         and left_padding.ndim > 0
         and left_padding.size == 1
+        and not hasattr(cache_entry, "bits")
     )
 
 
@@ -1419,10 +1422,20 @@ class Qwen3_5Attention(nn.Module):
         kv_seq_len = keys.shape[-2]
 
         if position_ids is None:
-            kv_seq_len += cache.offset + 1
-            position_ids = mx.arange(cache.offset, cache.offset + L)
-            position_ids = mx.expand_dims(position_ids, axis=0)
-            position_ids = mx.tile(position_ids, (3, 1, 1))
+            cache_offset = cache.offset
+            if isinstance(cache_offset, mx.array) and cache_offset.ndim > 0:
+                offsets = mx.maximum(cache_offset[:B], 0)
+                kv_seq_len = kv_seq_len + offsets + 1
+                position_ids = offsets[:, None] + mx.arange(L)[None, :]
+                position_ids = mx.expand_dims(position_ids, axis=0)
+                position_ids = mx.tile(position_ids, (3, 1, 1))
+            else:
+                if isinstance(cache_offset, mx.array):
+                    cache_offset = int(cache_offset.item())
+                kv_seq_len += cache_offset + 1
+                position_ids = mx.arange(cache_offset, cache_offset + L)
+                position_ids = mx.expand_dims(position_ids, axis=0)
+                position_ids = mx.tile(position_ids, (3, 1, 1))
         else:
             kv_seq_len += cache.offset + 1 if cache is not None else 0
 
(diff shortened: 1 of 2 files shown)


lucasnewman (approved):
LGTM

Issue #1315, opened 2026-06-06
20260606 -- Diagnostics Report — 1 failure(s), 5 harness issue(s) (mlx-vlm 0.6.2)


jrp2014:
Some models are actually pretty good.  Others are what I would call buggy (repetitious, irrelevantly verbose, etc).  Some of the Qwen models are excluded (-e mlx-community/Qwen3-VL-2B-Thinking-bf16 -e Qwen/Qwen3-VL-2B-Instruct -e mlx-community/Qwen3-VL-2B-Instruct-bf16) as they can't seem to be run in sequence with each other (see https://github.com/Blaizzy/mlx-vlm/issues/1306).  `LiquidAI/LFM2.5-VL-450M-MLX-bf16` doesn't load at all.  Quite a few of the models leak tokens or some other issue that is a problem with the harness / config rather than the model itself.   We do seem to be making progress! Thanks for all the hard work on this @Blaizzy 


<!-- markdownlint-disable MD013 -->


**Run summary:** 55 locally-cached VLM model(s) checked; 1 hard failure(s), 5 harness/integration issue(s), 0 preflight warning(s), 54 successful run(s).

Test image: `20260525-164635_DSC00024.jpg` (2.4 MB).

---

## Issue Queue

Root-cause issue drafts are generated in
[issues/index.md](https://github.com/jrp2014/check_models/blob/main/src/output/issues/index.md).
Each row is intended to become one focused upstream GitHub issue.

<!-- markdownlint-disable MD060 -->

| Target                                                   | Problem                                               | Evidence Snapshot                                                                                                                                                                                                                           | Affected Models                                            | Issue Draft                                                                                                                              | Evidence Bundle                                                                                                                                                                                      | Fixed When                                                |
|----------------------------------------------------------|-------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------|
[...]

PR #1316, opened 2026-06-06
Fix thinking enabled by default for chat templates

lucasnewman:
Right now there's a gap where a chat template can enable thinking to true when unspecified, but mlx-vlm api/cli/server defaults thinking to false unless `--enable-thinking` or equivalent is provided. This change aligns the default chat template behavior to match mlx-vlm defaults when the chat template supports it.

This fixes the `mlx-community/Qwen3.5-35B-A3B-bf16` issue reported in https://github.com/Blaizzy/mlx-vlm/issues/1315, where the generation cap is small and enable_thinking=False from the api level, but thinking is enabled by default in the chat template and the thinking token length exceeds the generation cap.


Blaizzy (approved):
LGTM! 🚀

Issue #1277, opened 2026-06-03
Gemma 4 12B Unified: vision_embedder fails with layer_norm size mismatch (6912) on --video/--image — preprocessor does not patchify raw pixels


artile:
## Summary

Running any video or image inference on `mlx-community/gemma-4-12B-it-*` (both `8bit` and `bf16` reproduced) raises:

```
ValueError: [layer_norm] weight must have the same size as the last dimension of x but has 6912 elements.
```

Text-only inference works correctly.

## Repro

```bash
hf download mlx-community/gemma-4-12B-it-8bit --local-dir ./gemma4-12b
mlx_vlm.generate \
  --model ./gemma4-12b \
  --video clip.mp4 --audio clip.wav --fps 1 \
  --prompt "Describe" --max-tokens 200
```

Tested with `--resize-shape 896 896` — same error.

## Environment

- macOS 26.4.1, Apple M1 Max, 64 GB
- mlx-vlm latest (installed today via `uv tool install mlx-vlm`)
- Python 3.13.7
- Model: `mlx-community/gemma-4-12B-it-8bit` (also `-bf16`)

## Stack trace

```
File "mlx_vlm/models/gemma4_unified/gemma4_unified.py", line 46, in __call__
    hidden_states = self.patch_ln1(pixel_values)
File "mlx/nn/layers/normalization.py", line 110, in __call__
    return mx.fast.layer_norm(x, weight, bias, self.eps)
ValueError: [layer_norm] weight must have the same size as the last dimension of x but has 6912 elements.
```

## Root cause analysis

In `gemma4_unified/gemma4_unified.py`, `VisionEmbedder.__call__`:

```python
def __call__(self, pixel_values, image_position_ids=None):
    if pixel_values.ndim == 4 and pixel_values.shape[-1] == self.patch_dim:
        pixel_values = pixel_values.reshape(pixel_values.shape[0], -1, self.patch_dim)
    hidden_states = self.patch_ln1(pixel_values)   # <- fails
```

`patch_ln1 = nn.LayerNorm(patch_dim)` where `patch_dim = 6912` (= `48 × 48 × 3`, since config has `model_patch_size: 48`).

The conditional `pixel_values.shape[-1] == self.patch_dim` guards re-shape, but the preprocessor delivers raw `[B, H, W, 3]` (last dim = 3), so the guard is False, the pixel tensor is **not patchified**, and `patch_ln1` receives a tensor whose last dim is 3 instead of 6912.

[...]

Issue #1008, opened 2026-04-10
mlx-community/gemma-4-31b-bf16 fails to describe image


jrp2014:
Here's the trace:

```
> python -m mlx_vlm.generate --model mlx-community/gemma-4-31b-bf16 --max-tokens 100 --temperature 0.0 --prompt "Describe this image." --image ~/Pictures/Processed/20260321-163922_DSC09457_DxO.jpg
<frozen runpy>:128: RuntimeWarning: 'mlx_vlm.generate' found in sys.modules after import of package 'mlx_vlm', but prior to execution of 'mlx_vlm.generate'; this may result in unpredictable behaviour
Calling `python -m mlx_vlm.generate ...` directly is deprecated. Use `mlx_vlm generate` or `python -m mlx_vlm generate` instead.
Fetching 18 files: 100%|█████████████████████| 18/18 [00:00<00:00, 13402.71it/s]
Download complete: : 0.00B [00:00, ?B/s]                 | 0/18 [00:00<?, ?it/s]
/Users/jrp/miniconda3/envs/mlx-vlm/lib/python3.13/site-packages/transformers/audio_utils.py:538: UserWarning: At least one mel filter has all zero values. The value for `num_mel_filters` (128) may be set too high. Or, the value for `num_frequency_bins` (257) may be set too low.
  warnings.warn(
==========
Files: ['/Users/jrp/Pictures/Processed/20260321-163922_DSC09457_DxO.jpg'] 

Prompt: <|image|> Describe this image.

==========
No text generated for this prompt

```

Library versions:

```
2026-04-10 14:13:07,047 - INFO - --- Library Versions ---
2026-04-10 14:13:07,047 - INFO - Pillow          : 12.2.0
2026-04-10 14:13:07,047 - INFO - huggingface-hub : 1.10.1
2026-04-10 14:13:07,047 - INFO - mlx             : 0.31.2.dev20260410+a33b7916
2026-04-10 14:13:07,047 - INFO - mlx-lm          : 0.31.3
2026-04-10 14:13:07,047 - INFO - mlx-metal       : 
2026-04-10 14:13:07,047 - INFO - mlx-vlm         : 0.4.4 (actually the latest built from source)
2026-04-10 14:13:07,047 - INFO - numpy           : 2.4.4
2026-04-10 14:13:07,047 - INFO - tokenizers      : 0.22.2
2026-04-10 14:13:07,047 - INFO - transformers    : 5.5.3
```

[...]

PR #1321, opened 2026-06-07
Fix Gemma 4 unified silently dropping image/video inputs (video processor rejects standard HF config keys)

sockeye44:
### Summary

On `main`, any `gemma4_unified` model whose `processor_config.json` contains a
`video_processor` block silently loses **all image and video inputs** — the
model receives only the text prompt and replies with e.g. *"I can't see any
image yet."* No error, no warning.

It's a regression from #1292 (*Add video input support for Gemma 4 12B*).

### Reproduction (no model download)

```python
from mlx_vlm.models.gemma4_unified.processing_gemma4_unified import Gemma4UnifiedVideoProcessor

# keys copied verbatim from google/gemma-4-31B-it processor_config.json -> "video_processor"
Gemma4UnifiedVideoProcessor(
    patch_size=16, pooling_kernel_size=3, max_soft_tokens=70, num_frames=32,
    do_rescale=True, rescale_factor=1/255, do_normalize=True,
    image_mean=[0, 0, 0], image_std=[1, 1, 1],
    do_convert_rgb=True, do_sample_frames=True, resample=3, return_metadata=False,
)
# TypeError: Gemma4VideoProcessor.__init__() got an unexpected keyword argument 'do_convert_rgb'
```

End-to-end (e.g. a `gemma4_unified` checkpoint that preserves the full processor config):

```
mlx_vlm.generate --model <gemma4_unified model> --image foo.png --prompt "What do you see?"
#  ->  "I can't see any image yet! Please upload or paste the image..."
```

### Root cause

`Gemma4UnifiedProcessor.from_pretrained` reads the `video_processor` block from
`processor_config.json` and builds the video processor from it:

```python
vp_config = dict(proc_config["video_processor"])      # standard HF keys
...
video_processor = Gemma4UnifiedVideoProcessor(**vp_config)
```

But `Gemma4VideoProcessor.__init__` (the parent, `models/gemma4/processing_gemma4.py`)
has a **fixed signature with no `**kwargs`** — unlike its siblings
`Gemma4ImageProcessor` and `Gemma4UnifiedAudioFeatureExtractor`, which both
accept and ignore extra keys. So standard HF keys this MLX reimplementation
doesn't consume — `do_convert_rgb`, `do_sample_frames`, `resample`,
`return_metadata` — raise `TypeError`.

That `TypeError` is then **silently swallowed** by the `AutoProcessor` patch in
`models/base.py` (`install_auto_processor_patch`), which falls back to the
default `AutoProcessor`. Since `transformers` has no `gemma4_unified` processor,
it degrades further to a bare `GemmaTokenizer` with no image processor.
`prepare_inputs` then returns only `input_ids` (no `pixel_values`), the single
`<|image|>` placeholder is never expanded into soft tokens, and the image is
dropped. The model loses its vision capabilities with no error surfaced.

### Fix

Give `Gemma4VideoProcessor.__init__` the same `**kwargs` tolerance the image and
audio processors already have, so unknown config keys are accepted and ignored.
The three processors become consistent.

### Why this scope

- Mirrors existing behavior of `Gemma4ImageProcessor` and
  `Gemma4UnifiedAudioFeatureExtractor` (both already take `**kwargs`).
- The offending keys are standard HF processor serialization — confirmed
  present in `google/gemma-4-31B-it`'s `processor_config.json` — so this affects
  any checkpoint that retains the full processor config.

### Tests

- Adds `test_video_processor_tolerates_extra_hf_config_keys` to
  `TestGemma4UnifiedProcessor`, constructing the video processor with the full
  real-world key set.
- `pytest mlx_vlm/tests/test_processors.py` → **93 passed, 1 skipped** (the
  prior 92 + the new regression test); the new test fails on `main` and passes
  with this change.
- Verified e2e on a `gemma4_unified` 12B: correctly describes a screenshot
  (OCR-level detail) and a synthetic shapes/colors/text image.

### Related

- Regression source: #1292
- Same model family, potentially related: #1277 (closed), #1008 (open — *"fails to describe image"*)

### DIFF
--- a/mlx_vlm/models/gemma4/processing_gemma4.py
+++ b/mlx_vlm/models/gemma4/processing_gemma4.py
@@ -236,7 +236,7 @@ class Gemma4VideoProcessor:
         default_fps: float = 2.0,
         **kwargs,
     ):
-        # Ignore extra HF processor-config keys this processor doesn't consume.
+        # Ignore extra HF processor's config keys this processor doesn't consume.
         if max_soft_tokens not in _SUPPORTED_SOFT_TOKENS:
             raise ValueError(
                 f"`max_soft_tokens` must be one of {_SUPPORTED_SOFT_TOKENS}, "
--- a/mlx_vlm/models/gemma4/processing_gemma4.py
+++ b/mlx_vlm/models/gemma4/processing_gemma4.py
@@ -234,7 +234,9 @@ class Gemma4VideoProcessor:
         image_mean: Optional[list] = None,
         image_std: Optional[list] = None,
         default_fps: float = 2.0,
+        **kwargs,
     ):
+        # Ignore extra HF processor-config keys this processor doesn't consume.
         if max_soft_tokens not in _SUPPORTED_SOFT_TOKENS:
             raise ValueError(
                 f"`max_soft_tokens` must be one of {_SUPPORTED_SOFT_TOKENS}, "
--- a/mlx_vlm/tests/test_processors.py
+++ b/mlx_vlm/tests/test_processors.py
@@ -276,6 +276,35 @@ class TestGemma4UnifiedProcessor(unittest.TestCase):
         )
         self.assertTrue(np.all(data["video_position_ids"][0, 2:] == -1))
 
+    def test_video_processor_tolerates_extra_hf_config_keys(self):
+        # Regression: extra HF config keys must be ignored, not rejected.
+        from mlx_vlm.models.gemma4.processing_gemma4 import Gemma4VideoProcessor
+        from mlx_vlm.models.gemma4_unified.processing_gemma4_unified import (
+            Gemma4UnifiedVideoProcessor,
+        )
+
+        # The edited base class, and the unified subclass that delegates to it.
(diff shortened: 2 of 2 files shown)


sockeye44:
1. **Re: test coverage.** Made it explicit: the test now constructs the bare Gemma4VideoProcessor directly as well.
2. **Re: whitelist/warn vs kwargs** Kept the bare **kwargs intentionally, to match the sibling Gemma4ImageProcessor and Gemma4UnifiedAudioFeatureExtractor (also HF's BaseImageProcessor), which all accept and ignore extra config keys.

lucasnewman:
@sockeye44 Thanks for the fix! Please don't use copilot for reviewing here -- you are welcome to use any tools you'd like on your side before you send the PR, but it makes it very difficult for human review when used here which is the bottleneck for getting changes merged.

lucasnewman (approved):

sockeye44:
@lucasnewman Thanks for the review! My bad, will keep that in mind.

sockeye44:
@lucasnewman I force-pushed to re-sign an earlier commit and clear the "Commits must have verified signatures" merge block. No code changes.

PR #1324, opened 2026-06-07
Fix APC exact disk-hit entries not promoted to in-memory LRU

mikeatlas:
## Summary

After a server restart with a warm disk cache, disk-restored exact cache entries were never written back to `_exact_cache` (the in-memory LRU). Every subsequent request for the same prefix paid disk-restore latency (~1-2s) instead of the fast in-memory clone path (~0.05s).

### Request flow before this fix

```mermaid
sequenceDiagram
    participant C as Client
    participant L as lookup_exact_cache
    participant M as _exact_cache (memory)
    participant D as DiskBlockStore

    Note over M: empty after restart
    C->>L: request 1 (same prefix)
    L->>M: miss
    L->>D: find_exact_prefix -> hit
    D-->>L: load prompt_cache
    L-->>C: return prompt_cache (no writeback to memory)

    C->>L: request 2 (same prefix)
    L->>M: miss (still empty)
    L->>D: find_exact_prefix -> hit again
    D-->>L: load prompt_cache
    L-->>C: return prompt_cache (pays disk cost every time)
```

### Request flow after this fix

```mermaid
sequenceDiagram
    participant C as Client
    participant L as lookup_exact_cache
    participant M as _exact_cache (memory)
    participant D as DiskBlockStore

    Note over M: empty after restart
    C->>L: request 1 (same prefix)
    L->>M: miss
    L->>D: find_exact_prefix -> hit
    D-->>L: load prompt_cache
    L->>M: clone + insert into LRU
    L-->>C: return prompt_cache

    C->>L: request 2 (same prefix)
    L->>M: hit
    M-->>L: clone from memory
    L-->>C: return prompt_cache (fast path)
```

### Why clone before storing

`generate_step` mutates the returned `prompt_cache` in-place as it appends generated tokens. Storing the same object in `_exact_cache` would corrupt the entry on the next lookup. A separate `_clone_prompt_cache_for_apc` call is required before storage.

### Why lazy promotion rather than eager loading at startup

Pre-loading disk entries at startup would block server init for several seconds per entry (each exact cache is ~40 KV tensor pairs deserialized from mmap), and `_exact_cache_max` is typically 2-16 slots while the disk store can hold gigabytes of entries across many prefixes. Choosing which entries to pre-load would require the same LRU logic anyway, applied to prompts that may never be requested again in this session. Lazy promotion is demand-driven: it pays the disk cost exactly once per session per prefix, and only for prefixes that are actually requested.

### Behavior when `_exact_cache_max == 0`

When in-memory caching is disabled, the disk-hit path is unchanged: stats are incremented and the entry is returned without any promotion attempt.

## Changes

- `mlx_vlm/apc.py`: after a successful disk hit in `lookup_exact_cache`, clone the restored cache and insert it into `_exact_cache` with LRU eviction. Falls through to the existing stats-only path when `_exact_cache_max == 0` or cloning fails.

## Tests

Six new tests in `mlx_vlm/tests/test_apc.py` (all pass, no regressions in the existing 31 tests):

| Test | What it verifies |
|---|---|
| `test_exact_disk_hit_is_promoted_to_memory` | Disk hit promoted; second request served from memory (`disk_hits` unchanged) |
| `test_exact_disk_hit_promotion_skipped_when_memory_disabled` | `_exact_cache_max=0` means every request hits disk |
| `test_exact_disk_hit_promotion_lru_eviction` | Promoting into a full cache evicts the LRU entry |
| `test_exact_lookup_memory_takes_priority_over_disk` | Memory entry wins over disk for the same key; `disk_hits=0` |
| `test_exact_disk_hit_promotion_clone_is_independent_of_returned_cache` | Mutating the returned cache does not corrupt the stored clone |
| `test_exact_disk_hit_promotion_with_nonzero_extra_hash` | Promotion key is correctly scoped by `extra_hash`; wrong hash gets no hit |

## Testing

```bash
python -m pytest mlx_vlm/tests/test_apc.py -q
# 37 passed
```

### DIFF
--- a/mlx_vlm/apc.py
+++ b/mlx_vlm/apc.py
@@ -2965,10 +2965,43 @@ class APCManager:
                         and len(stored_tokens) == disk_prefix_len
                         and token_tuple[:disk_prefix_len] == stored_tokens
                     ):
+                        # Promote the disk-restored entry to the in-memory LRU
+                        # so subsequent identical requests get the fast clone
+                        # path instead of paying disk-restore latency again.
+                        # We clone before storing because the caller's copy will
+                        # be mutated in-place by generate_step as it appends
+                        # generated tokens; the stored copy must stay pristine.
                         # Disk reads and warm-cache construction intentionally
                         # happen outside the manager lock. If clear()/reset_stats()
                         # races here, the restored tensors are still valid; only
                         # the hit counter lands in the new stats window.
+                        if self._exact_cache_max > 0:
+                            storage_copy = _clone_prompt_cache_for_apc(prompt_cache)
+                            if storage_copy is not None:
+                                promote_key = _sequence_hash(
+                                    stored_tokens, extra_hash, self.block_size
+                                )
+                                with self.lock:
+                                    self.stats.exact_hits += 1
+                                    self.stats.disk_hits += 1
+                                    self.stats.hits += 1
+                                    self.stats.matched_tokens += disk_prefix_len
+                                    if promote_key not in self._exact_cache:
+                                        self._exact_cache[promote_key] = (
+                                            APCExactCacheEntry(
+                                                token_ids=stored_tokens,
+                                                extra_hash=int(extra_hash),
+                                                prompt_cache=storage_copy,
+                                                last_used=time.time(),
+                                            )
+                                        )
+                                        self._exact_cache.move_to_end(promote_key)
+                                        while (
+                                            len(self._exact_cache)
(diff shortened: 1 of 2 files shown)


Blaizzy (approved):
LGTM, thanks! 🚀

mikeatlas:
let me see about fixing that linter/formatter error, just a sec!

lucasnewman (approved):

PR #1325, opened 2026-06-07
Align Qwen3-VL visual masks to the current prefill / decode window

lucasnewman:
Otherwise full image positions masks are out-of-bounds with respect to the shorter window hidden-state slices and can fault. Resolves https://github.com/Blaizzy/mlx-vlm/issues/1306

Issue #1323, opened 2026-06-07
Vision recall regression since #1210: fewer IDs extracted from images at greedy decoding (0.5.0 → 0.6.2)


gruenewald:
Thanks again for mlx-vlm — it powers our local document entity-extraction pipeline on Apple Silicon, and we really appreciate the work. This is the vision-specific half of a quality regression we observed going from 0.5.0 → 0.6.2 (the separate, fully self-contained text-only KV-quant half is **#1322**); filing it with everything we found so it's as easy as possible to act on. 🙏

## Context: entity-extraction recall dropped from 0.5.0 → 0.6.2

We run `mlx-community/Qwen3.6-35B-A3B-4bit` greedily (`temperature=0`) to extract IDs from document images via a JSON-schema response format. Going from 0.5.0 to 0.6.2, recall on our fixed 6-document / 28-ID suite dropped noticeably — and crucially this happens at **byte-identical input**, so it's a model/runtime change, not our app:

| mlx-vlm | extraction suite | single image (8 IDs) |
|---|---|---|
| 0.5.0 | 27–28 / 28 | 8 / 8 |
| 0.6.2 / main | 17–22 / 28 | 6 / 8 |

Text-only generation is unaffected between the two versions — the regression is **vision-specific**. (The companion KV-quant issue #1322 covers a separate, text-reproducible cause.)

## Environment (exact)

| | |
|---|---|
| Model | `mlx-community/Qwen3.6-35B-A3B-4bit` (public; `qwen3_5_moe`, 4-bit, group_size 64) |
| mlx-vlm | broken on current `main` (`6d8d5e3`) / 0.6.2; correct on 0.5.0 |
| mlx / mlx-lm / mlx-metal | 0.31.2 / 0.31.3 / 0.31.2 · Python 3.13 |
| Hardware / OS | Apple M4 Max, macOS 26.5.1 |

## Where it started: bisected to #1210

Using `git bisect` with a crash-free criterion (`enable_thinking=false`, greedy, the same image+schema request replayed against raw servers with the app layer bypassed):

```
parent  2bd1b19  -> 8 / 8 IDs   (GOOD)
        eb7537b  -> 4 / 8 IDs   (BAD)   #1210 "Fix Qwen MTP batched target-verify drift", +1461/−83 lines in qwen3_5/language.py
```

[...]

PR #1210, opened 2026-05-21
Fix Qwen MTP batched target-verify drift

Blaizzy:
## Summary

Fixes Qwen3.5/Qwen3.6 MTP batch drift by making target-verify paths match singleton numerics for left-padded and mixed-length batches.

Changes include:
- preserve left-padding position offsets during Qwen batched prefill
- make Qwen target-verify fallback projections row/time singleton exact when the dense helper cannot be used
- pass the sliced attention mask through target-verify attention chunks
- extend speculative tests to cover multi-row target-verify projections

## Root Cause

The remaining batch drift came from target-verify fallback projections that were only split by time, not by row. Small GDN projections fell back to batched GEMM across rows, which changed recurrent GDN state numerics for mixed batches. Left-padded batched prefill also needed singleton-equivalent position handling so no-drafter and MTP prefill states stay aligned.

## Validation

Focused tests:

```bash
PYTHONPATH=/tmp/codex-mlx-lm-target:. pytest \
  mlx_vlm/tests/test_speculative.py::test_qwen_target_verify_linear_matches_singleton_dense_gemv \
  mlx_vlm/tests/test_speculative.py::test_qwen_target_verify_small_projection_matches_singleton_dense_gemv \
  mlx_vlm/tests/test_speculative.py::test_qwen_target_verify_gated_norm_matches_singleton_path \
  mlx_vlm/tests/test_speculative.py::test_qwen_gdn_verify_conv_matches_singleton_windows -q
```

Result: `4 passed`.

Qwen3.6-35B-A3B AIME 2026 ids 1-4, `max_tokens=256`, `temperature=0`, `seed=42`, thinking enabled:

| Mode | Batch | Wall | Tokens | Tok/s | Exactness |
|---|---:|---:|---:|---:|---|
| No drafter | singleton x4 | 27.60s | 1024 | 37.10 | reference |
| No drafter | 4 | 25.34s | 1024 | 40.42 | exact vs singleton |
| MTP | singleton x4 | 11.97s | 1024 | 85.52 | reference |
| MTP | 4 | 9.02s | 1024 | 113.48 | exact vs singleton |

MTP batch-4 is `2.81x` faster than no-drafter batch-4 for this short run, and `1.33x` faster than sequential singleton MTP.

## Qwen3.5 9B 5-bit Temperature Sweep

AIME 2026 prompts, `max_tokens=2048`, `seed=42`, thinking enabled. All runs below were token-identical vs their no-drafter reference.

[...]

PR #1332, opened 2026-06-08
Fix Qwen3-VL deepstack visual embeds misaligned during chunked prefill

### LINKED ISSUE #856: Qwen3-VL-8B: Significantly degraded visual understanding compared to LM Studio with same weights
## Description

Using the same model weights (`lmstudio-community/Qwen3-VL-8B-Instruct-MLX-4bit`), same image, same prompt, and same hardware (M5 Max 40C, 64GB), mlx-vlm produces dramatically worse visual understanding than LM Studio's mlx-engine.

## Reproduction

Same engineering drawing with a visible As-Built stamp at 4 MP:

**LM Studio (mlx-engine v1.4.0):**
- Detects stamp, reads OCR text ("REDLINE AS-BUILT ☐ CHANGE ☑ NO CHANGE"), finds redlines
- `features_detected: ["As-Built Stamp", "Change/No Change Checkbox", "Redline Annotations", "Material Substitution Notes"]`
- `as_built_stamp_detected: true`

**mlx-vlm (v0.4.0) via vllm-mlx and Bodega:**
- Returns `features_detected: []`
- `as_built_stamp_detected: false`
- "No As-Built stamp or revision table found"

## Accuracy on full dataset (53 files at 4 MP)

| Server | Underlying Engine | Accuracy |
|--------|------------------|----------|
| LM Studio | mlx-lm + VisionAddOn | **72.0%** |
| vllm-mlx | mlx-vlm generate() | 52-55% |
| Bodega | mlx-vlm (assumed) | 41.5% |

## Investigation

We confirmed the image preprocessing is **identical** — `pixel_values` shape `(15600, 1536)` and `image_grid_thw` match exactly between HuggingFace processor and mlx-vlm's `prepare_inputs()`. The image reaches the model at full resolution in all cases.

The difference appears to be in how the text generation pipeline handles visual tokens after the vision encoder. LM Studio uses a "VisionAddOn" architecture where mlx-vlm is only used for vision embedding extraction, while text generation always goes through mlx-lm. This separation apparently produces significantly better results.

Specific areas that may differ:
1. **Chat template application** — how vision tokens (`<|vision_start|>`, `<|image_pad|>`, `<|vision_end|>`) are placed in the prompt
2. **Deepstack visual embedding handling** — Qwen3-VL injects intermediate vision features into early language model layers
3. **Sampling pipeline** — mlx-lm has more mature sampling (logits processors, etc.)

## Environment

- mlx-vlm: 0.4.0
- mlx: 0.31.1
- transformers: 5.0.0rc3
- Hardware: Apple M5 Max 40C, 64GB
- Model: lmstudio-community/Qwen3-VL-8B-Instruct-MLX-4bit
- macOS 26.3

## Expected Behavior

Same model weights should produce comparable visual understanding regardless of inference server.

sockeye44:
### Summary

Qwen3-VL / Qwen3-VL-MoE lose most of their visual detail on any prompt that
triggers **chunked prefill** — the default for prompts longer than                                                                                                                                                              
`prefill_step_size` (2048), i.e. essentially every high-res or multi-image
input. Symptoms range from degraded OCR / object recall to total collapse into
repetition.

### Reproduction

`mlx-community/Qwen3-VL-2B-Instruct-4bit` on the bundled
`examples/images/demo_pdf1_page1.png`. `--prefill-step-size 256` forces chunking
on a small image so it's obvious; it also fires at the default 2048 for any image
over ~2048 vision tokens (e.g. two images, ~3.5k tokens).

```
# chunked (buggy) — loops, reads nothing:
mlx_vlm.generate --model mlx-community/Qwen3-VL-2B-Instruct-4bit \
  --image examples/images/demo_pdf1_page1.png --temperature 0 \
  --prefill-step-size 256 --prompt "Transcribe all visible text."
# -> 108年8月 \n 108年11月 \n 108年10月 \n 108年11月 ...   (repeats forever)

# no chunking (correct) — reads the document:
mlx_vlm.generate ... --prefill-step-size 1000000 ...
# -> 高雄醫學大學 108 學年度行事曆 ...  | 選次 | 日 | 一 | 二 | ...
```

The added unit test reproduces it with no model download.

### Root cause

During chunked prefill, `LanguageModel.__call__` slices `visual_pos_masks` to the
current window but forwards the **full** `deepstack_visual_embeds` to every chunk:

```python
visual_pos_masks = visual_pos_masks[:, start : start + inputs.shape[1]]   # sliced
...
out = self.model(..., visual_pos_masks=visual_pos_masks,
                  deepstack_visual_embeds=deepstack_visual_embeds)          # NOT sliced
```

`_deepstack_process` reads `visual_embeds[offset : offset + n_visual]` with
`offset` reset to `0` on every call, so every chunk past the first re-applies the
*first* chunk's deepstack features to later vision tokens. The length guard
(`sample_embeds.shape[0] != n_visual`) doesn't catch it — the slice is the right
length, just the wrong rows.

### Fix

Realign `deepstack_visual_embeds` to each window by slicing it by the number of
vision tokens before the window (mirroring the existing `visual_pos_masks`
slice). Applied to both `qwen3_vl` and `qwen3_vl_moe`. Decode and non-chunked
prefill are unaffected (no-op).

### Scope

Fixes the single-sequence path (`mlx_vlm.generate`), which is what #856
exercises. The batched continuous-batching path slices the mask per row, but
realigning the concatenated per-row deepstack array under chunked prefill is
more involved — left as a follow-up; it's unaligned but safe today (guarded by
the existing length check).

### Tests

- Adds `test_qwen3_vl_deepstack_embeds_aligned_on_chunked_prefill` to `TestModels`
  (pure-Python tiny configs, no weights), covering both `qwen3_vl` and
  `qwen3_vl_moe`: drives `LanguageModel.__call__` through a mid-prompt chunk
  window and asserts the deepstack embeds are sliced to that window. Fails on
  `main` (`AssertionError: (5, 8) != (2, 8)` — the full embeds are forwarded),
  passes with this change.
- `pytest mlx_vlm/tests/test_models.py` → **148 passed, 26 subtests**.
- Verified e2e on `mlx-community/Qwen3-VL-2B-Instruct-4bit` (Apple M5) with
  `examples/images/demo_pdf1_page1.png`:
  - `--prefill-step-size 256`: on `main` the model collapses into a
    `108年8月 / 108年11月 / …` loop and reads nothing; with the fix it transcribes
    the document and matches the non-chunked (`--prefill-step-size 1000000`)
    reference.
  - Default `prefill_step_size=2048`, two images (~3.5k vision tokens, so it
    chunks): on `main` it emits empty table rows; with the fix it transcribes
    correctly.

### Notes

#1323 attributes the recall loss to #1210 / MRoPE, but the MRoPE position ids are
bit-identical to HF transformers for single- and multi-image; the root cause is
this deepstack misalignment. #1325 fixed the `visual_pos_masks` slice but left
`deepstack_visual_embeds` unaligned during prefill.

Fixes #856

### DIFF
--- a/mlx_vlm/models/qwen3_vl/language.py
+++ b/mlx_vlm/models/qwen3_vl/language.py
@@ -515,7 +515,18 @@ class LanguageModel(nn.Module):
                     if isinstance(cache_offset, mx.array)
                     else int(cache_offset)
                 )
-                visual_pos_masks = visual_pos_masks[:, start : start + inputs.shape[1]]
+                window = inputs.shape[1]
+                # Slice deepstack embeds to this window too; _deepstack_process reads from offset 0.
+                if deepstack_visual_embeds is not None:
+                    n_before = int(visual_pos_masks[:, :start].sum().item())
+                    n_window = int(
+                        visual_pos_masks[:, start : start + window].sum().item()
+                    )
+                    deepstack_visual_embeds = [
+                        embeds[n_before : n_before + n_window]
+                        for embeds in deepstack_visual_embeds
+                    ]
+                visual_pos_masks = visual_pos_masks[:, start : start + window]
             else:
                 rows = []
                 for b in range(visual_pos_masks.shape[0]):
--- a/mlx_vlm/models/qwen3_vl_moe/language.py
+++ b/mlx_vlm/models/qwen3_vl_moe/language.py
@@ -559,7 +559,18 @@ class LanguageModel(nn.Module):
                     if isinstance(cache_offset, mx.array)
                     else int(cache_offset)
                 )
-                visual_pos_masks = visual_pos_masks[:, start : start + inputs.shape[1]]
+                window = inputs.shape[1]
+                # Slice deepstack embeds to this window too; _deepstack_process reads from offset 0.
+                if deepstack_visual_embeds is not None:
+                    n_before = int(visual_pos_masks[:, :start].sum().item())
+                    n_window = int(
+                        visual_pos_masks[:, start : start + window].sum().item()
+                    )
+                    deepstack_visual_embeds = [
+                        embeds[n_before : n_before + n_window]
+                        for embeds in deepstack_visual_embeds
+                    ]
+                visual_pos_masks = visual_pos_masks[:, start : start + window]
             else:
                 rows = []
                 for b in range(visual_pos_masks.shape[0]):
--- a/mlx_vlm/tests/test_models.py
+++ b/mlx_vlm/tests/test_models.py
@@ -2053,6 +2053,115 @@ class TestModels(unittest.TestCase):
                 self.assertEqual(recorder.visual_pos_masks.shape, (1, 1))
                 self.assertEqual(recorder.visual_pos_masks.tolist(), [[False]])
 
+    def test_qwen3_vl_deepstack_embeds_aligned_on_chunked_prefill(self):
+        """Chunked prefill must realign deepstack embeds per window; otherwise later
+        chunks reuse the first chunk's embeds (offset resets to 0). #856 / #1323."""
+        from types import SimpleNamespace
+
+        from mlx_vlm.models import qwen3_vl, qwen3_vl_moe
+
+        cases = [
+            (
+                qwen3_vl,
+                qwen3_vl.TextConfig(
+                    model_type="qwen3_vl_text",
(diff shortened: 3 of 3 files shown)


lucasnewman (approved):
Thanks!

PR #1337, opened 2026-06-09
Respect generation_config sampling defaults

### LINKED ISSUE #1331: generation_config.json sampling params (temperature, top_p, top_k) silently ignored by server
## Bug

The server reads `generation_config.json` at model load time but only extracts `eos_token_id` from it. `temperature`, `top_p`, and `top_k` are silently discarded. The server then falls back to its own hardcoded default (effectively greedy/deterministic for the OpenAI-compatible endpoint), ignoring the model author's intended sampling configuration.

## Reproduction

Using `gemma-4-31b-it-qat-mxfp4`, which ships with this `generation_config.json`:

```json
{
  "do_sample": true,
  "temperature": 1.0,
  "top_k": 64,
  "top_p": 0.95
}
```

Start the server normally:

```bash
mlx_vlm.server --model /path/to/gemma-4-31b-it-qat-mxfp4 --port 8080
```

Send the same prompt three times with no explicit sampling params:

```bash
for i in 1 2 3; do
  curl -s http://localhost:8080/v1/chat/completions \
    -H "Content-Type: application/json" \
    -d '{"model":"...","messages":[{"role":"user","content":"Write one random haiku."}],"max_tokens":30}' \
    | python3 -c "import json,sys; print(json.load(sys.stdin)['choices'][0]['message']['content'].strip())"
done
```

**Result:** identical output on all three runs — deterministic/greedy behaviour, ignoring `do_sample: true` and `temperature: 1.0`.

Sending with `"temperature": 1.0` explicitly produces varied output as expected.

## Root cause

In `utils.py`, `load_config()` reads `generation_config.json` but only uses `eos_token_id`:

```python
generation_config_file = model_path / "generation_config.json"
if generation_config_file.exists():
    generation_config = json.load(f)
    if eos_token_id := generation_config.get("eos_token_id", False):
        config["eos_token_id"] = eos_token_id
    # temperature, top_p, top_k, do_sample are read but never applied
```

## Expected behaviour

When a request does not include explicit sampling parameters, the server should fall back to values from `generation_config.json` (`temperature`, `top_p`, `top_k`, `do_sample`) rather than its own hardcoded defaults. This matches `mlx-lm`'s behaviour and respects model authors' published sampling recommendations.

## Impact

Models like Gemma 4 ship with explicit sampling guidance (`temperature=1.0, top_p=0.95, top_k=64`). Without this fix, users running via the OpenAI-compatible server endpoint get deterministic outputs unless they manually pass sampling params in every request — a non-obvious footgun, especially when integrating with tools like Hermes that don't forward sampling params by default.

## Environment

- mlx-vlm v0.6.2
- Model: `mlx-community/gemma-4-31B-it-qat-mxfp4`
- Apple M5 Pro, macOS, mlx-vlm server mode

Blaizzy:
## Summary
- carry `temperature`, `top_p`, `top_k`, and `do_sample` from `generation_config.json` into loaded model config
- use model generation defaults for server requests when clients omit sampling params
- preserve explicit request sampling overrides

## Root cause
`load_config()` only copied `eos_token_id` from `generation_config.json`, and server request construction fell back to hardcoded defaults when sampling parameters were not provided. Models that ship sampling recommendations were therefore served greedily unless every request repeated those values.

Fixes #1331.

## Validation
- `uv run --with pytest python -m pytest mlx_vlm/tests/test_utils.py mlx_vlm/tests/test_server.py`
- `uv run --with pytest python -m pytest -s mlx_vlm/tests --ignore=mlx_vlm/tests/test_smoke.py --ignore=mlx_vlm/tests/test_utils.py`
- GitHub Actions `test` check passed on commit `814580ab`
- commit hooks: `black`, `isort`, `autoflake`

### DIFF
--- a/mlx_vlm/server/app.py
+++ b/mlx_vlm/server/app.py
@@ -110,11 +110,20 @@ def _build_gen_args(
         "enable_thinking",
         get_server_enable_thinking(),
     )
+    default_temperature = _model_config_field_or_default(
+        processor, "temperature", DEFAULT_TEMPERATURE
+    )
+    default_top_p = _model_config_field_or_default(processor, "top_p", DEFAULT_TOP_P)
+    default_top_k = _model_config_field_or_default(processor, "top_k", 0)
+    if _model_config_field_or_default(processor, "do_sample", None) is False:
+        default_temperature = 0.0
     args = GenerationArguments(
         max_tokens=max_tokens,
-        temperature=getattr(request, "temperature", DEFAULT_TEMPERATURE),
-        top_p=getattr(request, "top_p", DEFAULT_TOP_P),
-        top_k=getattr(request, "top_k", 0),
+        temperature=_request_field_or_default(
+            request, "temperature", default_temperature
+        ),
+        top_p=_request_field_or_default(request, "top_p", default_top_p),
+        top_k=_request_field_or_default(request, "top_k", default_top_k),
         min_p=getattr(request, "min_p", 0.0),
         seed=getattr(request, "seed", None),
         logprobs=bool(getattr(request, "logprobs", False)),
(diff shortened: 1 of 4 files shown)


lucasnewman (approved):
LGTM

PR #1342, opened 2026-06-09
Add server thinking defaults

Blaizzy:
## Summary

- Add server startup flags for thinking budget and thinking delimiter defaults.
- Apply those defaults when requests omit thinking fields while preserving request-level overrides.
- Document the startup flags and request parameter behavior.

## Testing

- `uv run pytest mlx_vlm/tests/test_server.py`


lucasnewman (approved):
Nice!

PR #1347, opened 2026-06-10
Add Gemma 4 DLM

Blaizzy:
Adds full support for the Gemma 4 DLM (block-diffusion language model) family: model implementation, checkpoint-parity generation, image inputs, quantized checkpoints, live terminal visualization, OpenAI-compatible server support, and long-context performance work. Also fixes masked-diffusion (LLaDA2) generation defaults and server routing that rode along with the server work.

## Model

- `diffusion_gemma4` model package: encoder/decoder backbone sharing one set of transformer weights (decoder adds the self-conditioning block), hybrid dense+MoE layers (top-8 of 128 experts), 5:1 sliding/full attention pattern, fused `SwitchLinear` experts.
- Generation matches the reference implementation: entropy-bound canvas sampler, linear temperature schedule, stable-and-confident adaptive stopping, self-conditioning via soft embeddings, defaults read from the checkpoint's generation config (48 steps, full 256-token canvas).
- Notable correctness fixes found during parity work: MoE router must route on the raw residual (routing on normalized states caused stray-token artifacts and prevented adaptive stopping from converging), fp32 logit softcap, generation-config EOS ids, `finish_reason` length/stop distinction.

## Features

- **Images**: Gemma 4 vision tower + multimodal embedder wired into the diffusion encoder, with bidirectional attention within image-token blocks (`use_bidirectional_attention="vision"`). Multi-image verified end-to-end.
- **Quantized checkpoints**: self-conditioning soft embeddings work with `QuantizedEmbedding` tables (dequantize-once-per-call + regular matmul, ~3x faster than `quantized_matmul(transpose=False)` at every bit width).
- **Live unmasking visualization**: nemotron/llada-style in-place canvas view (full sequence with `[Mask]` placeholders, display-width-exact wrapping, alternate-screen for tall outputs), on by default for verbose terminal runs, owned by the model package. Shared rendering primitives consolidated in `models/diffusion_visualizer.py`.
- **Server**: diffusion lane in the response generator — streams block-by-block (one SSE delta per denoised canvas) and aggregates for non-streaming, with correct `usage` token accounting. Masked-diffusion text models (LLaDA2, nemotron in diffusion mode) route through the same lane via a model-owned `on_block` callback.
- **Seeded generation**: `--seed` for reproducible canvas initialization.

## Performance

- Self-conditioning, entropy chain, and softcap fused/optimized: +23% step throughput on a 5-bit 26B (510 -> 627 token-steps/sec).
- Sliding-window decoder attention sliced to the window instead of masking the full encoder KV: per-step cost is now context-flat (was +120 ms from 1.7k -> 6.8k context); 8k-context generation +20%.
- bf16 26B benchmarks vs the AR `gemma-4-26b-a4b-it` baseline (max 500 tokens, temperature 0): diffusion wins prefill <=4k and generation at 2k context (31.8 vs 28.5 tok/s), trails at 8k (21.5 vs 27.9) where the full-attention layers' context cost dominates.

## Masked-diffusion fixes

- Sampler knobs now resolve as user arg -> config `default_diffusion_*` -> the model generate()'s own reference defaults, instead of forced shared constants. The forced values corrupted LLaDA2.0-mini output (token-editing oscillation); both LLaDA2.0 and 2.1 now generate cleanly with their reference semantics (threshold 0.95, no committed-token editing).
- Server routes masked-diffusion models away from the AR batch generator.

## Tests

201 passed, 1 skipped across the diffusion and server suites, including tiny-config coverage for image scatter, bidirectional masks, quantized embeddings, block-boundary streaming, visualizer rendering, and the LLaDA `on_block` callback.


mlx_vlm/models/diffusion_gemma4/README.md
```diff
@@ -0,0 +1,153 @@
...
+DiffusionGemma4 is a block-diffusion language model port for
+`gg-hf-st/test-checkpoint-26B`. It generates a canvas of noisy token ids and
+repeatedly denoises that canvas before appending the accepted tokens to the
+prefix.
+
+## Model
+
+| | |
+|---|---|
+| **Model ID** | `gg-hf-st/test-checkpoint-26B` |
```
lucasnewman: This doesn't look right?

lucasnewman (approved):
🚀

PR #1348, opened 2026-06-10
Fix DiffusionGemma long-context prefill

Blaizzy:
## Summary

Fix DiffusionGemma long-context generation by honoring the existing `prefill_step_size` setting in the block-diffusion prefill path.

## Root Cause

DiffusionGemma was prefilling the entire prompt in one encoder call. For ~100K-token prompts, the full-attention layers attempted to allocate a dense attention score workspace of roughly 313 GB, causing Metal allocation failure before generation started.

## Changes

- Add chunked prefill for safe text-only, no-padding DiffusionGemma prompts.
- Keep padded, static-cache, and visual-token prompts on the existing unchunked path to preserve attention semantics.
- Avoid materializing the all-ones attention mask for no-padding prefill.
- Add tiny-model tests for chunking, fallback behavior, and chunked-vs-unchunked token equivalence.

## Validation

- `uv run pytest mlx_vlm/tests/test_diffusion_gemma.py -q`
- `uv run pytest mlx_vlm/tests/test_diffusion_models.py::TestMaskedDiffusionServerLane::test_diffusion_generation_family_block -q`
- `uv run pre-commit run --all-files`
- Real generate smoke: ~99K rendered prompt tokens + `--max-tokens 500` completed with `--prefill-step-size 2048`.

### DIFF
--- a/mlx_vlm/generate/diffusion.py
+++ b/mlx_vlm/generate/diffusion.py
@@ -425,6 +425,64 @@ def _diffusion_static_cache_length(
     return prompt_length + cached_canvas_tokens
 
 
+def _diffusion_has_visual_token_types(mm_token_type_ids: Optional[mx.array]) -> bool:
+    if mm_token_type_ids is None:
+        return False
+    return bool(mx.any((mm_token_type_ids == 1) | (mm_token_type_ids == 2)).item())
+
+
+def _diffusion_should_chunk_prefill(
+    *,
+    prefill_step_size: Optional[int],
+    prompt_length: int,
+    has_padding: bool,
+    use_static_cache: bool,
+    pixel_values: Optional[mx.array],
+    mm_token_type_ids: Optional[mx.array],
+) -> bool:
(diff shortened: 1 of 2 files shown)


mlx_vlm/generate/diffusion.py
```diff
@@ -425,6 +425,64 @@ def _diffusion_static_cache_length(
...
+    mm_token_type_ids: Optional[mx.array],
+) -> bool:
+    if prefill_step_size is None or prompt_length <= prefill_step_size:
+        return False
+    if has_padding or use_static_cache or pixel_values is not None:
+        return False
+    # Visual spans can use bidirectional attention inside the whole block. Keep
+    # those prompts on the existing single prefill until chunking can split on
+    # visual-block boundaries without changing attention semantics.
+    return not _diffusion_has_visual_token_types(mm_token_type_ids)
```
Blaizzy, line 450: TODO: This should be policy inside the model just like in gemma4 AR

PR #1353, opened 2026-06-11
fix(server): honor per-request top_k/min_p (and seed) in the batched sampler

mmmugh:
## Problem

When `mlx_vlm.server` serves an LM/VLM, per-request `top_k` and `min_p` are silently ignored, and per-request `seed` doesn't vary output. A client sending `top_k=1` or `min_p=0.6` gets the same result as sending neither — a 200 OK with the wrong sampling.

## Root cause

The server runs continuous batching unconditionally for LM/VLM models, and its batched sampler `_PositionedTargetSampler` only implemented `temperature`/`top_p`. `_make_sampler` constructed it with only those (plus `seed`), so `top_k`/`min_p` never reached the sampler. The non-batched `stream_generate` path forwards them via `to_generate_kwargs`, but that path isn't reachable for an LM/VLM since batching can't be disabled. There were also two copies of the sampler — the one in `server/generation.py` had drifted from `generate/ar.py`.

## Fix

- Extend `_PositionedTargetSampler` to apply top-k (argpartition mask) and min-p (log-threshold mask) per row, vectorized over the batch, before the temperature/top-p draw — mirroring `mlx_lm.sample_utils`.
- Thread `top_k`/`min_p` through `_make_sampler`: seeded calls get the positioned (stateless, reproducible) sampler now covering them; unseeded calls keep `mlx_lm`'s sampler.
- Consolidate the duplicate sampler into `generate/ar.py`.
- Extend the generation_config sampling defaults (#1337) with `min_p`, and make `#1337`'s `top_k` default actually bind (it was inert without this sampler fix).

## Testing

- New regression tests in `mlx_vlm/tests/test_positioned_sampler.py` plus server/utils tests.
- Verified live against `Qwen3-VL-8B-Instruct-8bit` with continuous batching: `top_k=1` at `temperature=2.0` now matches greedy output byte-for-byte (previously multilingual token soup), `min_p=0.6` visibly constrains sampling, seeds reproduce/diverge as expected, and the default temperature/top_p-only path draws byte-identical tokens to the previous implementation (pinned in tests).

### DIFF
--- a/mlx_vlm/generate/ar.py
+++ b/mlx_vlm/generate/ar.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import contextlib
 import functools
 import logging
+import math
 import os
 import sys
 import time
@@ -45,6 +46,7 @@ DEFAULT_TEMPERATURE = 0.0
 DEFAULT_TOP_P = 1.0
 DEFAULT_TOP_K = 0
 DEFAULT_MIN_P = 0.0
+DEFAULT_SEED = 0
 DEFAULT_REPETITION_CONTEXT_SIZE = 20
 DEFAULT_PREFILL_STEP_SIZE = 2048
 DEFAULT_COMPLETION_BATCH_SIZE = 32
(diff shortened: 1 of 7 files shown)


lucasnewman:
This doesn't really work -- if the batch contains requests with different generation parameters, we need row-level sampling to handle that correctly. If you'd like to take another stab at it with that in mind, you're welcome to.

mmmugh:
@lucasnewman — you were right. Before I rewrite it, a sanity check on the shape.

mlx-lm's `GenerationBatch` already carries per-row samplers. mlx-vlm's fork collapsed the list into one object:

```python
# mlx_lm/generate.py
self.samplers: List[Callable]                   # one per row
def extend(self, batch): self.samplers.extend(batch.samplers)
def filter(self, keep):  self.samplers = [self.samplers[i] for i in keep]
```

On `main`, `extend()`/`filter()` never touch `self.sampler`, so a merging batch's sampler is dropped — that's "request #1 wins." Same root cause: `_make_sampler` drops `top_k`/`min_p` (`server/generation.py:1110`), and `greedy_sampling` is a batch-wide bool from request #1 (`:1371`).

**Proposal.** Carry `temperature/top_p/top_k/min_p/seed` as per-row `[B]` arrays through `insert → extend → filter`, mirroring `_rope_deltas`. Make the sampler row-aware, keep `sample_target`, and pass real `row_ids` — every call site passes `[0]*B` today, so the per-row RNG keying `_position_keys` implements is inert. Sampling becomes one sorted-space pass: `top_k` by rank, `top_p` by cumsum, `min_p` by ratio, greedy as `where(temp < eps, argmax, sampled)`. `apply_top_k` can't be reused per row — `mx.argpartition` rejects an array `kth`.

I'd pick that over restoring mlx-lm's per-row closures, which draw from the global RNG (and `speculative/mtp.py:368` duck-types on `sample_target`). But the plumbing is identical either way, so:

1. Vectorized `[B]` arrays, or mlx-lm's loop?
2. Plain AR + MTP now, `_run_speculative` (dflash/eagle3) as a follow-up?
3. `tests/test_server.py:4266` encodes one-sampler-per-batch — delete, or repurpose to assert co-batching?

First test I'd write: strengthen `test_positioned_target_sampler_is_batch_grouping_invariant` to two rows with distinct `row_ids` and distinct params, asserting the batched draw equals each row drawn alone. Fails on `main` today.

PR #1354, opened 2026-06-11
fix: guard cache_offset == 0 against array truth-value error in continuous batching

imi4u36d:
## Problem

Under continuous batching, `cache_offset` can be an `mx.array` (one element per sequence in the batch) rather than a scalar. Using `cache_offset == 0` directly in an `if` statement raises:

```
ValueError: [convert] Only length-1 arrays can be converted to Python scalars.
```

This crashes the engine loop on the second concurrent request for any affected VLM model.

## Root Cause

When `BatchKVCache` is used, `c0.offset` becomes an `mx.array`. While some models (e.g. ernie4_5_moe_vl, glm4v_moe, qwen3_vl_moe) already guard against this with `cache_offsets is None and cache_offset == 0`, several models do not.

## Fix

Guard all unguarded `cache_offset == 0` checks with a `cache_offsets is None` (or `cache_offset_array is None`) prefix, consistent with the pattern already used by other models.

### Affected files (5 files, 6 occurrences):

| File | Line | Guard variable |
|------|------|---------------|
| `qwen3_5/language.py` | 2450, 2472 | `cache_offsets` |
| `qwen2_5_vl/language.py` | 456 | `cache_offsets` |
| `qwen2_vl/language.py` | 449 | `cache_offsets` |
| `qwen3_vl/language.py` | 547 | `cache_offset_array` |
| `paddleocr_vl/language.py` | 428 | `cache_offsets` |

### Not affected:
- **hunyuan_vl** — already converts offset to scalar via `.item()`
- **ernie4_5_moe_vl, glm_ocr, glm4v_moe, qwen3_vl_moe, qwen3_omni_moe, glm4v** — already guarded

### Inheritance:
Fixing `qwen3_5/language.py` also fixes `qwen3_5_moe` and `minicpmv4_6` which inherit `LanguageModel.__call__` from it.

## Testing

Verified with a standalone reproduction script that:
1. Scalar `cache_offset == 0` still works correctly
2. Array `cache_offset` no longer crashes when guarded
3. Batched scenarios (all-zero, all-nonzero, mixed) all behave correctly


lucasnewman:
@imi4u36d Thanks for the fix! Please run `pre-commit run --all` to format and make sure your commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification).

lucasnewman (changes requested):
See comment.

imi4u36d:
Thanks for the guidance! I've run pre-commit to format the code and signed my commits. Please review again.

lucasnewman (approved):
🚀

PR #1363, opened 2026-06-12
Fix Qwen3-VL crash when cache offset is None

ptrpspeen:
## Problem

When a host application passes cache objects whose `offset` (or `_idx`) is still `None` before the first prefill, every Qwen3-VL request crashes with:

```
TypeError: unsupported operand type(s) for +=: 'int' and 'NoneType'
```

at `mlx_vlm/models/qwen3_vl/language.py` (`kv_seq_len += cache_size`). Encountered in the wild serving `typhoon-ai/typhoon-ocr1.5-2b` (arch `qwen3_vl`) via the vMLX engine, which uses custom cache implementations. Hit on `mlx_vlm==0.5.0`, reproduced against current `main`.

## Fix

Treat a `None` cache offset/`_idx` as an empty cache (0) at the four read sites in `qwen3_vl/language.py` and `qwen3_vl_moe/language.py`. No behavior change for stock int/`mx.array` offsets.

## Test

`test_qwen3_vl_none_cache_offset`: tiny Qwen3-VL with a cache whose `offset` starts as `None`. Fails with the exact `TypeError` above before the fix, passes after. `test_qwen3_vl` / `test_qwen3_vl_moe` still pass.


lucasnewman:
@ptrpspeen I don't know what vMLX is, but this seems to be working around a bug in a downstream project? Why not just fix the issue there to return 0 for the offset instead of None?

PR #1369, opened 2026-06-12
[Cohere] Fix streaming thinking parsing after prompt-opened

Terrencezzj:
## Summary
- Start chat-completions streaming in reasoning mode when the formatted prompt already ends with a known thinking start marker.
- Add test coverage.

## Test plan
- `python -m pytest mlx_vlm/tests/test_server.py -k "continues_prompt_thinking_block"`


lucasnewman:
@Terrencezzj Thanks for the fix! Can you make sure your latest commit is [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and then we can merge?

lucasnewman (approved):

PR #1372, opened 2026-06-12
Add preloading for multiple model kinds in the server

lucasnewman:
Adds the following options in addition to `--model` to allow preloading models for different model kinds:

```bash
mlx_vlm.server --model <language_model> \
  --image-model <image_generation_model> \
  --tts-model <text_to_speech_model> \
  --stt-model <speech_to_text_model>
```


Blaizzy (approved):
LGTM 🚀

PR #1374, opened 2026-06-13
Minimax m3 support

ivanfioravanti:
### Minimax m3 support experiment

Status: minimax-m3-support has substantial MiniMax M3 support implemented:
- Native minimax_m3_vl model implementation.
- Functional MLX MiniMax Sparse Attention with index-key side cache.
- Sparse/cache behavior for batch, rollback, APC, chunked prefill, padding/causal masks.
- MiniMax prompt formatting for image/video tokens, typed media counts, and video counts.
- MiniMax thinking mode mapping and server thinking parsing.
- MiniMax XML-style tool-call parsing.
- MXFP8/MoE sanitization paths and quantization config handling.
- Local MiniMax processor auto-trust plus image/video resize and metadata patches. 

Working on performance optimizations right now and some more testing.

4bit model for initial tests uploaded on mlx-community [here](https://huggingface.co/mlx-community/MiniMax-M3-4bit).


ivanfioravanti:
MSA/text/image/video paths implemented! 
Context Benchmark test in progress 👷🏻‍♂️

theuseless-ai:
Heads-up in case it's useful: the `mlx-community/MiniMax-M3-4bit` checkpoint linked here doesn't load on the current branch (`8b6e3c1`) — it errors with:

```
ValueError: Received 855 parameters not in model:
  ...block_sparse_moe.shared_experts.{gate,up,down}_proj.*
```

That checkpoint stores routed experts pre-stacked as `switch_mlp.{gate,up,down}_proj` with a separate `shared_experts`, but the packed path in `_sanitize_moe_weights` only handles the per-expert HF layout, so the `shared_experts.*` tensors get orphaned. It's also flagged `format=mlx`, so `load_model` skips `sanitize()` entirely (which I think is the actual blocker — the pack design alone isn't the whole story).

I took a small stab at a fix (2 files):
- a fallback in `_sanitize_moe_weights` that builds the packed `gate_up_proj`/`down_proj` from the pre-stacked `switch_mlp` + separate `shared_experts`, and
- an opt-in `always_sanitize` flag on the M3 `Model`, honored in `load_model`, so `sanitize()` runs even for this `format=mlx` checkpoint.

Branch / diff: https://github.com/ivanfioravanti/mlx-vlm/compare/minimax-m3-support...theuseless-ai:mlx-vlm:fix/m3-prestacked-checkpoint-loader

Verified on an M3 Ultra 512GB: loads + correct text & vision output, ~32–34 tok/s (vs ~23 on the early `3f70318` commit the checkpoint previously needed). First time poking around mlx-vlm internals so I may have missed context — happy to open a PR against the branch if it'd help.

trevorgordon981:
Confirming @theuseless-ai's diagnosis — and it bites the 8-bit checkpoint the same way. Two independent blockers:

1. **`format=mlx` short-circuits `sanitize()`.** `load_model` skips it for mlx-format checkpoints, so `_sanitize_moe_weights` never runs regardless of what it does. `always_sanitize` is the right lever.
2. **Pre-stacked *packed* MoE layout.** These checkpoints ship `block_sparse_moe.switch_mlp.{gate,up,down}_proj` already stacked (128 experts) with a separate `shared_experts.*`, and with `pack_shared_expert=True` the model wants a *packed* `switch_mlp.gate_up_proj` of **129** experts (gate/up concatenated, shared folded in as expert 128). `_sanitize_moe_weights` only builds from the unstacked HF `experts.N.w1/w2/w3` keys, so on these it's a no-op and `shared_experts.*` orphan -> the `855 parameters not in model`.

I have it loading clean on the 8-bit (`0 missing / 0 extra`, strict `load_weights`; offline key/shape level so far — full generation parity is the next check). Three pieces:
- a pre-stacked branch in `_sanitize_moe_weights` (rename router -> `block_sparse_moe.*`; under `pack_shared`, concat the stacked gate/up on the output axis + append the shared expert -> packed 129-expert `gate_up_proj`/`down_proj`; quant `weight`/`scales`/`biases` carry through the same concat),
- indexer naming (`self_attn.indexer.{q,k}_{proj,norm}` vs the flat `self_attn.index_*` the attention builds),
- stripping the VL extras when loading into the text-only `Model` (else strict `load_weights` raises on leftover `vision_tower.*`).

One engine-side gap too, independent of the checkpoint: `mlx_vlm/generate/ar.py::_make_cache.to_batch_cache` dispatches on `isinstance` and never calls `MiniMaxM3KVCache.to_batch()`, so the server raises `does not yet support batching` on the first batched request. A one-liner (honor a cache's own `to_batch()` before the isinstance chain) routes the sparse layers through the model's own converter — 86/86 M3 + 80/80 generate tests pass after it.

Glad to PR all of this onto your branch and close mine (#1375) — yours is well ahead on vision/video, MSA batching/APC/chunked-prefill, thinking, and tools; no reason to split the effort. Whatever keeps it to one clean MoE-stacking path.

trevorgordon981:
Pushed the lot on top of your branch in case it's useful: https://github.com/trevorgordon981/mlx-vlm/tree/m3-mlx-loading-fix — two commits (the pre-stacked/packed-MoE loading fix + the `to_batch_cache` one-liner), `166 passed / 0 failed` on `test_minimax_m3.py` + `test_generate.py`. Open a PR off it whenever, or just cherry-pick — whatever's least friction.

ivanfioravanti:
@Blaizzy what's the staus? It seems recent commits are not working with model uploaded on HF. Happy to convert it and reupload in case. Keep us posted.

ivanfioravanti:
Re-ran the converter on this branch and re-uploaded mlx-community/MiniMax-M3-4bit. It now ships the packed layout directly, block_sparse_moe.switch_mlp.gate_up_proj/down_proj at 129 experts (128 routed + shared folded in under pack_shared_expert), no orphaned shared_experts.* or experts.N.*, still format=mlx. Loads and generates correctly; text and vision both verified.

Since _sanitize_moe_weights now produces canonical packed layout at conversion time, the format=mlx skip-sanitize path works as intended. Fixed at the source rather than on every load. Thanks to both for catching the stale checkpoint.

@trevorgordon981  your second point is independent of the checkpoint and valid: to_batch_cache in generate/ar.py only dispatches on isinstance for the standard cache types and never calls MiniMaxM3KVCache.to_batch(), so the server raises does not yet support batching on the first batched request. I'd love to land that one-liner (plus the indexer-naming piece, if the re-conversion didn't already resolve it). Want to open a PR against this branch with just those, dropping the MoE-loading changes, since loading's handled now? Happy to review/merge fast. And agreed, let's keep it to one effort.

trevorgordon981:
Done — opened ivanfioravanti/mlx-vlm#3 against your branch with just the `to_batch_cache` one-liner (the engine-side batching fix), MoE-loading changes dropped since your re-conversion handles them.

Re: the indexer-naming piece — checked, and it's **not** needed anymore. The re-converted `mlx-community/MiniMax-M3-4bit` already ships flat `self_attn.index_*` keys (`index_q_proj`, `index_k_norm`, etc.), so the rename would be a no-op. Left it out to keep the PR to just the one independent fix.

Should be a fast review. Thanks for the quick re-conversion!

ivanfioravanti:
Merged! Super @trevorgordon981 
@Blaizzy up to you 🙌🏻

trevorgordon981:
I implemented MiniMax-M3 support independently too (the now-closed #1429), and since the vision patch-embed is exactly where my `nn.Conv3d`-based approach hit a channel-order bug, I ran a numerical-equivalence check of this PR's `MiniMaxVisionPatchEmbedding` against the HF `transformers` reference. Sharing in case it's useful:

**Result: bit-equivalent to the HF reference, no ordering issues.**

- Ran this PR's `MiniMaxVisionPatchEmbedding` (flatten-and-matmul, no Conv3d) vs the verbatim HF `MiniMaxM3VLVisionEmbeddings.forward` (`nn.Conv3d`, kernel = stride = patch) on the **same input + same weights**, real M3 patch geometry (C=3, T=2, patch=14 → patch_dim=1176).
- **max abs diff = 1.19e-6** — pure fp32 accumulation-order noise between Conv3d and matmul.
- **No channel/spatial scramble.** A channel-permutation probe is the decisive check: permuting the input channel axis shifts the output by ~3.85 (so it genuinely depends on channel order), and this PR's permuted output still matches HF's permuted output to **9.5e-7**. That's the precise failure mode a naive channels-last Conv3d hits — the flatten-and-matmul here avoids it by construction, flattening `(C, T, ph, pw)` in the same row-major order the conv kernel uses.

So the image patch-embed path looks correct. Happy to also run it end-to-end through `mlx_vlm convert -q --q-bits 8` + an image inference when I get a free window, if that'd be helpful. Nice work on this.

Blaizzy:
> I implemented MiniMax-M3 support independently too (the now-closed #1429), and since the vision patch-embed is exactly where my `nn.Conv3d`-based approach hit a channel-order bug, I ran a numerical-equivalence check of this PR's `MiniMaxVisionPatchEmbedding` against the HF `transformers` reference. Sharing in case it's useful:
> 
> **Result: bit-equivalent to the HF reference, no ordering issues.**
> So the image patch-embed path looks correct. Happy to also run it end-to-end through `mlx_vlm convert -q --q-bits 8` + an image inference when I get a free window, if that'd be helpful. Nice work on this.


Thanks @trevorgordon981, feel free to run tests and let me know if you find anything

PR #1378, opened 2026-06-14
Fix token_context desync that crashes continuous batching

### LINKED ISSUE #1441: Continuous batching: IndexError in GenerationBatch.filter() (token_context desync after json-schema request)
### Summary

Under continuous batching, a plain request admitted right after a
guided/json-schema (structured-output) request finishes can crash the generation
thread with `IndexError: list index out of range` in `GenerationBatch.filter()`.

### Error

```
IndexError: list index out of range
```

in `mlx_vlm/generate/ar.py`, `GenerationBatch.filter()`, when it reindexes
`token_context`.

### Root cause

`GenerationBatch.filter()` reindexes `token_context` with indices valid for
`uids`, relying on `len(token_context)` being either `0` or `len(uids)`. That
invariant breaks when a batch carries a populated `token_context` with **no
active logits processor** — the state left after a guided/json-schema request
finishes — and a plain request is admitted via `extend()`:

- `extend()` only realigns `token_context` when a side has an active processor;
- `_ensure_token_context()` returns early in the all-None (no-processor) case;

so the merged `token_context` ends up non-empty but **shorter than `uids`**. The
next eviction then raises the `IndexError` in `filter()`.

### Reproduction

With the server running and continuous batching enabled:

1. Send a request with `response_format` json_schema (a logits processor
   populates `token_context`); let it finish.
2. Admit a plain request (no `response_format`) into the same batch via
   `extend()`.
3. The next filter/eviction raises `IndexError: list index out of range`.

### Environment

- mlx_vlm `main` (also reproduces on 0.6.3)
- Apple Silicon, continuous batching enabled, mixed json-schema + plain requests

### Proposed fix

Make `_ensure_token_context()` realign a non-empty `token_context` to `uids`
even when no processor is active, so `extend()`'s existing normalization repairs
the desync (padding uses empty contexts, never read for sequences without an
active processor). Fix + regression tests are already up as **#1378**, which can
reference this issue.

rmaher001:
Fixes #1441

## Problem

Under continuous batching, `GenerationBatch.filter()` can raise `IndexError: list index out of range` at:

```python
self.token_context = [self.token_context[idx] for idx in keep]
```

It surfaces when the batch mixes a request that uses a logits processor (e.g. `response_format` / json-schema guided decoding, which populates `token_context` for the whole batch) with plain requests: once the guided request leaves the batch no processor is active, and a plain request is then admitted via `extend()`.

## Root cause

`GenerationBatch` keeps parallel per-sequence lists (`uids`, `token_context`, `logits_processors`, …) that must satisfy the invariant `len(token_context) ∈ {0, len(uids)}` — `filter()` reindexes `token_context` with indices valid for `uids` and relies on it.

`_ensure_token_context()` only pads/truncates `token_context` to `len(uids)` when `force` or an *active* processor is present; in the all-`None` branch it returns early, leaving a non-empty `token_context` unnormalized. `extend()` only realigns the two sides when at least one has an active processor, so when neither does, `self.token_context.extend(other.token_context)` can leave `token_context` non-empty but shorter than `uids`. The next eviction → `filter()` → `IndexError`.

## Fix

Make `_ensure_token_context()` also realign a *non-empty* `token_context` to `len(uids)` when no processor is active (fall through to the existing pad/truncate instead of returning early). The `_ensure_token_context()` call already at the end of `extend()` then repairs the desync. Padding with empty contexts is correct: a plain request has no constraint history, and that empty entry is never read for a sequence without an active processor (the `_step()` processor loop is gated on `any(self.logits_processors)` and skips `None` slots).

## Tests

Adds three model-free regression tests to `TestBatchGenerator` that reproduce the exact `extend()` + `filter()` desync (a leftover-guided batch + plain request(s)) and assert `len(token_context) == len(uids)` plus no `IndexError`. They fail before the fix and pass after; the full `test_generate.py` suite stays green.

### DIFF
--- a/mlx_vlm/generate/ar.py
+++ b/mlx_vlm/generate/ar.py
@@ -880,7 +880,13 @@ class GenerationBatch:
         if not (force or (self.logits_processors and any(self.logits_processors))):
             if not self.logits_processors:
                 self.token_context = []
-            return
+                return
+            if not self.token_context:
+                return
+            # token_context was populated while a processor was active and is
+            # still carried after that sequence left the batch. Fall through to
+            # keep it aligned with uids so filter() can reindex it safely even
+            # though no processor is currently active.
         if len(self.token_context) < len(self.uids):
             missing = len(self.uids) - len(self.token_context)
             self.token_context.extend([[] for _ in range(missing)])
--- a/mlx_vlm/tests/test_generate.py
+++ b/mlx_vlm/tests/test_generate.py
@@ -816,6 +816,121 @@ class TestBatchGenerator:
         assert [r.token for r in first] == [5, 6, 7]
         assert seen_contexts == [[30, 7]]
 
+    def test_generation_batch_extend_aligns_token_context_without_active_processor(
+        self,
+    ):
+        # A batch can carry a populated token_context while having no *active*
+        # processor: the state left behind after a guided/json-schema request
+        # (which populates token_context for the whole batch) has finished.
+        # Admitting a plain request must keep token_context aligned with uids;
+        # otherwise filter() later indexes past the end of token_context.
+        # (logits_processors=[None] is a reserved per-sequence slot with no
+        # active processor -- distinct from an empty [] processor list.)
+        sampler = lambda logprobs: mx.argmax(logprobs, axis=-1)
+        stop_criteria = lambda token: False
+        leftover = GenerationBatch(
+            model=MagicMock(),
+            uids=[0],
+            inputs=mx.array([5], dtype=mx.int32),
+            prompt_cache=[],
+            sampler=sampler,
+            stop_criteria=stop_criteria,
+            max_tokens=[8],
+            token_context=[[30]],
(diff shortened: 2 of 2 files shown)


rmaher001:
Friendly ping on this one 🙂 — CI is green and it still merges cleanly onto current `main`. This fixes a token_context desync that crashes continuous batching, which we're hitting in production serving Gemma-4. Happy to rebase or adjust if anything's needed. Thanks for all the work on mlx-vlm!

lucasnewman:
@rmaher001 Thanks! Can you fix the merge conflict and [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits and then we can merge?

rmaher001:
Closing this out — the same continuous-batching `token_context` desync is now fixed on `main` via #1451 (which closed the matching issue #1441). Thanks @Blaizzy for the quick turnaround on both this and the Gemma 4 mask fix (#1445)!

Issue #1340, opened 2026-06-09
gemma-4-e4b-it-4bit fails to load: load_model() drops strict kwarg before model.load_weights()


gierran:
## Summary

`mlx-community/gemma-4-e4b-it-4bit` fails to load via both the CLI and the Python API with `ValueError: Received 126 parameters not in model: language_model.model.layers.{24..41}.self_attn.{k_proj,v_proj,k_norm}.*`. The published 4bit checkpoint contains K/V/k_norm weights for the kv-shared layers (24-41 in E4B; `num_kv_shared_layers=18`, `num_hidden_layers=42`), but `models/gemma4/language.py` correctly skips creating those parameters on kv-shared layers (`kv_shared_only=True`). The strict mlx loader then rejects the extras.

The root cause is in `mlx_vlm/utils.py`: `load()` plumbs a `strict` kwarg through to `load_model()`, but `load_model()` calls `model.load_weights(list(weights.items()))` without forwarding it, so `mlx.nn.Module.load_weights` defaults to `strict=True` and raises.

## Versions

- mlx-vlm: 0.6.2 (also reproduced on `main` HEAD `c9a6743`)
- mlx: 0.31.2
- mlx-lm: 0.31.3
- Python: 3.12.13
- macOS 26.3.1 / Apple M4 (32GB)

## Repro

```bash
python -m venv venv && source venv/bin/activate
pip install -U mlx-vlm
python -m mlx_vlm.generate \
  --model mlx-community/gemma-4-e4b-it-4bit \
  --prompt "Hello" --max-tokens 20
```

Result:

```
File "mlx_vlm/utils.py", line 632, in load_model
    model.load_weights(list(weights.items()))
File "mlx/nn/layers/base.py", line 185, in load_weights
    raise ValueError(
ValueError: Received 126 parameters not in model:
language_model.model.layers.24.self_attn.k_norm.weight,
language_model.model.layers.24.self_attn.k_proj.biases,
language_model.model.layers.24.self_attn.k_proj.scales,
language_model.model.layers.24.self_attn.k_proj.weight,
language_model.model.layers.24.self_attn.v_proj.biases,
... (through layer 41)
```

Passing `strict=False` via the Python `load(..., strict=False)` API does **not** help, because `load_model` discards it before calling `model.load_weights`.

## Diagnosis

`text_config` of the checkpoint:
- `num_hidden_layers: 42`
- `num_kv_shared_layers: 18`
[...]

PR #1382, opened 2026-06-14
fix(gemma4): load legacy MLX checkpoints with obsolete shared-KV weights

mvance:
In the interest of full disclosure, I leveraged AI assistance for this PR; I'd classify it as "Level 8: Bots planned, human approved" on the [VisiData scale](https://www.visidata.org/blog/2026/ai/). The changes have gone through multiple rounds of review with CodeRabbit, Gemini, cubic.dev, Claude, and Codex to converge on a focused, low-risk patch.

## Background

The issue first surfaced when attempting to load the `mlx-community/gemma-4-e4b-it-4bit` model with a recent version of mlx-vlm. Following a performance optimization that introduced shared KV-cache modules for Gemma 4 models, the library began expecting a reduced set of weight tensors. However, many community checkpoints converted prior to this change still contain the "extra" legacy KV weights. Because `load_model` skips weight sanitization entirely for MLX-format checkpoints, these redundant weights were passed straight through to `model.load_weights`, triggering a `ValueError` for unexpected parameters and preventing the model from being used.

## Fix

Apply a narrow Gemma 4-specific filter at the end of `load_model`, gated on both `is_mlx_format` and `model_type == "gemma4"`:

```python
if is_mlx_format and model_type == "gemma4":
    model_keys = {k for k, _ in tree_flatten(model.parameters())}
    weights = {k: v for k, v in weights.items() if k in model_keys}
```

The filter runs after `nn.quantize` and `quantize_activations` so that the allowed-key set reflects the model's final, quantized parameter tree (i.e. `.scales` / `.biases` are preserved for quantized layers). It runs before `model.load_weights`, so strict loading receives only keys the model owns.

## What this PR does not change

- MLX format detection is unchanged (metadata check on the first shard).
- Non-MLX sanitization is unchanged for all models, including Gemma 4 (the existing `Gemma4.LanguageModel.sanitize()` path already strips obsolete shared-KV weights via `_is_unused_shared_kv_weight`).
- The existing `load_model` quantization predicate is unchanged.
- No public APIs change.

Earlier iterations attempted broader changes (multi-shard MLX detection, restructured non-MLX sanitization, extracted quantization predicate). Review feedback consistently pushed back on the blast radius of those changes, since shared sanitization paths affect many unrelated models. This patch is intentionally limited to the specific regression.

## A note on `gemma4_unified`

`gemma4_unified` imports the same `LanguageModel` and could in principle have the same shape of regression. However, `gemma4_unified` landed only two days before `_is_unused_shared_kv_weight` was added in #1301, so the window for legacy `gemma4_unified` MLX checkpoints is very narrow and no such failure has been reported. Expanding the gate to `model_type in ("gemma4", "gemma4_unified")` is a trivial follow-up if it ever becomes necessary.

## Tests

- `tests/test_gemma4_regression.py::test_load_mlx_format_filters_legacy_weights` — builds a minimal Gemma 4 model with `num_kv_shared_layers=2`, saves an MLX-format checkpoint containing the legacy per-layer KV weights, and verifies strict `load_weights` succeeds: obsolete shared-KV weights are dropped, valid weights (`layer_scalar` etc.) are preserved.
- `tests/test_gemma4_regression.py::test_load_non_mlx_format_sanitizes_legacy_weights` — regression guard confirming the non-MLX path still strips the same legacy weights via `Gemma4.LanguageModel.sanitize`.
- `tests/test_utils.py::test_load_model_does_not_filter_non_gemma_mlx_weights` — verifies the new filter is gated on `model_type == "gemma4"` and does not affect other model types on the MLX path.

## Validation

```
.venv/bin/python -m pytest mlx_vlm/tests/test_gemma4_regression.py -q
.venv/bin/python -m pytest mlx_vlm/tests/test_utils.py -q
```

Both pass locally, and `mlx-community/gemma-4-e4b-it-4bit` loads successfully with the patch.


mvance:
**Related: issue #1340 and PR #1345**

After opening this PR I found that issue #1340 describes the same root cause in detail (including the exact 7 weights × 18 layers = 126 unexpected parameters for the E4B checkpoint). The diagnostic work there confirmed what this PR is fixing.

PR #1345 takes a different but complementary approach: it fixes the `strict` kwarg propagation bug so that `strict=False` can be threaded through `load()` → `load_model()` → `model.load_weights()`. That's a legitimate bug fix on its own — the kwarg was accepted but silently dropped. This PR and #1345 are not mutually exclusive:

- **#1345** restores the expected behavior of the `strict` API so users can opt in to non-strict loading.
- **This PR** makes MLX-format Gemma 4 checkpoints load correctly by default, without users needing to know about `strict=False` or the architectural discrepancy between the checkpoint and the current model.

Happy to coordinate with the author of #1345 if the maintainer prefers to handle these together or in a specific order.

lucasnewman:
@Blaizzy Do you have the ability to update the weights? This issue keeps coming up.

Blaizzy:
Done ✅

PR #1384, opened 2026-06-15
Add an opt-in --diffusion-turbo inference engine for DiffusionGemma (block-diffusion)

programVeins:
## Add an opt-in `--diffusion-turbo` engine for DiffusionGemma

A high-throughput, **fully opt-in** generation path for DiffusionGemma-style block-diffusion
models (`mlx-community/diffusiongemma-26B-A4B-it-4bit`). On an M5 Pro (macOS 27 beta) it runs
**~4–7× the stock diffusion engine** at the same 512-token measurement, with **GSM8K exact-match
equal to the reference** at the recommended setting.

With `--diffusion-turbo` absent, nothing changes - the new module isn't imported and
`stream_diffusion_generate` is untouched.

Screen Recording 2026-06-15 at 3 17 33 PM

### Performance (M5 Pro, fresh process, 512 tokens)

| prompt | stock | turbo fast | turbo quality | speedup |
| --- | --- | --- | --- | --- |
| explain speculative decoding | 25.7 | **107.5** | 104.5 | ~4.2× |
| short story | 18.9 | **137.6** | 120.2 | ~7.3× |
| what is MLX | 18.9 | **108.9** | 88.9 | ~5.8× |

GSM8K[:20] averaged ~86 tok/s (short answers spend more steps/token). Quality at
`--turbo-steps 10 --turbo-repair`: **11/12 GSM8K, equal to the reference loop.**

### How it works

Block diffusion denoises a 256-token canvas over several forward passes, freezing positions as
they get confident. The stock loop is good but spends most of each step on work that's
unnecessary once you accept a small, controllable fidelity trade. The engine combines:

1. **Exact top-64 sampler chain** - softcap/temperature/entropy/sampling run on the top-K logits,
   not the full 262k vocab (~35 ms → ~3 ms/step). A two-stage chunk-max top-K is provably exact;
   softcap/temperature are monotonic, so top-K on raw logits is exact.
2. **Top-K self-conditioning** - gather 64 embedding rows instead of `softmax(logits) @ embed`
   (drops a ~378-GFLOP matmul and a dequantized ~1.5 GB table).
3. **Compacted active-set decoder** (`--turbo-compact`) - forward only live canvas positions,
   serving frozen ones from per-layer K/V buffers; bit-exact vs the stock decoder when nothing is
   frozen (documented runner invariant).
4. **Shape bucketing** - round live-position counts to ≤10 fixed sizes, bounding MLX's compiled-
   kernel cache. Fixes an in-process throughput decay (compiled-graph growth from variable shapes,
   not a memory leak).
5. **Scheduling & safety** - two-stage confidence schedule (no forced flush; the flush caused a
   multi-canvas "the the the" collapse), a repair pass that recovers reference final-step
   semantics, EOS tail-drop, and a repeat-guard degeneration stop.

### Files

`generate/diffusion_turbo.py` (engine), `generate/diffusion_turbo_runner.py` (compacted decoder),
`generate/diffusion.py` (+dispatch hook), `generate/dispatch.py` (+flags, +chat stats footer).
~1.7k lines, additive, 0 deletions. Validated against the reference loop on the live checkpoint
(GSM8K parity; bit-exact decoder; both RoPE variants checked numerically). Happy to add unit tests
for the shape-independent helpers (`_topk_logits` vs `mx.topk`, bucketing, acceptance mask) - these
need no model download.

### Scope & caveats

- **DiffusionGemma-only**, batch size 1, never auto-selected; raises a clear error otherwise.
- **macOS 27 beta:** engine uses int32 throughout to avoid a Metal-JIT int64-scatter bug (will raise separate
  upstream report to `ml-explore/mlx`, not fixed here).
- **MLX dependency:** full MoE throughput on M5 needs the `gather_qmm` `bk32`→`bk64` fix, already
  addressed by approved PR [`ml-explore/mlx#3632`](https://github.com/ml-explore/mlx/pull/3632). The engine runs correctly on stock MLX without
  it - just slower on that path. Out of scope here.

Can split into reviewable PRs if preferred (core loop → compacted decoder → bucketing → polish).

### Usage

```bash
# Recommended: reference-level quality, ~4–6×
python -m mlx_vlm generate \
  --model mlx-community/diffusiongemma-26B-A4B-it-4bit \
  --prompt "Explain how speculative decoding works in detail" \
  --max-tokens 512 --skip-special-tokens \
  --gen-kwargs '{"diffusion_turbo": true, "turbo_compact": true, "turbo_accept": "confidence", "turbo_threshold": [0.95, 0.80, 8], "turbo_repair": true}';
```

Dials (keys inside `--gen-kwargs`):
- `diffusion_turbo: true` — enable the engine (required to opt in).
- `turbo_compact: true` — compacted active-set decoder; the main speedup (recommended).
- `turbo_accept` — `"confidence"` (default) or `"entropy-bound"` (checkpoint-faithful rule).
- `turbo_threshold` — commit confidence. A float (e.g. `0.9`) or a two-stage
  `[hi, lo, switch_step]` like `[0.95, 0.80, 8]` (picky early, relaxed for stragglers;
  higher = more faithful/slower). The two-stage form is recommended.
- `turbo_repair: true` — final-canvas re-argmax recovering the reference sampler's
  final-step semantics (recommended).
- `entropy_bound` — only with `turbo_accept: "entropy-bound"`; `0.1` = reference.
- `turbo_steps` — optional hard per-canvas flush cap (aggressive; can degrade long-form,
  leave unset unless maximizing speed on short content).

Per-canvas step ceiling uses the existing flag `--max-denoising-steps N` (works alongside `--gen-kwargs`).

### DIFF
--- a/mlx_vlm/generate/diffusion.py
+++ b/mlx_vlm/generate/diffusion.py
@@ -4,15 +4,24 @@ import logging
 import os
 import shutil
 import time
-import unicodedata
 from typing import Any, Dict, Generator, List, Optional, Tuple
 
 import mlx.core as mx
 import mlx.nn as nn
 from transformers import PreTrainedTokenizer
 
+from ..models.diffusion_visualizer import (
+    clip_display_width,
+    display_width,
+    escape_carriage_returns,
+)
 from ..tokenizer_utils import make_streaming_detokenizer
-from .common import GenerationResult, generation_stream, wired_limit
+from .common import (
+    GenerationResult,
+    _chunked_prefill_enabled,
+    generation_stream,
+    wired_limit,
+)
 
 logger = logging.getLogger("mlx_vlm.generate")
 
@@ -20,41 +29,6 @@ DEFAULT_TEMPERATURE = 0.0
 DEFAULT_DIFFUSION_MIN_CANVAS_LENGTH = 64
 DEFAULT_DIFFUSION_MAX_DENOISING_STEPS = 48
 DEFAULT_DIFFUSION_UNMASKING_WIDTH = 0
-
-
-def _display_width(text: str) -> int:
-    width = 0
-    for char in text:
-        if unicodedata.combining(char):
-            continue
-        width += 2 if unicodedata.east_asian_width(char) in ("F", "W") else 1
-    return width
-
-
-def _clip_display_width(text: str, max_width: int) -> str:
-    if max_width <= 0:
-        return ""
-
-    out = []
-    width = 0
-    clipped = False
-    for char in text:
-        if unicodedata.combining(char):
-            char_width = 0
-        else:
-            char_width = 2 if unicodedata.east_asian_width(char) in ("F", "W") else 1
-        if width + char_width > max_width:
-            clipped = True
-            break
-        out.append(char)
-        width += char_width
-
-    if clipped and max_width >= 3:
-        while out and _display_width("".join(out)) > max_width - 3:
-            out.pop()
-        out.append("...")
-
-    return "".join(out)
 
 
 def _diffusion_display_limit(requested_width: Optional[int] = None) -> Optional[int]:
@@ -80,10 +54,11 @@ def _format_diffusion_draft_line(
     response: GenerationResult,
     requested_width: Optional[int] = None,
 ) -> str:
+    text = escape_carriage_returns(response.draft_text)
     width = _diffusion_display_limit(requested_width)
     if width is None:
-        return response.draft_text
-    return _clip_display_width(response.draft_text, width)
+        return text
+    return clip_display_width(text, width)
 
 
 def _print_diffusion_draft(
@@ -98,15 +73,15 @@ def _format_diffusion_live_text(
     text: str,
     requested_width: Optional[int] = None,
     *,
-    preserve_newlines: bool = False,
+    preserve_newlines: bool = True,
 ) -> str:
     width = _diffusion_display_limit(requested_width)
-    text = text.replace("\r", "\\r")
+    text = escape_carriage_returns(text)
     if not preserve_newlines:
         text = text.replace("\n", "\\n")
     if width is None:
         return text
-    return _clip_display_width(text, width)
+    return clip_display_width(text, width)
 
 
 def _print_diffusion_live_text(
@@ -131,7 +106,7 @@ def _terminal_rows_for_text(text: str, columns: Optional[int] = None) -> int:
     columns = max(1, columns)
     rows = 0
     for line in text.split("\n"):
-        width = _display_width(line)
+        width = display_width(line)
         rows += max(1, (width + columns - 1) // columns)
     return rows
 
@@ -197,36 +172,6 @@ def diffusion_kwargs_from_args(args: Any, config: Any) -> Dict[str, Any]:
         return {}
 
     kwargs = {}
-    if getattr(args, "diffusion_turbo", False):
-        kwargs["diffusion_turbo"] = True
-        if getattr(args, "entropy_bound", None) is not None:
-            kwargs["entropy_bound"] = args.entropy_bound
-        if getattr(args, "turbo_no_monotone", False):
-            kwargs["turbo_monotone"] = False
-        if getattr(args, "turbo_steps", None) is not None:
-            kwargs["turbo_steps"] = args.turbo_steps
-        if getattr(args, "turbo_accept", None):
-            kwargs["turbo_accept"] = args.turbo_accept
-        if getattr(args, "turbo_repair", False):
-            kwargs["turbo_repair"] = True
-        if getattr(args, "turbo_compact", False):
-            kwargs["turbo_compact"] = True
-        if getattr(args, "turbo_repeat_guard", None) is not None:
-            kwargs["turbo_repeat_guard"] = args.turbo_repeat_guard
-        # Two-stage confidence schedule: picky (threshold) for the first
-        # `relax_after` steps, then relaxed (relax_to) for stragglers. This
-        # keeps canvas structure clean without the all-at-once final flush
-        # that compounds errors across re-encoded canvases.
-        relax_to = getattr(args, "turbo_threshold_relax", None)
-        relax_after = getattr(args, "turbo_threshold_relax_after", None)
-        if args.threshold is not None and relax_to is not None:
-            kwargs["turbo_threshold"] = [
-                args.threshold,
-                relax_to,
-                int(relax_after if relax_after is not None else 8),
-            ]
-        elif args.threshold is not None:
-            kwargs["turbo_threshold"] = args.threshold
     if args.max_denoising_steps is not None:
         kwargs["max_denoising_steps"] = args.max_denoising_steps
     if args.diffusion_full_canvas:
@@ -368,6 +313,7 @@ def _diffusion_soft_embedding_weight(embed_tokens: nn.Module) -> mx.array:
             embed_tokens.biases,
             group_size=embed_tokens.group_size,
             bits=embed_tokens.bits,
+            mode=getattr(embed_tokens, "mode", "affine"),
         )
     return embed_tokens.weight
 
@@ -453,31 +399,6 @@ def _diffusion_static_cache_length(
         model_canvas_length
     )
     return prompt_length + cached_canvas_tokens
-
-
-def _diffusion_has_visual_token_types(mm_token_type_ids: Optional[mx.array]) -> bool:
-    if mm_token_type_ids is None:
-        return False
-    return bool(mx.any((mm_token_type_ids == 1) | (mm_token_type_ids == 2)).item())
-
-
-def _diffusion_should_chunk_prefill(
-    *,
-    prefill_step_size: Optional[int],
-    prompt_length: int,
-    has_padding: bool,
-    use_static_cache: bool,
-    pixel_values: Optional[mx.array],
-    mm_token_type_ids: Optional[mx.array],
-) -> bool:
-    if prefill_step_size is None or prompt_length <= prefill_step_size:
-        return False
-    if has_padding or use_static_cache or pixel_values is not None:
-        return False
-    # Visual spans can use bidirectional attention inside the whole block. Keep
-    # those prompts on the existing single prefill until chunking can split on
-    # visual-block boundaries without changing attention semantics.
-    return not _diffusion_has_visual_token_types(mm_token_type_ids)
 
 
 def _diffusion_prefill_cache(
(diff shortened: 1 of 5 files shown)


mlx_vlm/generate/diffusion.py
```diff
@@ -401,6 +455,31 @@ def _diffusion_static_cache_length(
...
+    mm_token_type_ids: Optional[mx.array],
+) -> bool:
+    if prefill_step_size is None or prompt_length <= prefill_step_size:
+        return False
+    if has_padding or use_static_cache or pixel_values is not None:
+        return False
+    # Visual spans can use bidirectional attention inside the whole block. Keep
+    # those prompts on the existing single prefill until chunking can split on
+    # visual-block boundaries without changing attention semantics.
+    return not _diffusion_has_visual_token_types(mm_token_type_ids)
```
Blaizzy: I think you need to rebase with main, we have a chunked prefill policy. 

It's what `_chunked_prefill_enabled` does

Imported from:

```python
from .common import (
    GenerationResult,
    _chunked_prefill_enabled,
    generation_stream,
    wired_limit,
)
```

cc: @lucasnewman

mlx_vlm/generate/diffusion.py
```diff
@@ -4,24 +4,15 @@
...
-    display_width,
-    escape_carriage_returns,
-)
 from ..tokenizer_utils import make_streaming_detokenizer
-from .common import (
-    GenerationResult,
-    _chunked_prefill_enabled,
-    generation_stream,
-    wired_limit,
-)
```
Blaizzy, line 24: Please revert the display stuff and keep only the turbo and chunked prefill changes

mlx_vlm/models/diffusion_gemma/diffusion_turbo_runner.py
```diff
@@ -0,0 +1,246 @@
...
+    x1 = x[..., :half].astype(mx.float32)
+    x2 = x[..., half:].astype(mx.float32)
+    c = cos[None, None, :, :]
+    s = sin[None, None, :, :]
+    out1 = x1 * c - x2 * s
+    out2 = x2 * c + x1 * s
+    return mx.concatenate([out1, out2], axis=-1).astype(x.dtype)
+
+
+class TurboCanvasRunner:
```
Blaizzy, line 64: Is this DiffusionGemma specific or it works on all models?

Blaizzy: Please test with Llada and Nemotron as well

mlx_vlm/generate/diffusion_turbo_runner.py
```diff
@@ -0,0 +1,246 @@
...
+        for layer in self.layers:
+            attn = layer.self_attn
+            self.buf_k.append(
+                mx.zeros((1, attn.n_kv_heads, canvas_length, attn.head_dim), dtype=dtype)
+            )
+            self.buf_v.append(
+                mx.zeros((1, attn.n_kv_heads, canvas_length, attn.head_dim), dtype=dtype)
+            )
+
+    def forward(self, tokens_f: mx.array, positions_f: mx.array, sc_f: Optional[mx.array]):
```
Blaizzy: I would use `__call__` for cohensiveness with the MLX standard

mlx_vlm/generate/dispatch.py
```diff
@@ -264,6 +264,86 @@ def parse_arguments():
         default="entropy-bound",
         help="Canvas update sampler for diffusion generation.",
     )
+    parser.add_argument(
```
Blaizzy: I would remove all new flags because we already have `--gen-kwargs` that can take custom flags. That way we can limit amount of options users see at once.

Then we can document these flags in the readme example section about diffusion turbo :)

mlx_vlm/generate/dispatch.py
```diff
@@ -1584,12 +1665,28 @@ def main():
...
                     continue
                 response += chunk.text
+                last_chunk = chunk
                 if not diffusion_output.handle_text(chunk.text):
                     print(chunk.text, end="")
 
             chat.append({"role": "assistant", "content": response})
             diffusion_output.finish(response)
             print()
+            if last_chunk is not None:
```
Blaizzy: Please revert all the changes to dispatch 👌🏽

We don't need it

mlx_vlm/models/diffusion_gemma/diffusion_turbo_runner.py
Blaizzy: If it's gemma specific I think it's best to move it to diffusion_gemma folder.

mlx_vlm/generate/diffusion_turbo.py
Blaizzy, line 1: @lucasnewman please access if we need this separate runner instead of just adding acouple steps/policy/attributes to diffusion.

Blaizzy (commented):
Super exciting performance gains!

Just left a few nits to align before we land this.

mlx_vlm/generate/dispatch.py
programVeins: Reverted all changes to dispatch!

mlx_vlm/generate/dispatch.py
programVeins: Makes sense! Removed.

mlx_vlm/generate/diffusion_turbo_runner.py
programVeins: Updated!

mlx_vlm/models/diffusion_gemma/diffusion_turbo_runner.py
programVeins: The turbo runner DiffusionGemma specific at the moment - moved to /models/diffusion_gemma. Hoping to test with other models like LLaDA, but don't want it in the scope of this PR yet.

mlx_vlm/generate/diffusion.py
programVeins: Done!

mlx_vlm/generate/diffusion.py
programVeins: Ah! Great - updated diffusion.py to remove all this - and just use the turbo engine.

mlx_vlm/generate/diffusion_turbo.py
programVeins: Hey! Lemme know if we want to reduce this down - or if the current runner is gtg. :)

lucasnewman: @programVeins First, I should say that this work is really cool and I think you've done a lot of good work getting this running as quickly as it is!

The main thing that concerns me is that your change basically duplicates the outer part of the loop for diffusion sampling, which means the implementations will diverge w.r.t. prefill policy and other changes we make along the way.

What do you think about factoring out the canvas denoising from the outer loop, and then you can swap in your turbo implementation and reuse the same setup logic? Here's an illustrative example of what I mean: https://gist.github.com/lucasnewman/ee6d792db0b5f23f8ad343c7a564944c

lucasnewman: I would also be fine handling this as a follow up fwiw, since we don't need to break the api surface area to do this later on.

programVeins: Hey @lucasnewman - thanks for suggesting. I've incorporated this change with [5813fa2](https://github.com/Blaizzy/mlx-vlm/pull/1384/commits/5813fa2f66a27f3e6f6d1fc7a1da5763475bfcce)

Please give it a check and lmk if this looks good! :)

programVeins:
Hey @Blaizzy @lucasnewman lmk how this change looks at the moment - if we're gtg :)

mlx_vlm/generate/diffusion_turbo.py
lucasnewman: I'm good with it since it's behind a flag for now and the code duplication is now pretty minimal. I think we'd eventually want to understand what evals look like compared to the baseline since the top-k-oriented sampling does result in materially different responses. @Blaizzy Are you good with merging this?

lucasnewman:
@programVeins Can you make sure all your commits are verified so we can clear the merge requirements?

Blaizzy:
Thanks @programVeins for the impressive work here — the performance numbers are genuinely cool, and you were responsive throughout the review.
 
After investigation, I've decided to close this PR. The core issue is that the ~1.2K extra lines don't buy us a clean, numerically faithful speedup, and we've already addressed the underlying speed problem in #1359 with a much smaller footprint.
 
The numbers, benchmarked against `google/diffusiongemma-26B-A4B-it`:
 
| Metric | Before (`main` 3ee75ba6) | After (#1359 `2a1db9b0`) | Delta |
|---|---|---|---|
| Generation tok/s (median of 6) | 77.651 | 111.353 | 1.434× / +43.4% |
| Generation tok/s range | 76.822–78.326 | 109.736–111.928 | 1.401×–1.457× |
| Peak memory | 18.696 GB | 17.058 GB | −1.638 GB / −8.8% |
 
On numerical fidelity, I ran a controlled comparison using the same HF prompt, canvas state, and previous-step self-conditioning logits at each step:
 
| Step | Candidate | Commit | PPL | KL to HF | Argmax agree |
|---|---|---:|---:|---:|---:|
| 0 | HF | reference | 83.95 | 0.000 | 256/256 |
| 0 | `pc/fix-diffusion-gemma-4bit-speed` | `46372225` | 97.72 | 0.152 | 236/256 |
| 0 | `diffusiongemma-turbo-pr` | `ba85a396` | 97.72 | 0.152 | 236/256 |
| 1 | HF | reference | 66.93 | 0.000 | 256/256 |
| 1 | `pc/fix-diffusion-gemma-4bit-speed` | `46372225` | 73.64 | 0.096 | 251/256 |
| 1 | `diffusiongemma-turbo-pr` | `ba85a396` | 76.77 | 0.137 | 244/256 |
| 2 | HF | reference | 4.73 | 0.000 | 256/256 |
| 2 | `pc/fix-diffusion-gemma-4bit-speed` | `46372225` | 5.03 | 0.062 | 256/256 |
| 2 | `diffusiongemma-turbo-pr` | `ba85a396` | 5.81 | 0.206 | 256/256 |
 
Both branches are identical at Step 0, but #1359 stays closer to HF at Steps 1 and 2 (lower KL, higher argmax agreement) — so we get most of the speedup *and* better numerical faithfulness, without the added surface area.
 
None of this takes away from the quality of the exploration here — the top-K sampler chain and compacted active-set decoder are clever, and I'd genuinely welcome a future contribution if you want to revisit any of those ideas in a smaller, more isolated form (e.g. validated against LLaDA/Nemotron as discussed). Thanks again 🙏

programVeins:
@Blaizzy Impressive to see the speed bump on https://github.com/Blaizzy/mlx-vlm/pull/1359 without much divergence and lesser lines of code! Thanks for the detailed investigation here - appreciate the feedback :)

PR #1386, opened 2026-06-16
fix(nemotron_labs_diffusion): load MLX-format checkpoints via opt-in sanitize flag

### LINKED ISSUE #1367: ValueError: 605 parameters not in model when loading MLX-quantized Nemotron checkpoint
## Bug

Loading `mlx-community/Nemotron-Labs-Diffusion-3B-4bit` (or any MLX-native quantized 
Nemotron checkpoint) fails immediately with:

ValueError: Received 605 parameters not in model:
diffusion_head.biases, diffusion_head.scales, diffusion_head.weight,
encoder.embed_tokens.biases, encoder.layers.0.mlp.down_proj.biases, ...

## Steps to reproduce

```python
from mlx_vlm import load
model, processor = load("mlx-community/Nemotron-Labs-Diffusion-3B-4bit")
```
##Environment
mlx-vlm version: 0.6.3
Python: 3.13
Hardware: Apple M4

## Cause
This checkpoint has `"format": "mlx"` in its safetensors metadata, so `is_mlx_format=True`
and `sanitize_weights()` is skipped. `Model.sanitize()` is what remaps the checkpoint's 
`encoder.*` keys to `language_model.model.*` — without it, all 605 weight keys are 
unrecognised at `load_weights()`.

luckyayush777:
Fixes #1367.

MLX-format checkpoints carry "format": "mlx" metadata, which skips the top-level sanitize_weights(model, weights) call. Nemotron's sanitize() remaps encoder.* keys onto the parameter tree, so skipping it makes the model fail to load.

Running sanitize unconditionally for MLX checkpoints isn't safe: for models with non-idempotent sanitize() (e.g. sam3, rt_detr_v2), it re-applies layout transforms like conv transposes and corrupts already-converted weights.

This PR adds an opt-in Model.sanitize_mlx_format flag. Nemotron sets it (its remap is pure, idempotent key-renaming); every other model defaults to False, so existing behavior is unchanged and there's no regression.

This is a deliberate band-aid. The real fix (probably) is for sanitize() methods that do layout transforms to guard against double-application (skip if already transformed). That's a larger change touching many models and does increase work when trying to support new models, and it raises a design question I'd like maintainer input on: should that guard be a required convention for new models, or handled some other way? Happy to follow up in a separate PR/issue once there's direction.

### DIFF
--- a/mlx_vlm/models/nemotron_labs_diffusion/nemotron_labs_diffusion.py
+++ b/mlx_vlm/models/nemotron_labs_diffusion/nemotron_labs_diffusion.py
@@ -9,6 +9,13 @@ from .language import LanguageModel
 
 
 class Model(nn.Module):
+    # MLX-format checkpoints ship encoder.* / diffusion_head.* keys that
+    # sanitize() remaps onto the parameter tree. That remap is pure key
+    # renaming (idempotent), so it is safe to run even for format="mlx",
+    # and must run or load_weights() fails with unrecognised parameters.
+    # See https://github.com/Blaizzy/mlx-vlm/issues/1367.
+    sanitize_mlx_format = True
+
     def __init__(self, config: ModelConfig):
         super().__init__()
         self.config = config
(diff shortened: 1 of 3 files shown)


lucasnewman:
Resolved in https://github.com/Blaizzy/mlx-vlm/pull/1498

PR #1387, opened 2026-06-16
Fix #1351: preserve tool-call/channel markers on the diffusion detokenization path

### LINKED ISSUE #1351: Diffusion path strips all special tokens at the token-id level, deleting the tool-call/channel markers its own gemma4 parser needs (tool_calls: null, 'thought' leaks into content)
## Summary

In mlx-vlm **0.6.3**, the diffusion generation path strips **all** special tokens at the token-id level during detokenization. This deletes the `<|tool_call>`/`<tool_call|>` and `<|channel>`/`<channel|>` markers that the server's **own** gemma4 tool parser (`mlx_vlm/tool_parsers/gemma4.py`, auto-selected — `/health` reports `loaded_tool_parser: "gemma4"`) and thinking splitter (`server/app.py` `_split_thinking`, which natively handles `<|channel>thought`) require.

Net effect for `mlx-community/diffusiongemma-26B-A4B-it-4bit`:

1. **Tool calls arrive as inert text content** — `tool_calls: null`, content is the raw payload like `call:get_weather{city:Austin}` with the surrounding markers deleted, so `process_tool_calls` never matches.
2. **The literal word `thought` leaks into content** after tool-result turns — the channel *name* is an ordinary string, so once the `<|channel>` marker around it is stripped, `_split_thinking` can't recognize the block and the prefix `thought\n` lands in `content`.

The server selects a tool parser whose markers its own detokenizer deletes before the parser ever sees them.

## Root cause

- `server/generation.py:1424` — the skip set is built from `tokenizer.all_special_ids` (every `special=True` token, no exclusions).
- `generate/diffusion.py:1053` — that set is passed per-token to `detokenizer.add_token(..., skip_special_token_ids=...)`, so the marker ids are dropped at the **token-id level**, before any decode call.
- `server/generation.py:1470` — the final decode additionally uses `skip_special_tokens=True`.

For DiffusionGemma the six ids the parsers need are `48`/`49` (`<|tool_call>`/`<tool_call|>`), `50`/`51` (`<|tool_response>`/`<tool_response|>`), and `100`/`101` (`<|channel>`/`<channel|>`) — all `special=True` in `tokenizer.json`, all in the skip set.

The autoregressive path doesn't have this problem (cf. #900, where markers leaked *through* to streamed content — the opposite failure; here they're deleted before `process_tool_calls` can fire). A decode-only patch is **not** sufficient — the stripping happens at the id level in `detokenizer.add_token`, before any `decode()` call.

## Reproduction

Serve the model:

```bash
python -m mlx_vlm server --model mlx-community/diffusiongemma-26B-A4B-it-4bit --port 8010
```

Send a standard tools request:

```bash
curl -s http://127.0.0.1:8010/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model": "mlx-community/diffusiongemma-26B-A4B-it-4bit",
  "messages": [{"role": "user", "content": "What is the weather in Austin? Use the tool."}],
  "tools": [{"type": "function", "function": {"name": "get_weather",
    "description": "Get current weather for a city",
    "parameters": {"type": "object", "properties": {"city": {"type": "string"}}, "required": ["city"]}}}]
}'
```

**Observed:**

```json
{"message": {"content": "call:get_weather{city:Austin}", "tool_calls": null}, "finish_reason": "stop"}
```

**Expected** (and what we get with the workaround below applied):

```json
{"message": {"content": "", "tool_calls": [{"function": {"name": "get_weather", "arguments": "{\"city\": \"Austin\"}"}}]}, "finish_reason": "tool_calls"}
```

Same defect in streaming mode (no `delta.tool_calls` ever emitted), and the `thought\n` leak reproduces by following up with a `role: "tool"` result turn — the chat template doesn't emit a pre-closed empty thought channel after `tool` turns, so the model emits `<|channel>thought ...` itself, and only the markers get stripped.

## Suggested fix

When a tool parser is loaded, exclude its marker token ids (and the channel/thinking markers used by `_split_thinking`) from the skip set built at `server/generation.py:1424`, so the diffusion path preserves exactly the special tokens the server's downstream parsers consume. Deriving the ids at runtime via `convert_tokens_to_ids` (rather than hardcoding 48/49/50/51/100/101) keeps it robust to id drift across model revisions.

## Workaround (verified)

We run the server through a launcher that wraps `load_model_resources` and, after model load:

1. **ID-level** (the one that matters): retype the tokenizer into a dynamic subclass whose `all_special_ids` property excludes the six marker ids — the engine's skip set then preserves them.
2. **Text-level**: wrap `tokenizer.decode(skip_special_tokens=True)` to still strip the remaining special tokens textually while keeping the six markers.

```python
import mlx_vlm.server.generation as _gen

KEEP = {"<|tool_call>", "<tool_call|>", "<|tool_response>", "<tool_response|>",
        "<|channel>", "<channel|>"}

def _patch_tokenizer(tokenizer):
    keep_ids = {tid for m in KEEP
                if (tid := tokenizer.convert_tokens_to_ids(m)) is not None and tid >= 0}
    cls = type(tokenizer)
    orig_prop = cls.all_special_ids
    tokenizer.__class__ = type(f"{cls.__name__}Patched", (cls,), {
        "all_special_ids": property(
            lambda self: [i for i in orig_prop.fget(self) if i not in keep_ids])})

    strip = {t for t in tokenizer.all_special_tokens if t not in KEEP}
    orig_decode = tokenizer.decode
    def decode(token_ids, skip_special_tokens=False, **kw):
        text = orig_decode(token_ids, skip_special_tokens=False, **kw)
        if skip_special_tokens:
            for s in strip:
                text = text.replace(s, "")
        return text
    tokenizer.decode = decode

_orig_load = _gen.load_model_resources
def load_model_resources(model_path, adapter_path):
    model, processor, config = _orig_load(model_path, adapter_path)
    _patch_tokenizer(getattr(processor, "tokenizer", processor))
    return model, processor, config
_gen.load_model_resources = load_model_resources

from mlx_vlm.server import main
main()
```

With this applied, non-streaming and streaming tool calls both produce proper `tool_calls` with `finish_reason: "tool_calls"` (the gemma4 parser even converts the model's unquoted `call:` syntax into valid quoted-JSON `arguments`), the `thought` leak is gone, and plain/streaming completions without tools are unchanged.

## Environment

- mlx-vlm 0.6.3, mlx 0.31.2, mlx-lm 0.31.3, transformers 5.11.0, tokenizers 0.22.2
- macOS 26.4.1, Apple Silicon (M5 Max, arm64), Python 3.12
- Model: `mlx-community/diffusiongemma-26B-A4B-it-4bit`

richgoodson:
Fixes #1351.

## Problem

On diffusion models (e.g. `mlx-community/diffusiongemma-26B-A4B-it-4bit`) the server returns `tool_calls: null` with the raw call leaking as text, and the literal channel name `thought` leaks into content.

## Root cause

The gemma4 tool-call markers (`<|tool_call>` / `<tool_call|>`, plus the `<|"|>` arg-string escape) and the channel markers (`<|channel>` / `<channel|>`) are themselves registered special tokens, so on the diffusion path they were deleted before detokenization. `process_tool_calls` gates on `tool_call_start in model_output`, and `_split_thinking` matches `<|channel>…<channel|>` — once the brackets are gone, neither fires, so the call arrives as inert `call:...` text and the channel name `thought` (plain text) leaks.

Verified on the real model tokenizer: those 5 markers are all single special tokens in `all_special_ids`.

## Fix

Narrow the diffusion skip set to `all_special_ids` **minus** the structural delimiters the server's parsers consume — derived from the active tool-parser module (so this works for gemma4 *and* cohere2_moe) plus the thinking/channel markers. Everything else (pad / mask / EOS / BOS sentinels the fixed-length diffusion canvas emits) is still stripped, so this is not a blanket "keep all special tokens".

- Includes an `<unk>` guard: a marker absent from a model's vocab maps to the `<unk>` id, which is itself usually in `all_special_ids`, so it must never be accidentally un-stripped.
- Also threads any request-configured custom thinking start/end tokens (`--thinking-start-token` / `--thinking-end-token`) through, so they're preserved on the diffusion path too.

## Tests

New `mlx_vlm/tests/test_diffusion_special_tokens.py` (mock tokenizer, no model download): narrowed skip-set contract, `<unk>` guard, custom thinking-token preservation, fallbacks, marker survival through `NaiveStreamingDetokenizer`, end-to-end recovery via `process_tool_calls` / `_split_thinking`, and a regression guard reproducing the old blanket-strip behavior.

Full CI test command passes locally on Python 3.10 (1097 passed, 3 skipped).

## Live verification (26B diffusion model)

A/B on the same model + prompts, non-streaming **and** streaming:

- **Without the fix:** `tool_calls: null`, content `thought\n…call:get_weather{city:Austin}` — reproduces the issue exactly.
- **With the fix:** `tool_calls: [get_weather({"city": "Austin"})]`, reasoning split out, content clean, no `thought` leak. Streaming (`stream: true`) emits the tool-call delta with no marker leakage.

### DIFF
--- a/mlx_vlm/server/generation.py
+++ b/mlx_vlm/server/generation.py
@@ -54,6 +54,72 @@ DEFAULT_ENABLE_THINKING = False
 METRICS_HISTORY_LIMIT = 100
 METRICS_RECENT_LIMIT = 32
 
+# Thinking / channel delimiter special tokens the server's thinking splitter
+# anchors on (see ThinkingStreamState._DEFAULT_OPEN_CLOSE_MARKERS and
+# app._count_thinking_tag_tokens). The channel *name* (e.g. "thought") is plain
+# text; only these bracket tokens are registered special tokens.
+_THINKING_MARKER_TOKENS = (
+    "<|channel>",
+    "<channel|>",
+    "<think>",
+    "</think>",
+    "<|START_THINKING|>",
+    "<|END_THINKING|>",
+)
+
+
+def _structural_marker_strings(processor) -> set:
+    """Special-token strings the server's parsers must still see after
+    detokenization: tool-call delimiters (from whichever tool parser this model
+    uses) and thinking/channel delimiters."""
+    markers = set()
+    try:
+        from ..tool_parsers import _infer_tool_parser_from_processor, load_tool_module
+
+        parser_type = _infer_tool_parser_from_processor(processor)
+        if parser_type:
+            module = load_tool_module(parser_type)
+            for attr in ("tool_call_start", "tool_call_end", "_ESCAPE"):
+                value = getattr(module, attr, None)
+                if isinstance(value, str) and value:
+                    markers.add(value)
+    except Exception:
+        logger.debug(
+            "Could not infer tool-call markers for diffusion detok", exc_info=True
+        )
+    markers.update(_THINKING_MARKER_TOKENS)
+    return markers
+
+
+def _diffusion_skip_special_token_ids(tokenizer, processor) -> set:
+    """Special-token ids to drop on the diffusion detokenization path.
+
+    The diffusion canvas is fixed-length and legitimately contains pad/mask/EOS
+    sentinel special tokens, so (unlike the autoregressive path) we strip
+    special tokens by id. But we must NOT strip the tool-call/channel delimiters
+    the server's tool parser and thinking splitter anchor on, or they arrive as
+    inert text (issue #1351). So: all_special_ids minus those marker ids.
+    """
+    all_special = set(getattr(tokenizer, "all_special_ids", None) or [])
+    if not all_special:
+        return all_special
+    convert = getattr(tokenizer, "convert_tokens_to_ids", None)
+    if convert is None:
+        return all_special
+    unk = getattr(tokenizer, "unk_token_id", None)
+    preserve = set()
+    for marker in _structural_marker_strings(processor):
+        try:
+            token_id = convert(marker)
+        except Exception:
+            continue
+        # Only spare genuine single special tokens. Unknown markers map to the
+        # <unk> id on many tokenizers, and <unk> is itself usually in
+        # all_special_ids -- never un-strip it.
+        if isinstance(token_id, int) and token_id in all_special and token_id != unk:
+            preserve.add(token_id)
+    return all_special - preserve
+
 
 class PromptTooLongError(ValueError):
     """Raised when a request exceeds the configured server context budget."""
@@ -1420,8 +1486,8 @@ class ResponseGenerator:
             raw_inputs.get("attention_mask"),
             max_tokens=args.max_tokens,
             temperature=args.temperature,
-            skip_special_token_ids=set(
-                getattr(tokenizer, "all_special_ids", None) or []
+            skip_special_token_ids=_diffusion_skip_special_token_ids(
+                tokenizer, self.processor
             ),
             mm_token_type_ids=raw_inputs.get("mm_token_type_ids"),
         )
--- a/mlx_vlm/tests/test_diffusion_special_tokens.py
+++ b/mlx_vlm/tests/test_diffusion_special_tokens.py
@@ -1,0 +1,144 @@
+"""Regression tests for issue #1351.
+
+The diffusion detokenization path used to strip *all* ``tokenizer.all_special_ids``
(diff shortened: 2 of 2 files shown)


Blaizzy:
Thanks @richgoodson, please open an issue and we'll handle the rest 🚀

PR #1392, opened 2026-06-17
Fix SFT masks for Gemma4 unified

kqw8:
This fixes two SFT masking issues:

1. Gemma4 unified should build its own causal/sliding attention masks during SFT. Passing the trainer's 2D padding attention mask as the model mask bypasses the model's internal causal mask construction.

2. train_on_completions should prefer a completion mask derived from the chat template instead of relying only on a hard-coded assistant token id.

Tests:
- python -m unittest mlx_vlm.tests.test_trainer

### DIFF
--- a/mlx_vlm/lora.py
+++ b/mlx_vlm/lora.py
@@ -187,6 +187,7 @@ def main(args):
             config,
             processor,
             image_resize_shape=args.image_resize_shape,
+            train_on_completions=args.train_on_completions,
         )
 
     # Setup model for training
--- a/mlx_vlm/tests/test_trainer.py
+++ b/mlx_vlm/tests/test_trainer.py
@@ -6,7 +6,12 @@ import mlx.nn as nn
 
 from mlx_vlm.trainer.datasets import VisionDataset
 from mlx_vlm.trainer.lora import LoRaLayer
-from mlx_vlm.trainer.sft_trainer import TrainingArgs, iterate_batches, train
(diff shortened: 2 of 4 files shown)


lucasnewman:
@kqw8 Thanks for the contribution! Overall this looks reasonable. A couple of changes needed before we can merge:

1) We're computing the completion_mask / completion prefixes even when train_on_completion = False, which will do extra work. We should ideally only do the work when we need to.

2) The loss calculation isn't quite right. This looks like a preexisting bug, but we should fix it now anyway if you're trying to train a model this way. The denominator here should be the number of masked elements included in the loss:

```
    ce = (
        nn.losses.cross_entropy(
            logits,
            labels,
            weights=weight_mask,
        )
        * length_mask
    )
    ntoks = length_mask.sum()
    ce = ce.sum() / ntoks
```

Ideally this would be something like:

```python
ce = nn.losses.cross_entropy(logits, labels)

loss_mask = length_mask  # default to the sequence mask

# your adjusted loss mask
if train_on_completions:
    if "completion_mask" in batch:
        loss_mask = loss_mask * batch["completion_mask"][:, 1:]
    else:
        loss_mask = <the assistant mask building you have>

# the denom here should be the masked elements
return (ce * loss_mask).sum() / mx.maximum(loss_mask.sum(), 1)
```

If you can make those changes we can merge this!

lucasnewman (changes requested):
.

lucasnewman (approved):
LGTM!

PR #1397, opened 2026-06-18
Preserve diffusion structural marker tokens

### LINKED ISSUE #1388: Diffusion detok path strips tool-call/channel marker tokens (root cause + fix) — follow-up to #1351
Follow-up to #1351 (and closed PR #1387) with a confirmed root cause, a reproduced A/B, and a proposed fix, per @Blaizzy's request to file an issue.

## Summary

On diffusion models (e.g. `mlx-community/diffusiongemma-26B-A4B-it-4bit`) the server returns `tool_calls: null` with the call leaking as raw text, and the literal channel name `thought` leaks into content.

## Root cause

The gemma4 tool-call markers (`<|tool_call>` / `<tool_call|>`, plus the `<|"|>` arg-string escape) and the channel markers (`<|channel>` / `<channel|>`) are themselves registered special tokens, so on the diffusion path they're deleted before detokenization. `process_tool_calls` gates on `tool_call_start in model_output`, and `_split_thinking` matches `<|channel>…<channel|>` — once the brackets are gone neither fires, so the call arrives as inert `call:...` text and the channel name `thought` (plain text) leaks.

Confirmed on the real tokenizer: all five markers are single special tokens in `all_special_ids` (ids 48/49/52/100/101 on `diffusiongemma-26B-A4B-it-4bit`).

## Reproduction (A/B on the 26B model, same prompt)

Request: chat completion with a `get_weather` tool, "What is the weather in Austin?"

- **Current behavior:** `tool_calls: null`, `content = "thought\n…call:get_weather{city:Austin}"`.
- **Expected:** `tool_calls: [get_weather({"city": "Austin"})]`, reasoning split out, content clean, no `thought` leak.

(Verified both non-streaming and streaming.)

## Proposed fix

Narrow the diffusion skip set to `all_special_ids` **minus** the structural delimiters the server's parsers consume — derived from the active tool-parser module (works for gemma4 and cohere2_moe) plus the thinking/channel markers, with:
- an `<unk>` guard (an unknown marker maps to the `<unk>` id, which is usually itself in `all_special_ids`, so it must not be accidentally un-stripped), and
- pass-through of any request-configured custom thinking start/end tokens.

Everything else (pad/mask/EOS/BOS sentinels the fixed-length diffusion canvas emits) is still stripped — this is **not** a blanket "keep all special tokens."

A working patch + mock-tokenizer unit tests (incl. a regression guard reproducing the old behavior) is on `richgoodson:fix/1351-diffusion-preserve-marker-tokens` (the code from the closed #1387). Full local test run on Python 3.10: 1097 passed / 3 skipped. Happy to adapt it to whatever form fits your process.

Blaizzy:
## Summary
- preserve diffusion detokenizer marker tokens needed by tool-call parsing and thinking/channel splitting
- keep ordinary special sentinels stripped, including unknown-marker encodings
- add regression coverage for tool/channel/custom thinking marker preservation

Fixes #1388

## Verification
- `uv run --with pytest python -m pytest mlx_vlm/tests/test_server.py::TestDiffusionSkipSpecialTokens -q`
- `uv run --with pytest python -m pytest mlx_vlm/tests/test_gemma4_tool_parser.py mlx_vlm/tests/test_cohere_tool_parser.py -q`
- `uv run --with pytest --with psutil python -m pytest -q` (1137 passed, 3 skipped, 41 subtests passed)
- `uv run --with pre-commit pre-commit run --all`

### DIFF
--- a/mlx_vlm/server/generation.py
+++ b/mlx_vlm/server/generation.py
@@ -43,6 +43,7 @@ from ..speculative.utils import (
 )
 from ..structured import ThinkingAwareLogitsProcessor
 from ..tokenizer_utils import _ServerTokenStreamer, make_streaming_detokenizer
+from ..tool_parsers import _infer_tool_parser_from_processor, load_tool_module
 from ..utils import ThinkingBudgetCriteria, load, prepare_inputs
 from .runtime import runtime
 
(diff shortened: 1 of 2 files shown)


Blaizzy:
Closing this as a duplicate of #1394. The earlier automation PR is still open, has the fuller #1388/#1351 patch, and its GitHub Actions test check is green.

PR #1401, opened 2026-06-19
fixes #1400: Remove mx.clear_cache()/gc.collect() from streaming

### LINKED ISSUE #1400: Remove `mx.clear_cache()` / `gc.collect()` from streaming `finally` blocks

**Files:** `mlx_vlm/server/openai.py`
**Lines (v0.6.3 / commit 4a30289):** 1103 (`/responses` endpoint), 1735 (`/chat/completions` endpoint)
**Symptom:** Server process exits silently (SIGSEGV / MLX fatal error) after
processing two concurrent streaming requests. No error is logged. The client
receives HTTP 200 but the connection drops mid-stream.

**Root cause:**
`stream_generator()` has a `finally` block that calls `mx.clear_cache()` and
`gc.collect()` on the asyncio event-loop thread as soon as one stream finishes.
When a second request is being served concurrently, the GPU thread
(`ResponseGenerator._run`) is still dispatching Metal kernels for that request.
`mx.clear_cache()` purges MLX's internal Metal buffer cache while those kernels
are in flight — a data-race on the Metal device heap that causes undefined
behaviour (SIGSEGV or `mx::Exception`), killing the process.

The sequence:

```text
asyncio thread                    │  GPU thread
                                  │
[Request A stream ends]           │  [Request B] Metal kernel still executing
  token_iter_A.close()  ──────► queues _cancel(uid_A), does NOT wait
  mx.clear_cache()  ◄── RACE ──► Metal kernel for B accesses same memory
  → process dies
```

**Fix:** Remove `mx.clear_cache()` and `gc.collect()` from both streaming `finally`
blocks. MLX's allocator manages memory without explicit cache clearing between
requests. The GPU thread already calls `mx.clear_cache()` in its own error handler
when appropriate.

royteeuwen:
Upstream v0.6.3 also added gc.collect() and a debug print alongside the clear_cache call; all three removed from both streaming finally blocks in /responses and /chat/completions.

### DIFF
--- a/mlx_vlm/server/openai.py
+++ b/mlx_vlm/server/openai.py
@@ -1100,9 +1100,6 @@ async def responses_endpoint(request: Request):
                             stream=True,
                             error="stream_closed_before_completion",
                         )
-                    mx.clear_cache()
-                    gc.collect()
-                    print("Stream finished, cleared cache.")
 
             return StreamingResponse(
                 stream_generator(),
@@ -1730,9 +1727,6 @@ async def chat_completions_endpoint(request: ChatRequest, http_request: Request)
                             stream=True,
                             error="stream_closed_before_completion",
                         )
-                    mx.clear_cache()
-                    gc.collect()
-                    print("Stream finished, cleared cache.")
 
             return StreamingResponse(
                 stream_generator(),


royteeuwen:
For me it's not clear if the provided fix is OK, there is an issue, but I want to be certain that there isnt a memory leak

Blaizzy:
Thanks a lot for catching @royteeuwen!

I fixed it here #1400

PR #1403, opened 2026-06-19
fixes #1402: Every first request after server startup causes a full model reload

### LINKED ISSUE #1402: Every first request after server startup causes a full model reload

**Files:** `mlx_vlm/server/app.py`
**Lines (v0.6.3 / commit 4a30289):** 584–586 (`get_cached_model` else-branch)
**Symptom:** Every first request after server startup causes a full model reload
(prints `"New text_generation model request, clearing existing cache..."` and
re-loads the model from disk). The pre-loaded model is discarded and the 45–60 s
startup load time is repeated on each request.

**Root cause:**
The `lifespan` startup hook pre-loads the model by calling
`get_cached_model(model_path, model_kind="text_generation")`, which stores the
cache entry under key `(path, None, "text_generation")`.

When a chat/completions request arrives, `openai.py` calls
`get_cached_model(request.model)` using the default `model_kind="auto"`. In the
`else` branch of `get_cached_model`, `effective_model_kind = model_kind` — so
the lookup key becomes `(path, None, "auto")`, which does **not** match the
pre-loaded `"text_generation"` key → cache miss → reload.

**Fix:** In the `else` branch, normalise `"auto"` → `"text_generation"`:
```python
effective_model_kind = "text_generation" if model_kind == "auto" else model_kind
```

This is safe because the `else` branch is only reached when the model is
confirmed to be neither image-generation, image-edit, nor audio — i.e., it is
always a text model regardless of whether the caller said `"auto"` or
`"text_generation"`.

royteeuwen:


lucasnewman (approved):
Thanks!

PR #1407, opened 2026-06-19
fixes #1406: move draft-kind/draft-block-size env vars outside draft-model guard

### LINKED ISSUE #1406: move draft-kind/draft-block-size env vars outside draft-model guard

Files: mlx_vlm/server/cli.py
Symptom: --draft-kind and --draft-block-size were silently ignored when --draft-model was not provided.  Needed for models with a built-in MTP head that don't require an external draft checkpoint.

royteeuwen:


lucasnewman (approved):
LGTM

PR #1409, opened 2026-06-19
Fix SAM 3.1 interactive point/box segmentation (prompt-insensitive noise)

FAbrahamDev:
## Summary

Interactive **point / box** prompts on **SAM 3.1** (`mlx-community/sam3.1-bf16`) currently return prompt‑insensitive noise — a scattered, whole‑frame speckle that ignores where you click. This PR makes single‑image interactive clicks produce correct, prompt‑localized, single‑object masks. Two independent bugs were responsible.

The text‑detection + memory‑propagation path is untouched.

## Root cause 1 — wrong decoder for interactive prompts

`MultiplexTrackerModel.track_step` encoded the prompt with `interactive_sam_prompt_encoder` but then **decoded with `self.sam_mask_decoder`** — the 16‑object multiplex *propagation* decoder. Its 80 learned multiplex tokens (16 iou + 48 mask + 16 obj) swamp the single sparse prompt token, so the output ignores the click. The checkpoint already ships a populated single‑object `interactive_sam_mask_decoder` (and `interactive_obj_ptr_proj`); the interactive path just never called it.

Fix (`sam3_1/tracker.py`): gate `track_step` on `is_interactive` and, for that case, decode with `interactive_sam_mask_decoder`, project the object pointer from the mask‑token output via `interactive_obj_ptr_proj`, add the learned `interactivity_no_mem_embed` for a cold (memory‑less) prompt, request `multimask_output` and argmax‑select the best of the 3 candidates, and reorder `high_res_features` to `[144, 288]` so the decoder's `conv_s1`/`conv_s0` skips land at the matching upscale stage instead of being silently dropped by the shape guard. `Sam3VideoPredictor._init_object` now sources point/box features from the **interactive** FPN (`need_interactive=True`) rather than the propagation FPN.

## Root cause 2 — high‑res skip added after LayerNorm/GELU

After routing to the correct decoder, masks became prompt‑responsive but were edge‑dominated with uniformly negative IoU. The mask decoders add the `conv_s0`/`conv_s1` high‑res skips **after** `upscale_layer_norm`+GELU; the reference (`facebookresearch/sam2` `mask_decoder.predict_masks`) adds them **before**:

```
upscaled = act1(ln1(dc1(src) + feat_s1))   # stage 1 (144)
upscaled = act2(dc2(upscaled)  + feat_s0)   # stage 2 (288)
```

Injecting the un‑normalized, high‑magnitude edge skip onto an already‑normalized/activated semantic path lets it dominate the hypernetwork dot‑product → whole‑frame speckle. Fix (`sam3_1/sam_components.py` `MultiplexMaskDecoder`, and the sibling `sam3/sam_components.py` `SAMMaskDecoder` for consistency): add each skip *before* the norm/activation. No weight shapes change; the checkpoint loads unchanged.

## Verification (on‑device, Apple Silicon)

Ran `sam3.1-bf16` on an M‑series Mac, single point prompts. Backbone (set‑image equivalent) ≈ 0.8 s once per image; interactive decode **≈ 6 ms warm**.

| click | before | after |
|---|---|---|
| standing figure | scattered whole‑frame speckle | clean full‑body silhouette, centroid Δ≈0.03 |
| bar stool | scattered speckle | clean single‑object mask, centroid Δ≈0.06 |

(Verified via a small harness that calls `track_step` directly and via `Sam3VideoPredictor` — happy to share it.)

## Scope / notes

- **Verified:** SAM 3.1 single‑image interactive **point** prompts.
- The op‑order fix also applies to the shared propagation decoder and to `sam3`'s `SAMMaskDecoder` — same reference‑aligned change; the non‑3.1 paths weren't separately exercised on device.
- **Box** prompts are routed correctly but the box‑corner coordinate normalization in `SAMPromptEncoder._embed_boxes` wasn't separately validated — left as follow‑up; not claimed here.
- Predicted‑IoU magnitudes read low/negative on stylized renders even for clean masks (likely out‑of‑distribution and/or pre‑sigmoid); mask quality and argmax candidate selection are unaffected.

Opened as a **draft** for maintainer review.

### DIFF
--- a/mlx_vlm/models/sam3/generate.py
+++ b/mlx_vlm/models/sam3/generate.py
@@ -708,12 +708,21 @@ class Sam3VideoPredictor:
             points = mx.array(prompt["points"])[None]
             labels = mx.array(prompt["labels"])[None]
 
-            tracker_fpn = self.model.tracker_neck(features)
-            track_features = tracker_fpn[2]
+            # Interactive clicks use the INTERACTIVE FPN, not the propagation FPN
+            # (the interactive decoder + its conv_s0/conv_s1 skips were trained on it).
+            _, inter_fpn, _ = self.model.detector_model.vision_encoder.neck(
+                features,
+                need_det=False,
+                need_interactive=True,
+                need_propagation=False,
+            )
+            track_features = inter_fpn[2]
+            high_res = [inter_fpn[0], inter_fpn[1]] if len(inter_fpn) > 1 else None
 
             result = self.model.tracker_model.track_step(
                 current_features=track_features,
                 prompt_points=(points, labels),
+                high_res_features=high_res,
             )
             mx.eval(result)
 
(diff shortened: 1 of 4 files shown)


lucasnewman:
Closing as abandoned.

PR #1411, opened 2026-06-20
Add reasoning_content to chat messages

### LINKED ISSUE #1408: reasoning field on message objects should be renamed to reasoning_content
The chat completions response object (both streaming deltas and final messages) currently exposes chain-of-thought tokens under a field named reasoning:
```
{ "role": "assistant", "content": "...", "reasoning": "<think>...</think>" }
```
This field name is not recognised by the dominant OpenAI-compatible client ecosystem. The de facto standard, established by DeepSeek-R1 and followed by tools such as the Vercel AI SDK and OpenCode, is reasoning_content. Clients that implement the interleaved reasoning pattern (feeding thinking tokens back into the next-turn context) validate the field name against a strict enum:
field: z.enum(["reasoning_content", "reasoning_details"])
"reasoning" is not in that enum. As a result:
- Streaming display of thinking tokens still works (the AI SDK falls back via delta.reasoning_content ?? delta.reasoning), but
- Multi-turn interleaved threading silently breaks — the client reads message["reasoning_content"] which is undefined, so no thinking context is carried forward.

## Proposed fix:
Rename the primary field to reasoning_content across all response construction sites, and optionally keep reasoning as a deprecated alias for backward compatibility:
{
  "role": "assistant",
  "content": "...",
  "reasoning_content": "<think>...</think>",
  "reasoning": "<think>...</think>"
}

## Affected files:
- mlx_vlm/server/schemas.py — ChatMessage field declaration
- mlx_vlm/server/openai.py — streaming delta construction (~2 sites) and final message construction (~2 sites)

Blaizzy:
## Summary
- emit reasoning_content on chat completion messages and streaming deltas
- keep reasoning as a deprecated compatibility alias
- preserve incoming reasoning_content in chat template messages

Fixes #1408

## Tests
- uv run --with pytest python -m pytest mlx_vlm/tests/test_server.py -k 'reasoning_content or preserves_assistant_reasoning_content or streaming_uses_custom_thinking_markers or streaming_splits_gemma_thinking_channel_content'
- uv run --with pre-commit pre-commit run --all-files
- cd mlx_vlm && uv run --with pytest python -m pytest -s ./tests --ignore=tests/test_smoke.py --ignore=tests/test_utils.py


mlx_vlm/server/openai.py
```diff
@@ -1374,8 +1374,9 @@ async def chat_completions_endpoint(request: ChatRequest, http_request: Request)
                 msg["tool_call_id"] = message.tool_call_id
             if message.name is not None:
                 msg["name"] = message.name
-            if message.reasoning is not None:
-                msg["reasoning"] = message.reasoning
+            if message.reasoning_content is not None:
+                msg["reasoning_content"] = message.reasoning_content
+                msg["reasoning"] = message.reasoning_content
```
lucasnewman, line 1377: I wonder if anyone was really relying on this or if we could drop the backwards compatibility? It could increase the response payload size quite a bit.

Blaizzy: Don't think a lot people were using it 

So we can deprecate it immediately

PR #1417, opened 2026-06-22
Fix gemma4 bidirectional-vision mask crash under continuous batching

rmaher001:
## Problem

With continuous batching enabled, serving a Gemma 4 multimodal model 500s on
image requests with:

```
ValueError: [broadcast_shapes] Shapes (2,1,16,583) and (1,1,583,583) cannot be broadcast
```

(also seen as `(4,1,16,6310)`/`(3,1,6310,6310)`, etc.). It only happens with
image inputs and only when ≥2 requests are coalesced into one batch — the whole
batch fails together.

## Root cause

`Gemma4TextModel._apply_blockwise_bidirectional_overlay`
(`mlx_vlm/models/gemma4/language.py`) builds the "same image block" overlay over
the **full key sequence** — `same_block` is `(B, kv, kv)` → `(B, 1, kv, kv)` —
and ORs it onto `base_mask`:

```python
return base_mask | mx.expand_dims(same_block, 1)
```

Its only shape guard checks the key dim (`mm_token_type_ids.shape[1] != base_mask.shape[-1]`).
Under continuous batching a step can carry a non-square `(B, 1, q, kv)` mask with
`q != kv` (chunked prefill against a KV prefix, or a coalesced decode step). The
boolean OR then broadcast-fails on the query axis (`q` vs `kv`).

Single-request prefill processes the whole prompt at once (`q == kv`), so this
path is never hit outside continuous batching — which is why it went unnoticed.

## Fix

The bidirectional overlay is only meaningful on a **square full-sequence prefill
mask** (`q == kv`). Skip it on non-square step masks and fall back to the causal
base mask:

```python
if base_mask.shape[-2] != base_mask.shape[-1]:
    return base_mask
```

The single-request prefill path (`q == kv`) is unchanged; only the previously
crashing batched non-square steps are affected, and they now degrade gracefully
to causal attention instead of throwing.

## Tests

`mlx_vlm/tests/test_gemma4_bidirectional_mask.py`:
- non-square `(2,1,4,12)` batched mask + image block — reproduces the crash,
  passes with the fix (returns the base mask unchanged);
- square `(1,1,12,12)` prefill — overlay still opens intra-image-block
  bidirectional attention (regression lock);
- end-to-end through `_make_masks` with a cached chunk (`KVCache.offset=8`,
  q=4, kv=12) and vision tokens — the realistic continuous-batching scenario.

The four existing `_make_masks` tests in `test_models.py` still pass.

Note: the end-to-end test constructs the model with `gemma4_unified.TextConfig`
(the unified variant shares this `Gemma4TextModel`), mirroring the existing
`test_models.py` gemma4 mask tests — not a mistake.

## Follow-up

This is the minimal, low-risk fix: on chunked-prefill steps the image tokens
attend causally rather than bidirectionally. A fuller fix would window
`block_sequence_ids` to the query span (`q_blocks` over the last `q` /
`offset:offset+q` slice, `k_blocks` over the full kv) so bidirectional vision
attention is preserved even when prefill is chunked. Happy to extend this PR in
that direction if preferred.

### DIFF
--- a/mlx_vlm/models/gemma4/language.py
+++ b/mlx_vlm/models/gemma4/language.py
@@ -474,6 +474,14 @@ class Gemma4TextModel(nn.Module):
             return base_mask
         if mm_token_type_ids.shape[1] != base_mask.shape[-1]:
             return base_mask
+        if base_mask.shape[-2] != base_mask.shape[-1]:
+            # Continuous batching can hand us a (B, 1, q, kv) step mask with
+            # q != kv (chunked prefill / coalesced decode). The overlay below is
+            # built over the full key sequence, so the boolean OR would
+            # broadcast-fail. The bidirectional vision overlay is only valid on a
+            # square full-sequence prefill mask; fall back to the causal base
+            # mask for non-square steps.
+            return base_mask
 
         block_sequence_ids = self._block_sequence_ids_for_mask(mm_token_type_ids)
         q_blocks = mx.expand_dims(block_sequence_ids, -1)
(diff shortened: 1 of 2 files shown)


Blaizzy:
Thanks @rmaher001 

Please open a issue first :)

Issue #1415, opened 2026-06-21
RFC: make MRoPE rope_deltas request-local under continuous batching


scyyh11:
## Problem

In continuous batching, multimodal MRoPE models may rely on per-sample `rope_deltas`.
Today several model wrappers compute `position_ids, rope_deltas` in `get_input_embeddings()`, but store them on shared language-model state such as:

- `language_model._position_ids`
- `language_model._rope_deltas`

When `mlx_vlm.server` accepts multiple multimodal requests, later requests can overwrite this shared state while an earlier active generation batch is still decoding. The decode batch may then use stale or wrong `rope_deltas`.

For PaddleOCR-VL this can surface as a hard shape error:

```text
ValueError: [broadcast_shapes] Shapes (8) and (2) cannot be broadcast.
```

Other MRoPE models may avoid a crash by broadcasting/tile behavior, but can still use the wrong per-row deltas silently.

## Current affected pattern

I found this during using PaddleOCR-VL. However, the risky pattern exists in multiple MRoPE VL wrappers that precompute `rope_deltas` and cache them on `language_model`.

## Proposed direction

Make `rope_deltas` request-local before entering `BatchGenerator`:

1. After `get_input_embeddings()` returns, capture the just-computed `rope_deltas` into prompt kwargs.
2. Let existing prompt kwargs merge logic concatenate/slice it per row.
3. Let `PromptProcessingBatch.generate()` initialize `GenerationBatch._rope_deltas` from merged prompt kwargs first.
4. Keep `language_model._rope_deltas` only as a compatibility fallback.

This avoids editing each model wrapper immediately, while preserving compatibility with future models that explicitly return `InputEmbeddingsFeatures(rope_deltas=...)`.

## Longer-term cleanup

Eventually MRoPE wrappers should explicitly return `rope_deltas` via `InputEmbeddingsFeatures`, and the fallback capture from shared model state can be removed.

Blaizzy:
Triage note: the shared-state pattern exists around server request embedding and later BatchGenerator admission, and the proposed request-local `rope_deltas` direction lines up with the current prompt-kwargs merge path. I marked this confirmed but also `needs-design` because the fix crosses normal and speculative server batching and should land with a focused regression test rather than a blind server-path patch.

PR #1422, opened 2026-06-23
Fix request-local MRoPE rope_deltas under continuous batching

scyyh11:
## Summary

Fix MRoPE `rope_deltas` handling so continuous batching does not depend on mutable language-model shared state during decode.

- Thread request-local `rope_deltas` through prompt kwargs into normal and speculative decode paths.
- Merge `rope_deltas` by batch row, padding missing text-only rows with zero instead of broadcasting a multimodal row delta.
- Preserve mutable `_rope_deltas` as a compatibility fallback for existing model wrappers.
- Have MRoPE wrappers that already compute `rope_deltas` return it via `InputEmbeddingsFeatures`.

## Problem

MRoPE VL models such as PaddleOCR-VL, Qwen VL, and GLM VL can compute per-sample `rope_deltas` during embedding preparation. Several wrappers stored those values only on `language_model._rope_deltas`.

Under continuous batching, later requests can overwrite that shared state while an earlier active batch is decoding. For PaddleOCR-VL this can surface as broadcast shape failures; for other models it can silently apply the wrong RoPE offset per row.

There is also a mixed-batch case where one multimodal request provides `rope_deltas` and another text-only request does not. The prompt-kwargs merge path must keep row alignment and use zero for missing text rows, not broadcast the multimodal delta across the batch.

## Changes

- Add public `normalize_rope_deltas()` in `mlx_vlm.generate.common`.
- Prefer row-aligned prompt `rope_deltas` in `PromptProcessingBatch.generate()`, `generate_step()` speculative decode, and the server speculative lane.
- Thread target kwargs through `SpeculativeGenerationBatch` and speculative verifier loops.
- Keep speculative fast paths for zero/default `rope_deltas`, but retain non-zero singleton/all-equal deltas.
- Merge missing `rope_deltas` rows as zeros in both cold prefill and APC mixed prefill.
- Return `rope_deltas` from the MRoPE wrappers that already compute it:
  - PaddleOCR-VL
  - Qwen2-VL / Qwen2.5-VL / Qwen3-VL / Qwen3-VL-MoE / Qwen3.5
  - GLM OCR / GLM4V / GLM4V-MoE

## Tests

- `env UV_CACHE_DIR=.uv-cache uv run --with pytest pytest mlx_vlm/tests/test_generate.py -q`
- `env UV_CACHE_DIR=.uv-cache uv run --with pytest pytest mlx_vlm/tests/test_server.py -k speculative -q`
- `env UV_CACHE_DIR=.uv-cache uv run --with pytest pytest mlx_vlm/tests/test_speculative.py -q`
- `env UV_CACHE_DIR=.uv-cache uv run --with pytest pytest mlx_vlm/tests/test_models.py::TestModels::test_mrope_embedding_wrappers_return_request_local_rope_deltas -q`
- `env UV_CACHE_DIR=.uv-cache uv run ruff check --select F401,F821 ...`
- `git diff --check`

Related: #1415


mlx_vlm/generate/ar.py
```diff
@@ -382,6 +383,7 @@ def _step(y, inputs_embeds=None):
                 if k != "inputs_embeds" and v is not None
             }
         )
+        prompt_kwargs_for_speculative_decode = dict(kwargs)
```
Blaizzy: Why make this temp variable name it `prompt_kwargs_for_speculative_decode` to then rename it back `decode_prompt_kwargs`? 

Which is it?

mlx_vlm/generate/ar.py
```diff
@@ -449,6 +451,22 @@ def _step(y, inputs_embeds=None):
...
+    if draft_model is not None:
+        decode_prompt_kwargs = dict(prompt_kwargs_for_speculative_decode)
+        rope_deltas = decode_prompt_kwargs.get("rope_deltas")
+        broadcast_singleton = False
+        if rope_deltas is None:
+            language_model = getattr(model, "language_model", model)
+            rope_deltas = getattr(language_model, "_rope_deltas", None)
+            broadcast_singleton = rope_deltas is not None
+        if rope_deltas is not None:
+            decode_prompt_kwargs["rope_deltas"] = normalize_rope_deltas(
```
Blaizzy: Why do we need to normalize? what's this `broadcast_singleton`?

Blaizzy (commented):
Thanks @scyyh11!

I like the direction, but I will need to make changes to merge this.

mlx_vlm/generate/ar.py
scyyh11: There are two different cases here:
- prompt-kwargs `rope_deltas` are row-aligned, so missing rows must be padded with zero
- the mutable LM-state fallback is legacy compatibility for models that expose a singleton delta for the whole batch

I’ll split/rename this so the call sites express the two cases directly.

mlx_vlm/generate/ar.py
scyyh11: This is meant to be the kwargs passed to the speculative target decode path after prefill.

I prefer renaming it to `target_decode_kwargs` and using that name consistently now.

Does `target_decode_kwargs` read right to you?

scyyh11:
@Blaizzy all done now, PTAL!

Blaizzy:
Hey @scyyh11, thanks a lot for flagging it!

I'm not sure we can proceed here, the change is too big and some parts seems like are unreleated changes.

Could you a reproducible example for the bug for all mrope model and we can design a better solution together or I can add you as co-author.

You can add them here, and once we can easily reproduce it, we'll design the soluton: https://github.com/Blaizzy/mlx-vlm/issues/1415

PR #1423, opened 2026-06-23
Qwen3.5 on CUDA: fix Metal-only crashes + chunked gated-delta prefill (8.5x faster prefill)

Blaizzy:
## Summary

Makes Qwen3.5 (hybrid gated-delta + attention) usable on non-Metal backends (CUDA), and removes the prefill bottleneck. Three commits:

1. **Guard the import-time target-verify Metal kernel.** `_TARGET_VERIFY_GEMV` was built with `mx.fast.metal_kernel(...)` at module import. On CUDA that raises `RuntimeError: [metal_kernel] No Metal back-end.` immediately, so the model couldn't even be imported. Built only when Metal is available, else `None`, with `_use_target_verify_dense` gated on it.

2. **Route ragged left-padded decode attention to the portable fallback on non-Metal.** The continuous-batching decode path builds Metal-only ragged-SDPA kernels, so every request with batch ≥ 2 returned HTTP 500. `_target_verify_left_padded_attention` already has a portable `scaled_dot_product_attention` fallback; the ragged helper now returns `None` on non-Metal so that path is used. Continuous batching works on CUDA (batch 2/4 verified).

3. **Chunked parallel gated-delta prefill.** 24 of 32 layers are gated-delta (linear attention). On backends without the fused Metal kernel, prefill ran as a Python `for t in range(T)` loop over the sequence (`mlx_lm.gated_delta_ops`) — ~T sequential ops per layer, ~49k sequential graph nodes for a 2048-token prompt, dominating prefill. Added `gated_delta_chunked`: parallel intra-chunk matmuls + a chunk-length unit-lower triangular solve + a T/C-step inter-chunk state scan. Used for prefill (T>1, scalar gating); decode (T==1) and vectorized gating keep the reference path.

Validated numerically vs the sequential reference to rel-err < 1e-3 (fp32, well within bf16), including the masked (left-padded) case.

## Performance

RTX PRO 6000 Blackwell (CUDA 13.3), Qwen3.5-4B bf16, greedy, CUDA graphs on.

### Prefill TTFT — before vs after (prefill-bound, 2048-token prompt)

| batch | before | after | speedup |
|---|---|---|---|
| 1 | 2090 ms | 247 ms | 8.5× |
| 2 | 4246 ms | 454 ms | 9.4× |
| 4 | crashed (HTTP 500) | 629 ms | fixed |

Decode is unchanged (141 / 420 tok/s at batch 1 / 4) — the fix only touches the T>1 prefill path.

### vs vLLM 0.20.2 and SGLang 0.5.12 (batch=1, context sweep, after fix)

Same checkpoint, same GPU, caching disabled on all engines, max_tokens=1024.

**Prefill — TTFT p50 (ms), lower is better:**

| context | mlx | vLLM | SGLang | mlx gap |
|---|---|---|---|---|
| 2K  | 248  | 62   | 59   | 4.2× |
| 4K  | 391  | 117  | 116  | 3.4× |
| 8K  | 704  | 251  | 258  | 2.8× |
| 16K | 1366 | 527  | 538  | 2.6× |
| 32K | 2916 | 1170 | 1204 | 2.5× |
| 64K | 6898 | 2853 | 2934 | 2.4× |

**Decode — tok/s (1000 / TPOT p50), higher is better:**

| context | mlx | vLLM | SGLang |
|---|---|---|---|
| 2K  | 146 | 153 | 155 |
| 4K  | 145 | 152 | 154 |
| 8K  | 141 | 150 | 151 |
| 16K | 133 | 147 | 147 |
| 32K | 121 | 140 | 140 |
| 64K | 102 | 129 | 128 |

Before these changes, prefill was ~34× behind vLLM/SGLang (and crashed at batch ≥ 2); after, it scales with the same shape and is ~2.4–4.2× behind, with the gap narrowing as context grows. Decode is within ~5% at short context and ~20% behind at 64K (the per-step full-attention over a large KV cache; the gated-delta layers keep constant-size state so they don't degrade). The remaining gap is general CUDA kernel efficiency (FlashAttention / fused gated-delta kernels), not algorithmic.

benchmark_all

## Notes

- `g` is clipped off 0 before `log` (`compute_g` can underflow to exactly 0.0 → `log(0) = -inf`).
- The triangular inverse uses forward substitution rather than repeated-squaring Neumann: the latter overflows when keys are correlated (`KK` off-diagonals ≈ 1), which is the real-model regime.
- The chunked prefill output is scheduled with `mx.async_eval` so this large/dynamic subgraph isn't folded into the server's captured CUDA graph (that capture otherwise hits a backend `cudaGraphAddDependencies` error). Decode never calls this path, so its CUDA graphs are untouched.

All changes are non-Metal-only fallbacks; macOS/Metal behavior is unchanged (Metal kernels still used when available).

### DIFF
--- a/mlx_vlm/models/qwen3_5/gated_delta.py
+++ b/mlx_vlm/models/qwen3_5/gated_delta.py
@@ -1,4 +1,3 @@
-import math
 from functools import partial
 from typing import Optional
 
@@ -18,17 +17,13 @@ def _compute_g_beta(A_log, a, b, dt_bias):
 
 
 def _invert_unit_lower(Tmat: mx.array, C: int) -> mx.array:
-    """Inverse of a batched unit lower-triangular matrix ``Tmat = I + N`` with
-    ``N`` strictly-lower (hence nilpotent). Uses the log-depth identity
+    """Inverse of a batched unit lower-triangular matrix ``Tmat`` (``[..., C, C]``,
+    unit diagonal) via forward substitution, batched over all chunks/heads.
 
-        (I + N)^-1 = (I - N) * prod_{k>=1} (I + N^(2^k))
-
-    which needs only ~log2(C) batched matmuls (vs a C-step forward
-    substitution). Fewer/shallower graph nodes — important so CUDA-graph
-    capture in the batched server path doesn't choke."""
-    # Numerically-stable forward substitution (the log-depth repeated-squaring
-    # Neumann inverse overflows when N has O(1) entries, which happens with
-    # correlated keys). Sequential in C but batched over all chunks/heads.
+    Sequential in C (small, fixed) but numerically stable. A log-depth
+    repeated-squaring (Neumann) inverse is faster but overflows when N has
+    O(1) entries — exactly the regime here, since correlated keys give KK
+    off-diagonals ~1."""
     X = mx.broadcast_to(mx.eye(C, dtype=mx.float32), Tmat.shape) + mx.zeros_like(Tmat)
     rows = [X[..., 0:1, :]]
     for i in range(1, C):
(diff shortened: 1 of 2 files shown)


lucasnewman (approved):
LGTM

PR #1426, opened 2026-06-23
[codex] Fix Gemma 4 variable image soft tokens

### LINKED ISSUE #1425: Gemma 4: no supported way to set vision_soft_tokens_per_image; non-default budget → empty vision features in masked_scatter
## Summary

There appears to be **no supported way to set the Gemma 4 vision soft-token budget** (`vision_soft_tokens_per_image` / the image processor's `max_soft_tokens`), and every best-effort attempt to set it to a non-default value crashes. The default (280) is the only value that works end-to-end, which is too low for dense document/slide OCR.

The model config and image processor clearly anticipate variable resolution — `Gemma4ImageProcessor` accepts `max_soft_tokens` (presets `(70, 140, 280, 560, 1120)`) and dynamically returns `num_soft_tokens_per_image`, and the vision config carries `position_embedding_size: 10240` (enough for the 1120 preset). But nothing wires it through, and forcing it hits two stacked failures.

## Environment

- mlx-vlm **0.6.3**, mlx 0.31.2, mlx-lm 0.31.3
- macOS 26.1, Apple Silicon (arm64), Python 3.12.13
- Model: `lmstudio-community/gemma-4-26B-A4B-it-QAT-MLX-4bit` (also reproduces conceptually for other Gemma 4 checkpoints; the vision path is shared)

## What I want

Transcribe a text-dense image at higher resolution by raising `vision_soft_tokens_per_image` (e.g. to 560 or 1120). At the default 280, fine print is missed. The README documents no API for this (no `generate()` kwarg, `load()` param, processor setting, or CLI flag), so I tried to set it directly.

## Reproduction

```python
from mlx_vlm import load, generate, apply_chat_template

model, processor = load("lmstudio-community/gemma-4-26B-A4B-it-QAT-MLX-4bit")
config = load_config("lmstudio-community/gemma-4-26B-A4B-it-QAT-MLX-4bit")

processor.image_processor.max_soft_tokens = 1120          # raise the budget
prompt = apply_chat_template(processor, config, "Transcribe all visible text.", num_images=1)
out = generate(model, processor, prompt, image=["slide.png"], max_tokens=64, verbose=False)
print(out.text)
```

### Failure 1 — `max_patches` desync (broadcast)

Setting only `image_processor.max_soft_tokens` produces:

```
ValueError: [broadcast_shapes] Shapes (1,9900,1152) and (1,2520,1152) cannot be broadcast.
```

Root cause: `VisionModel.max_patches` is fixed at construction from `default_output_length` (280), independent of the processor's `max_soft_tokens`:

```python
# mlx_vlm/models/gemma4/vision.py
self.max_patches = self.default_output_length * self.pooling_kernel_size**2   # 280 * 3**2 = 2520
```

So the processor emits ~9900 patches (1120 budget) but the tower pads/expects 2520.

### Failure 2 — empty vision features in `masked_scatter`

Working around Failure 1 by also syncing the tower (the position-embedding table is `10240`, so ≥1120 budgets fit):

```python
vt = model.vision_tower
vt.default_output_length = 1120
vt.max_patches = 1120 * vt.pooling_kernel_size**2   # 10080  (< 10240 pos-emb cap)
processor.image_processor.max_soft_tokens = 1120
```

removes the broadcast error but then fails deeper, with **empty** vision features:

```
File "mlx_vlm/generate/ar.py", line 372, in generate_step
    embedding_output = model.get_input_embeddings(...)
File "mlx_vlm/models/gemma4/gemma4.py", line 129, in get_input_embeddings
    inputs_embeds = _scatter(...)
File "mlx_vlm/models/gemma4/gemma4.py", line 124, in _scatter
    return masked_scatter(inputs_embeds, mask_expanded, features)
File "mlx_vlm/models/gemma4/gemma4.py", line 16, in masked_scatter
    aligned = source.flatten()[indices % source.size]
ValueError: [take] Cannot do a non-empty take from an empty axis.
```

i.e. `features` (the projected vision soft tokens) is zero-size when the budget differs from the default. The same image at the default budget (280) works end-to-end.

## Questions / ask

1. **Is there a supported way to set `vision_soft_tokens_per_image` for Gemma 4** (per-call, per-load, or processor-level) that I missed? If so, documenting it would resolve this.
2. If not, could the variable-budget path be wired so that setting `max_soft_tokens` (a) syncs `VisionModel.max_patches` / `default_output_length`, and (b) actually produces non-empty features into `get_input_embeddings` / `masked_scatter`? The image-processor side already returns dynamic `num_soft_tokens_per_image` and the placeholder expansion in `Gemma4Processor.__call__` already honors it — the gap looks like the vision-tower `max_patches` coupling plus the feature scatter at non-default budgets.

Happy to test patches against this checkpoint.

chiptoe-svg:
## Summary

Fixes Gemma 4 image runs that use a non-default soft-token budget such as `1120`.

- add `vision_soft_tokens_per_image` as a processor/generate alias for image `max_soft_tokens`
- derive the Gemma 4 vision tower output length from the processed image patch grid instead of the config default
- require image/video feature counts to match prompt image/video token slots before scatter, while preserving the existing audio truncation path
- document the CLI/Python usage and add regression coverage for the issue

## Root Cause

The image processor could resize and prompt-expand for a non-default soft-token budget, but the vision tower still padded/truncated around `default_output_length`. That made the tower produce the default number of pooled features even when the prompt contained more image token slots. A manual workaround that changed the tower default without changing the pooler default could then invert the pool mask and produce empty visual features.

## Validation

- `.venv/bin/python -m black --check --target-version py312 mlx_vlm/generate/dispatch.py mlx_vlm/models/gemma4/gemma4.py mlx_vlm/models/gemma4/processing_gemma4.py mlx_vlm/models/gemma4/vision.py mlx_vlm/tests/test_gemma4_variable_soft_tokens.py`
- `.venv/bin/python -m isort --check-only --profile=black mlx_vlm/generate/dispatch.py mlx_vlm/models/gemma4/gemma4.py mlx_vlm/models/gemma4/processing_gemma4.py mlx_vlm/models/gemma4/vision.py mlx_vlm/tests/test_gemma4_variable_soft_tokens.py`
- `git diff --check`
- `.venv/bin/python -m pytest mlx_vlm/tests -q`

Fixes #1425.

### DIFF
--- a/mlx_vlm/generate/dispatch.py
+++ b/mlx_vlm/generate/dispatch.py
@@ -165,6 +165,13 @@ def parse_arguments():
         help="Resize shape for the image.",
     )
     parser.add_argument(
+        "--vision-soft-tokens-per-image",
+        type=int,
+        choices=(70, 140, 280, 560, 1120),
+        default=None,
+        help="Gemma 4 image soft-token budget per image.",
+    )
+    parser.add_argument(
         "--prompt",
         type=str,
         nargs="+",
@@ -1547,6 +1554,13 @@ def main():
             }
             if args.resize_shape is not None:
                 stream_kwargs["resize_shape"] = args.resize_shape
+            vision_soft_tokens_per_image = getattr(
+                args, "vision_soft_tokens_per_image", None
+            )
+            if vision_soft_tokens_per_image is not None:
+                stream_kwargs["vision_soft_tokens_per_image"] = (
+                    vision_soft_tokens_per_image
+                )
             if args.prefill_step_size is not None:
                 stream_kwargs["prefill_step_size"] = args.prefill_step_size
             if is_masked_text_diffusion:
@@ -1617,6 +1631,11 @@ def main():
         }
         if args.resize_shape is not None:
             gen_kwargs["resize_shape"] = args.resize_shape
+        vision_soft_tokens_per_image = getattr(
+            args, "vision_soft_tokens_per_image", None
+        )
+        if vision_soft_tokens_per_image is not None:
+            gen_kwargs["vision_soft_tokens_per_image"] = vision_soft_tokens_per_image
         if args.prefill_step_size is not None:
             gen_kwargs["prefill_step_size"] = args.prefill_step_size
         if is_masked_diffusion_text_model(model):
--- a/mlx_vlm/models/gemma4/README.md
+++ b/mlx_vlm/models/gemma4/README.md
@@ -98,6 +98,17 @@ python -m mlx_vlm.generate \
   --temperature 1.0 --top-p 0.95 --top-k 64
 ```
 
+For text-dense images, increase the image soft-token budget:
+
+```sh
+python -m mlx_vlm.generate \
+  --model google/gemma-4-e4b-it \
+  --image path/to/slide.png \
+  --prompt "Transcribe all visible text." \
+  --vision-soft-tokens-per-image 1120 \
+  --max-tokens 500
+```
+
 ### Audio understanding (2B/4B only)
 
 ```sh
@@ -189,6 +200,7 @@ result = generate(
     processor=processor,
     prompt=prompt,
     image=image,
+    vision_soft_tokens_per_image=1120,
     max_tokens=500,
     temperature=1.0,
     top_p=0.95,
--- a/mlx_vlm/models/gemma4/gemma4.py
+++ b/mlx_vlm/models/gemma4/gemma4.py
@@ -10,10 +10,21 @@ from .language import LanguageModel, RMSNormNoScale
 from .vision import VisionModel
 
 
-def masked_scatter(input_tensor, mask, source):
+def masked_scatter(input_tensor, mask, source, allow_truncate=False):
     mask_flat = mask.flatten().astype(mx.int32)
+    n_mask = int(mask_flat.sum().item())
+    if source.size < n_mask or (source.size != n_mask and not allow_truncate):
+        hidden_size = input_tensor.shape[-1]
+        n_tokens = n_mask // hidden_size
+        n_features = source.size // hidden_size
+        raise ValueError(
+            "Multimodal features and tokens do not match: "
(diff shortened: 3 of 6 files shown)


chiptoe-svg:
Validated this branch on real hardware with actual models, beyond the unit tests — Apple Silicon (M-series, 64 GB), `gemma-4-26B-A4B-it-QAT-MLX-4bit` and `gemma-4-12B-it-qat-4bit`, on text-dense slides. The variable-soft-token path works end-to-end:

- Budgets **560** and **1120** — which previously hit the `broadcast_shapes` desync / empty-features `masked_scatter` (#1425) — now run cleanly.
- The image-token count **scales with the requested budget** (`prompt_tokens` 306 / 569 / 1142 for 280 / 560 / 1120) instead of being pinned at the default — the tower output length now tracks the processed patch grid, as intended.
- **Fine print resolves at higher budgets:** packaging text garbled at 280 (e.g. "Doritos / TANGY CHEESE / CHILI HEATWAVE / 200g") is read correctly at 560.

One orthogonal note for anyone testing on dense images: under pure greedy decoding gemma-4 tends to fall into a repetition loop on list-heavy slides (a token repeated 500+ times) — this is independent of this change (it reproduces at the default budget, and on the easy/non-dense slide it doesn't occur at all) and is fully eliminated by a small `repetition_penalty` (~1.15–1.3), which leaves the now-correct fine print intact and returns in ~2.5 s. Probably worth a one-line note in the README example to pair high soft-token budgets with a repetition penalty for dense documents.

Blaizzy:
Thanks @chiptoe-svg for catching this and the detailed write-up!

That said, I don't think wiring it as presented is the right call, it adds a lot of custom code around a single model arch. 

Let us take it from here :)

chiptoe-svg:
understood - was just desperately trying to get gemma 4 models to run with vision on this infrastructure. I look forward to seeing it done cleanly! thanks

PR #1427, opened 2026-06-23
Add Unlimited-OCR model support

shuuul:
## Summary

- add `baidu/Unlimited-OCR` support with a dedicated `unlimitedocr` model package
- implement the Unlimited-OCR processor, config loading, prompt remapping, and single-`<image>` multi-page prompt semantics used by upstream `infer_multi`
- add the Unlimited-OCR DeepSeek-V2 language wrapper with ring sliding-window attention (R-SWA) KV cache support
- add model docs, README index entry, processor patch registration, and targeted tests

## Verified

- loaded the real `baidu/Unlimited-OCR` Hugging Face checkpoint through `mlx_vlm.generate`
- rendered a local 44-page arXiv PDF to ordered page images and verified multi-page OCR starts correctly with paper title/author text
- verified the long multi-page path with `cropping=false, image_size=1024`, matching upstream's global-view multi-page workflow
- checked that the prompt template inserts one `<image>` token for multiple pages, matching upstream `infer_multi`

## Testing

- `uv run pytest -q mlx_vlm/tests/test_models.py -k unlimitedocr mlx_vlm/tests/test_processors.py -k UnlimitedOCR`
- `uv run python -m compileall -q mlx_vlm/models/unlimitedocr mlx_vlm/prompt_utils.py`
- `git diff --check`

## Notes

- The CLI still expects image files; PDF workflows should render pages to images before passing them in order via `--image`.
- R-SWA preserves the full prefill cache and bounds generated-token KV growth to `prefill + sliding_window_size`, matching Unlimited-OCR's decode-time ring cache behavior.


mlx_vlm/prompt_utils.py
Blaizzy, line 1: Please revert all changes and just use `image_token_newline` :)

Also for the model type, it's best to stick with the one of the config.json, which one is it?
- unlimited-ocr or unlimitedocr

mlx_vlm/utils.py
```diff
@@ -46,6 +46,7 @@
     "falcon-perception": "falcon_perception",
     "nemotronh_nano_omni_reasoning_v3": "nemotron_h_nano_omni",
     "cohere2moe": "cohere2_moe",
+    "unlimited-ocr": "unlimitedocr",
```
Blaizzy: I would prefer unlimited_ocr

Blaizzy (commented):
Overall LGTM, just a couple nits

mlx_vlm/utils.py
shuuul: changed to unlimited_ocr now

jason-ni:
I have trouble to run this PR in my local M4 pro macbook. Firstly it has tokenizer issue. It's fixed by AI. I'm not sure is it only for me. Therotically, every one should have hit this.  Fix pushed to: https://github.com/jason-ni/mlx-vlm/tree/unlimited

Then I found it failed to recognize a paper page that I used to test for document OCR models. I put it in the `examples/images/paper_test.png`.  The 4bit model have more hallucination and the 8bit model is better but still failed at last section, failing into repeat loop. I tested this image at https://huggingface.co/spaces/baidu/Unlimited-OCR it works well.

It seems the weight precision is more sensitive than other VLM models. Or maybe the impl has some numeric divergence some place.

Blaizzy:
Could you share a reproducible example?

jason-ni:
> Could you share a reproducible example?

In my forked branch, I pushed a `run_unlimited_ocr.py`.  And I use GLM 5.2 fixed a tokenizer issue and image preprocessor logic. But the critical finding is we must use bf16 for encoding. LLM decoding with Q8 is fine. Q4 may have small issue. So I tried to make a hybrid weight file with the `convert_hybrid.py`.

HuXioAn:
Hi, I had the same tokenizer issue when I was trying to enable the multi-page function of the model. @jason-ni fixed it in his fork by using the right tokenizer:

   The trigger words: "Multi page parsing."
    LlamaTokenizerFast      -> 'Mult','ip','age','par','sing','.'   # "Multipageparsing."
    PreTrainedTokenizerFast -> 'Multi',' page',' parsing','.'       # correct

  Fix in processing_unlimitedocr.py:
  \- tokenizer = LlamaTokenizerFast.from_pretrained(path)
  \+ tokenizer = PreTrainedTokenizerFast.from_pretrained(path)

However, Multi-page has to go through the `unlimited_ocr` model/processor (single <image> for all pages + single_prompt_multi_image placement). Routing it through the `deepseekocr` processor/model instead hits a multi-image placement error ((273,1280) vs (819,1280) broadcast) — the `unlimited_ocr` path in this PR handles it correctly.

 So I think only the tokenizer change is needed on the PR side.

I tried a MXFP8 model `sahilchachra/unlimited-ocr-mxfp8-mlx` which keeps the bf16 vision encoder. It worked well on simpel multi-page case, with no looping issue. But the `tokenizer.json` file that ships with this model has to be replaced by the official one in `baidu/Unlimited-OCR`, and change the `model_type` to `unlimited-ocr` instead of `deepseekocr`.

shuuul:
Hi @jason-ni @HuXioAn and @Blaizzy ，
Thanks for pointing out the bugs and sharing the feedback!

I have investigated the issues on my end (testing on an M2 Pro machine) and pushed updates to address them:

- Tokenizer Issue: Fixed. I have switched the implementation to use PreTrainedTokenizerFast as suggested.
- Repetition & Quantization Behavior: Fixed. I aligned the implementation with the upstream sliding-window no-repeat n-gram logits processor. The 4-bit quantized model should now work correctly without falling into infinite loops.

I also prepared a test script here for verification:
[run_unlimited_ocr_quantized_pdf.sh](https://github.com/user-attachments/files/29441098/run_unlimited_ocr_quantized_pdf.sh)

Please let me know if this works on your cases!

Blaizzy:
Thanks @shuuul 

Unfortunately that's not the issue. 

Please revert the ngram processor changes.

shuuul:
@Blaizzy The adjustments to the ngram processor were specifically introduced to handle issues arising from quantization. While the infinite loop problem does occur in bf16 as well, it is significantly worse in the 4-bit version.

jason-ni:
> @Blaizzy The adjustments to the ngram processor were specifically introduced to handle issues arising from quantization. While the infinite loop problem does occur in bf16 as well, it is significantly worse in the 4-bit version.

The ngram processor is required to avoid LLM sampler run into loop until max context size reached. It detects repeat pattern and stop earlier. But it won't help to correct the sampling content. It's helpfull in practice that let it fail earlier, without wasting time to run to end of context limit. In llama.cpp we already have this for long time.

I also reported same issue to llama.cpp unlimited-ocr PR at the same day. The author of that PR found the point. We have to use F32 for vision encoding. I tried that in my branch. After force vision encoding using F32 converted weight tensors, even the 4bit model can work without falling in loop, only with small hallucination. It's kind of surprising, as in other VLM model BF16/FP16 is good enough. And with convertion to F32, we need more fine grain memory management. That is release mlx cache memory inside encoding layer loop. Otherwise it will OOM in my 24GB macbook. 

I'm not going to push my branch for this, because the AI written code is kind of messy. But the final result is quite good, <20s for the page in my issue. So I suggest you guys have a verification of this. It's fun.

HuXioAn:
A bit of upstream context that might help: Baidu's reference code ships `SlidingWindowNoRepeatNgramProcessor`, but infer/infer_multi default it to off (no_repeat_ngram_size=0) and only the examples enable it (size 35, window 128 single / 1024 multi-page)passed through HF's logits_processor. So even upstream treats it as an opt-in, caller-supplied processor rather than always-on behavior.

mlx-vlm already exposes logits_processors, so this works without touching the generic dispatch: with just the tokenizer fix and passing the n-gram processor at call time, I get clean, loop-free output on three real 11–15 page papers (the key knob for multi-page is window=1024).

Suggestion: keep this PR scoped to model correctness (tokenizer + single-<image> multi-page prompt), drop the auto-injection / model-type special-casing in `dispatch.py/sample_utils.py`, and instead document the recommended `logits_processors` snippet in the model README, or in a demo script.

shuuul:
Hi @Blaizzy @jason-ni @HuXioAn ,
Thanks for the discussion. I just reverted the ngram processor.

Blaizzy (approved):
LGTM!

Blaizzy:
Thanks guys!

shuuul:
@Blaizzy I just fixed the format.

Blaizzy:
Thanks @shuuul!

PR #1434, opened 2026-06-25
adding hisa to dsa based models(currently only deepseek)

Lazarus-931:
<html><head></head><body><p>Implementing HISA (Hierarchical Indexing for Sparse Attention) for DSA models, currently only for DeepSeek.</p>
<p>HISA promises 2–8× improvements, and I believe this holds at ~685B parameters (fp8) or ~700B, as the paper tested with GLM-5.</p>
<p>For verification, I used small randomly initialized weights through the DeepSeek path (~7GB). Improvements still showed — up to 1.40× at 128K context — and I expect the full gains to materialize at the parameter scales the paper targets.</p>
<p>Results:</p>

context | flat ms/tok | HISA ms/tok | speedup | peak GB
-- | -- | -- | -- | --
16k | 8.50 | 8.88 | 0.96× | 2.6
32k | 9.68 | 9.34 | 1.04× | 3.2
65k | 12.18 | 10.18 | 1.20× | 4.4
128k | 17.03 | 11.98 | 1.42× | 6.7

This also includes tests for accuracy against vanilla indexer 

<p>We do get 0.96× for smaller context windows, but I think it's fine considering the likelihood of that usage, and this is an overall net positive change.</p>
<p>Paper: https://arxiv.org/abs/2603.28458</p>
<p>To verify against my fork:</p>
<pre><code class="language-bash">#!/usr/bin/env bash
# HISA perf: upstream main (flat) vs fork (HISA), random-init small deepseek_v4,
# swept over context. Apple Silicon &gt;=16GB, needs uv + git.
set -euo pipefail
W=$(mktemp -d)
git clone -q --depth 1 https://github.com/Blaizzy/mlx-vlm     "$W/main"
git clone -q --depth 1 https://github.com/Lazarus-931/mlx-vlm "$W/fork"
cat &gt; "$W/b.py" &lt;&lt;'PY'
import time, numpy as np, mlx.core as mx
from mlx_vlm.models.deepseek_v4 import Model, ModelConfig
N = 12
cfg = ModelConfig(vocab_size=2000, hidden_size=1024, intermediate_size=1024,
    moe_intermediate_size=512, num_hidden_layers=N, num_attention_heads=16,
    num_key_value_heads=1, n_routed_experts=16, num_experts_per_tok=2,
    q_lora_rank=256, head_dim=128, o_groups=4, o_lora_rank=256,
    index_n_heads=64, index_head_dim=128, index_topk=512,
    compress_ratios=[0] + [4] * (N - 2) + [0], max_position_embeddings=262144)
m = Model(cfg); m.eval(); mx.eval(m.parameters())
def ms(T):
    c = m.make_cache(); p = 0
    while p &lt; T:
        n = min(256, T - p); mx.eval(m(mx.array(np.random.randint(0, 2000, (1, n))), cache=c).logits); p += n
    for _ in range(4): mx.eval(m(mx.array([[0]]), cache=c).logits)
    t = time.perf_counter()
    for _ in range(20): mx.eval(m(mx.array([[0]]), cache=c).logits)
    return (time.perf_counter() - t) / 20 * 1e3
for T in (16384, 32768, 65536):
    print(T, round(ms(T), 2), flush=True)
PY
run() { uv venv -q "$1/.venv"; uv pip install -q -e "$1" --python "$1/.venv/bin/python"; "$1/.venv/bin/python" "$W/b.py"; }
run "$W/main" &gt; "$W/m.txt"
run "$W/fork" &gt; "$W/f.txt"
echo; printf "| context | main flat | fork HISA | speedup |\n|--:|--:|--:|--:|\n"
join "$W/m.txt" "$W/f.txt" | awk '{printf "| %d | %.2f ms | %.2f ms | %.2fx |\n", $1, $2, $3, $2/$3}'
</code></pre></body></html>

### DIFF
--- a/mlx_vlm/models/deepseek_v4/language.py
+++ b/mlx_vlm/models/deepseek_v4/language.py
@@ -17,6 +17,7 @@ from ..base import (
 )
 from ..cache import CacheList, PoolingCache, RotatingKVCache
 from .config import ModelConfig
+from .hisa_kernel import hisa_select
 from .hyper_connection import HyperConnection, HyperHead, hc_expand
 
 
@@ -568,6 +569,32 @@ class Indexer(nn.Module):
             and self.index_keep * self.index_block >= k
         ):
             return self._hisa_select(q, pooled, x, k)
+
+        # HISA batched path for L > 1 (prefill / speculative decode). Honors the
+        # causal mask via valid_len = #visible pooled positions per query (the
+        # pool mask is contiguous-causal, so the count is the visibility cutoff).
+        if (
+            L > 1
+            and self.index_block > 0
+            and Np >= self.index_block * self.index_keep
+            and self.index_keep * self.index_block >= k
+        ):
+            if pmask is None:
+                valid_len = mx.full((B, L), Np, dtype=mx.int32)
+            else:
+                pm = pmask if pmask.ndim == 3 else pmask[None]
+                valid_len = mx.broadcast_to(pm, (B, L, pm.shape[-1])).sum(-1)
+            weights = self.weights_proj(x).astype(mx.float32) * (self.n_heads**-0.5)
+            return hisa_select(
+                q,
+                pooled,
+                weights,
+                self.scale,
+                k,
+                self.index_block,
+                self.index_keep,
+                valid_len,
+            )
 
         scores = q.astype(mx.float32) @ pooled[:, None].swapaxes(-1, -2).astype(
             mx.float32
--- a/mlx_vlm/models/deepseek_v4/language.py
+++ b/mlx_vlm/models/deepseek_v4/language.py
@@ -492,6 +492,47 @@ class Indexer(nn.Module):
         self.weights_proj = nn.Linear(config.hidden_size, self.n_heads, bias=False)
         self.compressor = Compressor(config, compress_ratio, self.head_dim)
         self.scale = self.head_dim**-0.5
+        self.index_block = getattr(config, "index_block", 0)
+        self.index_keep = getattr(config, "index_keep", 0)
+
+    def _hisa_select(self, q: mx.array, pooled: mx.array, x: mx.array, k: int):
+        """HISA hierarchical decode selection: coarse block filter -> fine top-k.
+
+        q: (B, n_heads, 1, head_dim); pooled: (B, Np, head_dim). Returns (B, 1, k)
+        indices into the prefix, matching the flat path's output. Scans n_blocks
+        coarse reps + index_keep*index_block fine candidates instead of all Np.
+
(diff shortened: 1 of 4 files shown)


mlx_vlm/turboquant.py
Blaizzy, line 1: I think you need to rebase this branch because we merged this

mlx_vlm/models/deepseek_v4/language.py
```diff
@@ -510,20 +545,36 @@ def __call__(
         q = q.transpose(0, 2, 1, 3)
         q = position_rope(q, offset)
 
+        Np = pooled.shape[1]
+        k = min(self.index_topk, Np)
+        pmask = pool_cache.make_mask(L, offset) if pool_cache is not None else None
+
+        # HISA hierarchical decode fast-path happends only when there is no mask to honor
```
Blaizzy, line 559: Can we make it work with batching ?

mlx_vlm/tests/test_deepseek_v4_hisa.py
Blaizzy, line 1: I would move this to test_models 👌🏽

We will refactor and consolidate tests soon

Blaizzy (commented):
LGTM, just left a couple nits and a question 🚀

mlx_vlm/models/deepseek_v4/language.py
Lazarus-931: yeah, will keep this open and work on it

mlx_vlm/tests/test_deepseek_v4_hisa.py
Lazarus-931: k, will fold the test into tests/test_models

mlx_vlm/models/deepseek_v4/language.py
Blaizzy: Great, let me know when it’s ready for final review 👌🏽

Lazarus-931:
cc: @Blaizzy

Lazarus-931:
added fix for #1436: the kimi_k25 MoonViT tower was 2D-only, so a 3-column (t,h,w) video grid would crash or be silently mis-processed.

The fix teaches each stage about time: the position embeddings get a per-frame temporal signal, the rotary positions are repeated across frames, and the frames are averaged together at the merge step. Attention now runs on one image/clip at a time instead of building one big mask.

Blaizzy:
Thanks @Lazarus-931! 

Overall LGTM and a solid improvement 🔥

One nit: 

Let's revert the change for Kimi and open a separate PR. Helps us keep PRs focused and trace issues better 

After the above is done we can merge

Blaizzy (approved):
LGTM 🚀

PR #1438, opened 2026-06-26
kimi_k25: 3D MoonViT video support (Fixes #1436)

### LINKED ISSUE #1436: kimi_k25 (Kimi-K2.x) MoonViT vision tower is 2D-only — no video/temporal support
### Summary

The `kimi_k25` vision tower (Kimi-K2.5 / K2.6 / K2.7 — MoonViT) on current `main`
(`mlx_vlm/models/kimi_k25/vision.py`, commit e390667) is **2D-only**, so it handles single
images but not **video**, even though these models are natively image-text-to-text **and** video.

Despite the `grid_thw` naming, every stage treats the grid as 2-column `(h, w)`:

- `Learnable2DInterpPosEmb.__call__` compares each grid row to `self.weight.shape[:-1]` (a 2-tuple) and bicubic-interpolates to a 2D `size`;
- `Rope2DPosEmb.get_freqs_cis` does `for h, w in shapes` (a 3-column `(t, h, w)` row would fail to unpack);
- `cu_seqlens` in `VisionModel.__call__` is `grid_thw[:, 0] * grid_thw[:, 1]` (i.e. `h*w`, ignoring any temporal dim);
- `patch_merger` takes `height, width = x_shape[0], x_shape[1]` and does the 2×2 spatial merge with no temporal pooling;
- `Attention` materializes a full `[1, seq, seq]` mask matrix.

There is no temporal position embedding and no temporal pooling, so multi-frame input isn't represented as a temporal sequence.

The source model declares the 3D MoonViT path in its own config (`video_attn_type: spatial_temporal`, `merge_type: sd2_tpool`, divided 3D position embedding with a temporal table, `use_unified_vision_chunk`), so video is a first-class capability of the checkpoint that the current port doesn't yet implement.

### Proposed change (3D MoonViT, gated on a 3-column `(t, h, w)` grid; `(h, w)` image path unchanged)

1. **Temporal position embedding** — keep the learnable 2D spatial embedding and add InternVideo2-style 1D sin-cos temporal embeddings per frame, computed/cached at module level so they are **not** checkpoint parameters.
2. **RoPE** — tile the per-frame 2D `freqs_cis` `t` times for a `(t, h, w)` grid.
3. **cu_seqlens** — span `t·h·w` per item.
4. **Attention** — block-diagonal varlen form (loop over `cu_seqlens` segments) instead of the full `[seq, seq]` matrix, which (a) is the correct mask for multiple images / video chunks and (b) avoids OOM once more than one chunk is passed. Single image/chunk is mathematically unchanged.
5. **Merger** — temporal-mean-pool over `t` (`sd2_tpool`) before the 2×2 spatial merge, so each chunk collapses to one spatial token set.

The image (`t = 1`) path stays byte-identical by construction.

### Status / validation

I have a working implementation of the above against this `kimi_k25` tower:

- **Image regression:** identical, correct descriptions; decode unchanged (~23 tok/s on an M3 Ultra).
- **Video:** a 4-frame chunk → correct scene perception; **multi-chunk** (16 frames / 4 chunks) → the model reasons about motion/changes across frames; no OOM; ~23 tok/s decode.

It currently lives in a public 3.6-bit MLX build of Kimi-K2.7-Code (`avlp12/Kimi-K2.7-Code-Alis-MLX-Dynamic-3.6bpw-VLM`).

I'm happy to open a PR porting this onto current `main`. Before I do — would you welcome it, and do you have a preferred shape for it (e.g. keep it inside `kimi_k25/vision.py` gated on the grid's temporal dim, vs a separate path)? A native `--video` CLI/processor path could be a follow-up; this would land the tower + a reference inference helper.

Lazarus-931:
Fixes #1436. The `kimi_k25` MoonViT vision tower was 2D-only, so a 3-column `(t, h, w)` video grid would crash or be silently mis-processed.

This adds a 3D MoonViT path gated on the grid's temporal dimension; the `(h, w)` image path is unchanged (verified `max|Δ| = 7.45e-8` against the old attention).

- **Temporal position embedding**: 2D spatial embedding tiled across frames + a module-cached 1D sin-cos temporal embedding (not a checkpoint parameter).
- **RoPE**: per-frame 2D `freqs_cis` tiled `t` times.
- **cu_seqlens**: spans `t*h*w` per item (`h*w` for a 2-col grid — unchanged).
- **Attention**: block-diagonal varlen (per image/clip segment) instead of the full `[1, seq, seq]` mask — the correct mask for multiple images / video chunks, and avoids the O(seq²) blowup. A single segment is mathematically identical to before.
- **Merger**: temporal mean-pool (`sd2_tpool`) over `t` before the 2×2 spatial merge, so each chunk collapses to one spatial token set.

Tests in `mlx_vlm/tests/test_models.py::TestKimiK25VideoVision`: image-path identity vs the old full-mask attention, video grid runs end to end, temporal pooling collapses frames independent of `t`, and block-diagonal attention has no cross-chunk leakage. Full `test_models.py` suite passes.

A native `--video` processor/CLI path is a sensible follow-up; this lands the tower.

### DIFF
--- a/mlx_vlm/models/kimi_k25/vision.py
+++ b/mlx_vlm/models/kimi_k25/vision.py
@@ -1,3 +1,4 @@
+import math
 from typing import List, Optional
 
 import mlx.core as mx
@@ -5,6 +6,30 @@ import mlx.nn as nn
 
 from ..kernels import bicubic_interpolate
 from .config import VisionConfig
+
+_TEMPORAL_EMB_CACHE = {}
+
+
+def _temporal_sincos_emb(num_frames: int, dim: int) -> mx.array:
+    """InternVideo2-style 1D sin-cos temporal embedding, shape (num_frames, dim).
+
+    Computed and cached at module level so it is never a checkpoint parameter.
+    """
+    key = (num_frames, dim)
+    if key not in _TEMPORAL_EMB_CACHE:
+        pos = mx.arange(num_frames, dtype=mx.float32)[:, None]
+        half = max(dim // 2, 1)
+        freq = mx.exp(-math.log(10000.0) * mx.arange(half, dtype=mx.float32) / half)[
+            None
+        ]
+        ang = pos * freq
+        emb = mx.concatenate([mx.sin(ang), mx.cos(ang)], axis=-1)
+        if emb.shape[-1] < dim:  # odd dim
+            emb = mx.concatenate(
+                [emb, mx.zeros((num_frames, dim - emb.shape[-1]))], axis=-1
+            )
+        _TEMPORAL_EMB_CACHE[key] = emb[:, :dim]
+    return _TEMPORAL_EMB_CACHE[key]
 
 
 def check_array_shape(arr):
@@ -76,7 +101,25 @@ class Learnable2DInterpPosEmb(nn.Module):
     def __call__(self, x: mx.array, grid_hws: mx.array) -> mx.array:
         pos_embs = []
         for shape in grid_hws.tolist():
-            if shape == self.weight.shape[:-1]:
+            if len(shape) == 3:
+                # video (t, h, w): spatial emb tiled across frames + temporal emb
+                t, h, w = shape
+                spatial = (
+                    bicubic_interpolate(
+                        mx.expand_dims(self.weight.transpose(2, 0, 1), axis=0),
+                        size=(h, w),
+                    )
+                    .squeeze(0)
+                    .transpose(1, 2, 0)
+                    .flatten(end_axis=1)
+                )
+                temporal = _temporal_sincos_emb(t, spatial.shape[-1]).astype(
+                    spatial.dtype
+                )
+                pos_embs.append(
+                    mx.tile(spatial, (t, 1)) + mx.repeat(temporal, h * w, axis=0)
+                )
+            elif shape == self.weight.shape[:-1]:
                 pos_embs.append(self.weight.flatten(end_axis=1))
             else:
                 result = (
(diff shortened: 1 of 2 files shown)


mlx_vlm/models/kimi_k25/vision.py
```diff
@@ -6,6 +7,30 @@
 from ..kernels import bicubic_interpolate
 from .config import VisionConfig
 
+_TEMPORAL_EMB_CACHE = {}
```
Blaizzy: How about we use the a bounded `lru_cache` because it's a much safer default, it avoids retaining MLX/device buffers indefinitely.

Same behaviour but with eviction 👌🏽

Example: 

```python 
from functools import lru_cache

@lru_cache(maxsize=32)
def _temporal_sincos_emb(num_frames: int, dim: int) -> mx.array:
    pos = mx.arange(num_frames, dtype=mx.float32)[:, None]
    half = max(dim // 2, 1)
    freq = mx.exp(
        -math.log(10000.0) * mx.arange(half, dtype=mx.float32) / half
    )[None]
    ang = pos * freq
    emb = mx.concatenate([mx.sin(ang), mx.cos(ang)], axis=-1)
    if emb.shape[-1] < dim:
        emb = mx.concatenate(
            [emb, mx.zeros((num_frames, dim - emb.shape[-1]))], axis=-1
        )
    return emb[:, :dim]
```

Blaizzy (commented):
LGTM, just one nit and we can merge 🚀

Blaizzy (approved):
🚀

Issue #1446, opened 2026-06-26
qwen3_5: BatchTurboQuantKVCache misrouted into SSM rollback → IndexError in rollback_speculative_cache (batched TurboQuant speculative decode)


bonyohana:
### Summary

In `qwen3_5`, `LanguageModel.rollback_speculative_cache` misclassifies the **batched TurboQuant full-attention KV cache** (`BatchTurboQuantKVCache`) as an SSM/GDN cache. This corrupts the `ssm_caches` ↔ `gdn_states` alignment and crashes batched speculative decoding with TurboQuant-quantized KV.

This is a **separate, more fundamental** bug from the TurboQuant×MTP *tail-zeroing* family in #1428 / #1432. I originally thought the tail-zero fix was "systemic" and applied to qwen3_5 too, then [retracted that on #1432](https://github.com/Blaizzy/mlx-vlm/pull/1432#issuecomment-4810383098) when the surface proved the tail-zero line was unreachable for any TurboQuant cache in qwen3_5. Tracing *why* it was unreachable surfaced the real qwen3_5 bug below.

### Root cause

`rollback_speculative_cache` splits caches into "trimmable KV" vs "SSM" using `is_trimmable()` as the proxy for *"is this an SSM/GDN cache?"* (`mlx_vlm/models/qwen3_5/language.py`, ~L1988–2025):

```python
ssm_caches = []
for c in caches:
    if c is None:
        continue
    if c.is_trimmable():
        ...   # trim + tail-zero  (full-attention KV rollback)
    else:
        ssm_caches.append(c)   # assumed to be SSM/GDN
```

That proxy holds for the f16 path (`KVCache.is_trimmable() → True`, GDN `ArraysCache → False`), but breaks for the **batched** TurboQuant cache:

| cache | `is_trimmable()` | `zero_row_tail` | `_idx` |
|---|---|---|---|
| `KVCache` (f16 full-attn) | `True` | – | – |
| `TurboQuantKVCache` (quant full-attn) | `True` | ✓ | – |
| **`BatchTurboQuantKVCache`** (batched quant full-attn) | **`False`** | ✓ | ✓ |
| `ArraysCache` (GDN/SSM) | `False` | – | – |

[...]

PR #1447, opened 2026-06-26
fix for #1446

Lazarus-931:
fix for #1446 

`LanguageModel.rollback_speculative_cache` (qwen3_5) splits caches into trimmable-KV vs SSM/GDN using `is_trimmable()` as the SSM proxy. 

Instead,  changed to classify KV-rollback interface instead of `is_trimmable()`, also since  SSM/GDN caches are  non-trimmable and lack the tail-rollback method, this works for every full-attn variant. 

All tests pass


Blaizzy (approved):
🚀

Blaizzy:
LGTM!!

Blaizzy:
Please setup verified signatures for your future commits

PR #1452, opened 2026-06-27
[codex] Add TurboQuant rollback hook for DeepSeek V4

Blaizzy:
## Summary
- route DeepSeek V4 speculative rollback through the TurboQuant `zero_row_tail` cache hook before falling back to dense KV assignment
- add a regression for batched DeepSeek V4 rollback with `BatchTurboQuantKVCache`

## Stack
This PR is stacked on #1432, which introduces the shared TurboQuant cache hook.

## Root cause
DeepSeek V4 rollback uses the same dense `cache.keys[bi, ...] = 0` rejected-tail assignment pattern as Gemma 4. That is not valid for TurboQuant cache states, which are immutable `NamedTuple` containers with mutable array fields.

## Validation
- `uv run --with pytest pytest mlx_vlm/tests/test_speculative.py::test_gemma4_rollback_speculative_cache_handles_turboquant_batch_tail_zero mlx_vlm/tests/test_speculative.py::test_deepseek_v4_rollback_speculative_cache_handles_turboquant_batch_tail_zero`
- `uv run --with pytest pytest mlx_vlm/tests/test_speculative.py mlx_vlm/tests/test_turboquant.py`

### DIFF
--- a/mlx_vlm/models/deepseek_v4/language.py
+++ b/mlx_vlm/models/deepseek_v4/language.py
@@ -1373,8 +1373,12 @@ class LanguageModel(nn.Module):
                 for bi, valid_end in enumerate(valid_ends.tolist()):
                     start = verify_start + int(valid_end)
                     if start < kv_len:
-                        cache.keys[bi, :, start:kv_len, :] = 0
-                        cache.values[bi, :, start:kv_len, :] = 0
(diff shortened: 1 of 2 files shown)


Blaizzy:
Closing this stacked DeepSeek-only PR because the requested follow-up is Qwen3.5, and the DeepSeek change already landed through #1432 before the stack split.

PR #1458, opened 2026-06-27
Improve Responses API reasoning and tool stream events

justin-schroeder:
Hi! This tightens a few Responses API shapes in the OpenAI-compatible server.

What changed:
- Return thinking blocks as `reasoning` output items with `summary` text, while keeping visible text in the message item.
- Stream thinking through `response.reasoning_text.delta/done` instead of mixing it into `response.output_text.delta`.
- Emit `response.function_call_arguments.done` for streamed function calls once arguments are parsed.

I kept this small and covered the new behavior in the existing server tests.

Tests:
- `.venv/bin/python -m pytest -q mlx_vlm/tests/test_server.py`
- `.venv/bin/pre-commit run --files mlx_vlm/server/responses_state.py mlx_vlm/server/openai.py mlx_vlm/tests/test_server.py`

### DIFF
--- a/mlx_vlm/server/openai.py
+++ b/mlx_vlm/server/openai.py
@@ -887,6 +887,13 @@ async def responses_endpoint(request: Request):
                         if tool_module is not None and chat_tools
                         else None
                     )
+                    thinking_state = ThinkingStreamState(
+                        gen_args.enable_thinking,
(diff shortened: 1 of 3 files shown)


lucasnewman (approved):
This is great, thanks!

PR #1462, opened 2026-06-28
perf(qwen3-omni): early-exit the thinker hidden-state extraction

yuhuanowo:
### What

`extract_thinker_hidden_states` runs a full thinker forward
(`output_hidden_states=True` — all 48 layers, plus the final norm and `lm_head`)
only to read one layer's hidden state, `accept_hidden_layer` (24 by default),
which is all the Talker conditions on. Everything computed past that layer is
discarded.

This adds an opt-in `early_exit_layer` to the thinker `LanguageModel` that
returns as soon as that layer is computed, and points
`extract_thinker_hidden_states` at it instead of indexing the full
`output_hidden_states` list.

### Why it's safe

The returned tensor is exactly what `output_hidden_states[early_exit_layer + 1]`
would hold — same layers in the same order, just stopped before the unused tail.
A direct comparison on Qwen3-Omni-30B-A3B-Instruct-4bit gives
`max|Δ| = 0.0`. It defaults to `None`, so existing callers and `generate_step`
are untouched.

### Speedup

The extraction forward drops from 52 ms to 29 ms (~1.8x) on the same model /
M3 Max, since it skips half the decoder stack plus the `lm_head` projection. It
runs once per Talker turn, so it's on the audio path.

### DIFF
--- a/mlx_vlm/models/qwen3_omni_moe/language.py
+++ b/mlx_vlm/models/qwen3_omni_moe/language.py
@@ -247,6 +247,13 @@ class Qwen3VLMoEModel(nn.Module):
             mask = create_attention_mask(
                 h, cache[0] if cache and cache[0] is not None else cache
             )
+        if early_exit_layer is not None and (
+            early_exit_layer < 0 or early_exit_layer >= len(self.layers)
+        ):
+            raise ValueError(
+                "early_exit_layer must be in "
+                f"[0, {len(self.layers) - 1}], got {early_exit_layer}"
+            )
 
         all_hidden_states = [] if output_hidden_states else None
         position_embeddings = None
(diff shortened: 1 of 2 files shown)


Blaizzy:
Hey @yuhuanowo!

Thanks for the optimization. After comparing this with the Transformers implementation, I think we should skip this change for now. Transformers uses output_hidden_states=True and reads hidden_states[accept_hidden_layer] directly, while this PR adds a custom early-exit path and changes the language model flow. The speedup is nice, but the added complexity and indexing subtlety make it harder to maintain. I’d prefer we keep the implementation closer to Transformers unless this becomes a proven bottleneck we need to optimize.

PR #1463, opened 2026-06-29
fix(apc): enable APC prefix cache with MTP speculative decoding

### LINKED ISSUE #1443: APC (Automatic Prompt Caching) silently disabled when MTP speculative decoding is enabled (v0.6.3)
## Summary

Enabling **MTP speculative decoding** (`--draft-kind mtp`) **silently disables APC (Automatic Prompt Caching)**. With the MTP drafter loaded, repeated requests that share a long prompt prefix are re-prefilled from scratch every time — `prompt_tokens_details.cached_tokens` stays `0` and wall-clock prefill latency does not drop on the second identical-prefix request. Dropping the draft model restores APC fully (large `cached_tokens`, ~20x faster warm prefill).

There is no error and no warning — APC just no-ops. So a deployment running both MTP and `APC_ENABLED=1` believes it has prompt caching but doesn't, and pays full prefill cost on every turn (very expensive for agent workloads with a large reused system prompt + growing history).

This is **distinct from #1428** (TurboQuant×MTP crash). Here there is **no TurboQuant** — KV cache is full f16, the only variable is the MTP draft.

## Environment

- mlx-vlm **0.6.3**
- macOS, Apple Silicon (M-series, 64 GB), Python 3.11.15
- Model: `gemma-4-31b-it-mxfp4`; draft model `gemma-4-31b-it-assistant-bf16` (MTP)
- `APC_ENABLED=1`, `APC_DISK_PATH=<dir>`, `APC_DISK_MAX_GB=20` (disk tier active per startup log: `APC enabled (... disk=True)`)
- **No** `--kv-bits` / `--kv-quant-scheme` (full f16 KV)

## Reproduce

Serve **with** the MTP draft + APC:

```bash
APC_ENABLED=1 APC_DISK_PATH=/tmp/apc-cache APC_DISK_MAX_GB=20 \
mlx_vlm.server \
  --model <gemma-4-31b-it-mxfp4> \
  --draft-model <gemma-4-31b-it-assistant-bf16> \
  --draft-kind mtp \
  --port 8081
```

Send the **same** ~4K-token system prefix twice (different 1-token user tail), via `/v1/chat/completions`, and read `usage.prompt_tokens_details.cached_tokens` plus wall-clock.

Then restart **without** `--draft-model` (everything else identical) and repeat.

## Observed (A/B, same machine, same fresh APC disk dir, MTP the only variable)

| Config | warm-request `cached_tokens` | warm prefill wall-clock | speedup vs cold |
|---|---|---|---|
| **MTP draft ON**, APC on | `0` (of 4230) | 12.15 s (cold 12.64 s) | **1.0x — no caching** |
| **MTP draft OFF**, APC on | `4212` (of 4228) | 0.59 s (cold 12.50 s) | **21.1x** |

The wall-clock confirms it's a real prefill cache miss under MTP, not just a `cached_tokens` reporting gap — the warm request under MTP takes as long as the cold one.

## Expected

Either APC prefix reuse should function alongside MTP speculative decoding, or — if they're fundamentally incompatible in the current design — the server should **emit a warning at startup** when both are enabled (and/or in the docs), rather than silently disabling APC. The silent no-op is the trap: caching looks enabled but isn't.

## Workaround

Drop the MTP draft model. For agent-style workloads (large reused prefix), APC's prefill savings far outweigh MTP's generation-throughput gain (~+1 tok/s in our measurements), so APC-without-MTP is the better trade.

vedang:
Two changes to make APC block-level prefix caching work alongside speculative (MTP) decoding:

1. harvest_blocks_from_batch_cache: accept regular KVCache entries (which use `offset` instead of `_idx`) when the prompt cache was created by `make_speculative_prompt_cache` for MTP batch_size=1. Without this fallback the harvest returns [] silently because `getattr(c, "_idx", None)` is None on non-batch KVCache.

2. BatchGenerator.__init__: stop nulling `apc_manager` when `draft_model` is set. This was a defensive disable added in eb7537b ("Fix Qwen MTP batched target-verify drift", PR #1210)

Together these allow mlx_vlm.server to serve fast hot-TTFT via APC while retaining the ~2-3x decode speedup from MTP speculative decoding, rather than forcing users to choose one or the other.

Fixes: #1443

### DIFF
--- a/mlx_vlm/apc.py
+++ b/mlx_vlm/apc.py
@@ -3711,6 +3711,11 @@ def harvest_blocks_from_batch_cache(
         keys = getattr(c, "keys", None)
         values = getattr(c, "values", None)
         idx = getattr(c, "_idx", None)
+        # Regular KVCache (non-batch) — e.g. from make_speculative_prompt_cache
+        # used with MTP draft model + batch_size=1. These entries carry
+        # ``offset`` instead of ``_idx`` and have no ``left_padding``.
+        if idx is None and keys is not None and values is not None:
+            idx = getattr(c, "offset", None)
         left_padding = getattr(c, "left_padding", None)
         if keys is None or values is None or idx is None:
             return []
--- a/mlx_vlm/generate/ar.py
+++ b/mlx_vlm/generate/ar.py
@@ -2095,7 +2095,6 @@ class BatchGenerator:
         self.draft_block_size = draft_block_size
         self.greedy_sampling = greedy_sampling or sampler is None
         if self.draft_model is not None:
-            apc_manager = None
             compute_logprobs = False
             top_logprobs_k = 0
             self.compute_logprobs = False


Blaizzy:
Hey @vedang 

Thanks for the contribution! 

Could you share the before and after? And if this change works with batched results (b > 1)

vedang:
## 1. Before/After Performance Comparison

All tests on Apple M4 Pro, 64 GB unified memory, `mlx-community/Qwen3.6-27B-4bit`.

### Configurations tested

| Config | Server command | APC? | MTP? |
|---|---:|---:|---:|
| **A: APC only** | `APC_ENABLED=1 APC_NUM_BLOCKS=8192 APC_DISK_PATH=... APC_DISK_MAX_GB=24 mlx_vlm.server --model mlx-community/Qwen3.6-27B-4bit --host 127.0.0.1 --port 8011 --max-tokens 768 --prefill-step-size 4096` | ✅ | ❌ |
| **B: MTP only** | `APC_ENABLED=1 APC_NUM_BLOCKS=8192 APC_DISK_PATH=... APC_DISK_MAX_GB=24 mlx_vlm.server --model mlx-community/Qwen3.6-27B-4bit --draft-model mlx-community/Qwen3.6-27B-MTP-4bit --draft-kind mtp --draft-block-size 2 --host 127.0.0.1 --port 8011 --max-tokens 768 --prefill-step-size 4096` (APC disabled by code) | ❌ | ✅ |
| **C: Both (fixed)** | Same as B, with the two fixes applied | ✅ | ✅ |

### Performance table

| Metric | A: APC only | B: MTP only | C: Both (fixed) | Unit |
|---|---:|---:|---:|---|
| **Cold TTFT** (10-13k prompt) | 101-113 | 96-137 | 87 | seconds |
| **Hot TTFT** (cached prefix) | 0.6-0.8 | never hit | **1.2-4.4** | seconds |
| **Hot prefill tok/s** | 14,000+ | ~116 (always reprefill) | **1,051-11,000** | tok/s |
| **Tool-call TTFT** | 5-10 (partial hit) | 48-130 (cold) | **16-22** | seconds |
| **Decode tok/s** | 6-9 | 19-20 | 18-20 | tok/s |
| **Avg request elapsed** | 40 | 94 | **34** | seconds |
| **APC hit rate** | 100% | 0% | 100% | — |
| **APC stores** | active | zero | active | — |

### Key takeaways

- **Config A** (APC only): fast TTFT when cached, but decode is only 6-9 tok/s — too slow for interactive agent loops.
- **Config B** (MTP only): fast decode at 19-20 tok/s, but APC was disabled by a defensive guard, so every turn reprefills — 100+ seconds of waiting per turn.
- **Config C** (both fixed): fast TTFT (1-4s when cached) AND fast decode (18-20 tok/s). Total request time dropped from 94s average to 34s — a **2.8× improvement**. Best-case hot request: 5.5s total.

### Performance before vs after (hot cache)

```
Before (MTP only, no APC):      After (MTP + APC, fixed):
┌──────────────────────┐        ┌──────┐
│   95-137s TTFT       │        │ 1-4s │  TTFT  
│   (every turn cold)  │   →    │      │
│   19 tok/s decode    │        │ 19 tok/s decode
│   ~100s total/turn   │        │ ~5-35s total/turn
└──────────────────────┘        └──────┘
```

## 2. Batch KV (b > 1) Analysis

| Batch size | Cache type | `_idx`? | `offset`? | Fix activates? | `batch_idx` range | Result |
|---:|---|:---:|:---:|:---:|---:|---|
| b=1, MTP | `KVCache` | ❌ | ✅ | ✅ yes | always 0 | `keys[0:1, :, 0:offset, :]` ✅ |
| b>1, MTP | `BatchKVCache` | ✅ | ❌ | ❌ no | 0..b-1 | `keys[i:i+1, :, 0:_idx[i], :]` ✅ |
| b=1, no MTP | `BatchKVCache` | ✅ | ❌ | ❌ no | always 0 | `keys[0:1, :, 0:_idx[0], :]` ✅ |

Blaizzy (approved):
🚀

Issue #1460, opened 2026-06-28
Feature Status for Deepseeks DSpark


Lazarus-931:
Not an repo issue, but an open DSpark implementation status issue. Since deepseeks algorithm uses a model unfit for current setup, I have instead choosen the option of training a qwen3-14b DSpark model with mlx. Will keep posted & hopefully with final pr. 

Paper is promising!

https://github.com/deepseek-ai/DeepSpec

Blaizzy:
Awesome, been looking at DSpark.

Looking forward to it!

Flor1an-B:
@Blaizzy DSPark will be also compatible with Qwen models? 
https://github.com/deepseek-ai/DeepSpec

PR #1465, opened 2026-06-29
feat: Gemma 4 DSpark self-speculative decoding

popfido:
## Gemma 4 DSpark self-speculative decoding

Adds **DSpark** (DeepSeek's self-speculative decoding, `deepseek-ai/DeepSpec`) as a fourth in-tree drafter family for Gemma 4 targets, alongside `dflash` / `eagle3` / `mtp`. Output is **byte-identical to base greedy decoding**.

DSpark is EAGLE-style: the drafter conditions on the target's intermediate hidden states (the same `capture_layer_ids` mechanism DFlash uses), projects them, and predicts a block of tokens biased by a low-rank **Markov head**, with a **confidence head** that can adaptively shorten speculation. It slots into the existing propose → verify → walk → rollback machinery — so the whole verify/accept path is reused unchanged.

### What's added

- **`speculative/drafters/gemma4_dspark/`** — a standalone drafter (ships its own `embed_tokens`/`lm_head`, unlike DFlash which binds to the target). Gemma 4 draft layers (K=V sharing, attention scale 1.0, partial RoPE, sandwich norms, GeGLU, logit softcap) run over mlx-vlm's `KVCache` with the same context/proposal split as `DFlashAttention`. Includes the Markov + confidence heads.
- **`speculative/dspark.py`** — `_dspark_rounds` (single-sequence, confidence-adaptive) and `_dspark_rounds_batch` (continuous batching), reusing `_speculative_walk` / `rollback_speculative_cache`.
- **Auto-detection & routing** — the published checkpoint declares a generic `model_type=gemma4_text`, so it's identified by its `Gemma4DSparkModel` architecture tag (+ draft-hyperparameter cluster) in `get_model_and_args` and `resolve_drafter_kind`; an explicit wrong `--draft-kind` is corrected with a warning.
- **CLI** — `--draft-kind dspark` and `--draft-confidence-threshold`.
- **Tests + README**.

### Usage

```sh
mlx_vlm.generate --model google/gemma-4-12b-it \
  --draft-model deepseek-ai/dspark_gemma4_12b_block7 \
  --prompt "Explain speculative decoding in 3 sentences." \
  --max-tokens 256 --temperature 0

mlx_vlm.server --model google/gemma-4-12b-it \
  --draft-model deepseek-ai/dspark_gemma4_12b_block7
```

### Lossless + adaptive

The confidence head truncates the draft block where `P(accept)` drops below `--draft-confidence-threshold` (env `MLX_VLM_DRAFT_CONFIDENCE_THRESHOLD`). This only changes **how far ahead the drafter speculates** before the target verifies — never which tokens are emitted — so output stays lossless. Confidence gating is applied on the single-sequence path; batched continuous decoding uses a fixed block (also lossless).

### Testing

`mlx_vlm/tests/test_dspark.py` (14 tests) + no regressions in `test_speculative.py` (147):

- **Losslessness** — `_dspark_rounds` and `_dspark_rounds_batch` produce output bit-identical to plain greedy against a history-independent Markov target, at confidence thresholds 0 / 0.5 / 0.9 and batch size 3.
- **Real checkpoint** *(env-gated on `DSPARK_GEMMA4_CONFIG` / `DSPARK_GEMMA4_CKPT`)* — the trained `dspark_gemma4_12b_block7` weights load with **all 74 keys matching** the module, and `draft_block` produces non-degenerate drafts.
- Config parsing, drafter-kind auto-detection/override, `get_model_and_args` routing (normal models unaffected), and compatibility validation.

```
161 passed (14 DSpark + 147 existing speculative), no regressions
```

### Benchmark note

DSpark acceptance is largely target-size-independent and set by the draft + task. Reference figures for the Gemma 4 backbone (from the `dspark-mlx` eager loop / the DSpark paper's acceptance band) on GSM8K, greedy:

| target | accepted length (τ) | acceptance | speedup |
|---|---|---|---|
| Gemma 4-12B-it (bf16) | ~5.84 | ~69% | ~2.06× |

This PR verifies **correctness** (lossless decode + real-weight load + key parity); end-to-end throughput on an mlx-vlm Gemma-4 target depends on hardware and the ~24 GB instruct model, and is left to measure on-device.

### Scope / limitations

- Gemma 4 only for now. Other DSpark backbones (Qwen3, DeepSeek-V4) can follow the same pattern.
- Why not port Qwen3
  - Port a dense Qwen3 target into mlx-vlm with the two spec hooks — substantial, and duplicates what mlx-lm already owns (mlx-vlm deliberately defers dense text to mlx-lm).
  - A DSpark drafter trained for Qwen3.5 — doesn't exist yet.
- Confidence-adaptive truncation is single-sequence; batched decoding uses a fixed block (still lossless).
- The drafter must target the deployed **instruct** model.

Based on `deepseek-ai/DeepSpec` (`dspark/*`); companion to the `dspark-mlx` package.

### DIFF
--- a/mlx_vlm/server/cli.py
+++ b/mlx_vlm/server/cli.py
@@ -173,8 +173,9 @@ def main():
         "--draft-kind",
         type=str,
         default=None,
-        choices=["dflash", "eagle3", "mtp"],
-        help="Drafter family -- 'dflash', 'eagle3', or 'mtp' (Gemma 4). "
+        choices=["dflash", "dspark", "eagle3", "mtp"],
+        help="Drafter family -- 'dflash', 'dspark' (Gemma 4 self-speculative), "
+        "'eagle3', or 'mtp' (Gemma 4). "
         "Default: auto-detected from the drafter's HF model_type.",
     )
     parser.add_argument(
@@ -182,6 +183,14 @@ def main():
         type=int,
         default=None,
         help="Override the drafter's configured block size.",
+    )
+    parser.add_argument(
+        "--draft-confidence-threshold",
+        type=float,
+        default=None,
+        help="DSpark only: truncate the draft block where the confidence head's "
+        "P(accept) drops below this threshold (0 = off, lossless either way). "
+        "Maps to the MLX_VLM_DRAFT_CONFIDENCE_THRESHOLD env var.",
     )
     parser.add_argument(
         "--top-logprobs-k",
@@ -235,6 +244,10 @@ def main():
         os.environ["MLX_VLM_DRAFT_KIND"] = args.draft_kind
     if args.draft_block_size is not None:
         os.environ["MLX_VLM_DRAFT_BLOCK_SIZE"] = str(args.draft_block_size)
+    if args.draft_confidence_threshold is not None:
+        os.environ["MLX_VLM_DRAFT_CONFIDENCE_THRESHOLD"] = str(
+            args.draft_confidence_threshold
+        )
     if args.prefill_step_size:
         os.environ["PREFILL_STEP_SIZE"] = str(args.prefill_step_size)
     os.environ["MLX_VLM_MAX_TOKENS"] = str(args.max_tokens)
--- a/mlx_vlm/utils.py
+++ b/mlx_vlm/utils.py
@@ -382,8 +382,13 @@ def get_model_and_args(config: dict):
 
     model_type = MODEL_REMAPPING.get(model_type, model_type)
 
+    # DSpark drafters declare a generic model_type (e.g. ``gemma4_text``) and are
+    # identified by their architecture tag (see speculative.drafters._config_is_dspark).
+    is_dspark = "Gemma4DSparkModel" in (config.get("architectures") or [])
     is_dflash = config.get("dflash_config", None) is not None
-    if is_dflash:
+    if is_dspark:
+        model_type = "gemma4_dspark"
+    elif is_dflash:
         model_type += "_dflash"
 
     last_err: Optional[ImportError] = None
--- a/README.md
+++ b/README.md
@@ -128,8 +128,9 @@ Speed up generation by drafting several candidate tokens with a small "drafter" 
 | Flag | Description |
 |------|-------------|
 | `--draft-model` | HuggingFace repo or local path for the drafter |
-| `--draft-kind` | Drafter family — `dflash` (default), `eagle3`, or `mtp` (Gemma 4) |
+| `--draft-kind` | Drafter family — `dflash` (default), `dspark` (Gemma 4 self-speculative), `eagle3`, or `mtp` (Gemma 4) |
 | `--draft-block-size` | Override the drafter's configured block size |
+| `--draft-confidence-threshold` | DSpark only — truncate the draft block where the confidence head's `P(accept)` drops below this (0 = off; lossless either way) |
 
 See [docs/usage.md](docs/usage.md) for Python API examples including batch generation.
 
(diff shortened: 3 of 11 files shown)


Lazarus-931:
@popfido thanks for the pr! 

quick question, why not reuse some of the gemma 4 implementations already in  mlx_vlm/models/gemma4 for less code duplication? 

I think we should keep this open until qwen & deepseek are supported, so we can close #1460 once done! 

EDIT: also CI/CD issues!

Lazarus-931:
@popfido lgtm!

best we keep it open until qwen & deepseek support implemented, for #1460 

cc: @lucasnewman @Blaizzy

popfido:
I in fact also implemented qwen and deepseek in my mlx reproduction repo.
qwen is also fully tested by myself. I'll ship it. For deepseek I just
finished a theoretical validation version but can't test since I do not
have any capable device for running such a huge model locally.

popfido:
> @popfido lgtm!
> 
> best we keep it open until qwen & deepseek support implemented, for #1460
> 
> cc: @lucasnewman @Blaizzy

The way I did not raise Qwen support is because DS just released DSpark ckpt for Qwen3. Qwen3.5 and Qwen3 are quite different architectures.

So a "similar" mlx-vlm PR would require one of:

- Port a dense Qwen3 target into mlx-vlm with the two spec hooks — substantial, and duplicates what mlx-lm already owns (mlx-vlm deliberately defers dense text to mlx-lm); or
- A DSpark drafter trained for Qwen3.5 — doesn't exist yet.

Either way, unlike Gemma 4, I couldn't validate it end-to-end (no real target+checkpoint pair inside mlx-vlm).

Lazarus-931:
@popfido I see, i think then we should keep this pr for strictly gemm4 and another for qwen, ds respectively. as checkpoints get released and e2e tested, we'll merge them as well

cc: @lucasnewman @Blaizzy

mlx_vlm/server/cli.py
```diff
@@ -183,6 +184,14 @@ def main():
...
         help="Override the drafter's configured block size.",
     )
+    parser.add_argument(
+        "--draft-confidence-threshold",
+        type=float,
+        default=None,
+        help="DSpark only: truncate the draft block where the confidence head's "
+        "P(accept) drops below this threshold (0 = off, lossless either way). "
+        "Maps to the MLX_VLM_DRAFT_CONFIDENCE_THRESHOLD env var.",
+    )
```
Blaizzy, line 194: We should probably call `--confidence-threshold` instead

It will serve both here and for diffusion models.

mlx_vlm/utils.py
Blaizzy, line 1: There are going to be new dspark models beyond gemma4.

Please make a better more generalizable solution.

Blaizzy:
Hey @popfido,

Thanks a lot for the great contributions!

That said, there's quite a bit here that needs to be reverted and redesigned. To speed up the process, I'll either make those changes in this PR directly, or open a new one and add you as a co-author.

mlx_vlm/server/cli.py
popfido: I think draft model should be differentiated with diffusion model

popfido: Thus their confidence threshold should be treated differently. Or you mean confidence-threshold should be used by both diffusion llm and draft model? Could it be a little bit confusing?

weklund:
Ran the mlx-dspark benchmark on an M5 Max 128GB with gemma-4-12B-it-8bit target for reference:

mlx-dspark benchmark · Apple M5 Max 128GB  · mlx 0.31.2
target: mlx-community/gemma-4-12B-it-8bit

run                          tok/s
  baseline                  10.6 tok/s   accept 1.00
  dspark cap=2         16.7 tok/s   accept 2.67  (1.58x)
  dspark cap=auto    18.2 tok/s   accept 4.13  (1.72x)
  dflash cap=2           16.6 tok/s   accept 2.54  (1.57x)
  dflash cap=auto      18.6 tok/s   accept 4.11  (1.75x)

Also tested how DSpark stacks with APC (exact-mode, since Gemma 4 is hybrid attention). APC 
eliminates prefill on repeated prompts, DSpark accelerates decode:

| Context         | No optimization (2nd req) | APC only (2nd req) | APC + DSpark (projected) |
|-----------------|---------------------------|--------------------|--------------------------|
| 4k (4099 tok)   | 4.67s                     | 1.87s (2.5x)       | ~1.1s (~4.2x)            |
| 8k (8081 tok)   | 8.68s                     | 1.88s (4.6x)       | ~1.1s (~7.9x)            |
| 16k (16146 tok) | 16.88s                    | 2.08s (8.1x)       | ~1.2s (~14x)             |

The combined story for multi-turn chat on a hybrid model is compelling. Looking forward to this landing.

PR #1466, opened 2026-06-29
Fix silent text drop / image misalignment for multi-<image> prompts in prepare_inputs

SuperMarioYL:
## Problem

In `prepare_inputs`, the `BaseImageProcessor` path builds `input_ids` by hardcoding a single image token between the first two text chunks:

```python
ids = chunks[0] + [image_token_index] + chunks[1]
```

`chunks` comes from `prompt.split("<image>")`, so this only ever handles a prompt with **exactly one** `<image>` sentinel. Other cases fail silently or with an opaque error:

- **Two or more `<image>` sentinels** → every chunk after the second (`chunks[2:]`) is **silently dropped**, and only **one** image token is inserted even though `pixel_values = preprocess(images=images)` stacks **all** the images. The model then receives `input_ids` with one image slot but N image features — a silent shape/count misalignment, i.e. a wrong answer with no error.
- **An image but no `<image>` sentinel** → `chunks` has one element, so `chunks[1]` raises a bare `IndexError`.
- The `max_length = ... + 1` padding math also assumes exactly one inserted token.

The other (`process_inputs_with_fallback`) path already guards this via the processor, raising `ValueError("Number of image tokens in prompt_token_ids (X) does not match number of images (Y)")`. The `BaseImageProcessor` path had no such check.

## Fix

- **Interleave** every text chunk with an image token (one per `<image>` sentinel) so multi-image prompts keep all of their text and emit exactly one image slot per image.
- **Validate** the total sentinel count against the number of images up front, raising the **same** clear `ValueError` the fallback path already uses, instead of dropping text / misaligning features / `IndexError`-ing.
- Compute `max_length` from the actual interleaved sequences.

Single-image prompts are unaffected — the produced `input_ids` are byte-for-byte identical to before.

## Tests

Adds `test_prepare_inputs_base_image_processor_interleaves_and_validates`, which exercises the `BaseImageProcessor` branch (previously uncovered) across single-image, multi-image, too-few-images, and no-sentinel cases. The multi-image assertion is **red on `main`** (the old code emits one image token) and green here.

```
mlx_vlm/tests/test_utils.py::test_prepare_inputs_base_image_processor_interleaves_and_validates PASSED
205 passed, 2 skipped
```

`black` / `isort` / `autoflake` (pre-commit) all clean.

### DIFF
--- a/mlx_vlm/tests/test_utils.py
+++ b/mlx_vlm/tests/test_utils.py
@@ -8,10 +8,12 @@ from unittest.mock import MagicMock, patch
 
 import mlx.core as mx
 import mlx.nn as nn
+import numpy as np
 import pytest
 from mlx_lm.utils import quantize_model
 
 from mlx_vlm.convert import _preserve_existing_deepseek_v4_quantization
+from mlx_vlm.models.base import BaseImageProcessor
 from mlx_vlm.models.text_only import TextOnlyModel
 from mlx_vlm.utils import (
     StoppingCriteria,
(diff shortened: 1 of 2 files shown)


mlx_vlm/tests/test_utils.py
```diff
@@ -405,6 +407,87 @@ def test_prepare_inputs():
...
 
 
+class _DummyBaseImageProcessor(BaseImageProcessor):
+    """BaseImageProcessor subclass that skips the transformers-backed __init__."""
+
+    def __init__(self):
+        pass
+
+    def preprocess(self, images):
+        return [np.zeros((3, 2, 2), dtype=np.float32) for _ in images]
```
Lazarus-931, line 417: i think u should stick to mx.zeros(), not np, for consistency.

mlx_vlm/tests/test_utils.py
```diff
@@ -405,6 +407,87 @@ def test_prepare_inputs():
...
+        image_token_index=99,
+    )
+    assert mx.array_equal(inputs["input_ids"], mx.array([[1, 99, 1, 99, 1]]))
+    assert int((inputs["input_ids"] == 99).sum()) == 2
+    assert inputs["pixel_values"].shape[0] == 2
+
+    # Too few images for the sentinels -> clear error (was a silent mismatch).
+    with pytest.raises(
+        ValueError,
+        match="Number of image tokens in prompt_token_ids.*does not match number of images",
```
Lazarus-931, line 469: is it supposed to be *does not*? with a "*" on both sides?

Lazarus-931 (commented):
lgtm, i think comments can be condensed a bit!

thanks for the pr!

cc: @lucasnewman & @Blaizzy

Blaizzy:
Hey @SuperMarioYL 

Thanks for the contribution, please open a issue with a reproducible example, then we can ideate on a solution together.

PR #1476, opened 2026-07-01
Fix Qwen3-Omni repeated generation

Blaizzy:
## Summary

Fixes a reproducible Qwen3-Omni issue where only the first generation in a process decoded correctly and later generations collapsed to token 0.

## Changes

- Reset and reuse Qwen3-Omni MRoPE position state correctly between prompts and cached decode.
- Precompute Omni MRoPE positions/deltas when visual grids are present, and forward embedding metadata through the top-level model wrapper.
- Materialize Qwen3-Omni `position_ids` before cached decode rotary attention to avoid the lazy MLX/Metal scheduling failure seen on repeated generation.
- Load Qwen3-Omni processors through local MLX-compatible image/video processors so `AutoProcessor` does not require PyTorch/Torchvision video processing.
- Add focused tests for MRoPE state handling and the processor patch.

## Validation

- `uv run --frozen python -m unittest mlx_vlm.tests.test_processors.TestQwen3OmniMoeProcessor mlx_vlm.tests.test_processors.TestQwen3OmniMoePatch mlx_vlm.tests.test_models.TestGetInputEmbeddings.test_qwen3_omni_moe_text_only_clears_mrope_state mlx_vlm.tests.test_models.TestGetInputEmbeddings.test_qwen3_omni_moe_prefill_recomputes_stale_position_ids`
- `uv run --frozen python -m compileall -q mlx_vlm/models/qwen3_omni_moe mlx_vlm/tests/test_models.py mlx_vlm/tests/test_processors.py`
- `git diff --cached --check`
- Local repro: repeated `Qwen/Qwen3-Omni-30B-A3B-Instruct` generations now return matching sane token sequences instead of token 0 collapse.


lucasnewman (approved):
LGTM

PR #1477, opened 2026-07-01
Fix DiffusionGemma tool marker tokenization

### LINKED ISSUE #1388: Diffusion detok path strips tool-call/channel marker tokens (root cause + fix) — follow-up to #1351

Blaizzy:
## Summary

Fixes DiffusionGemma tool-call marker leakage by changing the DiffusionGemma processor/tokenizer setup rather than the generation path.

The tokenizer for `google/diffusiongemma-26B-A4B-it` registers tool/channel delimiters like `<|tool_call>`, `<tool_call|>`, `<|channel>`, `<channel|>`, and `<|"|>` as special tokens. The diffusion streaming path builds its skip set from `tokenizer.all_special_ids`, so those structural markers were stripped before the tool parser could see them. This caused tool calls to remain as raw text and allowed the `thought` channel label to leak into content.

This PR demotes those parser markers from tokenizer special-token metadata while keeping them tokenizable as normal vocabulary tokens. That lets the existing diffusion detokenizer and tool parser behavior work without generation-specific special cases.

## Validation

- `uv run --with pytest python -m pytest mlx_vlm/tests/test_diffusion_gemma.py mlx_vlm/tests/test_gemma4_tool_parser.py mlx_vlm/tests/test_server.py::test_responses_endpoint_returns_function_call_items mlx_vlm/tests/test_server.py::test_chat_completions_streaming_tool_calls_emit_usage_chunk -q`
- `git diff --check`
- Verified against the cached real `google/diffusiongemma-26B-A4B-it` tokenizer that marker IDs leave `all_special_ids` while `convert_tokens_to_ids(...)` still resolves them.

Fixes [#1388](https://github.com/Blaizzy/mlx-vlm/issues/1388)

### DIFF
--- a/mlx_vlm/models/diffusion_gemma/processing_diffusion_gemma.py
+++ b/mlx_vlm/models/diffusion_gemma/processing_diffusion_gemma.py
@@ -15,9 +15,51 @@ from transformers.tokenization_utils_base import PreTokenizedInput, TextInput
 
 from ..gemma4.processing_gemma4 import Gemma4Processor
 
+_TOOL_PARSER_TOKENS = {
+    "<|tool_call>",
+    "<tool_call|>",
+    '<|"|>',
+    "<|channel>",
+    "<channel|>",
+}
+
+_TOOL_PARSER_TOKEN_ATTRIBUTES = (
+    "stc_token",
+    "etc_token",
+    "escape_token",
+    "soc_token",
+    "eoc_token",
+)
+
+
+def _token_text(token):
+    return getattr(token, "content", token)
+
+
+def _make_tool_parser_tokens_non_special(tokenizer):
+    for attr in _TOOL_PARSER_TOKEN_ATTRIBUTES:
+        if _token_text(getattr(tokenizer, attr, None)) in _TOOL_PARSER_TOKENS:
+            try:
+                setattr(tokenizer, attr, None)
+            except AttributeError:
+                pass
+
+    additional_tokens = getattr(tokenizer, "additional_special_tokens", None)
+    if not additional_tokens:
+        return
+    tokenizer.additional_special_tokens = [
+        token
+        for token in additional_tokens
+        if _token_text(token) not in _TOOL_PARSER_TOKENS
+    ]
+
 
 class DiffusionGemma4Processor(Gemma4Processor):
     model_type = "diffusion_gemma"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


lucasnewman (approved):
LGTM

PR #1483, opened 2026-07-02
fix #1481: quantized KV cache crashes in mask-length checks

### LINKED ISSUE #1481: gemma3: every request 500s when --kv-bits is set (AttributeError: 'tuple' object has no attribute 'shape' in sliding-window mask check)
### Bug

With `--kv-bits` set, the gemma3 family crashes on **every** chat completion — the sliding-window mask check in `models/gemma3/language.py` assumes `keys` is an array, but with KV quantization it's the `(packed, scales, biases)` tuple from `QuantizedKVCache`.

Reproduced identically on **v0.6.3 (pypi)** and **main @ `de6fa2d`**.

### Repro

```bash
python -m mlx_vlm.server \
  --model mlx-community/gemma-3-4b-it-8bit \
  --host 127.0.0.1 --port 8002 --kv-bits 8 --max-kv-size 65536
```

Any `/v1/chat/completions` request → HTTP 500:

```
  File ".../mlx_vlm/models/gemma3/language.py", line 232, in __call__
    h = layer(h, local_mask, c)
  File ".../mlx_vlm/models/gemma3/language.py", line 158, in __call__
    r = self.self_attn(self.input_layernorm(x), mask, cache)
  File ".../mlx_vlm/models/gemma3/language.py", line 84, in __call__
    if mask.shape[-1] != keys.shape[-2]:
                         ^^^^^^^^^^
```

Removing `--kv-bits` makes the same server/model work fine.

### Environment

Mac Studio M2 Ultra 192GB · macOS 15 · Python 3.14.3 · mlx 0.31.2 · mlx-lm 0.31.3 · mlx-vlm 0.6.3 and main @ `de6fa2d` (both affected)

### Notes

- `keys.shape[-2]` (the current KV length) is available for quantized caches as `cache.offset` (or from the packed component's shape); the mask-length check just needs a quantized-aware path.
- Other model families (e.g. mistral3/mistral4, qwen2_5_vl) run fine with `--kv-bits` on the same builds, so this looks gemma3-local, not a general quantized-KV problem.

Lazarus-931:
With --kv-bits, there is an array and tuple mixup. KVCache returns a tuple (packed_uint32, scales, biases) which doesn't support .shape[], so doing keys.shape[-2] on that result to get the KV sequence length causes AttributionError. 

Solution: a function to handle sequence length out of whichever representation it's handed, a mx.array, tuple or list.

All tests pass + new TestQuantizedKVCacheMask 

fixes #1481

### DIFF
--- a/mlx_vlm/__init__.py
+++ b/mlx_vlm/__init__.py
@@ -3,7 +3,6 @@ import os
 os.environ["TRANSFORMERS_NO_ADVISORY_WARNINGS"] = "1"
 os.environ["TRANSFORMERS_VERBOSITY"] = "error"
 
-
 from .convert import convert
 from .generate import (
     BatchResponse,
--- a/mlx_vlm/__init__.py
+++ b/mlx_vlm/__init__.py
@@ -3,11 +3,7 @@ import os
 os.environ["TRANSFORMERS_NO_ADVISORY_WARNINGS"] = "1"
 os.environ["TRANSFORMERS_VERBOSITY"] = "error"
 
-# transformers >= 5.13 tightened AutoTokenizer.register to require a config
-# *class* (it reads ``key.__module__``); mlx_lm still registers a custom
-# tokenizer by string model_type, which raises at import time. Tolerate string
-# keys here — before mlx_lm is imported below — so mlx-vlm works across the
-# transformers 5.5–5.13 range without pinning the dependency down.
+
 try:
     from transformers.models.auto import auto_factory as _hf_auto_factory
 
--- a/mlx_vlm/__init__.py
+++ b/mlx_vlm/__init__.py
@@ -4,21 +4,6 @@ os.environ["TRANSFORMERS_NO_ADVISORY_WARNINGS"] = "1"
 os.environ["TRANSFORMERS_VERBOSITY"] = "error"
 
 
-try:
-    from transformers.models.auto import auto_factory as _hf_auto_factory
-
-    _hf_orig_register = _hf_auto_factory._LazyAutoMapping.register
-
-    def _hf_safe_register(self, key, value, exist_ok=False):
-        if isinstance(key, str):
-            self._extra_content[key] = value
-            return
-        return _hf_orig_register(self, key, value, exist_ok=exist_ok)
-
-    _hf_auto_factory._LazyAutoMapping.register = _hf_safe_register
-except Exception:  # pragma: no cover - never block import on a shim failure
-    pass
-
 from .convert import convert
 from .generate import (
     BatchResponse,
--- a/mlx_vlm/__init__.py
+++ b/mlx_vlm/__init__.py
@@ -3,6 +3,26 @@ import os
 os.environ["TRANSFORMERS_NO_ADVISORY_WARNINGS"] = "1"
 os.environ["TRANSFORMERS_VERBOSITY"] = "error"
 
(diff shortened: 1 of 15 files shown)


mlx_vlm/tests/test_models.py
```diff
@@ -8233,3 +8233,53 @@ def test_hisa_batched_l_gt_1_matches_flat(self):
...
+
+        keys = mx.zeros((1, 2, 5, 64))
+        self.assertEqual(kv_sequence_length(keys), 5)
+        quantized = mx.quantize(keys, group_size=64, bits=8)
+        self.assertEqual(kv_sequence_length(quantized), 5)
+
+    def test_gemma3_generation_with_quantized_cache(self):
+        from mlx_vlm.models import gemma3
+        from mlx_vlm.models.cache import KVCache, QuantizedKVCache
+
```
Blaizzy: I would remove this test, as it's for only one model. 

The test above should catch changes in kv_sequence_length and the testing test_gemma3 should catch the rest.

Blaizzy (commented):
LGTM, just one nit

Blaizzy:
@Lazarus-931 Pre-commit is failing

Could you please install the hook so it automatically check it before any commit

https://pre-commit.com/#3-install-the-git-hook-scripts

Lazarus-931:
done!

An updated transformer (tdy) broke the CI, addressed as well.

mlx_vlm/__init__.py
```diff
@@ -3,6 +3,22 @@
 os.environ["TRANSFORMERS_NO_ADVISORY_WARNINGS"] = "1"
 os.environ["TRANSFORMERS_VERBOSITY"] = "error"
 
+
+try:
```
Blaizzy: What's this for ?

Lazarus-931: repo always uses recent hf transformer, which was updated today, CI broke b/c of this. 

this recent version, 5.13 changes _LazyAutoMapping.register to require a config class but mlx-vlm calls it with str, thus tests failing

so I monkeypatched transformers' tokenizer-registry register method to tolerate a string key

Blaizzy: Could you share a reproducible example for this issue so I can understand it better?

Lazarus-931: Sure!

```
python3 -m venv /tmp/repro
/tmp/repro/bin/pip install "mlx-lm==0.31.3" "transformers==5.13.0"
/tmp/repro/bin/python -c "import mlx_lm"
```
This will raise an error like:

```
AttributeError: 'str' object has no attribute '__module__'
mlx_lm/tokenizer_utils.py:505 → transformers/.../auto_factory.py:680
```

to further check, u can downgrade transformers and see it worked before it's upgrade

```
/tmp/repro/bin/pip install "transformers==5.12.1"
/tmp/repro/bin/python -c "import mlx_lm; print('imports fine on 5.12.1')"
```

mlx_vlm/__init__.py
```diff
@@ -3,6 +3,7 @@
 os.environ["TRANSFORMERS_NO_ADVISORY_WARNINGS"] = "1"
 os.environ["TRANSFORMERS_VERBOSITY"] = "error"
 
+
```
Blaizzy:
```suggestion
```

Blaizzy (approved):
LGTM

PR #1484, opened 2026-07-02
fix #1480

### LINKED ISSUE #1480: Deterministic permanent output collapse to `"!!!!…"` (token id 0) after ~70–90
consecutive `generate()` calls with a 4-bit Qwen3-VL model on Apple Silicon
## Summary

Running back-to-back single-image `mlx_vlm.generate()` calls in one long-lived
process, output **permanently** degenerates to a run of `!` characters (`!` is
token id 0 in the Qwen vocabulary, i.e. the logits have collapsed to zeros/NaN and
argmax returns token 0) after a near-deterministic number of calls (~70–90). Once
it starts, **every** subsequent generation in that process is degenerate — new
prompts, new images, and a freshly reset conversation do not recover it. Only a
process restart does. Under concurrent GPU load the same workload instead
hard-crashes with a Metal command-buffer fault, which suggests the silent
collapse is the non-fatal flavor of the same underlying Metal-state corruption.

## Environment

| | |
|---|---|
| mlx | 0.31.2 (latest PyPI at time of writing) |
| mlx-vlm | 0.6.3 (latest PyPI at time of writing) |
| Python | 3.12.13 |
| macOS | 26.5 (build 25F71) |
| Hardware | Apple M4 Pro, 48 GB unified memory |
| Model | Qwen3-VL (`model_type: qwen3_vl`, `Qwen3VLForConditionalGeneration`), MLX **4-bit** quantization (converted with `mlx_vlm.convert`) |

## Reproduction

Minimal reproducer (no server, no threads, one fixed synthetic 640×480 JPEG,
greedy decoding). This is a portable rewrite of the wrapper we used in
production diagnosis; the call path is identical (`load` →
`apply_chat_template` → `generate`):

```python
"""Hammer: back-to-back generate() until the output collapses to '!!!!'."""
import io, os, tempfile, time

from PIL import Image
import mlx.core as mx
from mlx_vlm import generate, load

MODEL = "path/to/qwen3-vl-4bit"  # our model: a 4-bit mlx_vlm.convert of Qwen3-VL

# Fixed synthetic "room" frame (gradient + blocks so it's not flat).
im = Image.new("RGB", (640, 480), (200, 190, 180))
for y in range(480):
    for x in range(0, 640, 8):
        im.putpixel((x, y), (x * 255 // 640, y * 255 // 480, 120))
fd, FRAME = tempfile.mkstemp(suffix=".jpg")
with os.fdopen(fd, "wb") as fh:
    buf = io.BytesIO(); im.save(buf, "JPEG", quality=85); fh.write(buf.getvalue())

print("mlx", mx.__version__)
config = load_config(MODEL)

prompt = "<0.0 seconds> You are watching a room. Stay silent unless something is wrong."
t0, first_bad = time.monotonic(), None
for i in range(1, 151):
    formatted = apply_chat_template(processor, config, prompt, num_images=1)
    out = generate(model, processor, formatted, image=[FRAME],
                   max_tokens=128, temperature=0.0, top_p=0.9, top_k=40,
                   verbose=False)
    text = getattr(out, "text", str(out))
    bad = text.strip().startswith("!!")
    if bad and first_bad is None:
        first_bad = i
        print(f"*** FIRST DEGENERATE at call {i}, t={time.monotonic()-t0:.0f}s ***")
    if i % 10 == 0 or bad:
        print(f"call {i:3d} t={time.monotonic()-t0:4.0f}s {'DEGEN' if bad else 'ok'} {text[:40]!r}")
print("first degenerate:", first_bad)
```

Observed with the original wrapper version of this script: **first degenerate
output at call 78 of 150**, and every call after it degenerate, matching the
production service (below).

## Expected

`generate()` output quality is independent of how many times it has been called
in the process; call 90 behaves like call 1.

## Actual

- Output is normal for ~70–90 calls, then collapses to `"!!!!!!…"` (token id 0
  repeated to `max_tokens`) and **never recovers** within the process.
- The onset is strikingly **count-based, not time- or content-based**: in a
  production service doing 1 image/frame decisions plus periodic short
  "describe" generations, collapse landed at ~70 decision calls (+~18 describe
  calls ≈ 88 total) after **every** fresh start — 7 of 9 consecutive runs, and
  our service logs show the model reloading and collapsing again on a ~5–6 min
  cycle for hours (14 consecutive episodes on 2026-07-02).
- Once collapsed, sending a brand-new conversation/prompt (we fired a full
  session reset at first `!!!!`) still returns degenerate output ⇒ it is not
  prompt/KV/session poisoning; the corruption is process-level.
- If a **second** process is hammering the GPU at the same time, instead of the
  silent collapse we get a hard abort:
  `libc++abi: terminating due to uncaught exception of type std::runtime_error:
  [METAL] Command buffer execution failed: Caused GPU Address Fault Error
  (kIOGPUCommandBufferCallbackErrorPageFault)` — and the co-resident service
  logged the matching
  `(00000005:kIOGPUCommandBufferCallbackErrorInnocentVictim)`.

## Ruled out (all tested explicitly, onset unchanged at ~70–90 calls)

- **Memory starvation / swap**: 48 GB machine, ~74 % free during episodes,
  `vm.swapusage` total 0 throughout.
- **MLX buffer-cache config**: `set_cache_limit` at 1 GB vs 6 GB; per-inference
  `clear_cache()` on vs off — no change in onset.
- **Wired memory**: `set_wired_limit` 14 GB vs not set at all — no change.
- **Image size / count**: 256 px vs 640 px inputs, 1 vs 2 images per call — no
  change.
- **Sampling**: temperature 0 (greedy) reproduces it.

## Possibly related

- ml-explore/mlx#3393 — quantized garbage output, GPU-config-specific Metal kernels
- ml-explore/mlx-lm#1332 — per-decode Metal buffer residency growth

## Why it matters

This runs an on-device safety monitor (fall detection for an elderly person).
A silent, permanent logits collapse after ~4 minutes of inference means any
long-lived mlx-vlm service on this configuration goes blind on a timer; we
currently mitigate with an external watchdog that restarts the process every
~6 minutes.

Happy to run instrumented builds / provide more logs.

Lazarus-931:
raised in #1480, the async eval runner, `mx.async_eval` whenever initiated by the `evaluate()` function would queue the pre-fill.

done continuously this can silently  read blank (all-zero) memory instead of real results, so argmax over all-zero scores always  picks token 0 which in qwen's vocab it's "!", hence the !!!!! the pr reported

fix is to just to switch the mx.async_eval with mx.eval, which actually only ever matters for decode, when cpu needs to do something after calling mx.async_eval, it's not needed in prefill. 

all tests pass & new TestGenerateStepPrefillEval included


Blaizzy:
I suspect the issue is not async_eval but rather something else

Check #1461

Blaizzy (approved):
🚀

PR #1486, opened 2026-07-03
Make Qwen MRoPE request-local under continuous batching

### LINKED ISSUE #1415: RFC: make MRoPE rope_deltas request-local under continuous batching

Blaizzy:
## Summary

- Pass Qwen `position_ids` and `rope_deltas` through request-owned prompt kwargs instead of mutating shared language model state.
- Keep Qwen `position_ids` in their raw model layouts: text-only requests can carry `(B, L)`, while multimodal MRoPE requests carry native `(3, B, L)`.
- Teach the batcher to split, pad, concatenate, and chunk native Qwen MRoPE `position_ids` on their correct batch and sequence axes.
- Merge mixed text/image Qwen batches by expanding text-only 2D `position_ids` into native MRoPE rows only when needed for a mixed batch.
- Drop `None` embedding fields before server/generation prompt kwargs assembly.

## Root Cause

Qwen VL MRoPE metadata is request-specific, but the previous path could store it on shared `language_model` fields. Under continuous batching, one request could overwrite or clear that state while another request was still decoding.

Once that state moved into request kwargs, the generic batcher still assumed sequence-aligned tensors were batch-major `(B, L, ...)`. Qwen multimodal MRoPE `position_ids` are native `(3, B, L)`, and text-only Qwen requests can be `(B, L)`. During live testing with `Qwen/Qwen3.5-35B-A3B`, mixed text/image batching exposed that mismatch as a rank/axis concatenate failure.

## Reproducible Example

Start the live server from the repo root:

```bash
PYTHONUNBUFFERED=1 TOKENIZERS_PARALLELISM=false \
  uv run mlx_vlm.server \
    --host 127.0.0.1 \
    --port 18080 \
    --model Qwen/Qwen3.5-35B-A3B \
    --max-tokens 32 \
    --log-level DEBUG
```

In another terminal, run overlapping text-only and image requests against the OpenAI-compatible endpoint:

```bash
uv run python - <<'PY'
import asyncio
import json
from pathlib import Path

import httpx

base_url = "http://127.0.0.1:18080/v1/chat/completions"
model = "Qwen/Qwen3.5-35B-A3B"
image = str((Path.cwd() / "examples/images/cats.jpg").resolve())

cases = [
    ("image-a", 0.0, [
        {"role": "user", "content": [
            {"type": "text", "text": "Describe this image in exactly five words."},
            {"type": "input_image", "image_url": image},
        ]},
    ], 10),
    ("text-a", 0.2, [
        {"role": "user", "content": "Reply with exactly these words: alpha beta gamma delta."},
    ], 8),
    ("image-b", 0.4, [
        {"role": "user", "content": [
            {"type": "text", "text": "What animal is visible? Answer in one word."},
            {"type": "input_image", "image_url": image},
        ]},
    ], 6),
    ("text-b", 0.6, [
        {"role": "user", "content": "Reply with exactly: green"},
    ], 4),
]

async def run_case(client, name, delay, messages, max_tokens):
    await asyncio.sleep(delay)
    payload = {
        "model": model,
        "messages": messages,
        "stream": True,
        "max_tokens": max_tokens,
        "temperature": 0.0,
        "enable_thinking": False,
        "stream_options": {"include_usage": True},
    }
    text = []
    usage = None
    async with client.stream("POST", base_url, json=payload, timeout=None) as resp:
        resp.raise_for_status()
        async for line in resp.aiter_lines():
            if not line.startswith("data: "):
                continue
            data = line[6:]
            if data == "[DONE]":
                break
            event = json.loads(data)
            if event.get("error"):
                raise RuntimeError(event["error"])
            usage = event.get("usage") or usage
            for choice in event.get("choices", []):
                chunk = (choice.get("delta") or {}).get("content")
                if chunk:
                    text.append(chunk)
    output = "".join(text)
    if not output or usage is None:
        raise RuntimeError(f"{name}: empty stream output={output!r} usage={usage!r}")
    print(name, output, usage)

async def main():
    async with httpx.AsyncClient(timeout=None) as client:
        await asyncio.gather(*(run_case(client, *case) for case in cases))

asyncio.run(main())
PY
```

Before this fix, the mixed continuous batch can fail in the generation thread with a `position_ids` concatenate error because text-only rows are rank 2 while multimodal MRoPE rows are rank 3. After this fix, all requests should stream content and usage successfully.

## Validation

- `python3 -m py_compile` on touched implementation and test files
- `git diff --check`
- Focused generation/server tests for prompt kwarg splitting, merging, chunked slicing, APC mixed prefill, and embedding field filtering
- Focused Qwen input-embedding tests for Qwen2-VL, Qwen2.5-VL, Qwen3-VL, Qwen3-VL-MoE, and Qwen3.5
- Focused Qwen decode tests for explicit `rope_deltas` kwargs
- Live server test with `Qwen/Qwen3.5-35B-A3B`: 6 overlapping streaming requests, 3 image and 3 text-only, all completed with content and usage after the batcher change

Fixes #1415

### DIFF
--- a/mlx_vlm/generate/ar.py
+++ b/mlx_vlm/generate/ar.py
@@ -549,7 +549,21 @@ _SEQUENCE_ALIGNED_PROMPT_KWARGS = {
 APC_PRIVATE_PROMPT_KEYS = ("_apc_tenant", "_apc_image_hash")
 
 
-def _prompt_kwarg_row(v: mx.array, row_idx: int, batch_size: int) -> mx.array:
+def _is_mrope_position_ids_prompt_kwarg(key: str, v: mx.array) -> bool:
+    return key == "position_ids" and v.ndim == 3 and v.shape[0] == 3
+
+
+def _prompt_kwarg_batch_size(key: str, v: mx.array) -> int:
+    if _is_mrope_position_ids_prompt_kwarg(key, v):
+        return v.shape[1]
+    return v.shape[0] if v.ndim > 0 else 0
+
+
+def _prompt_kwarg_row(key: str, v: mx.array, row_idx: int, batch_size: int) -> mx.array:
+    if _is_mrope_position_ids_prompt_kwarg(key, v):
+        if v.shape[1] == batch_size:
+            return v[:, row_idx : row_idx + 1, :]
+        return v[:, :1, :]
     if v.shape[0] == batch_size:
         return v[row_idx : row_idx + 1]
     return v[:1]
@@ -569,9 +583,9 @@ def _split_prompt_kwargs_per_row(prompt_kwargs: dict, batch_size: int) -> List[d
 
     rows = [{} for _ in range(batch_size)]
     for k, v in (prompt_kwargs or {}).items():
-        if isinstance(v, mx.array) and v.ndim > 0 and v.shape[0] >= 1:
+        if isinstance(v, mx.array) and _prompt_kwarg_batch_size(k, v) >= 1:
             for i in range(batch_size):
-                rows[i][k] = _prompt_kwarg_row(v, i, batch_size)
+                rows[i][k] = _prompt_kwarg_row(k, v, i, batch_size)
         else:
             for row in rows:
                 row[k] = v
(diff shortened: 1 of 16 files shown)


lucasnewman (approved):
LGTM

Lazarus-931:
awesome work @Blaizzy , lgtm!

Issue #1490, opened 2026-07-03
20260703 -- Diagnostics Report — 4 failure(s), 5 harness issue(s) (mlx-vlm 0.6.3+)


jrp2014:
<!-- markdownlint-disable MD013 -->

**Run summary:** 61 locally-cached VLM model(s) checked; 4 hard failure(s), 5 harness/integration issue(s), 0 preflight warning(s), 57 successful run(s).

`diffusiongemma` models now failing.

Test image: `cats.jpg` (0.2 MB).

---

## Issue Queue

Root-cause issue drafts are generated in
[issues/index.md](https://github.com/jrp2014/check_models/blob/main/src/output/issues/index.md).
Each row is intended to become one focused upstream GitHub issue.

<!-- markdownlint-disable MD060 -->

| Target                                         | Problem                                                                                          | Evidence Snapshot                                                                                                                                                                                                                         | Affected Models                                            | Issue Draft                                                                                                                                                              | Evidence Bundle                                                                                                                                                                                             | Fixed When                                                |
|------------------------------------------------|--------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------|
| `mlx`                                          | Weight/config mismatch during model load                                                         | Weight Mismatch \| phase model_load \| ValueError                                                                                                                                                                                         | 1: `mlx-community/LFM2.5-VL-1.6B-bf16`                     | [issue draft](https://github.com/jrp2014/check_models/blob/main/src/output/issues/issue_001_mlx_mlx-model-load-weight-mismatch_001.md)                                   | [repro JSON](https://github.com/jrp2014/check_models/blob/main/src/output/repro_bundles/20260703T125909Z_004_mlx-community_LFM2.5-VL-1.6B-bf16_MLX_MODEL_LOAD_WEIGHT_MISMATCH_7574b1189.json)               | Load/generation completes or fails with a narrower owner. |
[...]

PR #1493, opened 2026-07-03
fix(lfm2_vl): only create projector LayerNorm when projector_use_laye…

Lazarus-931:
All LFM2.5-VL checkpoints set projector_use_layernorm: false and ship no layer_norm weights, but the projector created the nn.LayerNorm module unconditionally, so every LFM2.5-VL repo failed to load with Missing 2 parameters: 

multi_modal_projector.layer_norm.bias/weight

This PR only creates the module when the flag is enabled, and drops stale layer_norm weights in sanitize so older conversions that still carry them keep loading. Old LFM2-VL models are unaffected since they default the flag to true and ship the weights. 

Verified that LFM2.5-VL 4bit and bf16 now load and generate correctly, LFM2-VL-450M-8bit still loads, and the full test_models.py suite passes 

Fixes the LFM2.5-VL item of #1490.

### DIFF
--- a/mlx_vlm/models/lfm2_vl/lfm2_vl.py
+++ b/mlx_vlm/models/lfm2_vl/lfm2_vl.py
@@ -16,7 +16,10 @@ class Lfm2VlMultiModalProjector(nn.Module):
         super().__init__()
         in_channels = config.vision_config.hidden_size * (config.downsample_factor**2)
         self.projector_use_layernorm = config.projector_use_layernorm
-        self.layer_norm = nn.LayerNorm(in_channels)
+        # LFM2.5-VL checkpoints set projector_use_layernorm=False and ship no
+        # layer_norm weights, so only create the module when it is enabled.
+        if self.projector_use_layernorm:
+            self.layer_norm = nn.LayerNorm(in_channels)
         self.linear_1 = nn.Linear(
             in_channels,
             config.projector_hidden_size,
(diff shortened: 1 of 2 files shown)


lucasnewman:
@Lazarus-931 I think this may be the opposite of https://github.com/Blaizzy/mlx-vlm/pull/1328, which will break the 2.5 model loading. Can you verify your fix against that model too?

Lazarus-931:
hi @lucasnewman yeah, verified and actually it's the model this pr fixes. 

LFM2.5-VL is what's broken, if u run this:
```
python -c "from mlx_vlm import load; load('mlx-community/LFM2.5-VL-1.6B-4bit')"
```
U get:

```
ValueError: Missing 2 parameters: multi_modal_projector.layer_norm.bias, multi_modal_projector.layer_norm.weight
```
None of the current LFM2.5-VL checkpoints ship multi_modal_projector.layer_norm.* weights, tested that this pr loads and generates with LFM2.5-VL 4bit/bf16 load and loads LFM2-VL-450M-8bit.

lucasnewman (approved):

mlx_vlm/models/lfm2_vl/lfm2_vl.py
```diff
@@ -16,7 +16,10 @@ def __init__(self, config: ModelConfig):
         super().__init__()
         in_channels = config.vision_config.hidden_size * (config.downsample_factor**2)
         self.projector_use_layernorm = config.projector_use_layernorm
-        self.layer_norm = nn.LayerNorm(in_channels)
+        # LFM2.5-VL checkpoints set projector_use_layernorm=False and ship no
+        # layer_norm weights, so only create the module when it is enabled.
```
Blaizzy:
```suggestion
```

mlx_vlm/models/lfm2_vl/lfm2_vl.py
```diff
@@ -224,4 +227,16 @@ def transform_key(key):
 
             return key
 
-        return {transform_key(k): v for k, v in weights.items()}
+        weights = {transform_key(k): v for k, v in weights.items()}
+
+        # Some older conversions include projector layer_norm weights even
+        # though projector_use_layernorm is disabled; drop them so they do
+        # not show up as unexpected parameters during strict loading.
```
Blaizzy:
```suggestion
```

Blaizzy (approved):
🚀

PR #1494, opened 2026-07-04
Consolidate caches + RoPE handling

lucasnewman:
Pull in the cache, RoPE, and switch-layer related dependencies to consolidate dispatch, which makes it a bit easier for us to support additional quantization formats / KV compression schemes and the like.


Blaizzy (approved):
LGTM, thanks!

PR #1495, opened 2026-07-04
make eos_token_id a config field so model loading works

Lazarus-931:
mlx-community/MolmoPoint-8B-fp16 failed to load with AttributeError: 
property 'eos_token_id' of 'ModelConfig' 

object has no setter, because molmo_point declared eos_token_id as a getter-only property while load_model tries to apply the repo config's value onto the model config. 

This PR replaces the property with a plain Optional[List[int]] field matching sibling models molmo and molmo2, so loading works and the config's 151645 now correctly reaches the stopping criteria and the server stop-token set instead of being permanently None. 

It also hardens apply_generation_config_defaults to skip read-only properties instead of crashing, so any future config with the same shape degrades gracefully, with a unit test covering the guard. 

Verified MolmoPoint-8B-fp16 checkpoint now loads with stopping criteria [151645], test_utils.py passes 32/32, and Qwen3-VL/LFM2 regression loads are unaffected.

Fixes the MolmoPoint issue of #1490.

### DIFF
--- a/mlx_vlm/models/molmo_point/config.py
+++ b/mlx_vlm/models/molmo_point/config.py
@@ -85,6 +85,7 @@ class ModelConfig(BaseModelConfig):
     subpatch_token_id: int = 151948
     location_token_id: int = 151949
     use_frame_special_tokens: bool = True
+    eos_token_id: Optional[List[int]] = None  # falls back to the tokenizer EOS
 
     # Point prediction config
     patch_location: Optional[str] = "3x3"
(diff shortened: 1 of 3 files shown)


mlx_vlm/utils.py
Blaizzy, line 1: This looks like a hallucination from the agent :)

mlx_vlm/models/molmo_point/config.py
```diff
@@ -147,7 +148,3 @@ def head_dim(self):
     @property
     def vocab_size(self):
         return self.text_config.vocab_size
-
-    @property
-    def eos_token_id(self):
-        return None  # Let the tokenizer handle EOS
```
Blaizzy, line 153: I wouldn't delete this

Blaizzy:
Fixed by https://github.com/Blaizzy/mlx-vlm/pull/1500

PR #1496, opened 2026-07-04
fix(utils): merge eos_token_id into model config instead of overwriti…

Lazarus-931:
For Phi-3.5, after a prompt, and response i'd keep printing <|end|><|endoftext|><|end|><|endoftext|>... until it hit the token limit.

Why: the model stops when it emits one of the stop tokens in its config. Phi-3.5 has three of them, but while loading the model we replace that list with the single value from generation_config.json, which is a token this model never emits. So nothing ever matched and generation ran until the cap.

Fix: add the value from generation_config.json to the existing stop tokens instead of replacing them. Sampling settings like temperature are untouched.

All tests green and same prompt now stops 

this was raised in #1490

### DIFF
--- a/mlx_vlm/tests/test_utils.py
+++ b/mlx_vlm/tests/test_utils.py
@@ -168,6 +168,29 @@ def test_apply_generation_config_defaults_preserves_model_config_signature():
     assert not hasattr(model_config, "max_new_tokens")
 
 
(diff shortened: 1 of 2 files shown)


Blaizzy:
Fixed by #1501

PR #1497, opened 2026-07-04
feat(utils): actionable error and [torch] extra for processors that n…

Lazarus-931:
Pixtral-family models (pixtral, Apriel, Devstral) fail to load on a default mlx-vlm install because transformers v5 only ships torch-backed image processors for them, surfacing as a raw ImportError from deep inside transformers with no hint about what to install.

Catch the backend ImportError in load_processor and re-raise with the exact install command, and add a `torch` extra so
`pip install mlx-vlm[torch]` pulls torch + torchvision. 

All test_utils.py tests pass, including a new one for this error path.

### DIFF
--- a/mlx_vlm/tests/test_utils.py
+++ b/mlx_vlm/tests/test_utils.py
@@ -755,6 +755,19 @@ def test_load_processor_propagates_auto_processor_errors():
             load_processor(Path("/tmp/model"), eos_token_ids=2)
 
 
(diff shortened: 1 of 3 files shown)


Blaizzy:
Thanks @Lazarus-931!

This solution is not optimal. We have migrated away from using torch for a long time now.

The solution here is to either add a custom processor or fix the existing custom processor.

Read more here: https://github.com/Blaizzy/mlx-vlm/pull/821

PR #1498, opened 2026-07-04
Always sanitize loaded weights and simplify audio preprocessing

Blaizzy:
## Summary

- Always run load-time weight sanitization instead of skipping safetensors files marked as MLX format.
- Move Qwen3 Omni weight sanitization into the model-owned path so checkpoint-specific remapping stays local to the model.
- Simplify shared audio preprocessing by removing the Qwen3 Omni special case and the global lossy-audio feature normalization heuristic.
- Add a regression test that preserves processor-produced audio features for lossy source filenames.
- Keep the dependency lock bounded to the supported Transformers range.

## Why

The previous sanitizer gate could bypass model-specific cleanup for checkpoints that still need remapping. The shared audio path also carried model-specific behavior and an extension-based normalization step that could silently alter feature distributions. Processors and model-local feature extractors should own those details.

## Validation

- `uv run --with pytest --with pytest-asyncio --with httpx --with fastapi --with starlette python -m pytest -q --tb=short --disable-warnings mlx_vlm/tests/test_processors.py::test_prepare_inputs_preserves_processor_audio_features_for_lossy_sources mlx_vlm/tests/test_server.py mlx_vlm/tests/test_server_audio.py`
- Result: `200 passed, 3 warnings in 2.01s`
- Commit hooks: `black`, `isort`, `autoflake`


Lazarus-931:
@Blaizzy  I think this led to #1504, i'll restore, we can add the fix in this branch then merge

Blaizzy:
Yes, but I fixed it already in another PR #1506 

It was a reasonable trade off

Blaizzy:
Please don't restore this

PR #1508, opened 2026-07-05
Unify diffusion generation path

Blaizzy:
## Summary
- Route Diffusion Gemma and masked-diffusion models through one diffusion stream adapter.
- Remove separate block/masked branching from CLI and server generation paths.
- Keep model-specific denoising behind the shared adapter while preserving block streaming.

## Validation
- `uv run python -m unittest mlx_vlm.tests.test_diffusion_models mlx_vlm.tests.test_diffusion_gemma`
- `uv run python -m py_compile mlx_vlm/generate/diffusion.py mlx_vlm/generate/dispatch.py mlx_vlm/server/generation.py mlx_vlm/tests/test_diffusion_models.py`
- `git diff --check`

### DIFF
--- a/mlx_vlm/generate/diffusion.py
+++ b/mlx_vlm/generate/diffusion.py
@@ -141,31 +141,34 @@ class _DiffusionRedrawer:
         self.rows = _terminal_rows_for_text(text)
 
 
-def _is_diffusion_config(config: Any) -> bool:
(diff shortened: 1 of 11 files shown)


lucasnewman (approved):
LGTM

Lazarus-931 (approved):
lgtm!

PR #1509, opened 2026-07-05
migration from mlx-lm dependency

Lazarus-931:
A draft compilation for mlx-lm dependency migration. 

This begins with a isolated change to the parser mods we import from mlx-lm for the current models.

Verified byte-identical to the mlx-lm parsers


Lazarus-931:
I think the next steps would be the trainer refactoring, but, was thinking we should overlap model components instead of isolated per-model self-containment, I think a perfect balance is reusing when perf isn't impacted and vice versa. So lib needs to design models around them being able to use same + diff fwd components. 

cc: @Blaizzy @lucasnewman

lucasnewman (approved):
LGTM

Issue #1482, opened 2026-07-02
Decode performance optimizations for MLX-VLM inference


dahai80:
## Decode performance optimizations for MLX-VLM inference

We have ported several decode-path optimizations from mlx-lm to the VLM text generation pipeline. These changes target the single-token decode step that dominates VLM inference time after the initial prefill.

### Proposed optimizations

1. **Port skip-logsumexp to VLM decode** — The same single-token optimization from mlx-lm applies to VLM text generation. Skipping the full logsumexp when only one token is needed avoids an expensive reduction over the full vocabulary each step.

2. **Port fused sampler to VLM** — Combine top-k / top-p / temperature scaling / argmax into a single Metal kernel instead of round-tripping through Python between each operation.

3. **FP8 KV cache for VLM** — Store the text decoder KV cache in FP8, halving attention memory bandwidth during decode. This is especially impactful for VLM models with long context windows (image tokens + conversation).

4. **mx.compile for VLM decode** — Apply `mx.compile` to the per-layer decode step in the VLM text generation loop, letting the MLX compiler fuse and optimize the attention + FFN graph for the single-token case.

5. **Python dispatch overhead reduction** — Reduce per-token Python overhead in the VLM decode loop (same optimizations as mlx-lm): minimize Python-to-C round trips, batch small operations, and avoid redundant tensor materialization.

6. **Mixed-bit quantization recipes for VLM** — Support `--quant-recipe` flag for VLM models, same as mlx-lm. This enables per-layer bit-width assignment (e.g. attention in 3-bit, FFN in 4-bit) that preserves quality while maximizing decode throughput.

### Benchmark results

Qwen3.6-27B on M2 Ultra 137GB, text decode tok/s:

| Configuration | tok/s | Speedup |
|---|---|---|
| mxfp8 baseline | 18.5 | — |
| mixed_3_4 (+96%) | 36.2 | +96% |
| quant2_all (+162%) | 48.5 | +162% |

### Implementation notes

- Optimizations 1–5 are runtime-only changes (no model format change required).
- Optimization 6 requires a new quantization pass but produces standard MLX model files that load without code changes.
- All optimizations are composable; the benchmark numbers above reflect the combined effect of 1–6.
- The VLM decode path is structurally identical to the mlx-lm decode path after the vision encoder prefill, so the port is mostly mechanical with VLM-specific adjustments for the image token prefix in the KV cache.

Lazarus-931:
[...]

PR #1512, opened 2026-07-06
feat: single-sequence greedy decode fast path (skip logsumexp)

dahai80:
Refs #1532.

## Summary

Skip `mx.logsumexp` and `mx.argmax` → `mx.argmax` only for the most common decode case: single-sequence greedy sampling with no logprobs requested.

This eliminates the most expensive per-step operation in local serving (1 user, 1 request, temperature=0).

## Changes

- `mlx_vlm/generate/ar.py`: Add fast path in `GenerationBatch._step()` before logsumexp computation

## Conditions (all must be true)

- `len(self.uids) == 1` — single active sequence
- No logits processors active
- `compute_logprobs == False` — user did not request logprobs
- `top_logprobs_k == 0` — no top-logprobs requested
- `greedy_sampling == True` — temperature=0 / top_k=1

## Performance

Correctness-preserving: `argmax(logits) == argmax(logits - logsumexp(logits))` since `logsumexp` is a per-row scalar constant, so greedy token selection is unaffected.

Benefit is **overlap-dependent**. We measured the analogous fast path in mlx-lm:

- **Batched mlx-lm path: ~0–5% wall-clock** because `mx.async_eval` already overlaps `logsumexp` with the next forward.
- **Non-batched (sequential) mlx-lm path: ~5–8%** — see ml-explore/mlx-lm#1479 for that measurement.

I have **not** measured mlx-vlm's single-sequence path specifically; depending on whether its generate loop uses `async_eval`, the benefit should fall in the 0–8% range above. The `len(self.uids) == 1` guard targets the case where async_eval overlap is least likely to mask the saving. Happy to run VLM-specific benchmarks and report concrete numbers on request.

## Testing

- Existing tests pass (no behavior change when conditions are not met)
- Fast path produces identical tokens to the full path (argmax is deterministic)


Blaizzy:
Thanks, please open a issue first

dahai80:
Thanks for the nudge to open an issue first — done: #1532 now tracks this optimization (focused issue, split out from the omnibus #1482). This PR's body now references it.

I also corrected the Performance section to be honest about what we've actually measured: the analogous fast path in mlx-lm is **overlap-dependent** — ~0–5% on the batched path (async_eval already overlaps logsumexp with the next forward) and ~5–8% on the non-batched path (see ml-explore/mlx-lm#1479). I have **not** measured mlx-vlm's single-sequence path specifically; the `len(self.uids) == 1` guard targets where async_eval overlap is least likely to mask the saving, and I'm happy to run VLM-specific benchmarks and report concrete numbers if useful.

Could a maintainer reopen this PR so it can be reviewed against #1532? (I don't have permission to reopen it myself.)

PR #1513, opened 2026-07-06
port  LFM2 into a shared components module and fix per-layer quantization overrides

Lazarus-931:
This follows up from mlx-lm migration #1509, where I:

1. Add mlx_vlm/models/components/,  a home for language-model internals shared by multiple model families, with components/lfm2.py as the first module. It replaces the two near-duplicate upstream files (mlx_lm.models.lfm2 and mlx_lm.models.lfm2_moe) with one implementation for Attention, ShortConv, and Lfm2Model.

2. Point lfm2_vl and lfm2_moe at the shared module

4. Fix an issue with loading  mlx-community/LFM2-8B-A1B-4bit where mixed-quant checkpoints store override paths without the language_model, where predicate never matched and loading failed with a shape error.

I compared the new component against mlx-lm's classes with shared weights (dense and MoE), was bit-exact through prefill and cached decode. A/B'd real checkpoints three dense VL models and the 8B MoE all produce token-identical greedy output vs main, and the previously-unloadable LFM2-8B-A1B-4bit now loads with logits exactly matching mlx-lm's loader. 167 model tests all pass, and both model dirs import cleanly with all mlx_lm imports blocked.

### DIFF
--- a/mlx_vlm/utils.py
+++ b/mlx_vlm/utils.py
@@ -637,9 +637,6 @@ python -m mlx_vlm.convert --hf-path <local_dir> --mlx-path <mlx_dir>
             # Handle custom per layer quantizations
             if p in config["quantization"]:
                 return config["quantization"][p]
-            stripped = p.removeprefix("language_model.")
-            if stripped != p and stripped in config["quantization"]:
-                return config["quantization"][stripped]
             if not hasattr(m, "to_quantized"):
                 return False
             # Skip layers not divisible by 64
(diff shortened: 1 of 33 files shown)


Blaizzy:
This is great, thanks! 

Now, let's make lfm2 a model type because we will use it to run LM only models. 


- models/lfm2/language.py, lfm2.py, config.py... ✅
- models/components/lfm2 ❌

Blaizzy:
This PR is a bit too large. I'd prefer to keep the scope focused on LFM2 and per-model cases, so it's easier to review and track down bugs.

mlx_vlm/utils.py
Blaizzy, line 1: Please revert this then we can merge

Blaizzy (commented):
One nit and we are good to merge after

Blaizzy (approved):
LGTM 🚀

PR #1517, opened 2026-07-06
Add deepseek_v3 and deepseek_v32 model types

Lazarus-931:
## Summary
- Add `models/deepseek_v3/` and `models/deepseek_v32/` as model types, ported from mlx-lm 0.31.3, and `models/mla.py` (`MultiLinear`) plus `models/pipeline.py` (`PipelineMixin`) as shared infrastructure alongside `switch_layers.py` and `rope_utils.py`.
- Rewire the four borrowers: `kimi_k25` and `glm_moe_dsa` import the model classes (ported verbatim, so glm's subclassing and unbound `sanitize`/`shard` calls work unchanged); `deepseek_v4` and `youtu_vl` import `MultiLinear`/`PipelineMixin`. No mlx-lm imports remain in the DeepSeek family.

## Validation
I compared the ported classes against mlx-lm's with shared weights: bit-exact through prefill and cached decode for deepseek_v3 (both `q_lora_rank` variants) and deepseek_v32 with the sparse indexer active, plus `MultiLinear` including its quantized form. A tiny deepseek_v3 checkpoint (`yujiepan/deepseek-v3-tiny-random`) loads through the new model type with logits exactly matching `mlx_lm.load`, which also covers the `kv_b_proj` split in sanitize. The distributed `pipeline()` and `shard()` paths are bit-exact against mlx-lm under `mlx.launch` with 2 and 3 ranks. Model tests pass including new entries for both types, and all six affected dirs import with mlx-lm blocked. Not verified at real scale: production checkpoints (V3/V3.2/K2.5 exceed my hardware) and the quantized `kv_b_proj` re-quantize branch — a smoke run from anyone with an M3 Ultra setup would be appreciated.

## Context
Same model-type pattern as #1513, applied per-model as requested. Independent of the other model-type PRs; can merge in any order.

### DIFF
--- a/mlx_vlm/models/deepseek_v3/__init__.py
+++ b/mlx_vlm/models/deepseek_v3/__init__.py
@@ -1,0 +1,4 @@
+from .config import ModelConfig
+from .deepseek_v3 import Model
+
+__all__ = ["Model", "ModelConfig"]
--- a/mlx_vlm/models/deepseek_v3/config.py
+++ b/mlx_vlm/models/deepseek_v3/config.py
@@ -1,0 +1,37 @@
+from dataclasses import dataclass
+from typing import Dict, Optional
+
+from ..base import BaseModelConfig
+
+
+@dataclass
+class ModelConfig(BaseModelConfig):
+    model_type: str = "deepseek_v3"
+    vocab_size: int = 102400
+    hidden_size: int = 4096
+    intermediate_size: int = 11008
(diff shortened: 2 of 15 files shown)


Blaizzy:
LGTM, but please fix the merge conflicts first 👌🏽

Blaizzy (approved):
LGTM 🚀

PR #1518, opened 2026-07-06
Add nemotron_h model type and remove mlx-lm from nemotron_h_nano_omni

Lazarus-931:
## Summary
- Add `models/nemotron_h/` as a model type (`config.py` / `language.py` / `nemotron_h.py`), ported from mlx-lm 0.31.3, and `models/ssm.py` (SSD-SSM ops including the fused Mamba2 decode kernel) as shared infrastructure for hybrid architectures.
- Rewire `nemotron_h_nano_omni` (language model, config, and its unbound `sanitize` calls) to the new home — no mlx-lm imports remain in the Nemotron family.

## Validation
I compared the port against mlx-lm's classes with shared weights across all four block types (Mamba2 mixer, attention, MLP, MoE): bit-exact through prefill, cached prefill, and decode steps that exercise the fused SSM Metal kernel, plus a direct comparison of the SSM update functions themselves. The two tiny-random NemotronH checkpoints on the Hub are malformed (they fail under mlx-lm with the same missing-config errors), so I generated a valid tiny checkpoint from mlx-lm weights and loaded it through both stacks: logits exactly match, covering the full config-parse/sanitize/load path. Model tests pass including a new `nemotron_h` entry, and both Nemotron dirs import with mlx-lm blocked. Not verified: production-size Nemotron-H checkpoints (4B+).

## Context
Same model-type pattern as #1513, applied per-model as requested. Independent of the other model-type PRs; can merge in any order.


Blaizzy:
LGTM, but first please fix merge conflicts

Blaizzy (approved):

Blaizzy:
Please resolve the conflicts then we can merge 👌🏽

Blaizzy (approved):

Issue #1515, opened 2026-07-06
APC exact-mode correctness tests for hybrid attention models


weklund:
@Blaizzy alright based on your comment https://github.com/Blaizzy/mlx-vlm/issues/1510#issuecomment-4887740773  I've been digging into the codebase trying to understand where I can actually be useful.

My goals:

  1. Bring value as quickly as possible
  2. Learn the codebase progressively by actually touching it
  3. Accomplish your comment: "truly need help is the APC, especially for hybrid attn models like Qwen3.5, Gemma 4 and more As well its efficiency across them"

I want to validate my understanding of the current state and propose a small first step. Happy to be corrected on any of this.

Hybrid models (Gemma 4 with RotatingKVCache, Qwen 3.5 with ArraysCache) get routed to APC's exact mode via model_apc_mode(). Exact mode stores and restores the full cache as a blob instead of block-level reuse. So this is for caches that can't be decomposed.

So if we can get per layer hybrid APC approach (is that this? https://github.com/Blaizzy/mlx-vlm/issues/1471) , then we can get block mode for the KVCache layers,  like exact snapshots only for the RotatingKVCache/ArraysCache layers.  Huge memory overhead improvement.

So if this is true, this is kind of a big architectural change to `apc.py`

I don't feel super comfortable making changes there until there are a few more tests on the exact mode path that assert full store/restore path.  If I'm understand I could introduce issues like these? #1295 and #1270 

So I'd like to propose end to end correctness tests for exact-mode path as a precursor to architectural changes for per layer hybrid APC

- Tiny random-weight Gemma 4 and Qwen 3.5 (same pattern as test_models.py, runs in CI)
- Full flow: prefill → store cache → restore → generate → assert output consistency
- Compare warm-to-warm (two restores produce identical output) since cold vs warm may differ due to batch invariance
- Parametrized so adding new hybrid models is a one-liner

Few things this gives us:

- Hopefully reproduce these reports that are happening  #1505, #1398, #1371 .  
- Validates recent work from @lucasnewman on #1494 across the cache layer. These tests confirm exact-mode APC still produces correct output
- Foundation and safety net for #1174  and #1471 
- Runs in existing CI
- No behavioral changes, so low risk on my end as I'm figuring out what I'm doing

If this is correct, few questions:

1. Any gotchas with Qwen 3.5's RoPE state after exact-mode restore I should think about?
[...]

Issue #1295, opened 2026-06-05
v0.6.0 APC (Adaptive Prefix Cache) extremely low hit rate on Qwen3.6, works great on DeepSeek


NeoInBJ:
## Describe the issue
After upgrading mlx-vlm from **0.5 to 0.6**, the **APC cache hit rate dropped extremely low on Qwen3.6** models.

- **0.5.x**: APC works perfectly, high hit rate, stable speed(prompt_tps > 23000).
- **0.6.0**: APC seems only work for the first stored apc_cache_block, hit rate displays 1.0, but 'matched_tokens' only 258 (my apc_block_size is setting to 256)

- **DeepSeek-Coder-V2-Lite-Instruct-4bit-mlx**: APC works **fine** in 0.6.0.

Only Qwen3.6 is affected.

## Environment
- mlx-vlm: 0.6.0
- Model: Qwen3.6 (Qwen3.6-35B-A3B-mxfp4, Qwen3.6-35B-A3B-4bit, and others I have)
- Device: Apple Silicon M2 Max 64GB mem

## Key APC stats (Qwen3.6)
First query:
`matched_tokens: 100`

Second query (same prefix):
`matched_tokens: 358`
`token_hit_rate: 1.0`

Third query (same prefix):
`matched_tokens: 616`
`token_hit_rate: 1.0`

DeepSeek with same prompt, same params:
first round:
`matched_tokens: 0`
`token_hit_rate: 0.0`
second round:
`matched_tokens: 2048`
`token_hit_rate: 0.33`

## My guess
I think the **new APC block alignment / KV indexing logic in 0.6.0** does not match Qwen3.6's KV structure or attention layout.

0.5 used flexible block splitting and worked correctly.
0.6 strict block padding breaks Qwen3.6 prefix matching.

Please fix APC compatibility for Qwen3.6.
Thanks!

lucasnewman:
Thanks for reporting! This was fixed in https://github.com/Blaizzy/mlx-vlm/pull/1291

NeoInBJ:
I downloaded newest main branch， but the problem still the same. How do I use your fix? with any parameters or environment parameters to enable your fix?

NeoInBJ:
> Thanks for reporting! This was fixed in [#1291](https://github.com/Blaizzy/mlx-vlm/pull/1291)

 downloaded newest main branch， but the problem still the same. How do I use your fix? with any parameters or environment parameters to enable your fix?

PR #1291, opened 2026-06-04
Fix APC for single requests

lucasnewman:
Based on https://github.com/Blaizzy/mlx-vlm/pull/1287 but this is a more surgical fix for the issue. I ran a test with Qwen3.5-35B-A3B-4bit across 1k -> 128k using Ivan's contexts and verified APC is being used correctly in both single and batched requests:

```
[4k] chars=16149
  PASS single: tokens=4177 uncached_prefill_s=1.71 cached_prefill_s=0.19 speedup=9.12x delta_s=-1.52 seed_stores=2 hits=2 matched_tokens=3907 cached_progress=3907 uncached_matched=0
  PASS batch2: tokens=8358 uncached_prefill_s=2.24 cached_prefill_s=0.26 speedup=8.58x delta_s=-1.98 seed_stores=4 hits=4 matched_tokens=7818 cached_progress=7818 uncached_matched=0
Downloading 8k: https://raw.githubusercontent.com/ivanfioravanti/llm_context_benchmarks/refs/heads/master/8k.txt

[8k] chars=32649
  PASS single: tokens=8248 uncached_prefill_s=3.19 cached_prefill_s=0.26 speedup=12.31x delta_s=-2.93 seed_stores=2 hits=2 matched_tokens=7978 cached_progress=7978 uncached_matched=0
  PASS batch2: tokens=16500 uncached_prefill_s=4.63 cached_prefill_s=0.30 speedup=15.41x delta_s=-4.33 seed_stores=4 hits=4 matched_tokens=15960 cached_progress=15960 uncached_matched=0
Downloading 16k: https://raw.githubusercontent.com/ivanfioravanti/llm_context_benchmarks/refs/heads/master/16k.txt

[16k] chars=66197
  PASS single: tokens=16429 uncached_prefill_s=6.96 cached_prefill_s=0.24 speedup=29.06x delta_s=-6.72 seed_stores=2 hits=2 matched_tokens=16159 cached_progress=16159 uncached_matched=0
  PASS batch2: tokens=32862 uncached_prefill_s=10.82 cached_prefill_s=0.36 speedup=30.21x delta_s=-10.46 seed_stores=4 hits=4 matched_tokens=32322 cached_progress=32322 uncached_matched=0
Downloading 32k: https://raw.githubusercontent.com/ivanfioravanti/llm_context_benchmarks/refs/heads/master/32k.txt

[32k] chars=133813
[...]

PR #1519, opened 2026-07-06
Add end-to-end APC exact-mode correctness tests for hybrid models

weklund:
Validates that store_exact_cache/lookup_exact_cache produces correct, deterministic cache state for models with mixed cache types (Gemma 4 RotatingKVCache + KVCache, Qwen 3.5 ArraysCache + KVCache).

Tests cover:
- Mode detection (hybrid models route to exact mode)
- Store/restore round-trip (non-None, correct prefix_len)
- Warm-to-warm consistency (multiple restores produce identical tensors)
- No corruption (generation from a restore doesn't mutate stored entry)
- Multi-token generation from restored cache (8 tokens, no crashes)
- Negative case (wrong prefix misses correctly)
- Longer prefix (64 tokens)
- Layer type classification (block-eligible vs exact-only per layer)
- Clone preserves layer count and types
- Block-eligible layers contain extractable K/V data after prefill

Uses tiny random-weight model instances — no downloads needed, runs in CI on macos-14 with Metal in < 2s.

  Validation
  - `uv run --with pytest pytest mlx_vlm/tests/test_apc_exact_mode.py -q` 
  - `uv run --with pytest pytest mlx_vlm/tests/test_apc.py -q` (no regressions)

Ref: https://github.com/Blaizzy/mlx-vlm/issues/1515

### DIFF
--- a/mlx_vlm/tests/test_apc_exact_mode.py
+++ b/mlx_vlm/tests/test_apc_exact_mode.py
@@ -259,11 +259,11 @@ def test_apc_exact_restore_does_not_corrupt_stored_entry(model_factory):
     apc = APCManager(num_blocks=64, block_size=16)
     apc.store_exact_cache(prefix_tokens, cache)
 
-    # First restore + heavy generation (should not corrupt the stored entry)
+    # First restore + generation (should not corrupt the stored entry)
     restored_first, _ = apc.lookup_exact_cache(full_tokens)
     lm(suffix, cache=restored_first)
     mx.eval([c for c in restored_first if c is not None])
-    _generate_greedy(lm, mx.array([[42]]), restored_first, 20)
+    _generate_greedy(lm, mx.array([[42]]), restored_first, 3)
 
     # Second and third restores: compare the raw cache tensors to verify
     # the stored entry itself was not mutated by the generation above.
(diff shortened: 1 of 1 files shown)


weklund:
Did some mutation testing to see what these tests actually catch and where the gaps are.

Introduced a bug where _clone_cache_entry_for_apc skips copying keys/values for RotatingKVCache layers (returns an empty cache shell). Two tests immediately fail:
 - test_apc_exact_restored_cache_generates_tokens[gemma4] crashes with ValueError: Shapes (17,33) and (1,2,17,17) cannot be broadcast
 - test_apc_exact_restore_does_not_corrupt_stored_entry[gemma4] fails on tensor comparison

So data-loss bugs in the clone path are caught.

Removed _idx copy from RotatingKVCache clone (circular buffer position). All tests still pass. The tiny model uses sliding_window=32 and our prefix is 16 tokens.  the buffer never wraps, so _idx stays 0 whether copied or not. A prefix longer than the sliding window would exercise this.  I can add if we think this is valuable or too pedantic.

Looked at #1270 and #1295 the actual APC bugs users reported on Qwen models:

- #1270: stores: 0 — the store path never fires at all (routing/gating issue in the server layer)
- #1295: hit rate dropped after 0.5→0.6 because MRoPE position ID changes broke the hash chain for Qwen models

https://github.com/Blaizzy/mlx-vlm/pull/1291 looks like a great fix for that instance of it.  

I guess my super dumb question is, do these classes of bugs still pop up from either regressions or new models?  🤔

Lazarus-931:
hi @weklund, thanks for the pr! yes, seems like the idea of an isolated test for hybrid models seems doable. 

I think we 're doing some restructure of tests, so we'll keep this open for now.

Other than that, the CI fails because of pre-commit not being setup, follow here: https://pre-commit.com/ 

cc: @Blaizzy

weklund:
Yep should be green now 🙏, I realized the pre commit after your run yesterday. These was another failure on GPU hanging so reduced that test.

Blaizzy:
Thanks a lot and well done @weklund!

That was really fast  🚀

Running the CI and merging after it clears

Blaizzy (approved):
🚀

weklund:
Yay!  TY!  There's so many fun problems to solve it's hard to know where to put my free cycles 😁 

So ty for telling me to look at APC

Next CI workflow should pass

weklund:
Bah sorry new machine I haven't added signing keys yet, h/o lemme fix

weklund:
@Lazarus-931 @Blaizzy sorry one more workflow run should get this green to merge :)

PR #1522, opened 2026-07-07
Optimize Qwen3 Omni audio hidden-state capture

sxy-trans-n:
## Summary

This PR avoids a redundant Thinker full forward in the Qwen3-Omni audio-output path.

Previously, `generate(..., return_audio=True)` and `generate_stream(...)` first generated the Thinker token sequence, then ran another full Thinker forward over the completed sequence only to recover the Talker accept-layer hidden states. This change captures those hidden states during Thinker generation and passes them directly to the Talker.

To keep the decode path lightweight, this also adds a narrow `output_hidden_state_idx` option for Qwen3-Omni so callers can request a single hidden layer instead of collecting the full hidden-state stack.

## Changes

- Capture Talker accept-layer hidden states during Thinker generation.
- Reuse the captured hidden states in Qwen3-Omni audio generation and streaming.
- Add `output_hidden_state_idx` for single-layer hidden-state capture.
- Add a unit test comparing the new incremental hidden-state path with the previous full-forward extraction path.

## Benchmark

Model: `mlx-community/Qwen3-Omni-30B-A3B-Instruct-4bit`  
Mode: text input → audio output, greedy Thinker/Talker settings.

Correctness: old/new generated token sequences, decoded text, audio shapes, and audio hashes are identical.

### End-to-end text→audio

| case | input len | old mean | new mean | speedup |
|---|---:|---:|---:|---:|
| short | 20 | 2.722s | 2.614s | 1.041x |
| medium | 236 | 4.283s | 4.093s | 1.046x |
| long | 1216 | 8.127s | 7.101s | 1.144x |

### Thinker hidden extraction only

| case | input len | old mean | new mean | speedup |
|---|---:|---:|---:|---:|
| short | 20 | 0.270s | 0.190s | 1.422x |
| medium | 236 | 0.912s | 0.616s | 1.481x |
| long | 1216 | 2.324s | 1.369s | 1.698x |

The end-to-end gain is smaller because Talker and Code2Wav still account for a significant part of total generation time.


mlx_vlm/models/qwen3_omni_moe/qwen3_omni_moe.py
```diff
@@ -3,6 +3,8 @@
 import mlx.core as mx
 import mlx.nn as nn
 import numpy as np
+from mlx_lm.models import cache as kv_cache
+from mlx_lm.sample_utils import make_logits_processors, make_sampler
```
lucasnewman: Can you use `mlx_vlm.models.cache` and `mlx_vlm.sample_utils` here instead of `mlx-lm`?

lucasnewman:
@sxy-trans-n Very nice! See the one comment about the imports but otherwise we can merge.

sxy-trans-n:
@lucasnewman 
I pushed a follow-up commit that switches the Qwen3-Omni generation imports from `mlx_lm` to the vendored `mlx_vlm` modules:

- `mlx_lm.models.cache` -> `mlx_vlm.models.cache`
- `mlx_lm.sample_utils` -> `mlx_vlm.sample_utils`

I also re-ran the focused unit test successfully:

```bash
uv run python -m unittest mlx_vlm.tests.test_qwen3_omni_moe
``` 

One note: after main was merged into this PR branch, I can no longer run the real Qwen3-Omni audio e2e smoke test locally because model loading now fails before generation:
ValueError: Expected shape (480, 3, 3, 1) but received shape (480, 3, 1, 3)
for parameter thinker.audio_tower.conv2d1.weight
This appears unrelated to this PR’s hidden-state capture change. It looks introduced by b93707d / #1498 (Always sanitize loaded weights...), which removed the previous MLX-format metadata guard and now always sanitizes loaded weights. For the MLX-format Qwen3-Omni checkpoint I tested, thinker.audio_tower.conv2d1.weight is already stored as (480, 3, 3, 1), so the new unconditional audio-tower sanitize path appears to transpose it again to (480, 3, 1, 3).

lucasnewman:
> @lucasnewman I pushed a follow-up commit that switches the Qwen3-Omni generation imports from `mlx_lm` to the vendored `mlx_vlm` modules:
> 
> * `mlx_lm.models.cache` -> `mlx_vlm.models.cache`
> * `mlx_lm.sample_utils` -> `mlx_vlm.sample_utils`
> 
> I also re-ran the focused unit test successfully:
> 
> ```shell
> uv run python -m unittest mlx_vlm.tests.test_qwen3_omni_moe
> ```
> 
> One note: after main was merged into this PR branch, I can no longer run the real Qwen3-Omni audio e2e smoke test locally because model loading now fails before generation: ValueError: Expected shape (480, 3, 3, 1) but received shape (480, 3, 1, 3) for parameter thinker.audio_tower.conv2d1.weight This appears unrelated to this PR’s hidden-state capture change. It looks introduced by [b93707d](https://github.com/Blaizzy/mlx-vlm/commit/b93707d8bf67aec4c5e85e003e18ddee8a28f63d) / #1498 (Always sanitize loaded weights...), which removed the previous MLX-format metadata guard and now always sanitizes loaded weights. For the MLX-format Qwen3-Omni checkpoint I tested, thinker.audio_tower.conv2d1.weight is already stored as (480, 3, 3, 1), so the new unconditional audio-tower sanitize path appears to transpose it again to (480, 3, 1, 3).

Thanks! I'll fix that in a follow up in a moment.

lucasnewman (approved):

lucasnewman:
@sxy-trans-n Can you sign your commits so we can clear the merge requirements? The follow up is https://github.com/Blaizzy/mlx-vlm/pull/1541

sxy-trans-n:
@lucasnewman 
Thanks, done.

I re-signed the two PR commits with SSH commit signing and force-pushed the branch with `--force-with-lease`.

I also updated the branch with latest `main` and re-ran the focused test:

```bash
.venv/bin/python -m unittest mlx_vlm.tests.test_qwen3_omni_moe
```

It passes.

I also re-ran a real Qwen3-Omni audio e2e smoke test against the local `mlx-community/Qwen3-Omni-30B-A3B-Instruct-4bit` checkpoint. The latest `main` sanitize fix resolves the previous loader error, and the audio path now runs end-to-end:

```text
loaded has_talker=True
sequences_shape=(1, 25)
wav_shape=(1, 1, 184320)
peak_memory_gb=23.705
```

Issue #1526, opened 2026-07-07
Regression in 0.6.4: Qwen3.6 generates garbage and Gemma4 processor loading falls through to Transformers


ft-ioxcs:
## Summary

`mlx-vlm==0.6.4` appears to regress text generation / processor handling for at least two MLX community models that work under `mlx-vlm==0.6.3` with the same environment:

- `mlx-community/Qwen3.6-35B-A3B-4bit`
- `mlx-community/gemma-4-12B-it-qat-4bit`

The Qwen model generates valid text through `mlx_lm.generate`, but `mlx_vlm.generate` on `mlx-vlm==0.6.4` emits garbage tokens and can crash in streaming detokenization. The Gemma model fails during processor loading because the custom MLX-VLM processor path appears to fall through to upstream Transformers, which then requires a missing `video_preprocessor_config.json`.

Downgrading only `mlx-vlm` from `0.6.4` to `0.6.3` fixes both models.

## Environment

Hardware / OS:

- Apple Silicon Mac
- macOS 26.5.2 arm64
- Python 3.12.13

Python environment packages:

```text
mlx==0.31.2
mlx-lm==0.31.3
mlx-vlm==0.6.4          # broken
mlx-audio==0.4.4
transformers==5.12.1
datasets==5.0.0
fsspec==2026.4.0
torch==2.12.1
torchvision==0.27.1
```

`pip check` reported no broken requirements.

The working rollback is:

```text
mlx==0.31.2
mlx-lm==0.31.3
mlx-vlm==0.6.3          # working
mlx-audio==0.4.4
transformers==5.12.1
datasets==5.0.0
fsspec==2026.4.0
torch==2.12.1
torchvision==0.27.1
```

## Reproduction 1: Qwen3.6 works with mlx-lm

This command works and produces coherent output:

```bash
mlx_lm.generate \
  --model mlx-community/Qwen3.6-35B-A3B-4bit \
  --prompt "Hi" \
  --max-tokens 1024 \
  --temp 0.6 \
  --top-p 0.95 \
  --min-p 0.05 \
  --top-k 20 \
  --verbose True
```

Observed output excerpt:

```text
Hi! How can I help you today?
Prompt: 11 tokens
Generation: 223 tokens
Peak memory: 19.608 GB
```

This suggests the model cache, MLX core, and basic tokenizer/model assets are usable.

## Reproduction 2: Qwen3.6 broken with mlx-vlm 0.6.4

[...]

Issue #1505, opened 2026-07-04
04072026 -- Diagnostics Report — 3 failure(s), 9 harness issue(s) (mlx-vlm 0.6.4)


jrp2014:
**Run summary:** 61 locally-cached VLM model(s) checked; 3 hard failure(s), 9 harness/integration issue(s), 0 preflight warning(s), 58 successful run(s).

Test image: `cats.jpg` (0.2 MB).

---

## Issue Queue

Root-cause issue drafts are generated in
[issues/index.md](https://github.com/jrp2014/check_models/blob/main/src/output/issues/index.md).
Each row is intended to become one focused upstream GitHub issue.

| Target                                                   | Problem                                               | Evidence Snapshot                                                                                                                                                                  | Affected Models                                            | Issue Draft                                                                                                                                                              | Evidence Bundle                                                                                                                                                                                             | Fixed When                                                |
|----------------------------------------------------------|-------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------|
| `mlx`                                                    | Weight/config mismatch during model load              | Model Error \| phase model_load \| ValueError                                                                                                                                      | 1: `mlx-community/MiniCPM-V-4.6-8bit`                      | [issue draft](https://github.com/jrp2014/check_models/blob/main/src/output/issues/issue_001_mlx_mlx-model-load-model_001.md)                                             | [repro JSON](https://github.com/jrp2014/check_models/blob/main/src/output/repro_bundles/20260704T185505Z_004_mlx-community_MiniCPM-V-4.6-8bit_MLX_MODEL_LOAD_MODEL_b4b095ff6a33.json)                       | Load/generation completes or fails with a narrower owner. |
[...]

PR #1527, opened 2026-07-07
Fix MRoPE decode positions in plain generate for Qwen VLMs

Lazarus-931:
Since #1486, all Qwen VLMs generate broken output with plain generate() like "This is is is is…" at temperature 0 (see #1505 and #1526). 

The problem: generate() only passes rope_deltas with the first (prefill) call, and decode steps pass nothing. The model used to remember this value internally, but #1486 removed that, so during decode it recomputes positions from a single token and gets them wrong, and every token after that is generated at the wrong position. 

when `rope_delta` is being passed, if we save on the model, decode can keep using it. Everything else is unaffected, and all verifications show models produce exactly the same otuput as 0.6.3 version, reg. tests fail on main and pass with this fix. 

But, Qwen3.5/3.6 checkpoints have a second, unrelated bug that will be fixed in a upcoming pr, both needs this fix.

### DIFF
--- a/mlx_vlm/models/qwen2_5_vl/language.py
+++ b/mlx_vlm/models/qwen2_5_vl/language.py
@@ -431,6 +431,14 @@ class LanguageModel(nn.Module):
             self._rope_deltas = None
             self._position_ids = None
 
+        # Plain `generate()` passes `rope_deltas` (computed by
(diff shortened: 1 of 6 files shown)


Blaizzy (approved):
LGTM!

Issue #1521, opened 2026-07-07
0.6.4 regression: MLX-format Qwen3.6-27B-MLX-4bit generates corrupted text because sanitize_weights runs on already-converted MLX safetensors


fidodido587:
Title: 0.6.4 regression: MLX-format Qwen3.6-27B-MLX-4bit generates corrupted text because sanitize_weights runs on already-converted MLX safetensors

Summary

mlx-vlm 0.6.4 appears to regress MLX-format text-only Qwen models. With Qwen3.6-27B-MLX-4bit, generation produces corrupted/gibberish output and can hit a UnicodeDecodeError in tokenizer_utils.py.

The same model works correctly on mlx-vlm 0.6.3. The tokenizer encode/decode output is identical between 0.6.3 and 0.6.4.

A targeted local patch that detects safetensors metadata format == "mlx" and skips sanitize_weights(...) for already-converted MLX safetensors fixes the issue while preserving 0.6.4 behavior for non-MLX-format weights.

Environment

* macOS 26.5.1
* Python: 3.11.15
* mlx-vlm: 0.6.4
* mlx: 0.31.2
* mlx-lm: 0.31.3
* transformers: 5.12.1
* Model: Qwen3.6-27B-MLX-4bit
* Model config:
    * model_type = qwen3_5
    * architectures = ['Qwen3_5ForConditionalGeneration']
    * quantization = {'group_size': 64, 'bits': 4, 'mode': 'affine'}
    * eos_token_id = [248046, 248044]
    * tokenizer_class = TokenizersBackend

Minimal reproduction

mlx_vlm.generate \
  --model "/path/to/Qwen3.6-27B-MLX-4bit" \
  --prompt "Say exactly: hello" \
  --max-tokens 20 \
  --temperature 0 \
  --no-verbose

Expected output

hello

Actual output on clean mlx-vlm 0.6.4

Examples observed:

Bn
)"`
{ 、 “d
-h 
或者_cut

or:

B**subst02פת
....
by

In some runs, CLI crashes with:

UnicodeDecodeError: 'utf-8' codec can't decode byte 0xac in position 4: invalid start byte

Traceback points to:

mlx_vlm/tokenizer_utils.py", line 237, in add_token
    ).decode("utf-8")

Evidence that this is not tokenizer/config/Transformers

The following were checked:

* config.json, tokenizer_config.json, and generation_config.json are identical for both tests.
[...]

PR #1506, opened 2026-07-04
Fix MiniCPM-V 4.6 weight sanitization

Blaizzy:
## Summary

Fixes #1504.

MiniCPM-V 4.6 checkpoint loading now handles both current supported checkpoint layouts:

- `mlx-community/MiniCPM-V-4.6-8bit`, which stores already-nested direct language weights such as `language_model.model.*`.
- `openbmb/MiniCPM-V-4.6`, which stores HF-style weights under `model.language_model.*` and offset-style RMSNorm weights.

The sanitizer now preserves already-nested language model keys instead of rewriting them into non-existent `language_model.model.model.*` parameters. It also applies the `+1.0` norm offset only for HF-style `model.language_model.*` source weights, while leaving MLX-converted direct norm weights unchanged.

## Root Cause

The MiniCPM-V 4.6 sanitizer had two incompatible assumptions:

- It unconditionally rewrote every `language_model.*` key to `language_model.model.*`, double-prefixing current MLX-community checkpoint keys.
- Its norm handling needed to distinguish source formats: OpenBMB HF-style weights need the offset, but MLX-community converted weights already include direct norm values.

Without the prefix fix, strict loading failed. Without source-aware norm handling, generation either produced nonsense for the 8-bit checkpoint or blank output for the OpenBMB checkpoint.

## Validation

- `uv run mlx_vlm.generate --model mlx-community/MiniCPM-V-4.6-8bit --prompt "Hi" --max-tokens 100`
  - Output: `Hello! How can I help you today?`
- `uv run mlx_vlm.generate --model openbmb/MiniCPM-V-4.6 --prompt "Hi" --max-tokens 100`
  - Output: `Hello! How can I help you today?`
- `uv run python -m unittest mlx_vlm.tests.test_models.TestMiniCPMV4_6`
- Commit hook checks: `black`, `isort`, `autoflake`

PR #1528, opened 2026-07-07
Skip norm weight shift when loading MLX-format Qwen3.5 checkpoints

Lazarus-931:
Qwen3.5/Qwen3.5-MoE sanitize() adds +1.0 to 1-D norm weights to convert HF zero-centered RMSNorm gammas to the MLX convention. Since sanitize now runs unconditionally on load, MLX-format checkpoints (which already store shifted values) get shifted a second time, corrupting every norm and producing garbage output (#1521, #1526).

By detecting the mlx safetensor format in load_model, we can skip the norm shift for checkpoints, so both the hf checkpoints and convert path are unchanged. 

Verified models now work & specific weights match 0.6.3.

Also added a regression test that covers shift-for-HF / preserve-for-MLX for both model types, that would have failed in main but passes here. 

All tests green!


Blaizzy:
Thanks @Lazarus-931!

I don't think this is quite the right solution here though. Take a look at #1506, which addresses this differently:

https://github.com/Blaizzy/mlx-vlm/pull/1506

Lazarus-931:
Good call, thanks — reworked to follow the #1506 pattern. The shift is now gated on the original key namespace inside sanitize() itself: converted checkpoints store keys under `language_model.*` so they're left untouched, while source-format keys (`model.language_model.*`, and bare `model.*` for text-only checkpoints) still get the +1.0. The utils.py metadata detection is gone; the change is fully contained in the two model files.

Re-verified: norm values from `mlx-community/Qwen3.5-0.8B-MLX-8bit` load identical to 0.6.3 and greedy output matches character-for-character; regression tests cover source-format (VL + text-only) and converted namespaces for both model types.

Blaizzy:
Awesome, LGTM!

Could you please resolve the conflicts before we merge

weklund:
Nice fix!   The key-prefix is pretty clean

1 question that came up when I was looking at the bug: does the language_model. prefix distinguish MLX-format from HF-format across all checkpoint layouts? I was looking at Nemotron (#1367) where MLX-format safetensors store un remapped keys since sanitize_key hasn't run yet at that point, so the original key prefix might not reflect whether the norm offset was already applied.  It uses `encoder.*` keys there.

I've been experimenting with a value-based guard (`value.mean().item() < 0.5`) that makes the decision based on whether the values themselves look pre-offset vs post-offset. It's more expensive (one mean per norm weight during load) but works regardless of key naming convention. Would that be worth considering as a complement or alternative here?

Here's the idea:

```
  if any(key.endswith(sfx) for sfx in norm_keys):
      if value.ndim == 1 and value.mean().item() < 0.5:
          value += 1.0
```

We look at what the weight actually contains rather than what the key is named. HF norms average around 0 (range roughly -0.3 to +0.3). MLX norms average around 1.0 (range roughly 0.7 to 1.3). The threshold of 0.5 sits between them.

Now I will very much admit that, kind of making an assumption here, that the HF norm weights will never go above that threshold and using a heuristic, haven't validating against a corpus of model checkpoints so I should probably do that first.  Could heavily fine-tuned models go over this?  I'll have to check.  

Thoughts?  Happy to be corrected here. @Blaizzy @Lazarus-931 

My thinking here is that, since we've had to fix several sanitized weights issues already, I'm trying to think: is there a way to capture all of them in future model issues with one pass instead of whack-a-mole? 

Also does MiniCPM-V 4.6 and Step3p7 have the same unconditional +1.0 pattern? They weren't included in this PR but seem to have the same bug.

weklund:
Oh crap forget what I said haha.  I started doing some validation testing. Qwen3.5-0.8B HF-format norm weights have means ranging from 0.07 to 3.4.  Never mind!  😅 

I did see what `mlx-lm` does here.  Interesting: https://github.com/ml-explore/mlx-lm/blob/2ed22318cd6a2fcc5c2e0caa1e1fb0ddeb7cafd5/mlx_lm/models/qwen3_5.py#L348

```
    def sanitize(self, weights):
        has_mtp_weights = any("mtp." in k for k in weights)
        has_unsanitized_conv1d = any(
            "conv1d.weight" in k and v.shape[-1] != 1 for k, v in weights.items()
        )
        should_shift_norm_weights = has_mtp_weights or has_unsanitized_conv1d
        weights = {k: v for k, v in weights.items() if "mtp." not in k}

        if self.args.tie_word_embeddings:
            weights.pop("lm_head.weight", None)

        norm_keys = (
            ".input_layernorm.weight",
            ".post_attention_layernorm.weight",
            "model.norm.weight",
            ".q_norm.weight",
            ".k_norm.weight",
        )
        for k, v in weights.items():
            if "conv1d.weight" in k and v.shape[-1] != 1:
                weights[k] = v.moveaxis(2, 1)
            if should_shift_norm_weights and any(k.endswith(sfx) for sfx in norm_keys):
                if v.ndim == 1:
                    weights[k] = v + 1.0
        return weights
```

Blaizzy:
> 1 question that came up when I was looking at the bug: does the language_model. prefix distinguish MLX-format from HF-format across all checkpoint layouts? 

Not really.

We just prefer the format, it helps separate things.

Model
 - language_model
 - vision_model
 - audio_model
 - projector

Blaizzy (approved):

Issue #1322, opened 2026-06-07
Quantized KV cache (--kv-bits) degenerates into single-token repetition on hybrid MoE models (Qwen3.5/3.6)


gruenewald:
First off — thank you for mlx-vlm. It's been fantastic for running VLMs locally on Apple Silicon, and it's the backbone of our document entity-extraction pipeline (German business documents, served through an OpenAI-compatible gateway). We tracked a quality regression from 0.5.0 → 0.6.2 down to a concrete root cause and a reproducible red→green test, and wanted to share it in case it's useful. 🙏

## Context: entity-extraction quality regressed from 0.5.0 → 0.6.2

Our production setup runs `mlx-community/Qwen3.6-35B-A3B-4bit` greedily (`temperature=0`) with **KV-cache quantization enabled** (it both saves memory *and*, on 0.5.0, improved extraction accuracy). On our 6-document / 28-ID extraction suite:

| | 0.5.0 (KV-quant) | 0.6.2 (KV-quant) |
|---|---|---|
| IDs extracted | **28 / 28** | **degenerate output** — single-token repetition (`"9999…"`), effectively 0 usable |

(For completeness: a separate vision-specific recall regression, filed in #1323, accounts for the rest of the gap once this degeneration is out of the way.)

This issue is the cleaner, fully self-contained half of that regression. It reproduces **text-only**, so no model-specific data is needed.

## Summary

With KV-cache quantization (`--kv-bits N`) on the hybrid MoE model `mlx-community/Qwen3.6-35B-A3B-4bit` (`model_type: qwen3_5_moe`), greedy decoding collapses into single-token repetition — e.g. `"9999…"` until `max_tokens`. 0.5.0 produced correct output with the same quantization, so this is a regression.

## Environment (exact)

| | |
|---|---|
| Model | `mlx-community/Qwen3.6-35B-A3B-4bit` (public; `qwen3_5_moe`, 4-bit affine, group_size 64, hidden 2048, 40 layers) |
| mlx-vlm | **0.6.2** / current `main` @ `6d8d5e3` — broken; **0.5.0** — correct |
| mlx / mlx-lm / mlx-metal | 0.31.2 / 0.31.3 / 0.31.2 · Python 3.13 |
| Hardware / OS | Apple M4 Max, macOS 26.5.1 |

[...]

PR #1534, opened 2026-07-08
Enable APC with KV-cache quantization (single-request)

### LINKED ISSUE #1174: Support APC with KV-cache quantization (critical for TTFT & long Agent tasks for local-first AI)
Hi team,

First, thank you for the amazing work on `mlx-vlm`!

I noticed in the documentation that on the server side, **APC (Adaptive Paged Attention) is skipped when KV-cache quantization is enabled**.

I would like to request adding support for **APC together with KV-cache quantization**. Here's why this matters in practice:

### 1. TTFT (Time To First Token) — critical for Agent workflows
APC dramatically reduces TTFT, which is **essential for multi-turn Agent tasks**. Each Agent step forces a full prefill over the entire growing conversation history. This can take tens of minutes per step — not a delay, but a complete breakdown of usability.

### 2. Memory efficiency for long-context Agent loops
On a **32GB M4 Mac Mini**, running a **~35B model (e.g., Qwen3.6-35B)** is already near the memory limit. For long Agent tasks with:
- long system prompt
- long conversation history,

the KV-cache grows quickly → OOM / crashes.

### The combined benefit:
**APC + KV-cache quantization** would:
- Reduce TTFT significantly (better Agent responsiveness),
- Lower KV-cache memory footprint (prevent OOM),
- Enable running **powerful 30B–35B models in real-world Agent workloads** locally on a 32GB Mac Mini.

This would unlock practical local deployment of models like Qwen3.6-35B for Agent tasks — a major win for local-first AI.

Would you consider enabling APC when KV-cache quantization is enabled in `mlx-vlm` (or the underlying serving stack)?

Happy to test, benchmark, or help with contributions.

Thanks again for your great work!

weklund:
## Summary

- Fixes #1174: APC was unconditionally disabled when `--kv-bits` was set, forcing full prefill on every request even with identical prompts
- Adds protocol-based adapter layer (`dequantize_for_apc()`) so quantized caches participate in APC harvest/restore without changing APC internals
- Restores quantized caches on APC hit (not plain float), preserving `--kv-bits` memory savings for the cached prefix

## Approach: Store Dequantized, Restore into Quantized

Thin adapters at the APC boundary. No changes to block format, disk serialization, hashing, or eviction.

**Store:** dequantize quantized K/V at harvest time, store as raw float tensors in APC blocks (same format APC already uses)

**Restore:** load raw float tensors from APC blocks, feed through `update_and_fetch()` which re-quantizes, model gets properly quantized cache

## Changes

- `mlx_vlm/models/cache.py`: Add `_dequantize_uniform()` helper + `dequantize_for_apc()` on `QuantizedKVCache` and `BatchQuantizedKVCache`
- `mlx_vlm/turboquant.py`: Add `dequantize_for_apc()` on `TurboQuantKVCache` and `BatchTurboQuantKVCache`
- `mlx_vlm/apc.py`: Protocol dispatch in `harvest_blocks_from_batch_cache`, `kv_quant_config` param on restore functions (`make_warm_kv_cache`, `make_warm_batch_kv_cache`, `make_warm_batch_kv_cache_multi`, `make_warm_kv_cache_from_layers`), expanded support checks, quantized clone for exact-mode
- `mlx_vlm/generate/ar.py`: Remove guard that killed APC when `kv_bits` was set; wire `kv_quant_config` into batch restore path
- `mlx_vlm/generate/dispatch.py`: Wire `kv_quant_config` into single-stream restore path

## Known limitations

- Exact-mode APC with `--kv-bits` on hybrid models (Gemma 4, Qwen 3.5) in multi-row batches: the exact-mode clone dequantizes to KVCache for storage, which may cause type mismatch with cold quantized caches in concurrent-batch scenarios. Block-mode (standard models) is fully handled.
- TurboQuant restore path creates uniform `BatchQuantizedKVCache`, not `BatchTurboQuantKVCache`. TurboQuant users get correct results but lose the custom codec on the restored prefix.

## Validation
- `uv run --with pytest pytest mlx_vlm/tests/test_apc_quantized.py -q`
- `uv run --with pytest pytest mlx_vlm/tests/test_apc.py -q` (no regressions)
- `uv run --with pytest pytest mlx_vlm/tests/test_apc_exact_mode.py -q` (no regressions)
- `uv run --with pytest pytest mlx_vlm/tests/test_batch_quantized_cache.py -q` (no regressions)
- `uv run --with pytest pytest mlx_vlm/tests/test_turboquant.py -q` (no regressions)

Ref: https://github.com/Blaizzy/mlx-vlm/issues/1174

### DIFF
--- a/mlx_vlm/apc.py
+++ b/mlx_vlm/apc.py
@@ -221,6 +221,17 @@ def _clone_cache_entry_for_apc(
             return None
         return tuple(copied)
 
+    # Protocol fallback: any cache with dequantize_for_apc (e.g. QuantizedKVCache)
+    # is cloned by dequantizing into a plain KVCache for exact-mode storage.
+    if hasattr(c, "dequantize_for_apc"):
+        dk, dv = c.dequantize_for_apc()
+        out = lm_cache.KVCache()
+        out.keys = _copy_mlx_array(dk)
+        out.values = _copy_mlx_array(dv)
+        out.offset = dk.shape[-2]
+        eval_targets.extend([out.keys, out.values])
+        return out
+
     return None
 
 
@@ -282,6 +293,8 @@ def _cache_entry_supports_exact_apc(c: Any) -> bool:
         ),
     ):
         return True
+    if hasattr(c, "dequantize_for_apc"):
+        return True
     if isinstance(c, lm_cache.CacheList):
         return all(_cache_entry_supports_exact_apc(sub_c) for sub_c in c.caches)
     if isinstance(c, tuple):
@@ -292,7 +305,11 @@ def _cache_entry_supports_exact_apc(c: Any) -> bool:
 def _cache_entry_supports_block_apc(c: Any) -> bool:
     from .models import cache as lm_cache
 
-    return isinstance(c, lm_cache.KVCache)
+    if isinstance(c, lm_cache.KVCache):
+        return True
+    if hasattr(c, "dequantize_for_apc"):
+        return True
+    return False
 
 
 def _sequence_hash(token_ids: Sequence[int], extra_hash: int, block_size: int) -> int:
@@ -3386,18 +3403,38 @@ class APCManager:
 def make_warm_kv_cache(
     matched_blocks: List[APCBlock],
     min_capacity_tokens: Optional[int] = None,
+    kv_quant_config: Optional[dict] = None,
 ) -> List[Any]:
-    """Stitch matched blocks into per-layer ``KVCache`` instances pre-filled
+    """Stitch matched blocks into per-layer cache instances pre-filled
     with the cached prefix's K/V state. Used by the single-stream
     ``stream_generate`` path.
+
+    When *kv_quant_config* is provided (dict with ``bits`` and ``group_size``),
+    returns ``QuantizedKVCache`` instances that re-quantize the raw float
+    blocks via ``update_and_fetch``.
     """
-    from .models.cache import KVCache
+    from .models.cache import KVCache, QuantizedKVCache
 
     if not matched_blocks:
         return []
     num_layers = len(matched_blocks[0].keys)
     out: List[Any] = []
     prefix_len = sum(b.keys[0].shape[-2] for b in matched_blocks)
+
+    if kv_quant_config is not None:
+        for layer_idx in range(num_layers):
+            ks = [b.keys[layer_idx] for b in matched_blocks]
+            vs = [b.values[layer_idx] for b in matched_blocks]
+            merged_k = mx.concatenate(ks, axis=2)
+            merged_v = mx.concatenate(vs, axis=2)
+            c = QuantizedKVCache(
+                group_size=kv_quant_config["group_size"],
+                bits=kv_quant_config["bits"],
+            )
+            c.update_and_fetch(merged_k, merged_v)
+            out.append(c)
+        return out
+
     step_probe = KVCache()
     kv_step = int(getattr(step_probe, "step", getattr(KVCache, "step", 256)))
     capacity = prefix_len
@@ -3449,13 +3486,18 @@ def make_warm_kv_cache_from_layers(
 
 def make_warm_batch_kv_cache(
     matched_blocks: List[APCBlock],
+    kv_quant_config: Optional[dict] = None,
 ) -> List[Any]:
-    """Stitch matched blocks into per-layer single-row ``BatchKVCache``
-    instances pre-filled with the cached prefix's K/V state. Used by the
-    batched continuous-batching path; the resulting cache list can be
+    """Stitch matched blocks into per-layer single-row batch cache instances
+    pre-filled with the cached prefix's K/V state. Used by the batched
+    continuous-batching path; the resulting cache list can be
     ``extend()``-ed into a running batch.
+
+    When *kv_quant_config* is provided (dict with ``bits`` and ``group_size``),
+    returns ``BatchQuantizedKVCache`` instances that re-quantize the raw float
+    blocks via ``update_and_fetch``.
     """
-    from .models.cache import BatchKVCache
+    from .models.cache import BatchKVCache, BatchQuantizedKVCache
 
     if not matched_blocks:
         return []
@@ -3467,14 +3509,21 @@ def make_warm_batch_kv_cache(
         vs = [b.values[layer_idx] for b in matched_blocks]
         merged_k = mx.concatenate(ks, axis=2)  # [1, H, prefix_len, D]
         merged_v = mx.concatenate(vs, axis=2)
-        c = BatchKVCache(left_padding=[0])
-        # state setter: (keys, values, offset, left_padding) → also sets _idx
-        c.state = (
-            merged_k,
-            merged_v,
-            mx.array([prefix_len]),
-            mx.array([0]),
-        )
+        if kv_quant_config is not None:
+            c = BatchQuantizedKVCache(
+                [0],
+                group_size=kv_quant_config["group_size"],
+                bits=kv_quant_config["bits"],
+            )
+            c.update_and_fetch(merged_k, merged_v)
+        else:
+            c = BatchKVCache(left_padding=[0])
+            c.state = (
+                merged_k,
+                merged_v,
+                mx.array([prefix_len]),
+                mx.array([0]),
+            )
         out.append(c)
     return out
 
@@ -3708,6 +3757,23 @@ def harvest_blocks_from_batch_cache(
     layer_keys: List[mx.array] = []
     layer_values: List[mx.array] = []
     for c in batch_caches:
+        # Protocol dispatch: quantized caches expose dequantize_for_apc()
+        # which returns raw float arrays suitable for direct slicing.
+        if hasattr(c, "dequantize_for_apc"):
+            dk, dv = c.dequantize_for_apc()
+            idx = dk.shape[-2]
+            left_padding = getattr(c, "left_padding", None)
+            if left_padding is not None:
+                try:
+                    lp = int(left_padding[batch_idx].item())
+                except Exception:
+                    lp = 0
+            else:
+                lp = 0
+            layer_keys.append(dk[batch_idx : batch_idx + 1, :, lp:, :])
+            layer_values.append(dv[batch_idx : batch_idx + 1, :, lp:, :])
+            continue
+
         keys = getattr(c, "keys", None)
         values = getattr(c, "values", None)
         idx = getattr(c, "_idx", None)
--- a/mlx_vlm/apc.py
+++ b/mlx_vlm/apc.py
@@ -154,6 +154,36 @@ def _clone_cache_entry_for_apc(
             eval_targets.extend([keys, values])
         return out
 
+    # BatchKVCache is not a KVCache subclass. With --kv-bits the batch cache
+    # factory is used even for single-row requests, so hybrid exact-mode
+    # snapshots often see BatchKVCache for the unquantized last layer.
+    # Collapse B=1 into a row KVCache (same as extract(0)); multi-row must
+    # be extracted before store.
+    if isinstance(c, lm_cache.BatchKVCache):
+        idx = int(getattr(c, "_idx", 0) or 0)
+        if c.keys is None or c.values is None or idx <= 0:
+            return lm_cache.KVCache()
+        if int(c.keys.shape[0]) != 1:
+            return None
+        row = c.extract(0)
+        off = int(getattr(row, "offset", 0) or 0)
+        if row.keys is not None and row.values is not None and off > 0:
+            keys = _copy_mlx_array(row.keys[..., :off, :])
+            values = _copy_mlx_array(row.values[..., :off, :])
+            step = int(getattr(row, "step", getattr(type(row), "step", 256)) or 0)
+            keys, values = _pad_kv_for_capacity(
+                keys,
+                values,
+                offset=off,
+                min_capacity_tokens=min_capacity_tokens,
+                step=step,
+            )
+            row.keys = keys
+            row.values = values
+            row.offset = off
+            eval_targets.extend([keys, values])
+        return row
+
     if isinstance(c, lm_cache.RotatingKVCache):
         out = type(c)(
             max_size=int(getattr(c, "max_size")),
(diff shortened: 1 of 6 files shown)


weklund:
### Benchmark: APC + `--kv-bits 8`

Tested on block-mode models (standard transformers). Previously with `--kv-bits` set, APC was unconditionally disabled, so every request paid the full cold prefill cost regardless of shared prefix.

#### Short prefix (~160 tokens)

| Model | Prompt Tokens | Cold Prefill | Warm Prefill | Cached Tokens | Speedup |
|-------|:---:|:---:|:---:|:---:|:---:|
| `Qwen3-0.6B-4bit` | 161 | 58ms | 27ms | 144/161 | **2.1x** |
| `Qwen3-4B-4bit` | 161 | 108ms | 51ms | 144/161 | **2.1x** |
| `Qwen3-8B-4bit` | 161 | 142ms | 58ms | 144/161 | **2.5x** |

#### Long prefix (~1560 tokens, simulates agent system prompt)

| Model | Prompt Tokens | Cold Prefill | Warm Prefill | Cached Tokens | Speedup |
|-------|:---:|:---:|:---:|:---:|:---:|
| `Qwen3-4B-4bit` | 1561 | 1,153ms | 62ms | 1,536/1,561 | **18.6x** |
| `Qwen3-8B-4bit` | 1561 | 1,583ms | 77ms | 1,536/1,561 | **20.7x** |

**Config:** `APC_ENABLED=1 APC_NUM_BLOCKS=1024 --kv-bits 8`, warm = avg of 3 requests, `max_tokens=5`
**Hardware:** Apple Silicon M4 Max, 128GB

The speedup scales with prefix length: 2x for short prefixes, 18-20x for agent-style system prompts (~1500 tokens). For the 25k-60k token contexts reported in the issue, the improvement would be even more dramatic (minutes to seconds).

#### Notes on hybrid models (Qwen3.5, Qwen3.6-MoE)

These models use `ArraysCache` + `KVCache` (hybrid architecture) and route to exact-mode APC. Exact-mode store/restore with `--kv-bits` in multi-row concurrent batches is a known limitation documented in the PR description. The fix fully covers block-mode (standard models), which is the common case for the reported setups.

weklund:
Just to be explicit here, I'm not going to switch this to ready for review until a few folks have pulled the branch and tried it out. 

Big change, so looking for feedback :)

pinglin:
Validated this PR on an Apple-silicon 128GB dev machine against our production serving config: **Qwen3.6-35B-A3B-mxfp4** (hybrid cache: 30×`ArraysCache` + 10×`KVCache` → exact mode), continuous batching server, `--kv-bits 8`, `--max-kv-size 131072`, `APC_ENABLED=1`.

**Stability/correctness: clean.** Long-prompt (18.7k tok) exact repeats, growing-prefix multi-turn, and a 4-way concurrent shared-prefix burst all returned correct answers, no 500s.

**But APC never engages for this model — traced it:** `model_apc_mode` resolves `"exact"` ✓, the harvest checkpoint is computed ✓, `store_exact_cache` is called ✓ — and then rejects silently: `_clone_prompt_cache_for_apc` returns `None` because the batch engine's prompt cache is `[ArraysCache, BatchKVCache, BatchQuantizedKVCache]` and there's no clone branch for **`BatchKVCache`** (the unquantized batch entries; `BatchQuantizedKVCache` clones fine via the new protocol). The single-request path in `_apc_prompt_cache_for_store` returns `self.prompt_cache` (batch classes) directly rather than extracting row caches the way the multi-row path does.

Two possible fixes: always go through `extract_prompt_cache_from_batch` (drop the single-request special case), or add a `BatchKVCache` branch to `_clone_cache_entry_for_apc`. Also worth adding a `stores_rejected` counter — the silent `return False` made this invisible in `/v1/cache/stats` (everything reads 0 while the server logs nothing).

Happy to re-run this validation on a follow-up commit.

pinglin:
Follow-up with a sharper root cause: on the *pre-PR* base without `--kv-bits`, the same model stores exact-cache snapshots fine — because single-request cold batches there use plain unbatched caches. The rejection appears when the kv-quantized path builds batch-class caches: `BatchKVCache` is not a `KVCache` subclass (separate `_BaseCache` lineage, constructor requires `left_padding`), so the `isinstance(c, KVCache)` clone branch never matches it and `type(c)()` wouldn't construct anyway. So routing single-request stores through `extract_prompt_cache_from_batch` (rather than special-casing `self.prompt_cache`) looks like the cleaner fix of the two I mentioned.

trevorgordon981:
Validated on a 512GB M3-class Mac Studio, mlx-vlm 0.6.5 @ this branch (`cd571af`). **APC + KV-quant works — the combination is now genuinely active instead of a silent no-op.**

Model: `mlx-community/Qwen2.5-VL-3B-Instruct-4bit` (couldn't fit MiniMax-M3 alongside a resident 320GB+ prod model this pass — happy to add an M3 hybrid-MoE run in a maintenance window). Workload: ~10k-token static system prefix + growing multi-turn history, 6 sequential greedy turns.

`test_apc_quantized.py`: 33/33 pass. Server matrix (per-turn TTFT):

| config | cold TTFT | warm TTFT (mean) | APC active? |
|---|---|---|---|
| APC on, fp16 | 14.4s | 0.58s | hits=5 |
| **APC on, kv8** | 18.9s | **0.50s** | **hits=5** ✅ |
| APC off, kv8 | 16.1s | 16.4s | (re-prefills every turn) |
| APC on, kv4 | 18.5s | 0.51s | hits=5 |

`/v1/cache/stats` in the kv8 cell: `lookups_hit=5, stores=1619, token_hit_rate=0.83` — vs pre-PR where kv-bits made APC inert. **~33x warm-TTFT reduction at kv8, ~48x at kv4.** No crashes, no Metal command-buffer errors.

Two honest notes:
1. **Memory:** APC-on peak RSS was identical for fp16/kv8/kv4 — the dequant-at-harvest design stores *float* blocks, so the APC prefix pool doesn't shrink with kv-bits (expected given the approach; worth a docs line).
2. **Correctness:** block-APC reuse isn't bit-exact vs a fresh prefill *even in fp16* (greedy outputs drift after a few turns); kv8 sits in the same regime, so no additional regression from requantization. Separately, **kv4 triggers #1322-style single-token repetition — but it does so with APC *off* too** (including the cold first turn), so that's the 4-bit KV quant itself, not this PR. kv8 was clean throughout.

LGTM to merge for the TTFT win; suggest documenting the float-pool memory cost.

weklund:
Thanks @pinglin — root cause confirmed and addressed for the common path.

### What was wrong
With `--kv-bits`, even single-row batches use batch cache classes (`BatchKVCache` / `BatchQuantizedKVCache`). Hybrid exact-mode store walked `_clone_prompt_cache_for_apc`, which had no `BatchKVCache` branch (`BatchKVCache` is not a `KVCache` subclass). Clone returned `None` → `store_exact_cache` returned `False` with no log/stats, so APC looked idle.

### What this PR now does (surgical)
- Clone `BatchKVCache` (B=1) into a row `KVCache` for exact snapshots
- Warn on unclonable exact-store rejects (cache type names in the log)
- Regression test for the hybrid `[ArraysCache, BatchKVCache, BatchQuantizedKVCache]` layout

@pinglin — worth re-running your Qwen3.6-35B-A3B-mxfp4 + `--kv-bits 8` validation on this tip; single-request / sequential multi-turn exact stores should engage now.

### Still limited (follow-up PR)
Multi-row concurrent exact + kv-bits can still fail because `BatchQuantizedKVCache` has no `extract()`, and the single-row special case in `_apc_prompt_cache_for_store` is a footgun. Block-mode (standard transformers) remains the well-tested path.

---

### Follow-up PR: internal APC adapter boundary (no CLI change)

**Goal:** Make APC robust across model × batch × quant without growing `isinstance` ladders — **implementation-only**; same env flags, CLI, and generate APIs.

**Non-goals:** New user-facing knobs, on-disk format changes, CLI migrations.

**Plan:**

1. **Canonical port** — `snapshot_prompt_cache(caches, row) → List[APCLayerView] | reject reason`. APC core only sees single-row, storage-native views (float K/V + Arrays). Always normalize; delete the “B=1 is already unbatched” special case.
2. **Adapters on cache types** — `apc_snapshot_row` / extract parity for `BatchKVCache`, `BatchQuantizedKVCache`, TurboQuant batch, etc. Quant dequant stays on the adapter, not scattered in harvest/clone/disk.
3. **Structured outcomes + observability** — `APCOutcome(reason=...)`; additive stats (`rejects`, `rejects_by_reason`, `last_reject`); `APC_TRACE=1` for per-op debug (same spirit as `APC_DISK_TRACE`). No silent `False`.
4. **Startup capability check** — dry-run adapters against the real `_make_cache(..., kv_bits=)` layout when APC starts; fail loud if a layer can’t snapshot.
5. **Collapse duplicate ladders** — memory exact clone, disk exact snapshot, and block harvest share one inbound path.

**Success criteria:** Hybrid exact + kv-bits works for multi-row concurrent; rejects show up in `/v1/cache/stats` and logs; no new required CLI flags; existing block-mode golden paths unchanged.

Happy to open that as a stacked PR after this lands.

weklund:
Since you were the issue creator @Edsuns would you like to test?  @bonyohana ?

Also @Blaizzy @Lazarus-931 @lucasnewman I think we're ready for your review :)

weklund:
### Smoke note: Gemma 4 batch exact + `--kv-bits` still incomplete

While re-running an end-to-end smoke matrix on this tip (block-mode Qwen3, hybrid Qwen3.5/3.6 including 35B-A3B, and Gemma 4), I hit a remaining gap on **Gemma 4** under continuous batching:

- **Model:** `mlx-community/gemma-4-e2b-it-4bit` (exact mode — sliding-window + full-attention mix)
- **Path:** `BatchGenerator` + `APC` + `--kv-bits 8`
- **Symptom:** exact-mode store is rejected; `/v1/cache/stats`-style counters stay at 0 for stores/hits on the batch path
- **Log (from the new warning):**

```text
APC exact-cache store rejected: unclonable prompt cache types
['BatchRotatingKVCache', ..., 'BatchQuantizedKVCache', ..., 'BatchKVCache']
```

Root cause is the same *class* of issue as @pinglin’s hybrid report, but for a different cache dialect: with `--kv-bits`, Gemma’s sliding-window layers show up as **`BatchRotatingKVCache`**, which exact-mode clone still doesn’t handle. (`BatchKVCache` is covered by the surgical fix on this PR; rotating batch is not.)

**Workaround / current behavior:**
- Single-stream `generate()` with APC + kv-bits on Gemma 4 **does** exact-store and hit (smoke saw warm exact hits).
- Server / continuous-batching path for Gemma 4 + exact + kv-bits still will not engage APC until adapters cover `BatchRotatingKVCache`.

**Qwen block-mode and Qwen3.5/3.6 hybrid exact + kv-bits** (generate + batch, including 35B-A3B) looked good in the same smoke run — those remain the validated in-scope wins for this PR.

A fix for Gemma 4 (and other missing batch/quant dialects) will come in the **follow-up PR** already outlined above: internal APC adapter boundary (`snapshot_row` / materialize, full batch+quant coverage, structured rejects) — no new CLI contract. Not folding that redesign into this PR on purpose.

mlx_vlm/apc.py
```diff
@@ -154,6 +154,36 @@ def _clone_cache_entry_for_apc(
...
+    # BatchKVCache is not a KVCache subclass. With --kv-bits the batch cache
+    # factory is used even for single-row requests, so hybrid exact-mode
+    # snapshots often see BatchKVCache for the unquantized last layer.
+    # Collapse B=1 into a row KVCache (same as extract(0)); multi-row must
+    # be extracted before store.
+    if isinstance(c, lm_cache.BatchKVCache):
+        idx = int(getattr(c, "_idx", 0) or 0)
+        if c.keys is None or c.values is None or idx <= 0:
+            return lm_cache.KVCache()
+        if int(c.keys.shape[0]) != 1:
```
lucasnewman, line 166: This would be more clear using `empty()` and adding something like `is_single_row()` to BatchKVCache so the intent and limitations are plain, instead of reaching into the internals like it does.

lucasnewman:
@weklund Thanks! Overall this looks reasonable to me and is definitely an improvement for the single request case. It would be great to cover both the SWA layer caches and concurrent batched requests as well, as it's not really "feature complete" until we have those. I'm fine with them coming in as follow ups though, as well as the comment above which is just a nit.

lucasnewman (approved):

Edsuns:
I tried running Claude (connected to the API deployed via mlx-vlm) to test it, and the `lookups_hit` and `lookups_miss` metrics for the `/v1/cache/stats` endpoint have consistently been 0. It seems there's still an issue.

weklund:
> I tried running Claude (connected to the API deployed via mlx-vlm) to test it, and the `lookups_hit` and `lookups_miss` metrics for the `/v1/cache/stats` endpoint have consistently been 0. It seems there's still an issue.

@Edsuns sorry just seeing this!  

is your intent that this pr https://github.com/Blaizzy/mlx-vlm/pull/1576 fixes it?

Edsuns:
> > I tried running Claude (connected to the API deployed via mlx-vlm) to test it, and the `lookups_hit` and `lookups_miss` metrics for the `/v1/cache/stats` endpoint have consistently been 0. It seems there's still an issue.
> 
> 
> 
> @Edsuns sorry just seeing this!  
> 
> 
> 
> is your intent that this pr https://github.com/Blaizzy/mlx-vlm/pull/1576 fixes it?

@weklund This PR seems to address the issue I encountered. It looks like a very useful improvement, and I'll try it out when I have some free time recently.

Edsuns:
> > I tried running Claude (connected to the API deployed via mlx-vlm) to test it, and the `lookups_hit` and `lookups_miss` metrics for the `/v1/cache/stats` endpoint have consistently been 0. It seems there's still an issue.
> 
> @Edsuns sorry just seeing this!
> 
> is your intent that this pr #1576 fixes it?

@weklund This PR doesn't solve the issue I'm running into. Let me re-explain the problem: when the Agent's task context reaches roughly 100k tokens, it hits an OOM error, even with KV quantization enabled, and the `lookups_hit` and `lookups_miss` metrics from the `/v1/cache/stats` endpoint have consistently been 0.

PR #1536, opened 2026-07-08
add gemma3_text model type reusing gemma3 language internals

Lazarus-931:
Refactor to move gemma3 text models from mlx-lm's dependency to full mlx-vlm native support

Gemma3 text-only checkpoints (gemma-3-1b etc.) now load and run natively through mlx-vlm instead of routing through the text_only.py mlx-lm fallback.

Verified bit-exact parity against mlx-lm's gemma3_text with shared weights (prefill at T=1/8/20/33, crossing the sliding window, and cached decode across the window boundary).

All tests gtg!


Blaizzy (approved):
LGTM 🚀

PR #1540, opened 2026-07-08
Fix PaddleOCR-VL sanitize double-prefixing MLX-format checkpoints

### LINKED ISSUE #1537: PaddleOCR-VL fails to load on 0.6.4: sanitize double-prefixes already-namespaced weights
### Summary
  
  PaddleOCR-VL models fail to load on **mlx-vlm 0.6.4** (works fine on **0.6.3**) with:

  ValueError: Received 165 parameters not in model:
  language_language_model.model.language_model.model.layers.0.input_layernorm.weight, ...
  
  ### Repro

  ```python
  from mlx_vlm import load
  load("<any PaddleOCR-VL MLX model, e.g. a *-MLX-bf16 conversion>")

  Root cause

  models/paddleocr_vl/paddleocr_vl.py sanitize() maps raw HF names to the model's
  namespaced names via an unanchored substring match:
  
  elif "model" in key:
      key = key.replace("model", "language_model.model")   # matches "model" anywhere, replaces ALL 
  occurrences
  
  mlx-vlm's own convert saves weights under the model's final class-structure names
  (language_model.model...) — save_weights(mlx_path, model) serializes the loaded model's
  parameters. On load, #1498 ("always sanitize loaded weights") now runs sanitize
  unconditionally, so the transform is re-applied to already-namespaced keys. Because the match
  is unanchored, the model inside language_model is also replaced:

  language_model.model.layers.0...  ->
  language_language_model.model.language_model.model.layers.0...

  This breaks every PaddleOCR-VL model on 0.6.4 — the on-disk naming is correct/standard;
  re-converting produces identical names and the same failure. 0.6.3 worked because it did not
  run sanitize unconditionally on already-namespaced weights.
  
  Suggested fix (anchor the match + make it idempotent)

  elif key.startswith("model."):
      key = key.replace("model", "language_model.model", 1)
  elif key.startswith("lm_head"):
      key = key.replace("lm_head", "language_model.lm_head", 1)

  Environment

  - mlx-vlm 0.6.4 (regression vs 0.6.3)
  - macOS, Apple Silicon
  - Model: a PaddleOCR-VL MLX conversion (bf16)

Lazarus-931:
sanitize() mapped raw HF names with an unanchored substring replace, so re-running it on already-converted checkpoints (unconditional since the always-sanitize change) turned language_model.model.* keys into language_language_model.model.* and every PaddleOCR-VL model failed to load. Skip weights that are already namespaced and anchor the model/ lm_head mappings, which also makes the lm_head branch reachable.

Fixes #1537

### DIFF
--- a/mlx_vlm/models/paddleocr_vl/paddleocr_vl.py
+++ b/mlx_vlm/models/paddleocr_vl/paddleocr_vl.py
@@ -168,6 +168,9 @@ class Model(nn.Module):
         return logits
 
     def sanitize(self, weights):
+        if any(k.startswith("language_model.") for k in weights):
+            return weights
+
         _keys_to_ignore_on_load_unexpected = [
             "packing_position_embedding",
             "vision_model.head",
@@ -181,10 +184,10 @@ class Model(nn.Module):
                     key = key.replace("visual.vision_model.encoder", "visual")
             elif "mlp_AR" in key:
                 key = key.replace("mlp_AR", "visual.projector")
-            elif "model" in key:
-                key = key.replace("model", "language_model.model")
-            elif "lm_head" in key:
-                key = key.replace("lm_head", "language_model.lm_head")
+            elif key.startswith("model."):
+                key = f"language_model.{key}"
+            elif key.startswith("lm_head"):
+                key = f"language_model.{key}"
 
             return key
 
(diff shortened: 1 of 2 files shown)


lucasnewman (approved):
LGTM

PR #1541, opened 2026-07-08
Fix double-sanitize issue with Qwen3 Omni MoE

lucasnewman:
Fallout from https://github.com/Blaizzy/mlx-vlm/pull/1498

PR #1550, opened 2026-07-09
fix(sam3): make sanitize idempotent for pre-converted MLX conv weights

### LINKED ISSUE #1545: SAM3: sanitize() double-transposes conv weights when loading pre-converted MLX checkpoints (mlx-community/sam3-bf16)
**Version:** 0.6.4 (also present on current `main`)

## Repro

```python
from mlx_vlm.utils import load_model, get_model_path
from mlx_vlm.models.sam3.generate import Sam3Predictor
# any pre-converted MLX checkpoint:
mp = get_model_path("mlx-community/sam3-bf16")
model = load_model(mp)   # crashes in patch embedding
```

Fails with a shape error in the vision patch embed: the conv weight arrives as `(1024, 14, 3, 14)`.

## Root cause

`Model.sanitize()` in `mlx_vlm/models/sam3/sam3.py` unconditionally applies the PyTorch→MLX conv transposes by key pattern. The `mlx-community/sam3-*` checkpoints are **already stored in MLX layout** `(out, H, W, in)`, so sanitize transposes them a second time: `(1024, 14, 14, 3) → (1024, 14, 3, 14)` for the patch-embedding conv, and similarly for the other matched patterns.

## Suggested fix

Guard the transpose with a layout check — e.g. for Conv2d patterns, only transpose when the weight is in PyTorch layout (`w.shape[1] == in_channels` rather than `w.shape[-1] == in_channels`), or detect once via the patch-embed weight and skip sanitize entirely for already-converted checkpoints.

## Workaround we're shipping

```python
import mlx_vlm.models.sam3.sam3 as sam3mod
sam3mod.Model.sanitize = staticmethod(lambda w: w)  # checkpoint already MLX layout
```

Happy to turn this into a PR if a layout-guard approach sounds right to you.

Lazarus-931:
sanitize() unconditionally transposed every 4D conv weight, so loading a pre-converted MLX checkpoint (mlx-community/sam3-*) double-transposed the patch-embed conv to (1024,14,3,14) and crashed. Guard the transpose with the model's own parameter shapes and skip weights already in MLX layout.

Fixes #1545.

this is the same fix as #1541 for qwen omni

### DIFF
--- a/mlx_vlm/models/sam3/sam3.py
+++ b/mlx_vlm/models/sam3/sam3.py
@@ -365,7 +365,22 @@ class Model(nn.Module):
         Main conversions:
         1. Conv2d: PyTorch [out, in, H, W] -> MLX [out, H, W, in]
         2. ConvTranspose2d: PyTorch [in, out, H, W] -> MLX [out, H, W, in]
+
+        Conv transposes are skipped for checkpoints already in MLX layout
+        (e.g. pre-converted ``mlx-community/sam3-*``), so sanitize is idempotent
+        and does not double-transpose (see #1545).
         """
+        # Detect layout once via the image patch-embed conv, whose in_channels
+        # is unambiguously 3 (RGB): MLX layout ends in 3, PyTorch layout has it
+        # at axis 1. If already MLX, the whole checkpoint is pre-converted.
+        already_mlx = any(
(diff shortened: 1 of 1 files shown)


lucasnewman (approved):
LGTM

PR #1551, opened 2026-07-09
Fix logits processor alignment in mixed generation batches

### LINKED ISSUE #1555: Structured-output logits processors become misaligned when generation batches are extended
## Description

`GenerationBatch.extend()` can misalign per-request logits processors when structured and unstructured requests are admitted into a continuous generation batch in separate waves.

Batches without active processors use an empty `logits_processors` list as a compact representation. If multiple plain batches are merged and a structured request is appended afterward, the structured request's processor is appended without placeholders for the existing rows.

For example, three UIDs may end up with only one processor entry:

```text
uids: [0, 1, 2]
logits_processors: [[structured_processor]]
```

The processor belongs to UID 2, but occupies row 0. During decoding, the structured-output grammar can therefore be applied to an unrelated plain request. Subsequent rows may also encounter an out-of-range processor index. I initially encountered this through concurrent structured and unstructured requests using the mlx-vlm continuous server, but the problem reproduces directly with GenerationBatch.

## Reproduction

Run the following script from an mlx-vlm checkout:

```python
import mlx.core as mx

sampler = lambda logprobs: mx.argmax(logprobs, axis=-1)
stop_criteria = lambda token: False

def make_batch(uid, processor=None):
    return GenerationBatch(
        model=object(),
        uids=[uid],
        inputs=mx.array([uid + 1], dtype=mx.int32),
        prompt_cache=[],
        sampler=sampler,
        stop_criteria=stop_criteria,
        max_tokens=[2],
        token_context=[[30]] if processor is not None else None,
        logits_processors=[[processor]] if processor is not None else None,
    )

first_plain = make_batch(0)
second_plain = make_batch(1)

structured_processor = lambda tokens, logits: logits
structured = make_batch(2, structured_processor)

first_plain.extend(second_plain)
print("After plain + plain:", first_plain.logits_processors)

first_plain.extend(structured)
print("UIDs:", first_plain.uids)
print("Processor slots:", first_plain.logits_processors)

assert len(first_plain.logits_processors) == len(first_plain.uids)
assert first_plain.logits_processors == [
    None,
    None,
    [structured_processor],
]
```

## Actual behavior

The final processor list contains only the structured processor:

```
After plain + plain: []
UIDs: [0, 1, 2]
Processor slots: [[<function ...>]]
```

The first assertion fails:

`AssertionError: ([0, 1, 2], [[<function ...>]])`

The processor is incorrectly associated with row 0, although it belongs to UID 2.
The reverse admission order is also unsafe: appending a plain request to an active structured batch leaves the processor list shorter than the UID list.

## Expected behavior

When any request in the merged batch has an active processor, every UID should have a corresponding processor slot:

```
UIDs: [0, 1, 2]
Processor slots: [None, None, [structured_processor]]
```

Plain rows should retain None, and the structured processor should remain aligned with its originating request.

## Impact

Under continuous batching, concurrent structured and unstructured requests can affect one another:

- JSON grammar being applied to an unrelated plain request
- Structured parser errors caused by tokens from the wrong request
- Invalid or truncated output
- Generation failure from processor-list indexing

Requests created together with an already aligned processor list are unaffected. The issue occurs when active generation batches are extended with requests that have different processor requirements.

## Proposed fix

Before concatenating logits_processors, restore None placeholders on both batches whenever either side contains an active processor. This preserves the one-slot-per-UID invariant while retaining the compact empty representation when no processors are active.

This is addressed in the PR #1551

byoan:
### Summary

Fix logits processor alignment when structured and unstructured requests are merged into the same generation batch.

### Problem

When batches containing no active logits processors are merged, their `logits_processors` list is compacted to an empty list.

If a structured request is subsequently appended, its processor is added without placeholder entries for the existing plain requests.
This shifts the processor to the first row of the combined batch instead of keeping it aligned with the structured request's UID.

As a result, structured-output constraints may be applied to an unrelated request during continuous batching.

### Fix

Restore `None` placeholders for existing batch rows before concatenating logits processors whenever either batch contains an active processor.

This keeps `uids`, `token_context`, and `logits_processors` aligned without changing the compact representation used when no processors are active.

### Tests

Added a regression test covering this merge sequence:

1. Merge two unstructured generation batches.
2. Append a structured generation batch.
3. Verify that the structured processor remains assigned to its own row.

Tested with:

```bash
python -m pytest mlx_vlm/tests/test_generate.py::TestBatchGenerator -q
Result: 18 passed.
```

Let me know if you need anything else!

Fixes #1555

### DIFF
--- a/mlx_vlm/generate/ar.py
+++ b/mlx_vlm/generate/ar.py
@@ -1104,6 +1104,14 @@ class GenerationBatch:
         if self_has_processors or other_has_processors:
             self._ensure_token_context(force=bool(other_has_processors))
             other._ensure_token_context(force=bool(self_has_processors))
+            if len(self.logits_processors) < len(self.uids):
+                self.logits_processors.extend(
+                    [None] * (len(self.uids) - len(self.logits_processors))
+                )
+            if len(other.logits_processors) < len(other.uids):
+                other.logits_processors.extend(
+                    [None] * (len(other.uids) - len(other.logits_processors))
+                )
         else:
             self.token_context = []
             other.token_context = []
--- a/mlx_vlm/tests/test_generate.py
+++ b/mlx_vlm/tests/test_generate.py
@@ -857,6 +857,49 @@ class TestBatchGenerator:
         finished_structured.filter([1])
         assert finished_structured.uids == [1]
 
+    def test_generation_batch_extend_restores_plain_processor_slots(self):
+        sampler = lambda logprobs: mx.argmax(logprobs, axis=-1)
+        stop_criteria = lambda token: False
+        plain = GenerationBatch(
+            model=MagicMock(),
+            uids=[0],
+            inputs=mx.array([5], dtype=mx.int32),
+            prompt_cache=[],
+            sampler=sampler,
+            stop_criteria=stop_criteria,
+            max_tokens=[2],
+        )
+        second_plain = GenerationBatch(
+            model=MagicMock(),
+            uids=[1],
+            inputs=mx.array([6], dtype=mx.int32),
+            prompt_cache=[],
+            sampler=sampler,
+            stop_criteria=stop_criteria,
+            max_tokens=[2],
+        )
+        processor = lambda tokens, logits: logits
+        structured = GenerationBatch(
+            model=MagicMock(),
+            uids=[2],
+            inputs=mx.array([7], dtype=mx.int32),
+            prompt_cache=[],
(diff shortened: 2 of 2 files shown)


Blaizzy:
Thanks!

Please open a GH issue first detailing the problem with reproducible example

byoan:
@Blaizzy Issue opened with the repro example, I believe we can re-open this PR unless you prefer I create a new one?

PR #1553, opened 2026-07-10
Optimize Qwen3 Omni streaming code windows

sxy-trans-n:
## Summary

Optimize Qwen3-Omni audio streaming by avoiding repeated reconstruction of the complete codec-token history.

After the first streaming window, `generate_stream()` currently stacks all generated codec frames and calls `stream_decode()` for every new frame, even when the next decode window is not ready. This overhead grows with audio duration.

## Changes

- Check whether the next Code2Wav window is ready before stacking codec frames.
- Stack only the required left context and newly decodable frames.
- Apply the same bounded-window behavior when flushing remaining frames.
- Validate invalid chunk/context configurations that cannot make decode progress.
- Preserve the existing public API, Thinker/Talker scheduling, codec tokens, audio chunk boundaries, and vocoder behavior.

## Benchmark

Model: `mlx-community/Qwen3-Omni-30B-A3B-Instruct-4bit`  
Backend: MLX / Metal  
Method: fixed codec tokens, warmup followed by 5 alternating runs per arm.

| Audio duration | Chunk / context | Before | After | Speedup |
|---:|---:|---:|---:|---:|
| 20.48s | 10 / 5 | 0.8058s | 0.7917s | **1.018x** |
| 90s | 300 / 25 | 1.5322s | 1.3783s | **1.112x** |

For the 90-second case:

- Stack calls: `827 → 4`
- Total stacked frames: `589,650 → 1,200`
- Code2Wav calls: unchanged at `4`
- Emitted samples: unchanged at `2,160,000`
- Output checksum difference: `0`

The improvement grows with sequence length. Short outputs that do not cross a streaming-window boundary are expected to see little or no change.

## Testing

- Added tests for streaming-window readiness and cursor boundaries.
- Compared repeated incremental windows and final flush against the previous full-history implementation.
- Verified matching output lengths and numerically equivalent waveforms.
- Ran 3 end-to-end audio-in/audio-out generations with 51 audio chunks per run and no failures.
- `python -m unittest mlx_vlm.tests.test_qwen3_omni_moe`
- `pre-commit run --all-files`

### DIFF
--- a/mlx_vlm/models/qwen3_omni_moe/qwen3_omni_moe.py
+++ b/mlx_vlm/models/qwen3_omni_moe/qwen3_omni_moe.py
@@ -664,6 +664,63 @@ class Model(nn.Module):
 
         return thinker_result, talker_wavs.astype(mx.float32)
 
+    @staticmethod
+    def _stream_decode_bounds(
+        codes_count: int,
+        decoded_len: int,
+        chunk_size: int,
+        left_context_size: int,
+    ) -> Optional[tuple[int, int, int]]:
+        context_len = min(left_context_size, decoded_len)
+        new_tokens = chunk_size - context_len
+        if codes_count - decoded_len < new_tokens:
+            return None
+        end = decoded_len + new_tokens
+        return decoded_len - context_len, end, context_len
+
+    def _decode_stream_window(
+        self,
+        codes_list: list[mx.array],
+        decoded_len: int,
+        chunk_size: int,
+        left_context_size: int,
+    ):
+        bounds = self._stream_decode_bounds(
+            len(codes_list), decoded_len, chunk_size, left_context_size
+        )
+        if bounds is None:
+            return None, decoded_len
+
+        context_start, end, context_len = bounds
+        codes_window = mx.stack(codes_list[context_start:end], axis=1).transpose(
+            0, 2, 1
+        )
+        wav_chunk, _ = self.code2wav.stream_decode(
+            codes_window,
+            chunk_size,
+            left_context_size,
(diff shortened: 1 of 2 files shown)


Lazarus-931:
hi @ssy, thanks for the pr! While perf speedup is important,  for the speedup reported, I don't think it's worth a pr.

cc: @Blaizzy

Blaizzy:
@Lazarus-931 I'm curious as to why you think that? To be clear, my intuition says the same. 

Could you run a numerical accuracy check?

sxy-trans-n:
@Lazarus-931 @Blaizzy 
Totally fair — this came out of some experimental work I was doing. I noticed a small win and thought I'd send it upstream.

For numerical accuracy, I also ran a sample-wise comparison using the real `mlx-community/Qwen3-Omni-30B-A3B-Instruct-4bit` Code2Wav weights and fixed valid codec tokens:

- 256 frames, chunk/context 10/5: 491,520 samples
- 1,125 frames, chunk/context 300/25: 2,160,000 samples

Both cases were bit-identical: `array_equal=True`, 0 mismatched samples, max/mean absolute error 0, and matching SHA-256 hashes. The PR test also compares repeated windows and the final flush against the old full-history path.

That said, the wall-clock gain is modest because the decode network dominates. The main benefit is removing the per-token quadratic stacking, which mostly matters for long-form streaming. If that isn't worth the additional surface area, I'm completely fine closing it and happy to defer to your judgment.

Lazarus-931:
@sxy-trans-n yeah, i recommend closing as well

cc: @Blaizzy

Issue #1538, opened 2026-07-08
Qwen3.6 MTP drafter norm shift reduces speculative acceptance; BPE detokenizer still strict-decodes UTF-8


ft-ioxcs:
## Summary

Follow-up from #1526 after testing the current `main` fixes locally.

The main Qwen3.6 target model regression appears fixed by #1528, but the Qwen3.6 MTP drafter path seems to still have a related norm-shift issue. This lowers MTP draft acceptance and reduces speculative decoding speedup.

I also noticed a separate BPE streaming detokenizer robustness issue where `BPEStreamingDetokenizer.add_token()` can still strict-decode incomplete UTF-8 byte sequences.

## Environment

Tested locally on Apple Silicon / macOS with:

```text
mlx==0.32.0
mlx-lm==0.31.3
mlx-vlm==0.6.4 plus local patches equivalent to the relevant current main fixes
mlx-audio==0.4.4
transformers==5.12.1
Python 3.12.13
```

## Issue 1: Qwen3.6 MTP drafter appears to still double-shift norm weights

Affected file on `main`:

```text
mlx_vlm/speculative/drafters/qwen3_5_mtp/qwen3_5_mtp.py
```

Current logic:

```python
if any(key.endswith(suffix) for suffix in norm_suffixes):
    if value.ndim == 1 and mx.issubdtype(value.dtype, mx.floating):
        value = value + 1.0
```

This looks similar to the target-model norm-shift problem fixed by #1528, but in the MTP drafter sanitizer.

The standalone MTP checkpoint:

```text
mlx-community/Qwen3.6-35B-A3B-MTP-4bit
```

appears to already be in converted MLX format. Its checkpoint keys look like:

```text
fc.weight
layers.0.input_layernorm.weight
layers.0.mlp.gate.weight
pre_fc_norm_embedding.weight
pre_fc_norm_hidden.weight
```

It does not use source/HF-style prefixes such as:

```text
mtp.*
model.language_model.*
```

So the MTP drafter norm weights appear to be shifted a second time. The target model still verifies the output, so generation remains coherent, but the draft model quality drops and speculative acceptance gets worse.

## Reproduction commands

Plain target generation:

```bash
mlx_vlm.generate \
  --model mlx-community/Qwen3.6-35B-A3B-4bit \
  --prompt "Write a 1000-word beginner tutorial about speculative decoding. Use short paragraphs and a numbered list of 8 steps." \
  --max-tokens 1024 \
[...]

PR #1554, opened 2026-07-10
fix(qwen3_5_mtp): guard norm +1.0 shift to raw-HF layout only

Lazarus-931:
The MTP drafter sanitize() shifted RMSNorm weights unconditionally, so pre-converted MLX MTP checkpoints (mlx-community/Qwen3.6-*-MTP-*, bare keys) were double-shifted, corrupting the drafter and lowering speculative acceptance. Shift only mtp.-prefixed (raw-HF) weights, mirroring the target model fix in #1528. Idempotent.

Refs #1538.

All work on existing checkpoints, I got the norm tensors from the safetensors headers of a 35b sized model and verified that sanitize doesn't add extra 1.0!

all tests green


Blaizzy (approved):
LGTM, thanks !

PR #1557, opened 2026-07-10
Move Qwen3 Omni MRoPE to request-owned kwargs

hellopahe:
The qwen3_omni_moe thinker still keeps MRoPE state on the language model instance - get_input_embeddings resets and rewrites _position_ids/_rope_deltas on every call, so concurrent requests through the batch/server paths can clobber each other's positions. The other Qwen VL families moved to request-owned position kwargs in #1494. 

This ports the module to the same pattern: get_input_embeddings computes position_ids/rope_deltas unconditionally and returns them on InputEmbeddingsFeatures, like qwen3_vl_moe. Text-only and audio-only inputs collapse the three identical axes to 2D, grid inputs keep the 3-axis form.

Two adjustments fall out of this: the audio-path helper from #1522 drops the prompt-shaped position_ids on decode steps (rope_deltas handles continuation there), and Thinker.__call__ was unpacking the features dataclass as a tuple, which throws - it now uses the object.

qwen3_omni_moe now runs the shared request-owned mrope assertion in test_models.py. 

AI assisted with the implementation.

Tested on M3 Max.
```
python -m pytest mlx_vlm/tests/test_qwen3_omni_moe.py -q   # 1 passed
python -m pytest mlx_vlm/tests/test_models.py -q -k omni   # 3 passed
python -m pytest mlx_vlm/tests/test_rope.py -q             # 9 passed, 2 failed — the same two fail on main (qwen3_5 fully-masked rows, ernie empty rows)
```

### DIFF
--- a/mlx_vlm/models/qwen3_omni_moe/qwen3_omni_moe.py
+++ b/mlx_vlm/models/qwen3_omni_moe/qwen3_omni_moe.py
@@ -226,6 +226,8 @@ class Model(nn.Module):
                 if k in ["image_grid_thw", "video_grid_thw"]
             }
         )
+        decode_lm_kwargs = dict(lm_kwargs)
+        decode_lm_kwargs.pop("position_ids", None)
 
         prompt_cache = kv_cache.make_prompt_cache(self.thinker.language_model)
         outputs = self.thinker.language_model(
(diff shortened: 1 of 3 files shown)


lucasnewman (approved):
LGTM

lucasnewman:
@hellopahe Can you [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits and force push so we can clear the merge requirements?

hellopahe:
@lucasnewman Done, signed and force pushed. Thanks for the review!

Issue #1546, opened 2026-07-09
SAM3: box/exemplar prompts accepted but never used — geometry_encoder (incl. boxes_pool_project ROI path) unwired in DetectorModel forward


bhushit:
**Version:** 0.6.4 (also present on current `main`)

## What happens

`Sam3Predictor.predict(image, text_prompt, boxes=...)` accepts box prompts and passes them down, but `DetectorModel.__call__` never uses them — the step commented `# 3. Combine text + optional geometry prompts` sets `prompt = inputs_embeds` and combines nothing. Box/exemplar prompting (a headline SAM 3 capability, and the weights for it ship in the checkpoints) is a silent no-op: same output with and without `boxes`.

Related: in `GeometryEncoder`, `boxes_pool_project` (the ROI-pooled visual-identity path), `prompt_layer_norm`, and `vision_layer_norm` are constructed (so weights load) but never called from anywhere; `encode_boxes` is only the direct `Linear(4, d)` coordinate projection.

## What we tried (empirical)

We hand-wired the geometry path externally two ways to see if the shipped pieces suffice:

1. `geo = geometry_encoder(boxes_norm_cxcywh, memory=src, memory_pos=pos_flat)`, concatenated to the text prompt + mask extended;
2. same, plus adding `label_embed` (box label) to each box token before the encoder layers.

Results on `mlx-community/sam3-bf16`:
- box + generic text ("object"): **0 detections at any threshold** — box tokens alone carry no concept;
- box + working text prompt: tokens *do* flow (top score shifts 0.903 → 0.889) but the boxed instance is never re-ranked to the top.

Consistent with the ROI-pooling path being the missing piece: without `boxes_pool_project`, a box token encodes coordinates but no visual content, so it can't act as an exemplar.

## Ask

Wire the geometry path in `DetectorModel.__call__` per the HF reference (encode boxes incl. the ROI-pooled features, apply the prompt/vision layer norms, concatenate to the text prompt for the DETR encoder/decoder + dot-product scoring). Happy to test on Apple Silicon — we have a real workload (text-ungroundable ornament motifs) where exemplar prompts are exactly what's needed.

PR #1563, opened 2026-07-10
sam3: implement box/geometry prompt support in the detector

Lazarus-931:
for SAM3, geometric encoder was a stub, was never implemented, so box prompts were silently ignored. 

so this pr introduces a implementation of HF's Sam3GeometryEncoder (direct projection + MLX roi_align + sine position encoding + label embedding, CLS appended at the end, pre-norm cross-attention layers)

for verification, I made sure that it matched transformers Sam3GeometryEncoder to max abs diff 1.5e-5 on identical weights/inputs, and tested e2eon mlx-community/sam3-bf16 on a box prompt, seeing that it now changes detector output which verifies that geometric encoder now works!

all tests green!

for #1546

### DIFF
--- a/mlx_vlm/models/sam3/generate.py
+++ b/mlx_vlm/models/sam3/generate.py
@@ -163,7 +163,17 @@ class Sam3Predictor:
 
         box_input = None
         if boxes is not None:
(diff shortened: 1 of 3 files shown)


lucasnewman (approved):
🚀

PR #1560, opened 2026-07-10
feat(apc): adapter boundary for batch/SWA kv-bits (#1559)

weklund:
## Summary

Follow-up to **#1534** / **#1174**, tracking **#1559**.

Lands a durable **internal APC adapter** for exact-mode APC + `--kv-bits` on the **batch / continuous-batching** path (SWA + multi-row), without new CLI flags or on-disk format changes (dequant-at-harvest float APC pool remains).

### What this PR does
- **`snapshot_prompt_cache_row`**: always row-normalize before exact store (removes the B=1 special case in `BatchGenerator`)
- **`BatchQuantizedKVCache.extract`** + **`BatchTurboQuantKVCache.extract`**: multi-row concurrent exact path
- **`BatchRotatingKVCache`**: exact support + fix S=1 prefill while right-pad (`_lengths`) is pending (unequal multi-row SWA warm)
- Clear **`batch_size` / `is_single_row()`** on batch cache types
- Additive **reject stats**: `rejects`, `rejects_by_reason`, `last_reject` on `/v1/cache/stats`
- **`layer_kv_for_apc`**: dequant-aware harvest so post-decode block store does not dense-slice quantized tuple keys
- Unit tests: `test_apc_adapters.py`

### Explicitly out of scope (tracked elsewhere)
- **Block-mode concurrent multi-row + kv-bits** (mask / mixed prefill; fails even without APC) → **#1562**
- Full product claim that APC + kv-bits is complete for all agent/server multi-row block workloads (see #1174 clarification)
- Startup self-check, optional `APC_TRACE`, formal `APCLayerView` type, quantized APC pool / RSS shrink under kv-bits
- User-facing “fully supported” docs until #1562 is resolved or wontfix’d with limitations

### Related
- #1559 — parent follow-up issue
- #1534 — single-request APC + kv-bits (merged)
- #1174 — original request (partial after #1534)
- #1562 — block multi-row + kv-bits

## Test plan
- [x] `pytest mlx_vlm/tests/test_apc_adapters.py` (+ quantized/exact/batch suites)
- [x] CI `Test PRs` green
- [x] Smoke (local): Gemma4-e2b batch exact store/hit + multi-row (incl. unequal suffix / turbo 3.5 & 4)
- [x] Smoke (local): Qwen3.5-4B hybrid multi-row concurrent + kv8 / turbo
- [x] Smoke (local): Qwen3-0.6B block sequential warm no regression; multi-row concurrent remains #1562
[...]

PR #1565, opened 2026-07-10
docs: APC works with KV-cache quantization

weklund:
## Summary

- Remove the outdated README claim that APC is skipped when KV-cache quantization is enabled (fixed in #1534 / #1560)
- Document that APC + `--kv-bits` work together, with a short note that the live cache stays quantized and the APC pool stores float K/V
- Add a combined server example (`APC_ENABLED=1` + `--kv-bits 8`)
- Cross-link from the KV Cache Quantization section

## Test plan

- [x] Visual check of APC + KV quant README sections
- [ ] N/A code tests (docs only)

Refs: #1174, #1534, #1560

### DIFF
--- a/README.md
+++ b/README.md
@@ -724,6 +724,14 @@ APC_NUM_BLOCKS=4096 \
 mlx_vlm.server --model Qwen/Qwen3-VL-4B-Instruct --port 8080
 ```
 
+APC works with KV-cache quantization (`--kv-bits`):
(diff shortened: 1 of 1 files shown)


weklund:
cc @lucasnewman @Blaizzy

Blaizzy (approved):
LGTM, thanks!

Issue #1559, opened 2026-07-10
Follow-up: APC + kv-bits full batch/SWA coverage (adapter boundary)


weklund:
## Summary

Follow-up to **#1534** (merged) / **#1174**.

#1534 enabled APC together with `--kv-bits` for the **single-request** path (and Qwen hybrid exact B=1 after the `BatchKVCache` clone fix). It is **not feature-complete** until we cover:

1. **SWA / Gemma 4** batch exact + kv-bits (`BatchRotatingKVCache`)
2. **Concurrent multi-row** exact + kv-bits (`BatchQuantizedKVCache` extract / row normalize)
3. A durable **internal adapter boundary** so the next cache dialect is not another one-off `isinstance` branch

**Contract:** implementation-only — same CLI / env / generate APIs. No new required flags. Prefer no on-disk format change (keep dequant-at-harvest float blocks unless we explicitly version a new layout later).

### What #1534 delivered
- APC no longer unconditionally disabled when `kv_bits` is set
- Dequant at harvest / requant on restore for **block-mode**
- Surgical exact-mode clone for **`BatchKVCache` (B=1)** hybrid layouts
- Warning log on unclonable exact-store rejects

### Known gaps (review + smoke)

| Gap | Source |
|-----|--------|
| SWA layer caches (`BatchRotatingKVCache`) not clonable for exact store on batch path | @lucasnewman review on #1534; smoke on `gemma-4-e2b-it-4bit` |
| Concurrent batched requests not fully covered | @lucasnewman; multi-row needs row extract on all batch quant types |
| `BatchQuantizedKVCache` has no `extract()` | Analysis on #1534; multi-row exact fails at extract |
| `_apc_prompt_cache_for_store` B=1 special case still a footgun | @pinglin on #1534 |
| Clone peeks at `_idx` / `keys.shape[0]` instead of clear APIs | @lucasnewman nit: use `empty()`, add `is_single_row()` (or similar) |
| Post-decode block harvest on `generate()` + kv-bits errors (`keys` is quantized tuple) | Smoke: `APC store failed: tuple indices...` (prefix reuse still works) |
| Rejects not in `/v1/cache/stats` (only logs) | @pinglin `stores_rejected` suggestion |
| APC float pool RSS does not shrink with kv-bits | @trevorgordon981 — docs only |

### Related
- #1174 — parent feature request
- #1534 — single-request APC + kv-bits (merged)
[...]

PR #1566, opened 2026-07-11
feat(apc): APC_TRACE + non-fatal layout self-check

weklund:
## Summary

Hygiene follow-up after #1534 / #1560 (tracker leftover from #1559 §C/§E):

- **`APC_TRACE=1`**: greppable store/reject/self-check log lines (same spirit as `APC_DISK_TRACE`)
- **Layout self-check**: dry-run `make_cache` → clone path used by exact store; classify unsupported types
- **Server**: runs self-check at model load when APC is enabled (**log-only**, never blocks serve)
- **README**: document `APC_TRACE` + self-check only (APC + kv-bits docs remain in **#1565**)

Design: optional env only; stats stay separate from trace; self-check reuses `_clone_cache_entry_for_apc` so checks match runtime store behavior.

## Test plan

- [x] `pytest mlx_vlm/tests/test_apc_observability.py`
- [x] `pytest mlx_vlm/tests/test_apc_adapters.py` (no regression)

Refs: #1559, #1174

### DIFF
--- a/README.md
+++ b/README.md
@@ -775,11 +775,12 @@ Common APC environment variables:
 | `APC_HASH` | `fast` | Set to `sha256` for a stable cryptographic hash |
 | `APC_TRACE` | unset | Set to `1` for greppable store/reject/self-check log lines |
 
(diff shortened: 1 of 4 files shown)


weklund:
After we made the APC work with quantized and hybrid caches, failures went from "APC is off" to "APC is on but it silently rejects or uses a weird layout". This PR is bubbling up those issues at load and can be found in logs. This doesn't change any serving behavior.

Blaizzy (approved):
LGTM, thanks!

PR #1570, opened 2026-07-11
add glm4_moe_lite_mtp drafter (GLM-4.7-Flash MTP split tool)

samithaj:
## Summary

Adds `speculative/drafters/glm4_moe_lite_mtp/`, a splitter that extracts
GLM-4.7-Flash's trained nextn (MTP) layer into a standalone drafter, mirroring
the `deepseek_v4_mtp` / `qwen3_5_mtp` split convention.

GLM-4.7-Flash (`glm4_moe_lite`) ships one trained nextn layer inside the target
checkpoint at `model.layers.<num_hidden_layers>.*` — a dedicated `embed_tokens`,
the `enorm` / `hnorm` / `eh_proj` projections, an MLA `self_attn`, a 64-expert
MoE + shared expert, and an untied `shared_head`. `split.py` writes those into a
flat, post-sanitize drafter folder:

- absorbed-MLA `kv_b_proj` split into `embed_q` / `unembed_out`,
- experts stacked into `switch_mlp`,
- the dedicated nextn `embed_tokens` and untied `lm_head` kept (GLM's nextn is
  not tied to the target, unlike DeepSeek/Qwen MTP),
- the `noaux_tc` router correction bias kept fp32 and the router gate weight
  left full precision (casting/quantizing them breaks routing),
- optional `--q-bits` / `--q-group-size` affine quantization,
- `config.json` with `model_type: "glm4_moe_lite_mtp"` and the source text
  config nested under `text_config`, plus tokenizer files.

`glm4_moe_lite_mtp` is registered in `DRAFTER_KIND_BY_MODEL_TYPE` as an `mtp`
kind.

## Scope: split tool + config only (runtime model class deferred)

Unlike `deepseek_v4_mtp` / `qwen3_5_mtp`, this ships the split tool and config
but no runtime drafter model class. mlx-vlm has no `glm4_moe_lite` backbone in
its model zoo yet, so it can neither load GLM-4.7-Flash as a target nor build a
sibling-style drafter — both `Qwen3_5MTPDraftModel` and `DeepseekV4MTPDraftModel`
are built directly on their in-zoo decoder layers (`models/qwen3_5`,
`models/deepseek_v4`). Adding the full GLM-4.7-Flash architecture (MLA +
`noaux_tc` MoE) is a separate, larger change, and a runtime drafter with no
target to speculate against would be unreachable today. The split output is the
useful, verifiable deliverable now; the runtime model class drops in cleanly
once the backbone lands.

One other deviation from the sibling splitters: quantization is applied inside
`split.py` (per-weight `mx.quantize` with the gate/norm/fp32-bias skips) rather
than by round-tripping through `mlx_vlm.convert`, because there is no loadable
`glm4_moe_lite_mtp` model to run convert against yet.

## Validation

I verified the split against the extraction tool from vllm-project/vllm-metal#484 (the reference
integration this drafter format feeds).

- **Tensor parity (bit-exact):** bf16 (24 tensors) and 4-bit (54 tensors) split
  outputs match the independent extraction tensor-for-tensor — identical keys,
  shapes, and dtypes, `max_abs_float_diff = 0.0`, and the packed uint32 weights
  / scales / biases compare exactly equal. (Container SHA differs only because
  `mx.save_safetensors` here writes `metadata={"format": "mlx"}` and a different
  on-disk tensor order; contents are identical.)
- **Strict load + finite forward:** both outputs `load_weights(strict=True)`
  into an `mlx_lm` `Glm4MoeLiteDecoderLayer`-based module and produce finite
  logits of shape `(num_slots, vocab_size)` (`argmax[0]` agrees across bf16 and
  4-bit).
- **Tests:** `test_split_glm4_moe_lite_mtp_flattens_nextn_layer` (kv_b_proj
  split, expert stacking, fp32 bias preservation, buffer drop, config shape) and
  `test_kind_none_autodetects_mtp_for_glm4_moe_lite_mtp`, alongside the existing
  drafter-kind suite. `black` + `isort` clean.

Only the shards holding the nextn tensors are read from the source repo (3 of 48
for GLM-4.7-Flash).

## Context

Source model `zai-org/GLM-4.7-Flash` is MIT-licensed and ungated; a hosted,
revision-pinned split artifact (bf16 + 4-bit) will follow, mirroring the hosted
`qwen3_5_mtp` / `deepseek_v4_mtp` heads.


Blaizzy (approved):
LGTM, thanks!

Blaizzy (approved):
LGTM, thanks!

Issue #1558, opened 2026-07-10
Text-only continuous batching: fully-left-padded prefill row crashes GatedDeltaNet ([reshape] Cannot infer the shape of an empty array) when co-batched prompt lengths differ by more than prefill_step_size — please merge #1373


ronm92130:
### Summary

Under `mlx_vlm.server` continuous batching, a **pure-text** request mix crashed the
GPU generation thread on Qwen3.6-35B-A3B (`qwen3_5_moe`):

```
ValueError: [reshape] Cannot infer the shape of an empty array
```

at `mlx_vlm/models/qwen3_5/language.py:1610` (main; 1601 in 0.6.3), the
`z = z.reshape(B, S, -1, self.head_v_dim)` in `Qwen3_5GatedDeltaNet.__call__`.
All in-flight requests received the error (same blast-radius behavior as #1346);
the server loop recovered afterward.

This is the same defect class that #1346 reported for mixed-length **vision**
batches and that open PR **#1373** ("Handle fully padded rows in Qwen 3.5 chunked
prefill", approved 2026-06-27) fixes — but reachable with **text-only traffic**, and
**not** covered by the empty-input request validation that closed #1485 (#1491):
every request in our incident had a normal, non-empty prompt.

We root-caused it, built a tiny unit repro (no weights needed), and validated
PR #1373 against it — details below. **Request: merge #1373** (or equivalent).

### Environment

- mlx-vlm **0.6.3**, mlx 0.31.2, mlx-lm 0.31.3
- Python 3.12.10, macOS 26.4.1 (25E253), Mac Mini M4, 32 GB unified memory
- Model: Qwen3.6-35B-A3B, 4-bit MLX conversion (`model_type=qwen3_5_moe`), text-only usage
- Server: `mlx_vlm.server` (uvicorn), continuous batching enabled, APC **disabled**,
  no kv-bits, no draft model, default `prefill_step_size` (2048)
- One occurrence in ~34 h of production serving; trigger conditions below

### Traceback (production, tail)

```
  File ".../mlx_vlm/server/generation.py", line 1340, in _run
    self._step(batch_gen, active)
  File ".../mlx_vlm/server/generation.py", line 1813, in _step
    prompt_responses, responses = batch_gen.next(**kwargs)
  File ".../mlx_vlm/generate/ar.py", line 2806, in next
    return self._next(**kwargs)
  File ".../mlx_vlm/generate/ar.py", line 2780, in _next
[...]

PR #1571, opened 2026-07-11
fix(qwen3_5): skip fully-padded rows in chunked batch prefill (#1558)

### LINKED ISSUE #1555: Structured-output logits processors become misaligned when generation batches are extended

Lazarus-931:
this fixes #1555 

When requests are getting batched, the structured req joins later, and grammer processor lands in the wrong row. To fix, I padded logits_processors with Nones up to the row count whenever any processor is active, same as token_context already does. 

Checked the issue repro lines up ([None, None, [proc]]) and test_generate.py stays green with a new regression test.

### DIFF
--- a/mlx_vlm/models/qwen3_5/language.py
+++ b/mlx_vlm/models/qwen3_5/language.py
@@ -716,6 +716,20 @@ def _pad_row_time(x: mx.array, pad: int, target_length: int) -> mx.array:
     )
 
 
+def _restore_batch_padding_metadata(cache_entry, offsets, steps: int):
+    if offsets is None:
+        return cache_entry
+    if not (
+        hasattr(cache_entry, "offset")
+        and hasattr(cache_entry, "left_padding")
+        and hasattr(cache_entry, "_idx")
+    ):
+        return cache_entry
+    cache_entry.offset = offsets + steps
+    cache_entry.left_padding = cache_entry._idx - cache_entry.offset
+    return cache_entry
+
+
 def _qwen3_5_left_padding_info(cache):
     left_padding = getattr(cache, "left_padding", None)
     if not (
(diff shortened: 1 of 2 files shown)


Blaizzy (approved):
LGTM, thanks @Lazarus-931!

Issue #1562, opened 2026-07-10
APC + kv-bits: staggered continuous-batching join crashes (last-layer type mismatch)


weklund:
## Summary

**Residual after #1568 / #1567:** concurrent **server** continuous batching with **`APC_ENABLED=1` + `--kv-bits`** can still crash when a second request joins an in-flight generation batch (staggered join). Pure multi-row + kv-bits **without** APC is fixed.

This is **not**:
- “APC disabled under kv-bits” (#1174 / #1534)
- Hybrid exact / SWA multi-row adapters (#1559 / #1560)
- TurboQuant **quality** under batch (#1569)

It **is** a type mismatch between:
1. Live gen path: `_make_cache` → last full-attn layer is **unquantized** `BatchKVCache` when `n > 2`
2. APC warm restore: `make_warm_batch_kv_cache` / `make_warm_batch_kv_cache_multi` → **all** layers `BatchQuantizedKVCache` when `kv_quant_config` is set

On join, `_extend_cache` → `BatchKVCache.extend` does `other.keys.shape` while warm keys are a **quantized tuple** →:

```text
AttributeError: 'tuple' object has no attribute 'shape'
```

→ HTTP 500 on concurrent clients.

---

### Fixed by #1568 (was original #1562 / split #1567)

| Path | Result |
|------|--------|
| Quant SDPA mask vs GQA scores, B≥2 | ✅ |
| `BatchQuantizedKVCache.prepare` / `finalize` | ✅ |
| `BatchGenerator.insert([a,b])` + `kv_bits=8`, APC off | ✅ |
| Same + unequal lengths, B=3, long decode | ✅ |
| Multi-model (Qwen3-0.6B, Qwen2.5-0.5B, Llama-3.2-1B, Qwen3-1.7B block; Qwen3.5-0.8B exact sanity) | ✅ |
| Server concurrent long prompts, **kv8, APC off** | ✅ HTTP 200 |
| Server concurrent, **float + APC** | ✅ HTTP 200 |

Original mask error (closed via #1567):

```text
ValueError: [broadcast_shapes] Shapes (2,1,…) and (2,H,G,…) cannot be broadcast
```

### Still broken (this issue)

| Path | Result |
|------|--------|
| Server concurrent long shared prefix, **kv8 + APC on** | ❌ HTTP 500 `tuple` / `.shape` |
| Server concurrent short prompts, kv8 + APC | ✅ (often no warm mid-join race) |
[...]

PR #1568, opened 2026-07-11
fix: multi-row continuous batching with KV quant (#1567)

weklund:
## Summary

Fixes concurrent continuous batching with `--kv-bits` (#1567).

Two overlapping chat requests against `mlx_vlm.server --kv-bits 8` crashed during prefill:

```text
ValueError: [broadcast_shapes] Shapes (2,1,18,18) and (2,8,2,18,18) cannot be broadcast.
```

**Cause:** Quantized GQA expands attention scores to 5D `(B, n_kv_heads, n_repeats, L, K)` while batch left-pad masks are 4D `(B, 1, L, K)`. Right-align broadcast aliases batch with heads and fails for `B > 1`.

**Fix:**
- `align_attention_mask_to_scores` — expand mask rank before the last two dims so it maps to score layout (SoC: attention owns score geometry)
- Wire into `quantized_scaled_dot_product_attention`
- `BatchQuantizedKVCache.prepare` / `finalize` — same multi-row pad lifecycle as `BatchKVCache`

## Test plan

- [x] `pytest mlx_vlm/tests/test_quant_sdpa_mask.py`
- [x] `pytest mlx_vlm/tests/test_batch_quantized_cache.py`
- [x] Live: two concurrent curls vs server `--kv-bits 8` on `Qwen3-0.6B-4bit` → both **HTTP 200** (was 500)

Fixes #1567  
Related: #1562

PR #1572, opened 2026-07-11
fix(attention): correct multi-row quantized KV attention (#1562)

Lazarus-931:
Concurrent multi-row prefill with --kv-bits crashes on a mask/GQA broadcast mismatch, and even past the crash quantized_matmul leaks state between batched rows. 

If we route batch quantized kv cache through dequantize + mx.fast.scaled_dot_product_attention, it leaves us a single-row on the fast path. 

I confirmed batch row 0 is bit-identical to standalone run and within 8-bit error diff from truth. 

all tests green including new one 

this is for #1562

### DIFF
--- a/mlx_vlm/models/base.py
+++ b/mlx_vlm/models/base.py
@@ -12,7 +12,7 @@ from PIL import Image
 
 from ..turboquant import BatchTurboQuantKVCache, TurboQuantKVCache
 from ..turboquant import _state_length as _turboquant_state_length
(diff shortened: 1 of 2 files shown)


Blaizzy:
@Lazarus-931 please check if #1568 solves this issue 👀

Dequantizing kills the gains

Lazarus-931:
@Blaizzy correct, thanks, will close this!

PR #1573, opened 2026-07-11
fix(apc): validate runtime caches and trace block stores

Blaizzy:
## Summary

Follow-up to #1566 that makes APC observability exercise the same cache layout and store paths used by the server at runtime.

- run the non-fatal APC self-check inside the generation worker using the runtime `_make_cache` factory
- include KV bits, group size, and quantization scheme so uniform and TurboQuant layouts are validated
- validate block and exact APC modes through their respective adapter contracts
- materialize exact-cache clones during the check so deferred MLX failures are observed
- emit `APC_TRACE store` for block-mode, memory, disk, reused, and layer-major outcomes
- record and trace unsupported block-harvest adapters instead of returning silently
- report final exact-store outcomes, including disk-only stores and disk scheduling failures

## Root cause

The original startup check called `language_model.make_cache()` on the FastAPI thread and always exercised the exact-cache clone adapter. Continuous generation builds a different effective cache layout in the GPU worker, including batch conversion and optional uniform/TurboQuant caches. The normal block store path also had no trace event, and unsupported harvest adapters returned an empty list without a reject reason.

## Impact

Operators now see trace events for the normal block APC path, and startup diagnostics reflect the effective server cache configuration without making failures fatal. Cache construction remains on the worker that owns MLX/GPU work.

## Validation

- `pre-commit run --files mlx_vlm/apc.py mlx_vlm/server/app.py mlx_vlm/server/generation.py mlx_vlm/tests/test_apc_observability.py`
- focused observability/adapters: 62 passed
- APC + quantization + server suite: 325 passed
- full non-smoke suite: 1463 passed, 3 skipped, 49 subtests passed

Runtime coverage populates and harvests both `BatchQuantizedKVCache` and `BatchTurboQuantKVCache`, then verifies the block-store trace.

### DIFF
--- a/mlx_vlm/apc.py
+++ b/mlx_vlm/apc.py
@@ -51,7 +51,7 @@ import time
 from collections import OrderedDict
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Any, Dict, Iterable, List, Optional, Sequence, Tuple
+from typing import Any, Callable, Dict, Iterable, List, Optional, Sequence, Tuple
 
 import mlx.core as mx
 import numpy as np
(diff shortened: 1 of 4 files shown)


lucasnewman:
@Lazarus-931 Do we still need this after your APC changes?

Lazarus-931:
> @Lazarus-931 Do we still need this after your APC changes?

yep, i think the work done for runtime check/observability changes are needed

PR #1576, opened 2026-07-11
feat(apc): APC_EXACT_MAX_PROMPT_TOKENS — cap prompt length eligible for exact-store

pinglin:
### Problem

APC exact mode snapshots the **entire prompt KV** of every completed prefill, and the exact cache is LRU by **entry count**, not by size. Since a snapshot's size scales linearly with prompt length (~20KB/token at fp16 on a Qwen3.5-class hybrid), a single cache entry can weigh anywhere from tens of MB (a short chat prefix) to ~2.6GB (a 128k-token document prompt) — a two-orders-of-magnitude spread. A count bound therefore doesn't bound memory at all: the same `APC_EXACT_CACHE_ENTRIES` setting costs N × ~0.1GB under one workload and N × ~2.6GB under another, decided purely by what traffic arrives.

Two consequences on memory-constrained hosts (e.g. 32–36GB Apple Silicon, where a 30B-class model's weights already take ~18–19GB):

1. **OOM aborts.** A client that sends a stream of *distinct* long prompts (document analysis, batch evaluation, RAG over large files) fills the exact cache with maximum-size snapshots that will never be hit — the prompts don't repeat — while live KV of in-flight requests sits on top. The total can cross the Metal ceiling and the server dies with `libc++abi: terminating due to uncaught exception ... [METAL] Command buffer execution failed: Insufficient Memory` (SIGABRT). Reproducible with a handful of concurrent ~100k-token requests on a 36GB host.
2. **GPU-thread stalls.** The exact store deep-copies the row's prompt cache on the generation thread; at doc scale that is a multi-GB copy per completed prefill, serializing against every other sequence's decode. Thread dumps under concurrent long-prompt load repeatedly catch the generation thread inside `_apc_prompt_cache_for_store`.

Neither is tunable today: `APC_EXACT_CACHE_ENTRIES` bounds count, and disabling APC entirely gives up warm-prefix restores for interactive traffic — the workload APC serves best.

### Change

`APC_EXACT_MAX_PROMPT_TOKENS` (env, default `0` = unlimited = stock behavior): in `_store_apc_exact_checkpoints`, a row whose `checkpoint_len` exceeds the cap marks `checkpoint_done` and skips the store. **Lookups are untouched** — an over-cap prompt can still hit a previously stored snapshot. ~10 lines in `mlx_vlm/generate/ar.py`; no API or on-disk format change.

Value guidance (in the README): set the cap just above the interactive prompt scale of the deployment, so recurring chat prefixes keep their warm restores while one-shot long prompts stop leaving multi-GB residue. E.g. a cap of 32768 keeps a ~27k-token chat prefix cacheable (measured warm restore of ~23k tokens: sub-second) while excluding 100k+ one-shot prompts — bounding the worst-case per-entry cost at ~0.65GB instead of ~2.6GB, i.e. worst-case cache residency drops ~4× at any entry count.

### Validation

- Unit (`mlx_vlm/tests/test_apc_store_cap.py`): env parsing (invalid/negative → 0), over-cap rows skip the store and are marked done (no re-check per step), under-cap rows store, `0`/unset is byte-identical to stock, mixed batches cap only over-cap rows.
- Load: with the cap set, a sustained mix of concurrent ~100k-token requests plus short repeated prompts runs without the OOM abort on a 36GB host, exact hits for the short prompts keep working, and `exact_stores` stays flat during the long-prompt phase.

### Relationship to existing work

- Complements #1534/#1560 (quantized caches become *storable*; this bounds *what* is worth storing — they compose).
- The durable fix is presumably a **size-budgeted** exact cache (LRU by bytes); this env cap is the minimal safe lever until then, and remains useful afterwards as a per-deployment policy knob.
- Happy to also add a `stores_rejected` counter to `/v1/cache/stats` in this PR (matches the #1559 gap-table row) so capped skips are observable.


pinglin:
Hi @Blaizzy — gentle nudge on this one when you have a moment. Would this prompt-length cap for exact-store be useful to mlx-vlm? I'm happy to update it based on your direction, or close the PR if the feature isn't of interest. Thanks!

pinglin:
CI style failure is fixed in 5f04fae: I applied the repository’s pinned isort output, and locally verified pre-commit run --all plus all 6 APC store-cap tests under Python 3.10. The replacement Actions run is currently awaiting maintainer approval: https://github.com/Blaizzy/mlx-vlm/actions/runs/31948171940

PR #1581, opened 2026-07-12
Fix AutoProcessor.register string-first calls (transformers 5.13 issue)

### LINKED ISSUE #1564: Loading `mlx-community/GLM-4.6V-nvfp4` with `transformers==5.13.0` logs an error
## Summary

Loading `mlx-community/GLM-4.6V-nvfp4` with `transformers==5.13.0` logs:

```text
Error registering glm4v_moe processor: 'str' object has no attribute '__module__'
```

This occurs during MLX-VLM's import-time processor registration and prevents a clean model load.

## Environment

- macOS on Apple Silicon: Apple M5 Max
- MLX-VLM: 0.6.5
- Transformers: 5.13.0
- Model: `mlx-community/GLM-4.6V-nvfp4`

## Reproduction

```python
from mlx_vlm.models.glm4v_moe import glm4v_moe
```

or load the model through the normal MLX-VLM API with Transformers 5.13 installed.

## Actual Behavior

`mlx_vlm/models/glm4v_moe/glm4v_moe.py` calls:

```python
AutoProcessor.register("Glm46VMoEProcessor", Glm46VMoEProcessor)
```

Transformers 5.13 expects a configuration class as the first argument to
`AutoProcessor.register`, not a string. Its registration internals therefore
access `__module__` on the string and raise the reported error.

The same unsupported string-registration pattern is also present in the
non-MoE GLM-4V integration:

```python
AutoProcessor.register("Glm46VProcessor", Glm46VProcessor)
```

## Expected Behavior

GLM-4.6V should load without processor-registration errors when used with
Transformers 5.13.

## Proposed Direction

MLX-VLM already installs a model-type-based `AutoProcessor.from_pretrained`
patch from each GLM processor module:

- `install_auto_processor_patch("glm4v_moe", Glm46VMoEProcessor)`
- `install_auto_processor_patch("glm4v", Glm46VProcessor)`

That patch handles dispatch using the model’s `config.json` `model_type`, so the
legacy string `AutoProcessor.register(...)` calls appear redundant. Removing
them would avoid the incompatible API call while preserving the intended
processor dispatch.

After adding regression coverage, the package constraint can be widened from:

```text
transformers>=5.5.0,<5.13.0
```

for compatibility with the 5.13 transformers release series.  (And actually earlier ones, that should not have accepted strings anyway.)

Lazarus-931:
During debugging #1564, realized seven other model modules experience the same AutoProcessor.register with a processor-class name strings as first arg. 

This pr delete sthe broken string-first AutoProcessor.register("Name", cls) calls from 7 model files. 

Verified on Apple Silicon across both transformers versions: transformers==5.13.0, all seven modules import cleanly, all seven processor re-exports work, and GLM-4.6V-nvfp4 dispatches to Glm46VMoEProcessor.

This fixes #1564

### DIFF
--- a/mlx_vlm/models/deepseek_vl_v2/__init__.py
+++ b/mlx_vlm/models/deepseek_vl_v2/__init__.py
@@ -1,4 +1,5 @@
 import mlx_vlm.models.deepseek_vl_v2.processing_deepsek_vl_v2  # noqa: F401 (installs processor patch)
 
 from .config import MLPConfig, ModelConfig, ProjectorConfig, TextConfig, VisionConfig
-from .deepseek_vl_v2 import DeepseekVLV2Processor, LanguageModel, Model, VisionModel
+from .deepseek_vl_v2 import LanguageModel, Model, VisionModel
+from .processing_deepsek_vl_v2 import DeepseekVLV2Processor
--- a/mlx_vlm/models/deepseek_vl_v2/deepseek_vl_v2.py
+++ b/mlx_vlm/models/deepseek_vl_v2/deepseek_vl_v2.py
@@ -4,15 +4,11 @@ from typing import Optional
 import mlx.core as mx
 import mlx.nn as nn
 import numpy as np
-from transformers import AutoProcessor
 
 from ..base import InputEmbeddingsFeatures
 from .config import ModelConfig, ProjectorConfig
 from .language import LanguageModel
-from .processing_deepsek_vl_v2 import DeepseekVLV2Processor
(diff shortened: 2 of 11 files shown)


Blaizzy (approved):
LGTM!

Can we relax the transformers pin ?

Lazarus-931:
yeah we should be able to once we merge 2-3 prs I'll make, we still import mlx-lm for mlx-lm's qwen 3.5 gated delta for example

PR #1593, opened 2026-07-14
vendor gated_delta kernel/ops from mlx-lm into qwen3_5

Lazarus-931:
Removes the module-level 'from mlx_lm.models.gated_delta import ...' in qwen3_5/gated_delta.py, which is really the last import-time mlx-lm dependency, which made importing qwen3_5/qwen3_5_moe/minicpmv4_6 crash under transformers>=5.13 

Verified that byte-identical output to mlx-lm's kernel+ops, qwen3_5 imports with mlx_lm poisoned, 36 qwen3_5/gated_delta tests pass.


Blaizzy (commented):
LGTM

PR #1594, opened 2026-07-14
vendor quantize/dequantize_model into mlx_vlm.quant_utils

Lazarus-931:
this pr adds mlx_vlm/quant_utils.py with quantize_model/dequantize_model (+ helpers) vendored from mlx-lm and reusing mlx-vlm's own switch_layers, so convert.py no longer imports mlx_lm.utils. 

part of the mlx-lm removal series.

all tests green


Blaizzy (approved):
LGTM

PR #1597, opened 2026-07-14
Add 1-bit inference kernel (Python-hosted)

Blaizzy:
## Summary

- add a Python-hosted Metal kernel for packed affine 1-bit matrix-vector inference
- use a lower-register PrismML-derived SIMD layout where each lane processes one 16-bit half-word and reuses the activation block across four outputs
- fuse Qwen 3.5's four decode projections into one no-copy 1-bit dispatch
- automatically replace checkpoint-backed 1-bit `Linear` and `Embedding` modules during model loading
- support group sizes 32, 64, and 128 with float16, bfloat16, and float32 activations
- document the checkpoint format and add kernel, layer, embedding, and loader replacement tests

## Why

PrismML's 1-bit support currently requires changes to MLX's compiled C++, CPU, and Metal backends. Hosting the inference kernel as an `mx.fast.metal_kernel` source string allows mlx-vlm to load compatible 1-bit checkpoints while continuing to use the standard MLX Python package.

## Impact

Existing affine 1-bit checkpoints using MLX's packed `uint32` layout, scales, and biases load without an extra inference flag. This is inference-only and currently covers standard `Linear` and `Embedding` modules; conversion from floating-point weights and specialized MoE linear modules remain out of scope.

## Validation

- `uv run --with pytest --with psutil pytest -q` — 1,610 passed, 3 skipped, 60 subtests passed
- `uv run --with pytest pytest -q mlx_vlm/tests/test_one_bit.py mlx_vlm/tests/test_utils.py` — 39 passed
- Qwen 3.5 native quantized fusion regression — passed
- `uv run --with ruff ruff check mlx_vlm/one_bit.py mlx_vlm/tests/test_one_bit.py mlx_vlm/utils.py`
- repository pre-commit hooks: black, isort, and autoflake
- `git diff --check`

## Performance

`prism-ml/Bonsai-27B-mlx-1bit`, 44-token prompt, 500 generated tokens on the same machine:

| | Original 1-bit | Optimized 1-bit | 2-bit reference |
|---|---:|---:|---:|
| Prompt | 10.002 tok/s | 47.683 tok/s | 32.623 tok/s |
| Generation | 8.540 tok/s | 35.719 tok/s | 28.628 tok/s |
| Peak memory | 5.558 GB | 5.558 GB | 9.917 GB |

The optimized kernel is approximately 4.2x faster than the original for generation and 25% faster than the reported 2-bit result, without increasing peak memory.


lucasnewman (approved):
LGTM

PR #1601, opened 2026-07-15
Avoid thinking budget decode synchronization

Blaizzy:
Supersedes #1600, which GitHub automatically closed when its source branch was renamed.

## Summary

- remove the per-token `mx.eval()` and GPU-to-CPU `.tolist()` conversion from continuous-batch thinking budget handling
- consume forced token IDs from CPU-side criterion state and apply overrides with an asynchronous MLX `where` operation only when needed
- preserve compatibility with custom thinking budget criteria
- serialize tokenizer preprocessing and budget-marker resolution, then queue prebuilt criteria to avoid concurrent fast-tokenizer `Already borrowed` failures
- add regression coverage for forced-token behavior, synchronization, and concurrent budget setup

## Root cause

`GenerationBatch.next()` materialized the asynchronously generated next-token array on every token whenever a thinking budget criterion was active. That synchronization broke the decode-ahead pipeline even before the budget was reached and caused a repeatable ~20% decode slowdown.

The server also constructed `ThinkingBudgetCriteria` on the GPU worker. Its tokenizer calls could race concurrent request preprocessing through the shared mutable fast-tokenizer backend, causing HTTP 500 responses with `RuntimeError: Already borrowed`.

## Impact

Using `prism-ml/Bonsai-27B-mlx-1bit`, 500 generated tokens, batch size 1, and the server's built-in `/v1/metrics` endpoint:

| Mode | Before | After samples | After mean |
| --- | ---: | --- | ---: |
| No budget | 34.89 / 34.78 tok/s | 35.18 / 34.30 / 33.48 tok/s | 34.32 tok/s |
| Budget 512 | 27.84 tok/s | 34.59 / 32.58 tok/s | 33.59 tok/s |

The deterministic ~20% regression is removed. Post-fix means differ by ~2.1%, with overlapping run-to-run and thermal variation.

## Batching performance

True simultaneous requests were measured with 96 generated tokens. Rates are aggregate steady-state decode throughput over the last 32 tokens, reported directly by `/v1/metrics`:

| Batch size | No budget | Budget 512 | Difference | Peak memory |
| ---: | ---: | ---: | ---: | ---: |
| 1 | 35.69 tok/s | 35.54 tok/s | -0.4% | 5.57 GB |
| 2 | 40.20 tok/s | 39.87 tok/s | -0.8% | 5.95 GB |
| 3 | 41.77 tok/s | 41.72 tok/s | -0.1% | 6.15 GB |
| 4 | 42.74 tok/s | 42.69 tok/s | -0.1% | 6.74 GB |

All concurrent requests completed successfully:

- batch 4, no budget: 4/4 HTTP 200
- batch 4, budget 512: 4/4 HTTP 200
- batch 4, forced budget 4: 4/4 HTTP 200

## Validation

- `uv run --with pytest pytest -q mlx_vlm/tests/test_generate.py mlx_vlm/tests/test_server.py` — 287 passed
- real-server simultaneous batch-4 checks for no budget, budget 512, and forced budget 4
- commit hooks: Black, isort, and autoflake passed
- `git diff --check` passed

### DIFF
--- a/mlx_vlm/generate/ar.py
+++ b/mlx_vlm/generate/ar.py
@@ -483,7 +483,7 @@ def generate_step(
             mx.clear_cache()
 
         if thinking_budget_criteria is not None:
-            forced_token_id = thinking_budget_criteria.apply_forced_token()
+            forced_token_id = thinking_budget_criteria.pop_forced_token_id()
             if forced_token_id is not None:
                 next_y = mx.array([forced_token_id], dtype=next_y.dtype)
         y, logprobs = next_y, next_logprobs
@@ -1251,7 +1251,7 @@ class GenerationBatch:
             ):
                 criteria = self.thinking_budget_criteria[i]
                 criteria(tok)
-                forced_next_tokens[i] = criteria.apply_forced_token()
+                forced_next_tokens[i] = criteria.pop_forced_token_id()
 
             if self.stop_criteria(tok):
                 finish_reason = "stop"
--- a/mlx_vlm/generate/ar.py
+++ b/mlx_vlm/generate/ar.py
@@ -1238,7 +1238,8 @@ class GenerationBatch:
 
         keep = []
         responses = []
-        forced_next_tokens = None
+        forced_next_tokens = [None] * len(self.uids)
+        fallback_next_tokens = {}
         for i in range(len(self.uids)):
             finish_reason = None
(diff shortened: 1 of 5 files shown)


Lazarus-931 (approved):
lgtm!

PR #1602, opened 2026-07-15
make mlx-lm optional for text-only fallback + text-model LoRA

Lazarus-931:
The last two mlx_lm touchpoints (text_only._get_classes, trainer.load_adapters) are the fallback for text architectures mlx-vlm has not vendored natively; they inherently use mlx-lm's model/tuner registries. 

Part of the mlx-lm removal series (#1509, #1516, etc).

Verified: works with mlx-lm present; clear error (not raw ImportError) when absent.

### DIFF
--- a/mlx_vlm/trainer/utils.py
+++ b/mlx_vlm/trainer/utils.py
@@ -321,7 +321,13 @@ def apply_lora_layers(model: nn.Module, adapter_path: str) -> nn.Module:
         nn.Module: The updated model with LoRA layers applied.
     """
     if getattr(model, "_is_text_model", False):
-        from mlx_lm.utils import load_adapters
+        try:
+            from mlx_lm.utils import load_adapters
+        except ImportError as e:
+            raise ImportError(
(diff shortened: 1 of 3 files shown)


mlx_vlm/trainer/utils.py
```diff
@@ -321,7 +321,13 @@ def apply_lora_layers(model: nn.Module, adapter_path: str) -> nn.Module:
         nn.Module: The updated model with LoRA layers applied.
     """
     if getattr(model, "_is_text_model", False):
-        from mlx_lm.utils import load_adapters
+        try:
```
Blaizzy: I think here we already have this ported over

Blaizzy (approved):
LGTM

mlx_vlm/trainer/utils.py
Lazarus-931: yeah saw, updated!

lucasnewman (approved):

Lazarus-931:
closing since prs after ard included this, would regress refractoring

PR #1606, opened 2026-07-15
Vendor qwen & deepseek text models 

Lazarus-931:
text_only fell back to mlx_lm.utils._get_classes for text-only architectures mlx-vlm has no native implementation of, so loading these plain LLMs required mlx-lm at runtime. 

This is just moving for in-house support for these models: qwen, qwen2_moe, qwen3_moe, qwen3_next, deepseek, deepseek_v2, plus their shared helpers . 

all tests green and verified model paths work!

big change, but all models work identically if had been using mlx-lm


lucasnewman (approved):
LGTM

PR #1607, opened 2026-07-15
share MLP implementations getting refractored

Lazarus-931:
Some MLP variants are called multiple times by diff. models. This change makes it so that repeatidly called mlp variants get impl once and calls several times, for much cleaner organization.  

tests and model text generation are identical and all green!


Blaizzy (approved):
LGTM, thanks!

PR #1608, opened 2026-07-15
share the vision multi-head Attention used 6 times

Lazarus-931:
The generic vision-tower attention (dims/num_heads multi-head attention) was duplicated across gemma3, aya_vision, idefics3, lfm2_vl, paligemma and phi4_siglip vision towers. Add mlx_vlm/models/attention.py with VisionAttention and repoint those six vision modules to import it. Behavior-preserving pure extraction.

tests & model text generation green!

### DIFF
--- a/mlx_vlm/models/attention.py
+++ b/mlx_vlm/models/attention.py
@@ -1,0 +1,52 @@
+from typing import Optional
+
+import mlx.core as mx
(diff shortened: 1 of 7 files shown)


Blaizzy (approved):
LGTM, thanks!

PR #1592, opened 2026-07-14
Prevent cross-thread MLX stream errors in multimodal generation

byoan:
## Summary

Fixes #1591 

Multimodal preprocessing can produce lazy MLX arrays associated with the request caller's stream.

`ResponseGenerator` queues these arrays for consumption by its dedicated generation thread. When that thread evaluates an array associated with the caller's stream, MLX raises:

```text
RuntimeError: There is no Stream(gpu, N) in current thread.
```

This was reproduced with `Qwen3.5-9B-MLX-4bit` and a mixed text/image request.

## Changes

- Recursively collect MLX arrays from preprocessed request inputs using `mlx.utils.tree_flatten`
- Materialize the collected arrays with a single `mx.eval()` call
- Perform materialization after context validation and immediately before inserting the request into the generation queue
- Add a regression test that consumes a lazy preprocessing array from another thread through `ResponseGenerator.generate()`

## Why Materialize at the Queue Boundary

`ResponseGenerator.generate()` performs preprocessing on the caller thread, while `BatchGenerator` and model execution run on a dedicated generation thread.

The request queue is therefore the ownership boundary between those threads.

Materializing the arrays immediately before that boundary ensures that no lazy MLX graph carrying caller-thread stream state is transferred to the generation thread.

Materialization happens after context-budget validation, so rejected requests do not perform unnecessary MLX evaluation.

## Reproduction Before the Fix

A normal multimodal request fails in the generation thread:

```text
File "mlx_vlm/models/qwen3_5/language.py", line 2306, in get_rope_index
    row_mask = attention_mask[i].tolist()
RuntimeError: There is no Stream(gpu, 2) in current thread.
```

The caller then remains blocked waiting for the generation context.

## Result After the Fix

The same Qwen3.5 mixed-image request completes normally:

```text
This image is completely blank, showing only a solid white background with no visible content or details.
```

## Testing

The tests were run using the same VLM runtime environment used for the model reproduction.

```text
mlx_vlm/tests/test_server.py::TestResponseGenerator
52 passed
```

The end-to-end reproduction was also run with:

```text
Model: lmstudio-community/Qwen3.5-9B-MLX-4bit
Input: mixed text and generated PNG image
Prompt tokens: 614
Result: generation completed successfully
```

PR #1610, opened 2026-07-15
Preserve MLX attention masks during preprocessing

### LINKED ISSUE #1591: Multimodal ResponseGenerator requests can fail with cross-thread MLX stream errors
## Summary

Multimodal requests submitted through `ResponseGenerator.generate()` can fail when preprocessed MLX arrays are transferred from the caller thread to the dedicated generation thread.

With Qwen3.5 and mixed text/image input, the generation thread raises:

```text
RuntimeError: There is no Stream(gpu, N) in current thread.
```

The generation thread terminates the request processing, while the caller remains blocked waiting for the generation context.

Text-only requests using the same model do not reproduce the issue.

## Reproduction

The following script uses the regular `ResponseGenerator` and `apply_chat_template` APIs. It creates a temporary image, submits a multimodal request, and consumes the generated tokens.

```python
import sys
import tempfile
from pathlib import Path

from PIL import Image

from mlx_vlm.server import GenerationArguments, ResponseGenerator


if len(sys.argv) != 2:
    raise SystemExit(f"Usage: {sys.argv[0]} /path/to/Qwen3.5-9B-MLX-4bit")

generator = ResponseGenerator(model_path=sys.argv[1])

try:
    _, processor, config = generator.wait_until_ready(timeout=300)

    with tempfile.TemporaryDirectory() as directory:
        image_path = Path(directory) / "image.png"
        Image.new("RGB", (64, 64), "white").save(image_path)

        prompt = apply_chat_template(
            processor,
            config,
            "Describe this image in one sentence.",
            num_images=1,
            enable_thinking=False,
        )

        _, tokens = generator.generate(
            prompt=prompt,
            images=[str(image_path)],
            args=GenerationArguments(
                max_tokens=32,
                temperature=0,
            ),
        )

        print("".join(token.text or "" for token in tokens))
finally:
    generator.stop_and_join()
```

Run it with a downloaded Qwen3.5 model:

```bash
python reproduce.py /path/to/Qwen3.5-9B-MLX-4bit
```

## Actual Result

The dedicated generation thread fails while reading the multimodal attention mask:

```text
Error in generation thread
  File "mlx_vlm/server/generation.py", line 1377, in _run
    input_ids, gen_kwargs = self._gpu_embed(raw_inputs, images)
  File "mlx_vlm/server/generation.py", line 1226, in _gpu_embed
    embed = self.model.get_input_embeddings(
        input_ids, pixel_values, mask=mask, **data_kwargs
    )
  File "mlx_vlm/models/qwen3_5/qwen3_5.py", line 110, in get_input_embeddings
    position_ids, rope_deltas = self.language_model.get_rope_index(
        input_ids, image_grid_thw, video_grid_thw, mask
    )
  File "mlx_vlm/models/qwen3_5/language.py", line 2306, in get_rope_index
    row_mask = attention_mask[i].tolist()
RuntimeError: There is no Stream(gpu, 2) in current thread.
```

The exact stream number may vary.

After this error, `ResponseGenerator.generate()` remains blocked waiting for the generation context and the script must be interrupted.

## Expected Result

The preprocessed inputs should be usable by the dedicated generation thread and the request should complete normally.

For example:

```text
This image is completely blank, showing only a solid white background with no visible content or details.
```

## Cause

`ResponseGenerator.generate()` performs request preprocessing on the caller thread:

```python
raw_inputs = self._preprocess_request(prompt, images, audio, videos)
```

It then places those inputs into `QueuedGenerationRequest` for consumption by the dedicated generation thread.

Multimodal preprocessing can return lazy MLX arrays. In this reproduction, the `attention_mask` remains associated with an MLX stream belonging to the caller thread.

When the generation thread evaluates that array through `attention_mask[i].tolist()`, the original stream is unavailable in the current thread and MLX raises the runtime error.

The thread ownership boundary is the insertion of the preprocessed inputs into the generation request queue.

## Environment

```text
macOS: 26.4.1
Python: 3.13.12
MLX: 0.32.0
mlx-vlm: 0.6.5
Upstream commit: dab4cb455b2c529993754e5b19b8310643c65558
Model: lmstudio-community/Qwen3.5-9B-MLX-4bit
```

Blaizzy:
## Summary

Closes #1591.

- Preserve an `attention_mask` that a model processor already returned as an MLX array instead of wrapping it in another `mx.array()` operation.
- Add a regression test that consumes the processed attention mask from a different thread.

## Root cause

Qwen3.5 uses the custom `Qwen3VLProcessor`, whose direct `input_ids`, `attention_mask`, `pixel_values`, and `image_grid_thw` outputs are materialized and safe to consume across threads.

The generic `prepare_inputs()` path subsequently called `mx.array(inputs["attention_mask"])` even when the value was already an MLX array. That redundant conversion created a lazy graph associated with the request caller's MLX stream. When `ResponseGenerator` handed the result to its dedicated generation thread, Qwen3.5 evaluated the mask in `get_rope_index()` and raised `RuntimeError: There is no Stream(gpu, N) in current thread`.

Preserving existing MLX arrays removes the lazy re-wrap at its source without forcing eager evaluation of every preprocessed request input.

## Impact

Mixed text/image requests using Qwen3.5 through `ResponseGenerator` now complete instead of failing in the generation thread and leaving the caller blocked waiting for a generation context.

This supersedes #1592 with a narrower fix based on tracing the custom processor's direct outputs.

## Validation

- Reproduced the failure with `Qwen/Qwen3.5-0.8B` before the fix.
- Confirmed the same multimodal request completes after applying only this `prepare_inputs()` change.
- `230 passed`: utility tests, `TestQwen3VLProcessor`, and the full server test module.
- Black, isort, autoflake, and `git diff --check` pass.

### DIFF
--- a/mlx_vlm/tests/test_utils.py
+++ b/mlx_vlm/tests/test_utils.py
@@ -3,6 +3,7 @@ import json
 import struct
 from io import BytesIO
 from pathlib import Path
+from threading import Thread
 from types import SimpleNamespace
 from unittest.mock import MagicMock, patch
 
@@ -404,6 +405,37 @@ def test_prepare_inputs():
     )
     assert "input_ids" in inputs
     assert mx.array_equal(inputs["input_ids"], mx.array([[1, 2, 3]]))
+
+
+def test_prepare_inputs_preserves_mlx_attention_mask_for_thread_handoff():
+    attention_mask = mx.array([[1, 1]], dtype=mx.int32)
+
+    class Processor:
+        tokenizer = SimpleNamespace(pad_token="[PAD]", eos_token="[EOS]")
+
+        def __call__(self, text=None, images=None, padding=None, return_tensors="mlx"):
+            return {
+                "input_ids": mx.array([[1, 2]], dtype=mx.int32),
+                "attention_mask": attention_mask,
+                "pixel_values": mx.zeros((1, 2), dtype=mx.float32),
+            }
+
+    inputs = prepare_inputs(
+        Processor(),
+        prompts="test <image>",
+        images=mx.zeros((3, 8, 8)),
+    )
+    consumed = []
+
+    def consume_attention_mask():
+        consumed.append(inputs["attention_mask"].tolist())
+
+    worker = Thread(target=consume_attention_mask)
+    worker.start()
+    worker.join(timeout=1)
+
+    assert inputs["attention_mask"] is attention_mask
+    assert consumed == [[[1, 1]]]
 
 
 def test_process_inputs_with_fallback():
--- a/mlx_vlm/utils.py
+++ b/mlx_vlm/utils.py
@@ -1760,8 +1760,11 @@ def prepare_inputs(
             inputs["pixel_values"] = inputs["images"]
             inputs.pop("images")
 
+        attention_mask = inputs.get("attention_mask")
         model_inputs["attention_mask"] = (
-            mx.array(inputs["attention_mask"]) if "attention_mask" in inputs else None
+            attention_mask
+            if attention_mask is None or isinstance(attention_mask, mx.array)
+            else mx.array(attention_mask)
         )
 
         # Convert inputs to model_inputs with mx.array if present


lucasnewman (approved):

Lazarus-931 (commented):
lgtm!

Issue #1604, opened 2026-07-15
Feature: OpenAI tool/function calling support in mlx_vlm.server (currently ignored; sending `tools` wedges the server)


dvictor357:
## Summary
`mlx_vlm.server`'s `/v1/chat/completions` endpoint does not support OpenAI-style **tool / function calling**. It doesn't read the `tools` / `tool_choice` request fields and never emits `tool_calls` in responses (`server.py` contains no tool/function handling). Worse, sending a `tools` payload currently returns an **empty response and can wedge the server** (subsequent requests then return HTTP 000).

This makes the server unusable as an OpenAI-compatible backend for agent frameworks that drive multi-step tool use, even though the underlying models (e.g. Qwen3.5 family) support tool calling natively via their chat templates.

## Environment
- mlx-vlm **0.6.3**, mlx 0.32.0, Python 3.13, macOS (Apple Silicon)
- Model: `qwen3_5`-arch models (e.g. `Qwen3.6-27B-MLX-4bit`, `prism-ml/Ternary-Bonsai-27B-mlx-2bit`)

## Current behavior
```
curl http://localhost:8080/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model":"...",
  "messages":[{"role":"user","content":"Weather in Jakarta? Use get_weather."}],
  "tools":[{"type":"function","function":{"name":"get_weather",
            "parameters":{"type":"object","properties":{"city":{"type":"string"}}}}}],
  "tool_choice":"auto","max_tokens":80
}'
```
→ empty body / no `tool_calls`; a follow-up plain request then returns HTTP 000 (server wedged).

## Requested behavior
1. Parse `tools` and `tool_choice` from the request.
2. Render them into the prompt via the model's chat template (tools kwarg / tool role).
3. Parse the model's tool-call output back into OpenAI `tool_calls`, with `finish_reason: "tool_calls"` (and streaming deltas).
4. At minimum, if tool calling won't be implemented, the server should **not hang** when `tools` is present — return a clear error instead.

## Why it matters
Agent frameworks (pi, and other OpenAI-tools clients) rely on tool calling to drive multi-step workflows. Without it, mlx_vlm.server can only do single-shot chat/vision generation, and local models can't be used for agentic/tool-using tasks even when the model itself is capable.

Blaizzy:
[...]

PR #1611, opened 2026-07-15
Add Chat Completions tool choice support

Blaizzy:
## Summary

Add OpenAI-compatible `tool_choice` handling to `/v1/chat/completions`.

- declare `tools` and `tool_choice` on the chat request schema
- support `auto`, `none`, `required`, and forced named functions
- disable tool exposure and parsing when `tool_choice` is `none`
- restrict forced choices to the selected function
- validate invalid choices, missing tools, and unknown function names
- forward `tool_choice` to model chat templates
- add fallback prompt guidance for templates that do not consume `tool_choice`

This addresses the missing `tool_choice` portion of #1604.

## Root cause

The Chat Completions endpoint forwarded `tools` into the model chat template but did not declare, validate, forward, or enforce `tool_choice`. Some templates, including Qwen3.5/Bonsai, expose tools but do not consume a `tool_choice` template argument.

## Behavior

- `none` is enforced server-side by omitting tools and disabling tool-call parsing.
- forced named choices expose only the selected function.
- `required` and forced choices receive explicit prompt guidance when the model template lacks native enforcement.

Prompt-guided enforcement remains model-dependent for small models under directly conflicting user instructions; native constrained decoding is outside this change.

## Validation

- `301 passed`: server, Gemma 4 tool parser, and MiniMax M3 tests
- live non-streaming and streaming checks with:
  - `prism-ml/Ternary-Bonsai-27B-mlx-2bit`
  - `Qwen/Qwen3.5-0.8B`
  - `google/gemma-4-E4B-it`
- pre-commit hooks: Black, isort, and autoflake

### DIFF
--- a/mlx_vlm/server/openai.py
+++ b/mlx_vlm/server/openai.py
@@ -157,6 +157,110 @@ def _ensure_effective_input(messages, *, images=None, audio=None):
     raise HTTPException(status_code=400, detail=_MISSING_INPUT_DETAIL)
 
 
+def _tool_function_name(tool: Any) -> Optional[str]:
+    if hasattr(tool, "model_dump"):
+        tool = tool.model_dump(exclude_none=True)
+    if not isinstance(tool, dict) or tool.get("type") != "function":
+        return None
+    function = tool.get("function")
+    if hasattr(function, "model_dump"):
+        function = function.model_dump(exclude_none=True)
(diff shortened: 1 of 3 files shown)


lucasnewman (approved):
LGTM

Lazarus-931 (commented):
lgtm!

PR #1613, opened 2026-07-15
Fix blocking continuous-batch prompt prefill

### LINKED ISSUE #1603: Server: continuous-batching prefill deadlocks (hangs at ~0% CPU, never returns) on prompts needing multiple prefill steps
# Server: continuous-batching prefill deadlocks (hangs at ~0% CPU, never returns) on prompts that need multiple prefill steps

## Summary
`mlx_vlm.server` hangs indefinitely on a `/v1/chat/completions` request when the prompt is large enough to require **multiple prefill steps** at the default `--prefill-step-size 2048`. The request never returns, **no error is logged**, and the server process sits at **~0% CPU** (it is not computing — it is deadlocked, not slow). Small prompts, streaming, and concurrent small requests all work fine.

## Environment
- mlx-vlm **0.6.3**, mlx 0.32.0
- Python 3.13, macOS (Apple Silicon)
- Model: `prism-ml/Ternary-Bonsai-27B-mlx-2bit` (`qwen3_5` arch). Behavior appears model-independent — the trigger is prompt length / number of prefill steps.

## Steps to reproduce
1. Start the server with defaults:
   ```
   mlx_vlm.server --model <qwen3_5 model> --port 8080
   ```
2. Send a `chat/completions` request whose prompt is ~8k+ tokens (≥4 prefill steps at the default 2048), `stream: false`:
   ```
   curl http://localhost:8080/v1/chat/completions -H 'Content-Type: application/json' \
     -d '{"model":"...","messages":[{"role":"user","content":"<~8000 tokens of text>"}],"max_tokens":20}'
   ```
3. The request **hangs forever**. Server CPU stays ~0%. No completion/error line is logged.

Small prompts (single prefill step) return normally; streaming and two concurrent small requests also work.

## Evidence it is a deadlock, not slow prefill
- `ps` shows the server at ~0.5% CPU during the hang; `utime` grows only ~0.4s over 110s (idle, not GPU-bound).
- `sample <pid>` shows the main asyncio thread parked in `select_kqueue_control_impl → kevent` (event loop idle) while the generation worker thread is blocked — the generation loop never steps.

## Workaround
Set `--prefill-step-size` ≥ the prompt token length so it prefills in one (or very few) steps:
```
mlx_vlm.server --model ... --port 8080 --prefill-step-size 16384
```
With this, large prompts complete (verified: ~10k-token prompt in ~80s, ~20k in ~158s). The deadlock only appears when the prompt spans many prefill steps.

## Expected
Large prompts should prefill across multiple steps and complete, the same way small prompts do.

Blaizzy:
## Summary

- schedule intermediate continuous-batching prompt-cache evaluation asynchronously
- restore asynchronous final-prefill evaluation in `generate_step`, reverting #1484
- keep final continuous-batch transitions and exact APC snapshots materialized at their ownership boundaries
- require MLX 0.32.0 and refresh the lockfile while retaining `mlx-lm`

## Root cause

`PromptProcessingBatch.prompt_step()` synchronously evaluated every cache state after every prefill chunk. For prompts spanning multiple chunks, that blocks the continuous-batching worker at each intermediate boundary instead of allowing Metal work and scheduler progress to overlap. This can leave the server generation loop unable to advance long prefill requests.

Later prompt chunks consume the same cache arrays and preserve their dependency order. Exact APC is the path that copies checkpoints into storage, so its clone path synchronously materializes those copied tensors before they cross the storage boundary. The final continuous-batch prefill-to-decode transition remains synchronous as well.

This PR also restores `mx.async_eval(y, logprobs)` at the single-request `generate_step` final-prefill boundary, reverting #1484 after direct long-lived CLI and server stress tests could not reproduce its token-0 collapse with public 4-bit Qwen3-VL or Qwen3.5 checkpoints on either MLX 0.31.2 or 0.32.0.

Fixes #1603.

## Validation

- `uv run --locked --with pytest python -m pytest -q mlx_vlm/tests/test_generate.py::test_generate_step_schedules_final_prefill_async mlx_vlm/tests/test_generate.py mlx_vlm/tests/test_apc_exact_mode.py --disable-warnings --maxfail=1` — 116 passed
- MLX 0.32.0, Qwen3.5-0.8B 4-bit, direct CLI with async final prefill — 500 consecutive image generations, zero collapse or errors
- MLX 0.32.0, Qwen3.5-0.8B 4-bit, server with async final prefill — 500 consecutive image requests, zero collapse or errors
- MLX 0.31.2 async control, Qwen3.5-0.8B 4-bit — 150 consecutive image generations, zero collapse or errors
- MLX 0.32.0, Qwen3-VL 2B 4-bit, direct CLI and server async stress — zero collapse or errors


lucasnewman (approved):

Lazarus-931 (approved):
lgtm

PR #1616, opened 2026-07-16
vendor Gemma/Phi/GLM/Cohere text models and reuse shared MLPs

Lazarus-931:
follows the same principle as having single attn and mlp variants called instead of identical layers implemented in each model, also vendoring some more models

tests & model text generation works just fine


Blaizzy:
Thanks! 

Could we not add a text_models folder  :)

Blaizzy (changes requested):
.

Blaizzy:
The text models don’t currently follow the structure we use elsewhere. Could you please take a look at Laguna, DSV4, and glm_moe_dsa as references?

Also, we use ModelConfig rather than ModelArgs.

Blaizzy (approved):
LGTM, thanks!

Issue #1619, opened 2026-07-17
Qwen3-Omni: decode steps lose the multimodal RoPE delta — generation degenerates instantly ("the the the…")


pwikstrom:
## Summary

With `qwen3_omni_moe` models (tested: `mlx-community/Qwen3-Omni-30B-A3B-Instruct-8bit`
and `-4bit`), any multimodal generation (image or video input) degenerates into
repetition from the first generated token — e.g. `the the the the…`, `8888888…`,
or an immediate empty EOS. Text-only prompts work fine. This reproduces with the
official CLI command from the model card, on both v0.6.4, v0.6.5 and current git main.

## Root cause (traced by instrumenting the model)

The generic generate loop calls the thinker's `LanguageModel` directly during
decode, bypassing the code that computed the correct multimodal M-RoPE positions
at prefill:

1. **The rope delta is never carried into decode.** At prefill,
   `get_rope_index` computes position_ids with a multimodal delta (in our trace:
   positions 0–144 with delta −5588 for a 5,739-token multimodal prompt), but
   the language model never stores that delta in `self._rope_deltas`. The first
   decode step then receives position **0**, and subsequent steps receive the raw
   `cache_offset` (5734, 5735, …) instead of `cache_offset + delta`
   (145, 146, …). The rotary embeddings are wrong from the first generated token.

2. **Chunked prefill passes the full-prompt position tensor to every chunk.**
   Each 2048-token prefill chunk receives the entire 5,739-length position_ids,
   so most of the multimodal context is rotated with wrong positions as well.

3. (Aggravating, video path) at the default `fps=2.0` a ~64 s video becomes 127
   tiny frames, which still breaks generation even with corrected positions;
   `fps=0.5` (fewer, higher-resolution frames, similar total tokens) works.

## Reproduction

```sh
pip install mlx-vlm==0.6.5
python -m mlx_vlm generate \
  --model mlx-community/Qwen3-Omni-30B-A3B-Instruct-4bit \
  --image any_photo.jpg \
  --prompt "Describe this image." --max-tokens 100 --temperature 0
# → "the the the the…" (or empty output / immediate EOS)
```

Text-only with the same model generates normally.

## Workaround (monkey-patch, verified)

Wrapping `qwen3_omni_moe.language.LanguageModel.__call__` to (a) stash the rope
delta whenever a multimodal prefill happens (recomputing via `get_rope_index`
when needed, reset-proof against the `pixel_values is not None` clear), (b) slice
[...]

PR #1621, opened 2026-07-17
fix(qwen3-omni): keep mRoPE positions correct across prefill and decode

Lazarus-931:
Two multimodal M-RoPE bugs raised in #1619:

When the prompt has an image or video, the model works out the special token positions during the first pass but forgets to save the offset it needs afterward. This fixes it by saving the offset in the first pass. 

Also, for long prompts the model reads the input in chunks instead of all at once. Since it handed the position list for the entire prompt, second prompt and onwards began counting from 0 and got rotated with wrong position. To fix, this pr trims the pos. list down to each chunk's window so each sees it's correct position.

All test green!


Blaizzy (approved):
LGTM, thanks!

PR #1623, opened 2026-07-17
Fix APC exact cache short prompt poisoning

### LINKED ISSUE #1617: APC exact mode: a short first prompt stores a 1-token snapshot that permanently disables checkpointing for that tenant
## Summary

When the first prompt for an APC tenant is shorter than `exact_cache_guard_tokens` (default 16), `adjust_prefix_to_text_suffix_boundary` clamps the desired prefix length up to **1**, so APC stores a **1-token snapshot**.

That snapshot matches the start of *every* subsequent prompt (they all begin with the same token, e.g. `<|im_start|>`), which makes `reused_prefix_len = 1`. Since new checkpoints are only stored when `reused_prefix_len == 0`, **APC never stores a useful snapshot for that tenant again** — for the lifetime of the process, and beyond, because the 1-token entry is also written to the disk tier and survives restarts.

Net effect: a chat session that begins with a short message (`"hi"`, `"привет"`, `"thanks"` — i.e. the most common way to open a conversation) pays a **full prefill on every single turn**, with no error and no log line.

Affected: **0.6.3 and 0.6.4** (verified: `adjust_prefix_to_text_suffix_boundary` is byte-identical between the two).

## Mechanism

1. **Guard default is 16** — `apc.py:2782`:
   ```python
   self.exact_cache_guard_tokens = max(
       1, int(os.environ.get("APC_EXACT_PREFIX_GUARD_TOKENS", "16"))
   )
   ```

2. **Checkpoint length is computed as `len(prompt) - guard`** — `generate/dispatch.py:1004`:
   ```python
   if apc_manager is not None and apc_mode == "exact" and reused_prefix_len == 0:
       exact_checkpoint_len = _apc.adjust_prefix_to_text_suffix_boundary(
           full_input_ids_list,
           len(full_input_ids_list) - apc_manager.exact_cache_guard_tokens,
           multimodal_token_ids,
           max_prefix_tokens=len(full_input_ids_list) - 1,
       )
       if exact_checkpoint_len <= 0:
           exact_checkpoint_len = None
   ```

3. **The `<= 0` guard above never fires**, because `adjust_prefix_to_text_suffix_boundary` clamps a non-positive desired length up to 1 — `apc.py:435`:
   ```python
   desired = max(1, int(desired_prefix_len))      # ← -1 becomes 1
   prefix_len = max(desired, media_safe_prefix_min(token_ids, media_token_ids))
   if prefix_len > max_len:
       return 0
   return prefix_len
   ```
   For a 15-token prompt: `desired = max(1, 15 - 16) = 1`, `max_len = 14`, so it returns **1** — a snapshot of a single token.

4. **The degenerate match is then accepted on lookup** — `generate/dispatch.py:876`:
   ```python
   exact_prompt_cache, exact_prefix_len = apc_manager.lookup_exact_cache(
       full_input_ids_list,
       extra_hash=apc_extra_hash,
       min_prefix_tokens=apc_safe_prefix_lookup_min,
   )
   ```
   `min_prefix_tokens` looks like it should protect against this, but for a **text-only** prompt it is 0 — `dispatch.py:820-824`:
   ```python
   apc_safe_prefix_min = _apc.media_safe_prefix_min(full_input_ids_list, multimodal_token_ids)
   apc_safe_prefix_lookup_min = max(0, apc_safe_prefix_min - 1)
   ```
   `media_safe_prefix_min` returns 0 when there are no media spans (`apc.py:392-394`). So the guard only enforces *media safety*, never *usefulness*, and a 1-token match passes.

5. **`reused_prefix_len = 1 != 0` closes the store gate forever** (`dispatch.py:1004`, and again at `:1043`). No new checkpoint is ever written for that tenant.

## Reproduction

Model: `Qwen3.6-35B-A3B` (VLM, 3-bit), `model_apc_mode()` → `"exact"`. Two fresh tenants, identical 3-turn dialog; the only difference is the length of the *first* prompt.

```python
# A: first prompt is long (~494 tokens)
# B: first prompt is "привет" (15 tokens), then the same long dialog
for i, prompt in enumerate(prompts, 1):
    for r in stream_generate(model=model, processor=processor, prompt=prompt,
                             image=None, max_tokens=6, sampler=sampler,
                             apc_manager=apc, apc_tenant=tenant):
        last = r
    print(last.prompt_tokens, last.cached_tokens, last.prompt_tokens / last.prompt_tps)
```

**A — first prompt long:**

| turn | prompt_tokens | cached_tokens | prefill |
|---|---|---|---|
| 1 | 494 | 0 | 2.172 s |
| 2 | 522 | 478 (92%) | 0.181 s |
| 3 | 554 | 478 (86%) | 0.278 s |

**B — first prompt is `"привет"` (15 tokens < guard 16):**

| turn | prompt_tokens | cached_tokens | prefill |
|---|---|---|---|
| 1 | 15 | 0 | 0.099 s |
| 2 | 494 | **1** | 0.920 s |
| 3 | 522 | **1** | 0.921 s |
| 4 | 554 | **1** | 0.965 s |

`stats_snapshot()` for B confirms the mechanism — the lookups "hit" but serve nothing, and nothing is ever stored:

```
{'exact_hits': 5, 'matched_tokens': 5, 'served_tokens': 0, 'exact_stores': 0,
 'stores': 0, 'token_hit_rate': 1.0, 'pool_used': 0}
```

Note `token_hit_rate: 1.0` — the metric counts *lookups that matched*, not the *share of the prompt served*, so this failure mode looks healthy in the stats.

## Secondary observation: degradation even in the happy path

In case A, reuse decays across turns (92% → 86% → …) because the checkpoint is only written when `reused_prefix_len == 0`, i.e. only on the cold turn. APC keeps replaying turn 1's snapshot while the dialog grows, so prefill climbs with conversation length.

I understand this is intentional given the documented split of responsibilities (`dispatch.py:875`):

```python
# APC: cross-request, hash-based prefix lookup. Only consulted if a per-turn
# PromptCacheState didn't already produce a hit.
```

…and passing a `PromptCacheState` does fix the per-turn case for us (94% flat). Mentioning it only because the 1-token poison above still costs the **cold-start / cross-restart** reuse that APC is specifically there to provide.

## Suggested fix

Refusing to store a snapshot that cannot be useful looks like the smallest change. Either:

- in `adjust_prefix_to_text_suffix_boundary`, don't clamp a non-positive `desired_prefix_len` up to 1 — return `0` instead and let the existing `if exact_checkpoint_len <= 0: exact_checkpoint_len = None` branch in `dispatch.py` handle it (`media_safe_prefix_min` should still be able to raise the floor when media tokens are present); or
- in `dispatch.py`, skip the checkpoint entirely when `len(full_input_ids_list) <= apc_manager.exact_cache_guard_tokens`.

Additionally, it may be worth rejecting degenerate matches on lookup — e.g. requiring `exact_prefix_len` to exceed some useful minimum (a block, or the guard) before it counts as a hit and sets `reused_prefix_len`. That would also stop stale 1-token entries already written to existing disk tiers from poisoning tenants after an upgrade.

## Environment

- mlx-vlm 0.6.3 (defect verified unchanged in 0.6.4 by diffing the sdist)
- mlx-lm 0.31.3, mlx 0.31.2
- Apple M4 Pro, 48 GB, macOS 15 (Darwin 25.5.0)
- Python 3.14

Blaizzy:
## Summary

Fixes #1617.

APC exact-mode checkpoint selection treated short text prompts as reusable one-token prefixes because `adjust_prefix_to_text_suffix_boundary()` clamped non-positive desired prefix lengths up to `1`. That allowed a short first prompt to create a degenerate exact-cache entry, and subsequent prompts could match that one token and skip writing a useful checkpoint.

This changes exact-mode APC to:

- return `0` for text-only guarded checkpoint lengths at or below zero
- keep the media-safe floor behavior for multimodal prompts that need media spans included in the prefix
- require exact-cache lookups during generation to exceed `exact_cache_guard_tokens`, so stale one-token entries already on disk do not keep poisoning tenants
- cover the short-prompt and media-floor cases with focused APC tests

## Validation

- `uv run --with pytest pytest mlx_vlm/tests/test_apc.py -q` -> 40 passed, 2 warnings
- pre-commit during commit: `black`, `isort`, `autoflake` passed
- `python3 -m compileall -q mlx_vlm/apc.py mlx_vlm/generate/dispatch.py mlx_vlm/generate/ar.py` passed using the bundled Python runtime

### DIFF
--- a/mlx_vlm/apc.py
+++ b/mlx_vlm/apc.py
@@ -494,8 +494,11 @@ def adjust_prefix_to_text_suffix_boundary(
     )
     if max_len <= 0:
         return 0
-    desired = max(1, int(desired_prefix_len))
-    prefix_len = max(desired, media_safe_prefix_min(token_ids, media_token_ids))
+    desired = int(desired_prefix_len)
+    media_min = media_safe_prefix_min(token_ids, media_token_ids)
+    if desired <= 0 and media_min <= 0:
+        return 0
+    prefix_len = max(desired, media_min)
     if prefix_len > max_len:
         return 0
     return prefix_len
--- a/mlx_vlm/generate/ar.py
+++ b/mlx_vlm/generate/ar.py
@@ -2251,13 +2251,17 @@ class BatchGenerator:
         if not ids_list or len(ids_list) < 2:
             return None
         safe_lookup_min = self._apc_safe_prefix_lookup_min(ids_list)
+        exact_lookup_min = max(
+            safe_lookup_min,
+            self.apc_manager.exact_cache_guard_tokens,
+        )
         extra_hash = self._apc_extra_hash(prompt_kwargs or {})
         apc_mode = getattr(self, "apc_mode", "block")
         if apc_mode == "exact":
             exact_cache, exact_prefix_len = self.apc_manager.lookup_exact_cache(
                 ids_list,
                 extra_hash=extra_hash,
-                min_prefix_tokens=safe_lookup_min,
+                min_prefix_tokens=exact_lookup_min,
             )
             if (
                 exact_cache is not None
@@ -2287,7 +2291,7 @@ class BatchGenerator:
             exact_cache, exact_prefix_len = self.apc_manager.lookup_exact_cache(
                 ids_list,
                 extra_hash=extra_hash,
-                min_prefix_tokens=max(prefix_len, safe_lookup_min),
+                min_prefix_tokens=max(prefix_len, exact_lookup_min),
             )
         warm_cache = None
         disk_prefix_len = 0
--- a/mlx_vlm/generate/dispatch.py
+++ b/mlx_vlm/generate/dispatch.py
@@ -822,6 +822,7 @@ def stream_generate(
         multimodal_token_ids,
     )
     apc_safe_prefix_lookup_min = max(0, apc_safe_prefix_min - 1)
+    apc_exact_lookup_min = apc_safe_prefix_lookup_min
 
     def _apc_suffix_is_text_only(prefix_len: int) -> bool:
         return _apc.prefix_leaves_text_only_suffix(
@@ -841,6 +842,11 @@ def stream_generate(
         apc_mode = _apc.model_apc_mode(model.language_model)
         if apc_mode is None:
             apc_manager = None
+        else:
+            apc_exact_lookup_min = max(
+                apc_exact_lookup_min,
+                apc_manager.exact_cache_guard_tokens,
+            )
 
     if apc_manager is not None:
         image_hash = _apc.hash_image_payload(pixel_values=pixel_values, image_ref=image)
@@ -877,7 +883,7 @@ def stream_generate(
             exact_prompt_cache, exact_prefix_len = apc_manager.lookup_exact_cache(
                 full_input_ids_list,
                 extra_hash=apc_extra_hash,
-                min_prefix_tokens=apc_safe_prefix_lookup_min,
+                min_prefix_tokens=apc_exact_lookup_min,
             )
             if (
                 exact_prompt_cache is not None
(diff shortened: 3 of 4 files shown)


lucasnewman:
Fixed in https://github.com/Blaizzy/mlx-vlm/pull/1901

PR #1625, opened 2026-07-17
vendor 24 more text models as native packages

Lazarus-931:
batch of common mlx-lm models vendered into native per-model packages. 

all 24 verified bit-exact vs mlx-lm, all test_modules suit passes as well


lucasnewman (approved):
LGTM

PR #1631, opened 2026-07-18
fix(lfm2_vl): make the training forward pass runnable for LoRA/SFT

chintan-projects:
LoRA/SFT on LFM2-VL crashes on the first training step. Inference is fine, so this only shows up once you try to train. Both errors come from `Model.__call__` in `models/lfm2_vl/lfm2_vl.py`:

1. `get_input_embeddings` gets `spatial_shapes` and `pixel_attention_mask` passed positionally, but the signature only takes `input_ids` and `pixel_values` positionally (everything else rides on `**kwargs`). So you get:
   ```
   TypeError: get_input_embeddings() takes from 1 to 3 positional arguments but 5 were given
   ```
2. After that's fixed, the `InputEmbeddingsFeatures` wrapper is handed to `language_model` as `inputs_embeds` instead of the array inside it, so the next layer trips on:
   ```
   AttributeError: 'InputEmbeddingsFeatures' object has no attribute 'shape'
   ```

The fix passes the two vision kwargs by keyword and unwraps `.inputs_embeds`, which is what `idefics3` and `gemma3` already do in their `__call__`.

### Repro

```bash
python -m mlx_vlm.lora \
  --model-path mlx-community/LFM2-VL-450M-bf16 \
  --dataset <any image+messages dataset> --split train \
  --train-mode sft --epochs 1 --batch-size 2
```

Crashes on the first step on `main` and in 0.6.5. With this change it trains through.

### Verification

Ran a small LoRA SFT on `mlx-community/LFM2-VL-450M-bf16` (image to JSON task, 48 examples) on an M4 Max:

- Training completes, loss drops from ~6.4 to ~3.0, ~1.9 GB peak.
- Generation output is byte-identical before and after the change, so the inference path isn't touched.
- `black --check` leaves the file unchanged.

Diff is 8 lines in one function. Happy to add a tiny training smoke test for `lfm2_vl` if you'd want one.


Blaizzy (approved):
LGTM, thanks!

PR #1635, opened 2026-07-19
Fix Qwen3 Omni deepstack: visual embeds were computed but never injected

hellopahe:
Ran into this while working on Qwen3 Omni serving: the deepstack embeds never reach the decoder. `get_input_embeddings` drops them on return, and `LanguageModel.__call__` overwrites its argument with `kwargs.get("deepstack_visual_embeds")`.

`_deepstack_process` has its own bug once it does run: np.where over the 2D mask takes row indices, and the write hits every batch row. Replaced it with the per-row version from qwen3_vl.

Also forwarded quant_predicate to the top-level Model.

Checked against transformers with a tiny thinker on shared weights: main matched HF only with deepstack disabled on the HF side, while this branch matched it as-is.

Tested on M3 Max
```
python -m pytest mlx_vlm/tests/test_qwen3_omni_moe.py -q        # 4 passed
python -m pytest mlx_vlm/tests/test_models.py -q -k qwen3_omni  # 3 passed
```

Claude assisted with the implementation.


Lazarus-931:
lgtm

cc: @Blaizzy

Blaizzy (approved):
LGTM, thanks!

PR #1637, opened 2026-07-19
Add Inkling from Thinky support

Lazarus-931:
Text backbone (banded relative-position attention via a fused Metal mask kernel, depthwise short-conv over a shared per-layer conv cache, sigmoid-gated MoE with shared experts, hybrid sliding/global layers), HMLP vision tower, dMel audio tower, and the top-level multimodal Model with masked-scatter merge. Reuses the shared SwiGLUMLP and CacheList; adds TestModels cases.

I verified it against HF InklingForConditionalGeneration on the 0.6B checkpoint: strict weight load, bit-identical logits (corr 1.0), matching greedy generation, and real image/audio processor I/O parity. 

MTP/spec-decode deferred since I'd have no way to verify to my knowledge.  Once smaller model get's released, we can hopefully use quantized version for verification + locally testing it

edit: added a standalone inkling_mtp drafter (--draft-kind mtp) mapping the checkpoint's model.mtp.* head onto a reused InklingDecoderLayer, with target-side verify/rollback hooks and a split.py to extract the head from a full Inkling checkpoint. Correctness is verified on tiny-random/inkling by a greedy-invariant test (speculative output identical to non-speculative) and an exact short-conv-cache rollback test.


lucasnewman (approved):
LGTM. There aren't any usable quants at the moment, right? If it works with a synthetic model that's probably good for now though.

Lazarus-931:
actually there was inference-optimization/Inkling-0.6B-A0.6B  checkpoints which i used for verifying model path works and tiny-random/inkling for the mtp component

Issue #1629, opened 2026-07-18
Redesign APC around pluggable cache components for custom attention and hybrid models


Blaizzy:
## Summary

Redesign Automatic Prefix Caching (APC) around a capability-driven, component-based cache plan so that all model cache layouts—including custom attention, sliding/windowed attention, recurrent/linear attention, and composite caches—have a correct reuse path.

"All models" should mean:

- every cache can fall back to a correct checkpoint snapshot when it exposes a snapshot/restore contract;
- block/page-level reuse remains an optimization for cache components that explicitly support splitting and concatenation;
- adding a new cache implementation does not require modifying APC's central clone, merge, disk, and mode-detection switches.

This follows the useful part of SGLang's current direction: one logical prefix index with independent state components and a match boundary accepted only when every required component is restorable.

## Motivation

APC currently centralizes knowledge of concrete cache classes:

- `_clone_cache_entry_for_apc()` switches on every supported cache type.
- `model_apc_mode()` reduces the model to `"block"`, `"exact"`, or disabled.
- `_merge_exact_cache_entries()` repeats concrete-type dispatch for continuous batching.
- Lookup/restoration/storage selection is duplicated between `generate/dispatch.py` and `generate/ar.py`.

This makes custom caches fragile and creates correctness risks:

1. **RingSlidingKVCache is structurally misclassified**

   `RingSlidingKVCache` inherits `KVCache`, so the current `isinstance(KVCache)` check classifies it as block-compatible. After the ring wraps, its absolute offset can exceed the physically retained sequence. Generic KV harvesting is therefore not a sufficient correctness contract.

2. **MiniMaxM3KVCache is rejected despite exposing the needed operations**

   MiniMax M3 stores ordinary K/V plus index-attention keys and already exposes `state`, `meta_state`, and `merge`. APC rejects it because the concrete class is not in its allowlist. With an adapter, its K/V and index keys could be captured as one pageable composite component.

3. **Exact snapshots do not scale as the universal architecture**

   Exact mode is a useful safety fallback, but cloning an entire prompt cache per checkpoint duplicates long prefixes and cannot express that some layers are pageable while others require recurrent checkpoints.

4. **Multimodal identity is not general enough**

   The request key is primarily salted with image data, while other semantic inputs and model-specific state may also affect the cache: audio, video, embeddings, masks, position/RoPE deltas, processor configuration, and custom attention metadata.

## SGLang reference

Relevant parts of SGLang's design:

[...]

PR #1638, opened 2026-07-19
implement apc as a pluggable, capability driven cache adapter for #1629

Lazarus-931:
This reworks Automatic Prefix Caching (Blaizzy/mlx-vlm#1629) around pluggable, capability-driven cache adapters so APC no longer hard-codes a list of concrete cache classes.

If every cache is described by a registered adapter or the generic checkpoint fallback, it fixes the two bugs:

1. RingSlidingKVCache is no longer mistaken for a pageable KV cache (it can serve wrong state after the ring wraps)
2. MiniMaxM3KVCache can now be reused instead of rejected

Single req and cont. batch paths share one lookup and one commit cord. instead of duplicates, where block storage is split in logical index + per-component handles.

With this pr, cache key now covers all non-token inputs, like audio, vid, emb, mask and  model/quant/schema fingerprints rather than only img.

Full APC and generate test suites pass


Blaizzy:
A few nits:

### 1. RingSliding exact-cache hits crash continuous batching

`RingSlidingKVCache` falls through to the generic merge at `apc_adapters.py:555`. It inherits `KVCache.merge`, which only accepts `caches`, but the new code passes `entries, prefix_lens`.

Repro from this PR's head:

```python
import mlx.core as mx

from mlx_vlm.apc_adapters import merge_cache_entries
from mlx_vlm.models.unlimited_ocr.language import RingSlidingKVCache

cache = RingSlidingKVCache(window_size=4)
cache.update_and_fetch(
    mx.ones((1, 2, 3, 4)),
    mx.ones((1, 2, 3, 4)),
)

merge_cache_entries([cache], [3])
```

Observed:

```text
TypeError: KVCache.merge() takes 2 positional arguments but 3 were given
```

`BatchGenerator._build_mixed_prompt_batch()` reaches this through `make_warm_batch_exact_cache_multi()` on an exact APC hit, so the headline RingSliding case fails in continuous batching. This needs a RingSliding-aware batch adapter, or the hit should be rejected and fall back to a cold prefill.

### 2. Disk APC namespaces collide across adapters

The server model cache distinguishes `(model_path, adapter_path, model_kind)`, but `apc_disk_namespace()` only receives `model_path` and KV settings. Thus two separately loaded LoRA adapters for the same base model use the same disk namespace.

Minimal illustration:

```python
from mlx_vlm.apc import apc_disk_namespace

common = dict(
    model_path="org/base-model",
    kv_bits=4,
    kv_group_size=64,
    kv_quant_scheme="uniform",
    quantized_kv_start=0,
)

# These represent get_cached_model(..., adapter_path="adapter-a")
# and get_cached_model(..., adapter_path="adapter-b").
adapter_a_namespace = apc_disk_namespace(**common)
adapter_b_namespace = apc_disk_namespace(**common)

assert adapter_a_namespace == adapter_b_namespace
```

The adapters can produce different K/V for identical tokens while reading and writing the same APC disk directory. Please include an adapter identity/fingerprint (and ideally the resolved model revision/weights fingerprint) in the namespace.

### 3. Semantic tensor hashing has reproducible collisions

`_hash_payload()` delegates every tensor to `hash_image_payload()`, which casts to FP16 and hashes only flattened bytes. Shape and dtype are not included.

```python
import mlx.core as mx
from mlx_vlm.apc import _hash_payload

# Same flattened values, different semantic shape.
a = mx.array([[1, 2, 3, 4]])
b = mx.array([[1, 2], [3, 4]])
assert a.shape != b.shape
assert _hash_payload(a) == _hash_payload(b)

# Distinct float32 embeddings collapse after the FP16 cast.
x = mx.array([1.0], dtype=mx.float32)
y = mx.array([1.0001], dtype=mx.float32)
assert bool(mx.any(x != y))
assert _hash_payload(x) == _hash_payload(y)
```

This can restore the wrong cache for distinct masks, embeddings, or other semantic tensors. The digest should include shape and dtype and should hash a lossless representation of the values.

### 4. The single-request key omits processed audio/video features

In `dispatch.py`, the key hashes the original `audio`/`video` references, and checks `kwargs["inputs_embeds"]` before `generate_step()` constructs embeddings. Processor outputs such as `input_features` and `pixel_values_videos` are passed to the model but not included in the key.

This reproduces the effective key construction for pre-tokenized audio inputs:

```python
import mlx.core as mx
from mlx_vlm.apc import semantic_extra_hash

def current_dispatch_key(input_features):
    kwargs = {"input_features": input_features}
    return semantic_extra_hash(
        image_hash=0,
        media={
            "audio": None,
            "video": None,
            "embeddings": kwargs.get("inputs_embeds"),
            "masks": None,
        },
    )

assert current_dispatch_key(mx.zeros((1, 80, 10))) == current_dispatch_key(
    mx.ones((1, 80, 10))
)
```

Those feature tensors produce different model state but the same APC key. The key should be built from the processed semantic inputs actually forwarded to the model, or from a canonical model/processor-provided dependency hook that covers all of them.

Lazarus-931:
should be good to go now

lucasnewman (approved):
Nice work!

weklund:
Sick!

PR #1647, opened 2026-07-21
fix(server): row-level per-request sampling in the continuous batch (supersedes #1353)

mmmugh:
Supersedes #1353. That PR added `top_k`/`min_p` to the *single shared* batch sampler; @lucasnewman closed it because a continuous batch needs **row-level** sampling when requests have different generation parameters. This does that: every row in the batch is sampled with its own `temperature / top_p / top_k / min_p / seed`.

## The framing

mlx-vlm's `GenerationBatch` was forked from mlx-lm's, and **mlx-lm already carries per-row samplers** (`self.samplers: List[Callable]`, threaded through `extend`/`filter`). The fork collapsed that list into one shared sampler built from the first request — which is the bug. On `main` today, `GenerationBatch.extend()`/`filter()` never touch `self.sampler`, so a merging batch's sampler is dropped ("request #1 wins"); `_make_sampler` drops `top_k`/`min_p`; and `greedy_sampling` is a batch-wide bool from request #1. This PR restores per-row sampling.

## Approach

- A frozen `SamplingConfig(temperature, top_p, top_k, min_p, seed)` carried as a per-row `List[SamplingConfig]` through `insert → extend → filter`, mirroring the existing `logits_processors` list exactly.
- A row-aware `_PositionedTargetSampler(configs)` that builds `[B]` `mx.array`s and applies **temperature → top_k → top_p → min_p in one sorted-space pass**, with fused greedy `where(temp < eps, argmax, sampled)`. `top_k` is a rank-mask (exactly-k), because `apply_top_k`'s `argpartition(kth=int)` can't take a per-row `k`. The vendored `sample_utils.py` is untouched.
- The two duplicated `_PositionedTargetSampler` copies (which had drifted on seed handling) are unified.
- Per-request seeds stay **stateless**: `_position_seed(seed, position)` → `mx.random.key`, so a seeded request is reproducible and invariant to batch composition (vLLM/SGLang semantics: same seed = same stream). Per-row differentiation comes from each row's seed.
- Out-of-domain `top_k`/`min_p` are rejected at admission (422) via Pydantic field constraints.

## The proof

The load-bearing test is a strengthened `test_positioned_target_sampler_is_batch_grouping_invariant`: **two rows with distinct `row_ids` and distinct `temperature/top_p/top_k/min_p/seed`**, asserting the batched draw equals each row drawn alone. That property is what #1353 lacked, and it fails on `main` today. Plus regression tests for the batch lifecycle (extend/filter alignment, greedy reconciliation, first-token per-row), MTP staggered completion, and speculative no-drop. **499 tests green.**

## Live verification

Verified end-to-end against a real `mlx_vlm.server` (Qwen3-VL-8B-Instruct-8bit, MLX 0.32.0, Apple Silicon), on the continuous-batching path.

| Check | Result | Detail |
|---|---|---|
| Greedy (temp=0) deterministic solo | ✅ | identical across runs |
| `top_k=1 @ temp=2.0` tracks greedy (top_k binds) | ✅ | shares a 305-char prefix with the greedy answer |
| Unconstrained `temp=2.0` diverges from greedy (incoherent) | ✅ | shares only 17 chars with greedy |
| Stochastic `temp=1.3, top_k=60, min_p=0.02` genuinely samples | ✅ | ≠ greedy |
| Seeded stochastic config reproducible (same seed twice, solo) | ✅ | identical |
| **Co-batched `top_k=1` row stays greedy-coherent every time** | ✅ | prefix-match with greedy across 5 co-batched pairs: `[120,120,120,120,120]` |
| **Co-batched `temp=2.0` row stays incoherent every time** | ✅ | prefix-match with greedy across the same pairs: `[17,17,17,17,17]` |
| The two co-batched rows never collapse to one mode | ✅ | greedy row ≠ hot row in every pair |

**The discriminator (the bar #1353 could not pass):** a `top_k=1` (greedy) request and an unconstrained `temperature=2.0` (incoherent) request were **co-batched**. The greedy row tracked the greedy answer *every* time while the hot row stayed wild *every* time, in the same batches — they retained distinct per-config behavior. Under the old shared-sampler behavior the whole batch uses one parameter set, so the greedy row would have gone wild whenever the hot request opened the batch; it never did.

**Why behavioral, not byte-identity:** the batched model forward on this GPU is floating-point non-deterministic across batch composition — two *identical* greedy requests, co-batched, share a long common prefix then diverge (reduction order depends on batch position). That is a pre-existing model property on the untouched greedy path, unrelated to sampling. So byte-exact per-row isolation is proven by the deterministic unit test `test_positioned_target_sampler_is_batch_grouping_invariant` (identical logprobs in → identical token out); the live test proves param adherence, which is robust to that FP noise.

## Scope

- **This PR: plain AR + MTP** — full per-row. MTP rides the same `BatchGenerator`; its walks call `sample_target` on the full batch, so per-row composes (one gate keeps positioned-sampler MTP batches at full width, since `_mtp_rounds_batch` otherwise compacts mid-round).
- **dflash/eagle3 (`_run_speculative`): interim explicit degradation, full per-row in a follow-up PR.** A heterogeneous batch is serialized-by-config (never the silent last-wins), with a `TODO(PR2)` marker. Full per-row there is tractable (no accept-math change — verification is exact-match-to-target-sample, not q/p rejection) but has its own `sampler_is_greedy` surface + test matrix, so it's a separate reviewable PR.

## An open question for you

The plumbing (per-row config through `insert/extend/filter`) is the same whether the sampler applies params vectorized (this PR) or as mlx-lm's per-row loop-of-closures. I went vectorized because it preserves the `sample_target` grouping-invariance that mlx-vlm's positioned sampler adds over mlx-lm's global-RNG closures (and `speculative/mtp.py` duck-types on `sample_target`). Measured ~1.6–3.1× faster than the loop at B=4–16, though that's a couple percent of a decode step at small B — happy to switch to the loop shape if you'd prefer parity with mlx-lm; it drops into the same seam.

## Known limitations (follow-ups)

- The model-dependent `top_k >= vocab_size` check raises `HTTPException(422)` but some pre-existing broad `except Exception` handlers in `openai.py`/`anthropic.py` downgrade it to 400/500. No crash either way (the sampler is robust to out-of-domain params); it's a status-code nicety I left rather than reworking unrelated error handling.
- `_run_speculative`'s serialize-by-config can defer a lone off-config request under sustained majority-config load (interim path; resolved by the PR-2 per-row work).

## Credits

Restores the per-row `samplers` pattern from **mlx-lm**. Vectorized-sampler shape (rank-mask top_k, fused greedy, per-row `[B]` params) follows **vLLM**, **SGLang**, and **TGI**. The vendored-in-tree `sample_utils`/`BatchKVCache` this builds on came from @lucasnewman's #1494.


Lazarus-931:
hey @mmmugh , the CI fails b/c of pre-commit, run `pre-commit run --all-files ` and you should be good to go

mlx_vlm/server/generation.py
```diff
@@ -1150,6 +1124,15 @@ def generate(
             raise ValueError(
                 "thinking_budget is not supported with speculative decoding in the server."
             )
+        vocab_size = self._model_vocab_size()
```
lucasnewman: Is this just a drive-by fix? I don't object to it, but it seems only tangentially related.

mlx_vlm/speculative/mtp.py
```diff
@@ -1092,7 +1092,16 @@ def _mtp_rounds_batch(
...
+        #
+        # Positioned per-row samplers are built once at the full batch width
+        # (one config per original row); mid-round compaction would feed the
+        # walk a shrunk logprobs and break the sampler's configs-length
+        # invariant. Keep all rows (finished rows are masked when emitting) so
+        # the sampler always sees the full width. Costs some compute on
+        # finished rows until the whole batch drains; correctness > that.
+        cache_filterable = all(
+            hasattr(c, "filter") for c in prompt_cache
+        ) and not _sampler_supports_positioned_target(sampler)
```
lucasnewman: This is a non-trivial change from a performance perspective, since we won't be dropping finished sequences. Can you benchmark batched inference before and after?

lucasnewman:
@mmmugh Thanks, see the comments for some questions. Please make sure to run `pre-commit run --all` to clear the formatter so we can run tests.

mlx_vlm/server/generation.py
mmmugh: You're right—removed it. It was tangential, and `batched_row_sample` already degrades an
out-of-range `top_k` safely (it just acts as "disabled"), so it wasn't guarding against a crash
anyway. Kept the `top_k`/`min_p`/`temperature` Pydantic field constraints, since those validate the
new per-row params directly. (Pushed in the latest commit.)

mlx_vlm/speculative/mtp.py
mmmugh: Good call—I reworked it so we *do* keep dropping finished sequences. Instead of disabling
mid-round compaction for positioned samplers, the sampler now compacts its per-row configs in
lockstep with the batch via `_PositionedTargetSampler.select(keep_slots)`, so `active_idx` shrinks
as before and the configs stay aligned to the shrunk `logprobs`. The regression test now asserts
compaction *happens* (`cache.filter_calls == [[1]]`) rather than being suppressed, and reverting the
one `sampler.select(...)` line reproduces the width-mismatch it guards against.

Benchmarked it end-to-end. I split `Qwen/Qwen3.5-4B` into a native `qwen3_5_mtp` drafter
(`python -m mlx_vlm.speculative.drafters.qwen3_5_mtp.split`) and served with
`--draft-kind mtp --draft-block-size 4` on an M-series Mac Studio. Workload: 8 concurrent chat
requests with `max_tokens` spread 48→256 so rows finish at staggered rounds and mid-round compaction
fires repeatedly, `temperature=0.7` so the batch runs the vectorized per-row sampler. I instrumented
`_mtp_rounds_batch` to confirm the change is actually exercised—it runs at B=8 with the positioned
sampler, and compaction fires on every drop (`n_active` 8→7→…→1, `select()` each time). 3-trial
median aggregate decode throughput (per-run variance <1%):

| | shared sampling config | distinct per-row config |
|---|---|---|
| main (c9e27b0)   | 135.6 tok/s | 134.6 tok/s |
| this branch      | 135.2 tok/s | 119.1 tok/s |

So the compaction rework itself is free: with a shared config the branch matches `main` (135.2 vs
135.6, within noise) while dropping finished rows exactly like `main` (verified firing 8→1).

Being upfront about the other cell—when every row carries a *different* sampling config, per-row
MTP decode runs ~12% slower than the uniform batch (119 vs 135). It's the same `_mtp_rounds_batch`
path (not the `serialize_speculative_batch_by_config` gate, which is dflash/eagle3-only): the cost of
actually honoring distinct seeds/params across a speculative batch, which `main` doesn't do at all.
It only appears when configs genuinely differ; shared-config batches are unaffected. I can chase that
down in the per-row-speculative follow-up if it's worth it. (Batched **AR** per-row overhead, for
reference, was ~1%: branch 272 vs main 275 tok/s.)

mmmugh:
Thanks @Lazarus-931—good catch. Ran `pre-commit run --all` (black/isort/autoflake) and pushed the formatting fix in 21d3002; it's on the current head. 🙏

lucasnewman (approved):

lucasnewman:
@mmmugh Can you [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits and force push so we can clear the merge requirements?

mmmugh:
Done—signed all 8 commits and force-pushed. The branch head is now a343cd6 and GitHub shows every commit as Verified, so that should clear the signed-commits requirement. 🙏

mmmugh:
@lucasnewman before I finish the merge, I'd like your steer on one integration point, since main has moved a lot in the sampler area since I opened this (the top-nσ / p-less / typical-p modes, plus _PositionedTargetSampler itself).

This PR generalizes _PositionedTargetSampler to carry a per-row SamplingConfig, which meant replacing the batch generator's per-request sampler= slot with per-row sampling_configs threaded through insert/extend/filter. Your new modes route through that same sampler= slot (_make_sampler(args)) and aren't expressible in the current SamplingConfig (temperature/top_p/top_k/min_p/seed)—so they collide in server/generation.py. (The other two conflicts are trivial: schema field constraints and a co-located test.)

How would you like the two to coexist?

Extend SamplingConfig + the per-row sampler to honor top_n_sigma/p_less/typical_p per-row (architecturally consistent, a bit more surface).
Keep the new modes on a batch-level sampler path alongside per-row (applied batch-global, not per-row).
Happy to implement whichever fits where you're taking this.

lucasnewman:
> @lucasnewman before I finish the merge, I'd like your steer on one integration point, since main has moved a lot in the sampler area since I opened this (the top-nσ / p-less / typical-p modes, plus _PositionedTargetSampler itself).
> 
> This PR generalizes _PositionedTargetSampler to carry a per-row SamplingConfig, which meant replacing the batch generator's per-request sampler= slot with per-row sampling_configs threaded through insert/extend/filter. Your new modes route through that same sampler= slot (_make_sampler(args)) and aren't expressible in the current SamplingConfig (temperature/top_p/top_k/min_p/seed)—so they collide in server/generation.py. (The other two conflicts are trivial: schema field constraints and a co-located test.)
> 
> How would you like the two to coexist?
> 
> Extend SamplingConfig + the per-row sampler to honor top_n_sigma/p_less/typical_p per-row (architecturally consistent, a bit more surface). Keep the new modes on a batch-level sampler path alongside per-row (applied batch-global, not per-row). Happy to implement whichever fits where you're taking this.

@Lazarus-931 Can you take a look?

Lazarus-931:
hi @mmmugh, yes great point, I think we'll merge this first with extended SamplingConfig, we'll edit those prs need be. Feel free to make changes here however for that!

mmmugh:
Merged current main; the branch is no longer conflicting. Per @Lazarus-931's call above, `SamplingConfig` now carries `top_n_sigma` / `p_less` / `typical_p`, honored per row: `_new_modes_keep()` builds the per-row keep mask with each mode gated per row, so a row that doesn't use one is untouched. One open question for you at the end.

Conflicts: `schemas.py` keeps this PR's field constraints plus main's three new fields; both adjacent `test_speculative.py` tests survive; `_make_sampler` is deleted (per-row admission leaves it with no callers, and `make_sampler` is no longer imported, so keeping it would `NameError`). A second merge picks up Mage-Flow and the black 26.3.1 bump, so the formatter check here runs the same version CI will.

Two bugs a pre-push audit turned up, both in main's own features, fixed and regression-tested:

- **APC.** `_apc_pick_for` destructures the queued-sequence tuple as 6 elements; `insert()` now appends a 7th (the config). Separate hunks, so git merged both sides cleanly—a semantic-only conflict that would have raised `ValueError` on every APC-enabled prefill. The new test feeds real `insert()` output to the real consumer so the two can't drift; one upstream test that hand-built a 6-tuple is updated to match.
- **float16.** `top_n_sigma` took max/std in the raw dtype, and `mx.std` over an fp16 logprob vector overflows to inf at any real vocab size, silently no-opping the mode. Upstream's `_top_n_sigma` casts to fp32 unconditionally and the deleted `_make_sampler` routed through it. Now fp32: 0 mismatches vs `apply_top_n_sigma` at V=8192/32000 in fp16, bf16 and fp32.

Four smaller ones in the new code: upstream's "drop if x < t" phrasing, so a single `-inf` logprob (grammar masking, `logit_bias`) can't NaN the threshold and pin a row to argmax; modes chained ahead of the rank filters per `make_sampler`'s order, or `typical_p` can empty a Qwen-default `top_k=20` window and collapse the row to greedy; the extra mask work skipped unless a row opts in; `typical_p` range-gated, and a non-bool `p_less` coerced rather than raising in the GPU thread and tearing down the batch.

**Your call.** Combined with `top_p` I renormalize over the survivors instead of reproducing `make_sampler`'s chain. `apply_top_p` compares cumulative mass against `1 - top_p` on a distribution that no longer sums to 1 once a mode has masked it, so it can reject every token: `top_n_sigma=1.0` with `top_p=0.9` leaves a row of all `-inf` upstream that then gets sampled from. Looks like a genuine upstream bug worth its own look, but happy to match upstream exactly here instead. With no mode active, behaviour is unchanged and still matches `apply_top_p` / `apply_min_p` exactly.

Masks are checked token-for-token against `sample_utils.apply_*`, plus per-row activation, grouping invariance, fp16 parity, NaN tolerance, `typical_p`+`top_k` and server wiring; every fix is mutation-tested (8/8 fail when reverted). 1826 passed, 4 skipped, plus one unrelated pre-existing failure (`test_qwen_target_verify_quantized_linear_matches_singleton_batch_path`, an exact bf16 quantized-matmul equality check) that fails identically on pristine `origin/main` in a clean worktree, so not caused by this PR. pre-commit clean under 26.3.1.

@lucasnewman your approve predates the merge and these fixes. The composition order is the part worth re-reading.

mmmugh:
Rebased onto current main—back to mergeable, commits signed, suite green.

Whenever one of you has a moment: CI has never run on this branch (fork PRs need a maintainer to approve the workflow), so it's sitting at "blocked" on a check that hasn't been triggered rather than on anything failing. A run would tell us if it's actually good on your infra.

Still open from the last update, whenever you get to it:

- `top_p` combined with one of the new modes: I renormalize over the survivors rather than reproducing `make_sampler`'s chain, because that chain can reject every token once a mode has masked mass away (`top_n_sigma=1.0` with `top_p=0.9` leaves an all `-inf` row). Happy to match upstream exactly instead if you'd rather keep them identical.
- @lucasnewman your approve predates the merge and the fixes that came with it, so it'll need a fresh look.

No rush—just flagging that it's ready and only needs a CI trigger to move.

Lazarus-931:
conflicts and gtg :)

mmmugh:
Conflicts resolved and rebased onto current main—mergeable again, commits signed, suite green.

main had moved a fair way (0.6.8-dev to 0.6.15), but only two files actually conflicted:

- `server/generation.py`: took your `greedy_sampling=args.temperature == 0` and `_effective_prefill_step_size()` verbatim. Worth noting the greedy flag is inert on this path since no `sampler=` is passed (`greedy_sampling or sampler is None` is already true), and the batch derives its greedy flag per row from the actual configs—so no behaviour change either way.
- `tests/test_server.py`: kept the two speculative-config tests from this branch alongside your updated `_run_speculative_prefill_once` signature.

- `tests/test_generate.py`: your new `test_precomputed_semantic_hash_reuses_actual_growing_apc_prefix` hand-builds a queued-sequence tuple, so it needed the `SamplingConfig` this branch appends as the 7th element.

Nothing touched the sampler itself, so the per-row work is unchanged from what you looked at.

One thing still open from before, in case it matters to you: combined with `top_p`, the new modes renormalize over the surviving tokens rather than reproducing `make_sampler`'s chain, because that chain can reject every token once a mode has masked mass away (`top_n_sigma=1.0` with `top_p=0.9` leaves an all `-inf` row). Happy to match upstream exactly if you'd rather they behave identically—say the word and I'll push the change.

mmmugh:
Converting to draft—please hold off merging. Sorry for the churn after saying it was good to go.

A final verification pass on my side turned up real problems, all of them mine rather than anything from the merge:

- The row sampler only accepts 2-D `[B, V]` logits, but the dflash/eagle3 verify loops call it with 3-D `[B, T, V]` (`dflash.py:153/276`, `eagle3.py:199/324`). Upstream's `_PositionedTargetSampler` handled 3-D, so a server started with `--draft-model` for those kinds now fails every request at any temperature. This one is entirely on me and predates your approve.
- An explicit `sampler=` passed to `BatchGenerator` / `batch_generate` is silently ignored and the rows fall back to greedy, which also affects `evals/ocrbench.py`.
- Greedy decode pays a full-vocab argsort per step that it doesn't need, so the default path got measurably slower. I'll bring numbers rather than a claim this time.
- Plus a couple of smaller ones: `top_k=-1` now 422s where it used to mean "disabled", and a stale length check in the MTP acceptance walk.

Worth flagging why the suite didn't catch these: every test touching the speculative path monkeypatches the round loop, so the real sampler never meets real verify logits. I'm adding tests that drive those loops unmocked so this class can't hide again.

I'll fix these, re-verify end to end, and flip it back to ready. Happy to split the PR if you'd rather review the per-row core separately from the sampling-mode work.

mmmugh:
## Ready for review again—what the hold-off found, and what changed

Apologies for the long silence. I converted this to a draft because a pre-merge audit of my own
branch turned up five blockers plus a compat regression, all of them mine and all of them predating
@lucasnewman's approval. Rather than patch and re-push, I worked the list properly and then kept
auditing until the audits stopped finding code defects. That took six rounds. Here is the whole
story, because some of it changes behavior you'd want to know about.

**12 new commits.** Net diff vs `main`: **656 lines across 3 source files** (`generate/ar.py`,
`server/generation.py`, `speculative/mtp.py`) and **2,588 lines of tests**. `server/schemas.py` is no
longer in the diff at all—see "Scope" below.

---

### The blockers

| | What was broken |
|---|---|
| **3-D verify logits** | `_PositionedTargetSampler` only accepted rank-2 `[B, V]`, but the dflash/eagle3 verify loops hand their sampler the whole block and index the result along axis 1. `server/generation.py` builds that sampler as the *only* sampler for those loops, so **a server started with a dflash or eagle3 drafter raised `too many values to unpack` on every request**, at temperature 0 as well as above it. Two shapes had to give: the rank, and the width—the drafter drafts one row at a time and continuous batching drops finished rows without rebuilding the sampler. |
| **`sampler=` was ignored** | `GenerationBatch._step` and `PromptProcessingBatch.generate` built a sampler from the per-row configs and never consulted `self.sampler`. Since `_generate_batch` inserts without per-row configs, every caller passing `sampler=` silently got greedy output. Control: the same harness calls a custom sampler **4 times on `main` and 0 times here**. `evals/ocrbench.py` was an in-repo victim. |
| **`sample_target` width check** | Validated against `len(configs)` where upstream validated against `len(row_ids)`. The ragged MTP path passes `[T, V]` for one row, so a single-stream MTP request with a seed and temperature > 0 raised before emitting a token. |
| **Frozen RNG key** | `__call__` derived its key from the row's *index in the batch* and position 0—a pure hash of two constants. The key never advanced (six consecutive calls returned the same tokens) and a row's draw depended on who it was batched with. |
| **Greedy paid for sorted space** | Every row went through a full-vocab argsort that `mx.where` then discarded for greedy rows. `DEFAULT_TEMPERATURE` is 0 and all three request schemas default to it, so this was the server's default path. |

---

### Sampling semantics—please read this part

While hardening the above I built a differential contract against `sample_utils.make_sampler` and
found that `batched_row_sample` had **drifted from the chain it replaces**. It was computing the
nucleus on the *temperature-scaled* distribution, while `make_sampler` filters unscaled logprobs and
applies temperature only at the draw. Nucleus size at V=4096, median of 10 rows:

| temp | `top_p` | ours (before) | `make_sampler` | |
|---|---|---|---|---|
| 0.5 | 0.9 | 37 | 960 | **0.04x** |
| 0.7 | 0.9 | 276 | 960 | 0.29x |
| 1.0 | 0.9 | 960 | 960 | 1.00x |
| 1.5 | 0.9 | 1962 | 960 | 2.04x |

Identical only at temperature exactly 1.0—which is why the branch's own parity test never saw it:
that test pinned temperature to 1.0 and asserted a subset rather than equality.

Also fixed in the same area: `top_p=1.0` (the server default) was truncating the tail on peaked
distributions through float32 cumsum drift—3,416 of 4,096 tokens dropped on one row—and
`top_p=0.0` collapsed to argmax where `make_sampler` reads it as "off".

`mlx_vlm/tests/test_sampler_parity.py` now states the contract and sweeps it: every rank-filter
combination × five temperatures, each mode alone and chained, three vocab sizes, float32/float16/
bfloat16. **160 tests.** It captures the mask the production code actually hands to the draw rather
than reimplementing the sampler.

<details>
<summary><b>Seven documented divergences from <code>make_sampler</code></b> (click to expand)</summary>

Four of these are places this behaves *better* than upstream; two are numerically neutral; one is an
open question for you (below).

1. **Empty rows**—a rank filter chained after a mode can reject every token upstream, and
   `make_sampler` then draws from an all `-inf` row, i.e. an arbitrary token. This keeps the top
   survivor.
2. **Input normalization / `typical_p`**—see the question below.
3. **Tie ordering at any cutoff**—upstream sorts ascending, this sorts descending, so equal-valued
   tokens straddling a threshold fall on opposite sides. Same count, same mass, different members.
   Common on quantized logits (28% of bfloat16 rows).
4. **`min_p` exactly on the boundary**—upstream's `@mx.compile`'d `_apply_min_p` drops a token whose
   logprob equals the threshold to the bit, while the same formula run eagerly keeps it. This matches
   upstream's *uncompiled* semantics.
5. **Reduced precision**—`p_less`/`typical_p`, and the rank filters on tie-free rows, differ in
   fp16/bf16 (1 of 155 fp16 rows, 2 of 86 bf16 rows at `top_p=0.9`).
6. **`top_k >= vocab_size`**—`apply_top_k` raises; this keeps everything and clamps into int32, so
   an absurd `top_k` can't crash the GPU worker and take every co-batched request with it.
7. **Temperature edges**—greedy is `temperature < 1e-5` rather than `== 0`; negative temperature is
   greedy here and inverts the distribution upstream.

</details>

**One question for you.** `batched_row_sample` log-normalizes its own input, because the speculative
loops (`dflash.py`, `eagle3.py`, `mtp.py`) hand the sampler **raw `lm_head` logits** while the AR path
hands it logprobs—and `exp()` of an unnormalized row isn't a probability vector, so without it
`top_p` keeps 4027 of 4096 tokens instead of 90 on every speculative request. `make_sampler` tolerated
raw logits only by accident, because `top_p_sampling` softmaxes and is scale-invariant.

The cost: normalization also fires on rows that are masked *on purpose*—a grammar mask or
`logit_bias=-inf`—and `typical_p` then measures typicality against the renormalized distribution
rather than the residual mass. Measured 8 tokens kept upstream vs 1 here on a 64-token row with 56
masked.

So: **should those loops be passing raw logits to a sampler at all?** The docstring already says
"log-normalized". Normalizing at the call sites instead would match upstream everywhere and let me
drop this divergence—but that's a change to three of your speculative files, and I didn't want to
make it unilaterally inside a PR about the server's continuous batch. Happy to do it either way.

---

### Behavior changes worth calling out

- **The server now honors `top_k` and `min_p`.** `main`'s `_make_sampler` passes only `temp` and
  `top_p` to `make_sampler` and never forwards either, so both were silently dropped. Requests that
  explicitly send them will now sample differently—that's the feature, but it is a change.
- **`min_p > 1` now decodes greedily** where `main` silently ignored it. Fallout of the above.
- **`BatchGenerator(sampler=…, greedy_sampling=True)`** now honors the sampler where `main` took the
  fused-greedy path and ignored it.

### Scope

An earlier revision of this PR added `ge`/`gt`/`le` bounds to the sampling fields on `OpenAIRequest`
and `VLMRequest`. That 422'd six request shapes `main` accepts—including `top_k=-1`, the
conventional "disabled" spelling—on two of three endpoints, for a change that is supposed to be
about sampling. **Reverted entirely**: `server/schemas.py` is byte-identical to `main` and no longer
appears in the diff. What replaces it is a test asserting every shape `main` admitted still reaches
the sampler and produces a valid in-range token, since with no validation the sampler is the only
thing between a nonsensical request and a crash.

### Performance

Sampler cost at V=151936, median of 7 × 30 reps:

| shape | before | after | bare `argmax` |
|---|---|---|---|
| `[1, V]` | 1.488 ms | 0.297 ms | 0.290 ms |
| `[16, V]` | 3.174 ms | 0.290 ms | 0.292 ms |
| `[8, 16, V]` verify block | **17.163 ms** | **0.413 ms** | 0.418 ms |

Greedy went from 5.1x–39.5x a bare `argmax` to 1.0x. Temperature > 0 also got ~2x faster, because
the parity fix replaced `softmax(sl/t)` with `exp(sl)`.

End to end, Qwen3-VL-8B-Instruct-8bit, 8 concurrent × 192 tokens at temperature 0, median of 7:

```
branch before   162.5 tok/s
branch after    165.5 tok/s
origin/main     165.1 tok/s     <- at parity
```

### Verification

- **Suite: 2619 passed / 7 skipped.** pre-commit clean over all files.
- **Every fix is mutation-tested**—revert the fix, confirm its test fails, restore. Where a mutant
  turned out to be semantically equivalent I've said so rather than inventing a test to "kill" it.
- **The speculative tests are no longer mocked.** The blindness that hid these bugs was that every
  test touching the speculative path monkeypatched the round loop. The new ones drive the real
  `run_speculative_server_rounds` / `_mtp_rounds` with only the model forward stubbed.
- **Live dflash** (`z-lab/Qwen3.5-4B-DFlash`): temperature 0 and 0.8, plus 4 concurrent requests
  co-batching at `batch=4`, `accept=2.55`, **0 errors**—the path that raised on every request before.
- **Live MTP** (`Qwen3.5-4B` split to a `qwen3_5_mtp` drafter): 4 concurrent *heterogeneous* rows
  (distinct temperature/top_k/top_p/min_p/seed) decoding together, same seed reproducing byte-identical
  output, and the ragged single-row path confirmed entered (`sample_target` receiving 3 rows against
  1 config). 192.1 tok/s with the drafter vs 160.9 without, so speculation genuinely engaged.
- **125,194 independent differential cases** re-derived parity without reusing the test file's own
  helpers. Zero code defects; the six discrepancies it found were all errors in my *documentation* of
  the divergences, now corrected.

@lucasnewman—your approval is pinned to `a343cd6` and is 12 commits stale, so it would be worth a
fresh look if you have time; the sampler internals changed substantially. @Blaizzy—the
normalization question above is the one decision I'd like your call on before this merges.

Issue #1648, opened 2026-07-21
Support modern samplers and higher-quality quantization options


Blaizzy:
## Summary

Add support for modern decoding samplers beyond the current min-p option, and expose higher-quality quantization options where they fall within mlx-vlm's scope.

## Motivation

Feedback from users running models on-device noted that sampling and quantization choices can materially affect output quality. llama.cpp already exposes newer samplers such as top-n-sigma, while mlx-vlm currently appears limited to min-p among modern alternatives.

Source discussion:
- https://news.ycombinator.com/item?id=48990705
- https://news.ycombinator.com/item?id=48991935

## Proposed scope

- Evaluate and implement top-n-sigma.
- Evaluate the additional samplers referenced in the source discussion.
- Expose sampler configuration through the Python API, server API, and CLI where applicable.
- Document valid ranges, defaults, and interactions between samplers.
- Evaluate whether higher-quality quantization algorithms belong in mlx-vlm or an upstream MLX package; document or track the appropriate integration point.
- Add deterministic tests covering parameter validation and sampling behavior.

## Acceptance criteria

- At least one modern sampler beyond min-p is available through the public generation APIs.
- Unsupported or conflicting sampler combinations return clear validation errors.
- Server and library documentation list all supported samplers and their parameters.
- The quantization portion has either an implementation path or a clearly documented upstream dependency.

PR #1653, opened 2026-07-22
feat(sampler): add p-less sampling

Lazarus-931:
this is adding much more sampling support to mlx-vlm, specifically p-less, where each step keeps tokens whose probability ≥ the collision probability Σp² (= exp(−H₂)) of the temperature-scaled distribution, then renormalizes and samples. 

I've exposed it via make_sampler, generate_step, and the server API (p_less, default off). 

verified against deterministic + numpy-reference tests. 

part of #1648.


mlx_vlm/sample_utils.py
```diff
@@ -151,6 +158,29 @@ def apply_top_k(
     return masked_logprobs
 
 
+@partial(mx.compile, inputs=mx.random.state, outputs=mx.random.state)
```
lucasnewman: No need to capture random state if we have no random sampling.

lucasnewman (approved):

Issue #1654, opened 2026-07-22
Validation errors raised inside `@mx.compile` samplers corrupt MLX trace state


Lazarus-931:
In `mlx_vlm/sample_utils.py`, `apply_min_p` and `apply_top_k` validate their parameters and `raise ValueError` from **inside** an `@mx.compile`-decorated function. Raising during an `mx.compile` trace corrupts the compile state, so **every subsequent compiled call in the same process** fails with `RuntimeError: [eval] Attempting to eval an array without a primitive`.

### Repro
```python
import mlx.core as mx
from mlx_vlm.sample_utils import apply_min_p, make_sampler

x = mx.array([[1.0, 2.0, 3.0, 4.0, 5.0]])

# 1) trigger the (expected) validation error
try:
    mx.eval(apply_min_p(x, -1.0))
except ValueError:
    pass

# 2) any compiled sampler afterwards now breaks:
make_sampler(temp=1.0, min_p=0.1)(x)   # RuntimeError: eval an array without a primitive
```

### Fix
Validate in a thin, non-compiled wrapper and keep only the math in the `@mx.compile` inner function:

```python
def apply_min_p(logprobs, min_p, min_tokens_to_keep=1):
    if not (0 <= min_p <= 1.0):
        raise ValueError(...)
    return _apply_min_p(logprobs, min_p, min_tokens_to_keep)

@partial(mx.compile, inputs=mx.random.state, outputs=mx.random.state)
def _apply_min_p(logprobs, min_p, min_tokens_to_keep):
    ...
```

Same applies to `apply_top_k`.

Lazarus-931:
Fixed by #1655.

PR #1655, opened 2026-07-22
Fix compile validation

Lazarus-931:
Problem: two samplers (min-p, top-k) validated their settings inside a compiled function. On MLX, throwing an error there corrupts the compiler's state, so the next sampler call crashes with eval an array without a primitive. 

Fix: move the checks into a plain wrapper and keep the math compiled. 

Verification: the crash repro now passes, and the full sampler suite plus two new regression tests are all green!


Lazarus-931:
fixes: #1654

lucasnewman (approved):
LGTM

Issue #1651, opened 2026-07-21
qwen3_omni_moe audio path: three stacked bugs for non-generate use (batched chunker slice, sample-domain feature mask -> 300GB allocs, ~80x audio token over-expansion)


FreazeMood:
### Summary

Extracting hidden states (or any non-generate use) of `qwen3_omni_moe` with real-length audio (>~5s) fails or silently degrades through three stacked bugs in the audio path. Found while running last-token feature extraction over 16 kHz mono WAVs on `mlx-community/Qwen3-Omni-30B-A3B-Instruct-4bit` (mlx-vlm 0.6.5, M4 Max). Repro + working patches for all three below.

### Bug 1 — audio chunker assumes unbatched `(mel, T)` input

`models/qwen3_omni_moe/audio.py`, chunk-padding loop:

```python
padded_feature[i, :, :chunk_len] = input_features[:, start_idx:end_idx]
```

`input_features` arrives batched `(1, 128, T)` from the standard path, so the slice cuts the **mel axis** instead of time and the assignment broadcasts garbage (`(1,1,100,200)` vs `(1,128,100)` shape error). Fix that worked for us:

```python
if input_features.ndim == 3:
    input_features = input_features[0]
```

(or slice `input_features[0, :, start_idx:end_idx]`).

### Bug 2 — `feature_attention_mask` is sample-domain, consumed as mel-domain

`thinker.py::get_audio_features` computes `feature_lens = feature_attention_mask.sum(-1)`, but the mask produced by the processor path is over **raw audio samples** (e.g. 320,000 for 20 s), not mel frames (2,000). The tower then builds `320000/100 = 3200` chunks and attempts absurd allocations — we saw `metal::malloc` requests of **138–311 GB** for 20–60 s clips. The same function's unpad block (`mx.take` over the flattened features by that mask) corrupts the features for the same reason.

Workaround that works: bypass the mask entirely and pass `audio_feature_lengths` in mel frames (`mx.array([input_features.shape[-1]])`). A real fix should normalize the mask domain (`mask.sum(-1) // hop_length`) or document that `audio_feature_lengths` is required.

### Bug 3 — audio placeholder expansion mints ~80x too many tokens

[...]

PR #1656, opened 2026-07-22
fix(qwen3_omni_moe): accept batched (1, mel, T) audio features in encoder

Lazarus-931:
In qwen3_omni_moe, the audio encoder's chunk-padding loop ([audio.py](<http://audio.py>)) does padded_feature\[i, :, :chunk_len\] = input_features\[:, start_idx:end_idx\], which assumes an unbatched (mel, T) input. The standard path hands it a batched (1, 128, T) tensor, so the slice cuts the mel axis instead of time and the write fails with ValueError: \[broadcast_shapes\] Shapes (1,1,100,200) and (1,128,100) cannot be broadcast (issue Blaizzy/mlx-vlm#1651, Bug 1).

This fixes by unbatching a leading dim of 1 at the top of AudioModel.**call** (if input_features.ndim == 3: input_features = input_features\[0\]) so the rest of the function always sees (mel, T). Unbatched input is unaffected, so it's fully backward-compatible.

All tests green

EDIT: removed regression test

### DIFF
--- a/mlx_vlm/models/qwen3_omni_moe/audio.py
+++ b/mlx_vlm/models/qwen3_omni_moe/audio.py
@@ -196,6 +196,11 @@ class AudioModel(nn.Module):
         feature_lens: Optional[mx.array] = None,
         aftercnn_lens: Optional[mx.array] = None,
     ):
+        # Accept the batched (1, mel, T) layout from the standard path as well as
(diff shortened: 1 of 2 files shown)


lucasnewman (approved):
Good catch!

Blaizzy (approved):
LGTM, thanks!

PR #1657, opened 2026-07-22
fix(qwen3_omni_moe): treat audio feature mask as sample-domain (mel-frame lengths)

Lazarus-931:
feature_attention_mask is over raw audio samples (e.g. 320,000 for 20s), but both thinker.get_audio_features (the audio tower's feature_lens) and the processor (the <|audio_pad|> placeholder count) consumed sum(mask) as if it were mel frames. For any real clip (>~5s) this made the tower build ~160× too many chunks, the 138–311 GB metal::malloc failures,  and the processor mint ~160× too many audio tokens (issue #1651, Bugs 2 & 3).

I convert the mask sum to mel frames via the mask/mel ratio (mask.sum // hop) in both places, using the identical normalization so the minted placeholder count always equals the encoder's output length. get_audio_features was also rewritten to unbatch and select each item's valid mel frames before concatenation.

When verifying, the tower's feature_lens drops from 320,000 → 2,000 for 20s audio (261 tokens / ~20 chunks, no giant allocation), and the processor mints 260 tokens for 20s and 390 for 60s (down from 41,600 / 62,400, about ~160× fewer). 

edit: removed the regression test

### DIFF
--- a/mlx_vlm/models/qwen3_omni_moe/processing_qwen3_omni_moe.py
+++ b/mlx_vlm/models/qwen3_omni_moe/processing_qwen3_omni_moe.py
@@ -107,11 +107,15 @@ class Qwen3OmniMoeProcessor(ProcessorMixin):
                 "attention_mask", None
             )
             audio_inputs["input_features"] = audio_inputs.pop("input_features", None)
-            audio_lengths = iter(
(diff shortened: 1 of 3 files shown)


lucasnewman (approved):
LGTM

PR #1659, opened 2026-07-22
chore(deps): make training-only datasets an optional [train] extra

Lazarus-931:
configures it so that mlx-vlm for lora is when other libraries are downloaded, inference-only regime therefore stays lean

verified the full inference surface imports without datasets; lora/evals raise the friendly error when not downloaded properly.

docs updated as well


Blaizzy (approved):
LGTM, thanks!

johnking0099:
Good for inference-only users.

PR #1660, opened 2026-07-22
fix(utils): resolve model module before AutoProcessor in load_processor

andyrost:
## Problem

A standalone `load_processor()` call fails for models whose processor classes are registered by mlx-vlm itself, while `load()` on the same path works:

```python
from mlx_vlm.utils import load_processor
load_processor("<local qwen3_5 quant written by mlx_vlm.convert>")
# ValueError: Unrecognized image processor in ...
```

Reproduced on current `main` (torch-free venv, transformers 5.14.1, macOS/Apple Silicon) with a Qwen3.5-based quant produced by `mlx_vlm.convert` (microsoft/Fara1.5-9B → 4-bit). `mlx_vlm.load()` on the identical path returns `Qwen3VLProcessor` fine.

## Root cause

30+ model modules install their `AutoProcessor` registrations as an **import side-effect** (`install_auto_processor_patch` in `qwen3_5`, `gemma3`/`gemma4`, `kimi_k25`, `molmo`, `phi3_v`, `deepseekocr`, …). `load()` triggers those imports via `load_model()` → `get_model_and_args`, but `load_processor()` calls `AutoProcessor.from_pretrained` directly — so whether it works depends on whether the caller happened to load weights first in the same process. Anyone using `load_processor` standalone (chat-template rendering, request validation in a server, tokenizer-only tooling) hits the raw transformers error.

## Fix

`_ensure_model_module_registered(model_path)`: best-effort resolution of the model module before the `AutoProcessor` call, mirroring what `load_image_processor` already does — but reading **only `config.json`** (local dirs directly; hub repos via `hf_hub_download`, never a weights snapshot). Every failure mode falls through to the previous behavior, so this can only widen the set of models that load.

Also normalizes an existing-local-dir `str` to `Path` at the top of `load_processor`: the `add_detokenizer` branch does `model_path / "tokenizer.json"`, which previously crashed with `TypeError` for plain-string local paths (second latent bug on the same standalone path).

## Verification

- Before: `load_processor(<qwen3_5 quant>)` raises `ValueError: Unrecognized image processor`; after: returns `Qwen3VLProcessor` with detokenizer + stopping criteria attached (str and `Path` args both).
- `load()` on the same model unregressed.
- `pytest mlx_vlm/tests/test_utils.py` — 36/36 pass (34 existing + 2 new regression tests: registration ordering, and best-effort fall-through when no config is reachable).

### DIFF
--- a/mlx_vlm/tests/test_utils.py
+++ b/mlx_vlm/tests/test_utils.py
@@ -781,6 +781,48 @@ def test_load_processor_propagates_auto_processor_errors():
             load_processor(Path("/tmp/model"), eos_token_ids=2)
 
 
+def test_load_processor_registers_model_module_before_auto_processor(tmp_path):
+    """Standalone load_processor must import the model module first so
+    import-time AutoProcessor registrations (install_auto_processor_patch)
+    are visible — load() got this implicitly via load_model."""
+    (tmp_path / "config.json").write_text(json.dumps({"model_type": "qwen3_5"}))
+    call_order = []
+
+    with (
+        patch(
+            "mlx_vlm.utils.get_model_and_args",
+            side_effect=lambda cfg: call_order.append(("register", cfg["model_type"])),
+        ),
+        patch(
+            "mlx_vlm.utils.AutoProcessor.from_pretrained",
+            side_effect=lambda *a, **k: call_order.append(("auto_processor",))
+            or MagicMock(),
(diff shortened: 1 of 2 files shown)


Blaizzy:
Thanks @andyrost for the contribution!

However, closing it for now because this is not the core issue.

PR #1662, opened 2026-07-22
Add DRY (Don't Repeat Yourself) repetition sampler

Lazarus-931:
Adds the DRY sampler as a logits processor, which works when tokens that would extend a repeated suffix of length L >= allowed_length are penalized by multiplier, with sequence breakers and a searchable-history window. Wired through make_logits_processors, generate_step, and the server, with unit tests and validation coverage.

Sampling series of #1648


lucasnewman:
@Lazarus-931 Do we still want this? I seem to recall it came from a HN comment (aka rant) although I've never really heard of it being used in practice.

Lazarus-931:
in retrospect,  i was just trying to check a box from the list,  I do see it used in llama.cpp, but I think this un-needed. can be re-visited ltr on

PR #1663, opened 2026-07-22
Add typical-p (locally typical) sampler

Lazarus-931:
Adds locally typical sampling (arXiv 2202.00666): keeps the most typical tokens (surprisal closest to the distribution entropy) until their cumulative probability reaches typical_p. Wired through make_sampler, generate_step, and the server (schemas/generation/request), with a numpy-reference test and validation coverage.

Sampler series from #1648


Blaizzy (approved):
LGTM, thanks!

PR #1666, opened 2026-07-23
Awq quant

Lazarus-931:
This pr adds calibration-aware weight quantization (AWQ, arXiv 2306.00978) with `-quant-method awq`

A short calibration pass records per-channel activation magnitude; each decoder layer's weights are rescaled to protect high-activation channels (grid-searched to minimize quantized-output error) and the inverse scale is folded into the preceding op so the fp model is unchanged before quantization. 

this is scoped to the language model, vision/audio support is under testing and will be included 

folding is gated to standard nn.RMSNorm (x·weight) decoders like Qwen 2/2.5/3(-VL), Llama, etc where it recovers ~20–80% of the RTN→bf16 perplexity gap (measured on Qwen3-VL-2B, Qwen2.5-0.5B, Llama-3.2-1B). 

Architectures with non-standard norms (the Gemma family, x·(1+weight)) safely fall back to plain RTN rather than being corrupted, also this has a regresssion test

series of #1648

edit: added the opt-in multimodal calibration so the LM's AWQ scales reflect media-token activations

edit 2: marking this ready since majority of audio models at large don't support AWQ


Blaizzy (approved):
LGTM!

PR #1702, opened 2026-07-23
Vendor standard models

Lazarus-931:
Continues the mlx-lm removal series (after #1606/#1616/#1625): 17 models vendored (10 dense + 7 MoE), each bit-exact Δ=0.0 vs mlx-lm 0.31.3, all recovered by test_models.py


Blaizzy (approved):
LGTM, thanks!

PR #1704, opened 2026-07-24
Shared transformer block

Lazarus-931:
Most of our text backbones are the same decoder with small variations — same
attention, same gated MLP, same pre-norm residual — so the stack gets reimplemented
in 20+ places. This replaces that with one config-driven `TransformerBlock` and moves
19 models onto it.

The block is a `BlockSpec` plus a pluggable token mixer. Attention covers GQA/MHA,
QK-norm (per-head/full-dim/post-rope), qkv and output biases, sliding windows, logit
softcapping, and per-layer NoPE, with the model's own RoPE passed in. The MLP is a
gated dense MLP or a MoE path on the existing `SwitchGLU` (simple top-k and grouped
routers, per-layer so dense/sparse mix). Norm can be RMSNorm, Gemma's (1 + weight),
or LayerNorm; residual can be pre-norm, Gemma2 sandwich, or OLMo2 post-norm. The
mixer sits behind a small `self_attn(x, mask, cache)` seam, so MLA/SSM mixers can
drop in later.

Migrated: llama, qwen2, qwen3, gemma2, olmo2, olmo3, helium, internlm3, exaone4,
ernie4_5, seed_oss, mimo, telechat3, hunyuan_v1_dense (dense) and ernie4_5_moe,
dots1, exaone_moe, qwen3_moe, olmoe (MoE). smollm3 comes along free via llama.

Pure refactor — weight and submodule names are unchanged, so checkpoints load as-is.
Every model is bit-exact: max|Δ logits| is 0.0 on prefill and cached decode vs the
old implementation, and the full test suite passes. A follow-up will extend the block
(fused gate/up projection, residual scaling, parallel residual) and bring more models over.

---

This is a simplification, so it removes far more than it adds: **+1,330 / −2,126,
net −796 lines**. The shared block costs +357 once and tests +260; the 19 model files
drop ~1,400 lines net (713 of config replacing 2,126 of hand-rolled decoder code).
Every model added after this is almost pure deletion.


Blaizzy:
Please resolve conflicts

lucasnewman:
I understand the temptation to do this but I think it might be a long-term maintenance problem:

1) Any changes to the shared block now need to revalidate all 20+ models that use it.
2) Any kind of dumb agent workflow will almost certainly reach for the helper even if there's very subtle differences in behavior in architecture that deviate from the standard construction or already-provided options. If they do identify subtle behavior differences correctly, it will need to modify the helper and then we're back to 1), i.e. the helper itself is now a "dangerous" core file to modify.
3) This makes it harder for humans to understand how a model works because you can't just read the code linearly anymore and have to jump around to get the full picture.
4) Likewise, porting the model to a different language (Swift or whatnot) now needs to reconstruct the behavior of the helper + model from multiple files instead of a direct port.

Just my 2¢ though!

Lazarus-931:
> I understand the temptation to do this but I think it might be a long-term maintenance problem:
> 
> 1. Any changes to the shared block now need to revalidate all 20+ models that use it.
> 2. Any kind of dumb agent workflow will almost certainly reach for the helper even if there's very subtle differences in behavior in architecture that deviate from the standard construction or already-provided options. If they do identify subtle behavior differences correctly, it will need to modify the helper and then we're back to 1), i.e. the helper itself is now a "dangerous" core file to modify.
> 3. This makes it harder for humans to understand how a model works because you can't just read the code linearly anymore and have to jump around to get the full picture.
> 4. Likewise, porting the model to a different language (Swift or whatnot) now needs to reconstruct the behavior of the helper + model from multiple files instead of a direct port.
> 
> Just my 2¢ though!

reading, it totally agree, this would cause those problems, closing!

PR #1708, opened 2026-07-24
Add Mage-Flow image model

lucasnewman:
New image generation / editing model from Microsoft, with various model flavors for editing and few-step generation.

[Model Card](https://huggingface.co/microsoft/Mage-Flow)

Sample from the 4-step Turbo model:
> An editorial product photograph of a cobalt-blue ceramic teapot on a coral-red table, soft window light, crisp shadows, highly detailed.
teapot


lucasnewman:
Note I updated the black version for pre-commit to the latest as we were pinned to an older version, hence the minor formatting changes in other places.

Lazarus-931 (approved):

Lazarus-931:
sgtm!

PR #1710, opened 2026-07-24
fix(rope): apply declared rope_scaling in models that ignored it (Llama-3.2-Vision and 9 others)

1qh:
## Problem

Ten models declare `rope_scaling` on their config dataclass and then build a plain `nn.RoPE` without passing it, so any checkpoint that ships scaling runs **unscaled** positions. The error is exactly zero at position 0 and grows with context length, so it does not show up in short-prompt testing.

**Llama-3.2-Vision is affected today.** `meta-llama/Llama-3.2-11B-Vision-Instruct` ships:

```json
"rope_scaling": {"rope_type": "llama3", "factor": 8.0, "high_freq_factor": 4.0,
                 "low_freq_factor": 1.0, "original_max_position_embeddings": 8192}
```

while `mllama/language.py` builds `nn.RoPE(..., scale=1)`. Comparing the required `Llama3RoPE` against that plain RoPE on identical inputs with the real config:

| position | cosine(correct, current) |
|---|---|
| 0 | 1.000000 |
| 1 000 | 0.959 |
| 8 192 | 0.757 |
| 20 000 | 0.619 |
| 60 000 | 0.503 |

`llama3` scaling is frequency-dependent, not a uniform divide, so it cannot be approximated by the unscaled path — the model's long-context behaviour is what suffers.

## Fix

Route each through the existing `rope_utils.initialize_rope`, already the pattern in `gemma4` and `diffusion_gemma`. Where a config declares no `rope_scaling`, `initialize_rope` returns the same unscaled `nn.RoPE`, so behaviour is unchanged for every current checkpoint of the other nine.

Models changed: `mllama`, `glm4_moe`, `qwen3_moe`, `granite_vision`, `granite4_vision`, `mixtral`, `molmo2`, `molmo_point`, `phi4mm`, `locateanything`.

## Deliberately not changed

`internvl_chat` declares `{"type": "dynamic", "factor": 2.0}` (e.g. InternVL2-8B). `initialize_rope` raises on `dynamic`, so routing it there would turn a silent inaccuracy into a load-time failure. It needs dynamic-NTK support first — flagging it here rather than papering over it.

## Tests

- `test_mllama_rope_scaling_is_applied` — asserts a `llama3` config yields `Llama3RoPE`. Fails on `main`.
- `test_mllama_without_rope_scaling_is_plain` — asserts the no-scaling path stays unscaled.

`python -m unittest mlx_vlm.tests.test_models` passes.

Companion to #1709, which fixes the same class in `gemma3`/`gemma3n` (found first, verified end-to-end against a real gemma-3 checkpoint).

### DIFF
--- a/mlx_vlm/models/glm4_moe/language.py
+++ b/mlx_vlm/models/glm4_moe/language.py
@@ -12,6 +12,7 @@ from ..base import (
 from ..mlp import DeepseekMLP as MLP
 from ..pipeline import PipelineMixin
 from ..switch_layers import SwitchGLU
+from ..rope_utils import initialize_rope
 from .config import ModelConfig
 
 
@@ -36,10 +37,12 @@ class Attention(nn.Module):
             self.q_norm = nn.RMSNorm(head_dim, eps=args.rms_norm_eps)
             self.k_norm = nn.RMSNorm(head_dim, eps=args.rms_norm_eps)
 
-        self.rope = nn.RoPE(
-            int(head_dim * args.partial_rotary_factor),
-            traditional=False,
+        self.rope = initialize_rope(
+            dims=int(head_dim * args.partial_rotary_factor),
             base=args.rope_theta,
+            traditional=False,
+            scaling_config=args.rope_scaling,
+            max_position_embeddings=args.max_position_embeddings,
         )
 
     def __call__(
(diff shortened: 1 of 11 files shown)


lucasnewman:
Thanks for reporting this! We'll resolve it in https://github.com/Blaizzy/mlx-vlm/pull/1927

PR #1711, opened 2026-07-25
Support mixed-precision compressed-tensors checkpoints (fp8 + NVFP4)

### LINKED ISSUE #1706: Mixed-precision compressed-tensors checkpoints (fp8 + NVFP4) fail to load — "parameters not in model"
## Summary

Loading a `compressed-tensors` checkpoint whose top-level `format` is `mixed-precision` fails: the NVFP4 fold is skipped, the packed tensors reach `model.load_weights(strict=True)`, and startup aborts (status 3).

## Root cause

`_transform_compressed_tensors_weights` (`mlx_vlm/utils.py:258`) only dispatches when the **top-level** `format` is `"nvfp4-pack-quantized"` (`:284`) or `"pack-quantized"` + int (`:288`); anything else returns the weights unchanged (`:291`). A mixed-precision export sets top-level `format: "mixed-precision"` and puts the real formats in **per-group** `config_groups`, so the fold never runs and the NVFP4 tensors (`weight_packed`, `weight_scale`, `input_global_scale`, `weight_global_scale`) survive to `model.load_weights(..., strict=True)` (`:673`).

## Reproduce

* Model: `unsloth/Qwen3.6-27B-NVFP4` (`model_type: qwen3_5`, `Qwen3_5ForConditionalGeneration`)
* `mlx-vlm 0.6.5+`, `transformers 5.14.1`, Apple silicon.

Its `quantization_config` (`quant_method: compressed-tensors`, `format: mixed-precision`) has two groups:

* **group_1** -> `format: "nvfp4-pack-quantized"`, 4-bit, group_size 16, `targets: mlp.(gate|up|down)_proj` -- this is what crashes.
* **group_0** -> `format: "float-quantized"`, **8-bit fp8,** `strategy: channel`, `targets: self_attn.*_proj`, `linear_attn.*`, `lm_head`, layers 56-63 MLP.
* Large `ignore` list: all `visual.blocks.*`, every `linear_attn` layer, `re:^mtp.*`.

```
...mlp.up_proj.weight_packed, ...up_proj.weight_scale.
ERROR:   Application startup failed. Exiting.
```

## What's needed

1. In the dispatcher, when `format == "mixed-precision"`, iterate `config_groups` and route each group by its own `format`/`targets` (the existing NVFP4 fold already handles the `nvfp4-pack-quantized` group).
2. Add a path for **channel-wise fp8** `float-quantized` (num_bits 8, `strategy: channel`) -- mlx modes are affine/mxfp4/nvfp4/mxfp8, none is per-channel fp8; simplest is to dequantize those layers to dense.
3. Apply quantization **per** `targets`**/**`ignore` so vision, `linear_attn`, and `mtp` stay unquantized.

## Related

Downstream: Blaizzy/nativ#85. Sibling of Blaizzy/mlx-vlm#1703 (gpt-oss `MXFP4-Q8`, also a mixed compressed-tensors export).

Note: verifying end-to-end needs hardware that holds the ~23 GB model (does not fit a 16 GB machine).

Lazarus-931:
Fixes #1706.

A `format: "mixed-precision"` compressed-tensors export keeps the real per-group formats in `config_groups`, so the NVFP4 fold was skipped and the packed tensors reached `model.load_weights(strict=True)`, aborting startup with "parameters not in model".

Route `mixed-precision` to a new transform that dispatches each quantized Linear by the tensors it actually carries (the group assignment already encodes this, so no fragile regex target matching):
- `.weight_packed` + `.weight_global_scale` -> NVFP4 fold
- `.weight_packed` alone -> INT4 pack-quantized fold
- `.weight_scale` without a packed sibling -> channel-wise fp8 `float-quantized`, dequantized to a dense weight (MLX has no per-channel fp8 mode)

Ignored layers (vision, linear_attn, mtp) carry a plain `.weight`, pass through untouched, and stay dense since the later nn.quantize pass only touches modules that have a `.scales` tensor. Compressed-tensors metadata with no MLX consumer (weight_shape, weight_global_scale, weight_zero_point, input_scale/input_global_scale, kv-cache k/v scales) is dropped before the strict load. A checkpoint mixing more than one MLX-native quantized mode raises rather than silently mis-quantizing.

Verification
Added TestMixedPrecisionCompressedTensors covering routing/metadata-drop, exact fp8 dequant values, an end-to-end nn.quantize → load_weights(strict=True) → forward (the line that used to abort), and the multi-mode guard; the full CI suite as well

### DIFF
--- a/mlx_vlm/tests/test_models.py
+++ b/mlx_vlm/tests/test_models.py
@@ -10985,3 +10985,189 @@ class TestVendoredDenseModels(unittest.TestCase):
 
     def test_afmoe(self):
         self._run("afmoe")
+
+
+class TestMixedPrecisionCompressedTensors(unittest.TestCase):
+    """``format: mixed-precision`` compressed-tensors (NVFP4 + channel fp8)."""
+
+    OUT = 32
+    IN = 32
+
+    @staticmethod
+    def _config():
+        # Mirrors unsloth/Qwen3.6-27B-NVFP4: fp8 float-quantized attn/lm_head
+        # in group_0, NVFP4-packed MLP in group_1, one top-level format.
+        return {
+            "quant_method": "compressed-tensors",
+            "format": "mixed-precision",
+            "config_groups": {
+                "group_0": {
+                    "format": "float-quantized",
+                    "weights": {
+                        "num_bits": 8,
+                        "type": "float",
+                        "strategy": "channel",
+                    },
+                },
+                "group_1": {
+                    "format": "nvfp4-pack-quantized",
+                    "weights": {"num_bits": 4, "group_size": 16},
+                },
+            },
+        }
+
+    def _weights(self):
+        out, inp = self.OUT, self.IN
+        rng = np.random.default_rng(0)
+        weights = {}
+        # NVFP4-packed MLP projection (group_1).
+        weights["model.layer.mlp.gate_proj.weight_packed"] = mx.array(
+            rng.integers(0, 256, (out, inp // 2), dtype=np.uint8)
+        )
+        weights["model.layer.mlp.gate_proj.weight_scale"] = mx.full(
+            (out, inp // 16), 0x38, dtype=mx.uint8  # E4M3 1.0
+        )
+        weights["model.layer.mlp.gate_proj.weight_global_scale"] = mx.array(
+            [1.0], dtype=mx.float32
+        )
+        # Metadata with no MLX consumer -- must be dropped.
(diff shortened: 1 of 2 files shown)


Blaizzy:
Please check #1664

Lazarus-931:
yeah, i thought initially tha'll fix it, but this pr touches _transform_compressed_tensors_weights dispatcher in utils.py, which runs before any model sanitize, so don't relate

Blaizzy (approved):
LGTM, thanks!

PR #1713, opened 2026-07-25
Preserve APC prefix reuse for growing prepared prompts

### LINKED ISSUE #1712: APC semantic hash changes when a prepared prompt is extended
## Description

Automatic Prefix Caching does not reuse a cached prefix when a prepared prompt is extended with additional text.

The server preprocessing path creates `inputs_embeds` and `attention_mask` tensors for the complete prompt. These internally derived, sequence-shaped tensors are included in `BatchGenerator._apc_extra_hash()`. Extending the prompt changes their shapes and contents, which changes the global semantic hash used for APC lookup.

As a result, the longer request is placed in a different APC namespace even though its initial token sequence is an exact prefix match for the earlier request.

Explicit caller-provided embeddings and masks can affect model semantics and must remain part of the hash. The issue concerns tensors produced internally from the same token sequence that APC already indexes.

## Reproduction

Run the following script against the current `main` branch:

```python

import mlx.core as mx

from mlx_vlm.generate import BatchGenerator
from mlx_vlm.server.generation import (
    GenerationArguments,
    GenerationContext,
    ResponseGenerator,
)


class EmbeddingOutput:
    def __init__(self, sequence_length: int):
        self.inputs_embeds = mx.ones((1, sequence_length, 4))

    def to_dict(self):
        return {"inputs_embeds": self.inputs_embeds}


class Model:
    def __init__(self):
        self.language_model = SimpleNamespace()

    def get_input_embeddings(self, input_ids, pixel_values, mask=None, **kwargs):
        return EmbeddingOutput(input_ids.shape[1])


class RequestSink:
    def __init__(self):
        self.requests = []

    def put(self, request):
        self.requests.append(request)
        request.rqueue.put(
            GenerationContext(uid=len(self.requests), prompt_tokens=request.prompt_tokens)
        )


def admitted_prompt_kwargs(sequence_length: int):
    generator = ResponseGenerator.__new__(ResponseGenerator)
    generator.wait_until_ready = lambda: None
    generator.draft_model = None
    generator.apc_manager = object()
    generator.apc_mode = "block"
    generator.model = Model()
    generator.processor = SimpleNamespace()
    generator._cancel = lambda uid: None
    generator._preprocess_request = lambda *args, **kwargs: {
        "input_ids": mx.arange(sequence_length)[None, :],
        "attention_mask": mx.ones((1, sequence_length), dtype=mx.int32),
    }
    generator.requests = RequestSink()
    args = GenerationArguments(
        max_tokens=1,
        tenant_id="repro-tenant",
    )

    _, token_iter = generator.generate("prompt", args=args)
    token_iter.close()
    request = generator.requests.requests[0]

    semantic_hash = getattr(request, "apc_semantic_hash", None)
    if semantic_hash is None:
        _, prompt_kwargs = generator._gpu_embed(request.raw_inputs, request.images)
        prompt_kwargs["_apc_tenant"] = args.tenant_id
    else:
        _, prompt_kwargs = generator._gpu_embed(
            request.raw_inputs,
            request.images,
            apc_semantic_hash=semantic_hash,
        )
    return prompt_kwargs


batch_generator = SimpleNamespace(apc_manager=object())
short_hash = BatchGenerator._apc_extra_hash(
    batch_generator, admitted_prompt_kwargs(8)
)
extended_hash = BatchGenerator._apc_extra_hash(
    batch_generator, admitted_prompt_kwargs(12)
)

print(f"short prompt hash:    {short_hash}")
print(f"extended prompt hash: {extended_hash}")

assert short_hash == extended_hash, (
    "Appending text changed APC's global semantic hash, so the cached token "
    "prefix cannot be looked up."
)
```

## Actual behavior

On `upstream/main` at commit `20fad4fd`, the script prints:

```text
short prompt hash:    2146270371181291635
extended prompt hash: 374550709281470985
```

It then fails with:

```text
AssertionError: Appending text changed APC's global semantic hash, so the cached token prefix cannot be looked up.
```

## Expected behavior

The internally prepared 8-token prompt and its 12-token extension should retain the same global semantic hash. APC can then search for the longest cached token prefix within that namespace.

Explicit custom embeddings, masks, media, tenant scope, and model dependencies should continue to affect the semantic hash.

byoan:
## Summary

This change prevents internally derived full-prompt tensors from changing the APC semantic namespace whenever text is appended to a prompt.

It covers both generation paths:

- Continuous batching through `BatchGenerator`
- Direct generation through `stream_generate`

## Problem

VLM preprocessing derives `inputs_embeds` and `attention_mask` from the complete prompt. APC previously included those tensors in its extra semantic hash.

Appending text changes the derived tensor shapes and contents, producing a new hash before APC performs its token-prefix lookup. The extended request therefore cannot find the cached state of the shorter prompt.

At the same time, caller-provided embeddings and masks must remain in the hash because they can change model semantics independently of the token IDs.

## Changes

- Compute the complete server-side APC semantic hash once before queue admission, after tenant and processed media inputs are available.
- Carry that precomputed integer through the request and return it directly from `BatchGenerator._apc_extra_hash()`.
- Strip the private APC metadata before prompt kwargs reach model forward calls, including speculative server prefill.
- Keep the existing semantic hash calculation as a fallback for direct `BatchGenerator` callers.
- Capture explicit caller-provided embeddings and masks before direct-stream preprocessing and continue hashing them.
- Fingerprint processed image content once before queue admission and carry only the resulting hash into the generation thread.
- Preserve image tensor shape, dtype, and exact bytes in that fingerprint.
- Reject requests that exceed the configured context budget before hashing processed media.
- Cache the model's APC mode during model initialization and skip semantic hashing entirely for unsupported cache layouts.
- Preserve existing media, tenant, model, and other semantic hash dependencies.
- Add regression coverage for growing prepared prompts and explicit custom tensors in continuous and direct generation paths.

## Reproduction results

Before this change, the reproduction produces different hashes for the shorter prompt and its extension:

```text
short prompt hash:    2146270371181291635
extended prompt hash: 374550709281470985
```

With this change, both requests remain in the same APC semantic namespace:

```text
short prompt hash:    7088136067003016882
extended prompt hash: 7088136067003016882
```

APC can therefore perform its normal longest-prefix lookup and reuse the cached state of the shorter request.

## Safety

Only the server path uses the precomputed hash. Direct `BatchGenerator` callers still hash explicit custom `inputs_embeds` and `attention_mask` values, so distinct caller-controlled model inputs do not collide.

Processed image content, audio, video, tenant, and model dependencies also continue to participate in cache isolation. Reusing a mutable path or URL with different processed pixels therefore produces a different image hash.

## Performance

The server computes the semantic hash once when an eligible request is admitted and passes only the resulting integer to the generation thread. APC lookup no longer re-hashes full prompt embeddings or masks.

Requests rejected by the configured context limit are rejected before media fingerprinting. Models with unsupported APC cache layouts are identified once during model initialization and skip semantic hashing entirely.

## Tests

The focused APC, generation, and server suites pass:

```text
python -m pytest mlx_vlm/tests/test_apc.py mlx_vlm/tests/test_generate.py mlx_vlm/tests/test_server.py -q
363 passed
```

The added tests cover:

- Stable APC hashes when text is appended after server preprocessing
- An actual stored APC prefix hit when the prompt is extended
- Stable APC hashes for prepared direct-stream inputs
- Distinct hashes for explicit custom embeddings
- Distinct hashes for explicit custom attention masks
- Distinct image hashes when the content behind the same reference changes
- Distinct image hashes for different tensor shapes and dtypes
- No media hashing for over-budget requests or unsupported APC model layouts
- Preservation of existing media and cache-isolation dependencies

Fixes #1712


bnsd55:
I reproduced this on an M5 Max 128GB using `mlx-community/Qwen3-Coder-30B-A3B-Instruct-8bit`. APC stored 64,384 tokens across 4,024 blocks, but the next growing Pi agent request matched zero tokens and re-prefilled the full ~64K prompt, taking roughly 200 seconds at 100% GPU utilization. This blocks practical multi-turn local-agent use. I support merging this fix and can validate the branch on this setup.

Lazarus-931:
hey @byoan, sry, didn't see this!

yeah, reproduced this and your fix works!

lgtm!

Lazarus-931 (approved):

Lazarus-931:
CI issues, can you do pre-commit first?
thx!

byoan:
@Lazarus-931 Should be good now, thanks

Lazarus-931 (approved):

Issue #1703, opened 2026-07-23
mlx-vlm not honoring gpt-oss's per-layer mixed quantization


Lazarus-931:
gpt-oss-20b-MXFP4-Q8 is a mixed-quant conversion: the MoE experts are MXFP4 (no biases) and everything else is 8-bit affine (which does carry scales + biases). mlx-vlm rebuilt those non-expert layers without affine QuantizedLinear slots (built them as MXFP4/plain), so the affine biases have nowhere to land causing "parameters not in model."

Follow up to fix downstream issue in Nativ: https://github.com/Blaizzy/nativ/issues/69

PR #1728, opened 2026-07-26
gpt_oss: honor per-layer mixed quantization (mxfp4 experts + 8-bit affine)

Lazarus-931:
`mlx-community/gpt-oss-20b-MXFP4-Q8` is a mixed per-layer quant (mxfp4 experts + 8-bit affine non-experts), but its config keys the per-layer overrides by the unwrapped path (`model.layers.N…`, `lm_head`) while `gpt_oss.Model.sanitize` moves everything under `language_model.` — so the overrides never match, the non-expert layers fall back to mxfp4 (no `biases` slot), and load aborts with "Received 122 parameters not in model". This adds `sanitize_quantization_config` to gpt_oss (as qwen3_5 and siblings already do), remapping the per-layer quant keys onto the `language_model.` paths. Verified end-to-end via `load_model` on a tiny checkpoint of the same shape (fails before, loads after) plus a new `TestGptOssMixedQuant`; full test suite green.

for #1703, closing nativ issue downstream(https://github.com/Blaizzy/nativ/issues/69)

### DIFF
--- a/mlx_vlm/models/gpt_oss/config.py
+++ b/mlx_vlm/models/gpt_oss/config.py
@@ -1,7 +1,34 @@
 from dataclasses import dataclass
-from typing import Any
+from typing import Any, Dict, Optional
 
 from ..base import BaseModelConfig
+
+
+def sanitize_quantization_config(quantization):
+    """Remap per-layer quantization keys onto the wrapped module paths.
+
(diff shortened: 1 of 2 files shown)


lucasnewman (approved):
LGTM

PR #1729, opened 2026-07-26
Allow image editing with FLUX.2-klein-4b

lucasnewman:
This was an oversight in the original implementations, but it's fully supported.


Blaizzy (approved):
LGTM, thanks! 🚀

PR #1730, opened 2026-07-26
Support quantized FLUX.2 models

lucasnewman:
Some of the existing quantized checkpoints for FLUX.2 in `mlx-community` have conversions with custom/unique metadata, so this adds support to the image model loader to handle those converted models.


Blaizzy (approved):
LGTM 🚀

Blaizzy (approved):

PR #1731, opened 2026-07-27
Honor explicit trust_remote_code in processor from_pretrained

### LINKED ISSUE #1724: --trust-remote-code is silently ignored by 13 processors (I think)
I was reviewing the code related to the `--trust-remote-code`  argument and might have come across an inconsistency among the handling of the flag across the supported models.

In these 9 files it appears like a provided argument is dropped and replaced with a hardcoded True:

- `mlx_vlm/models/gemma4/processing_gemma4.py` — lines 903, 911
- `mlx_vlm/models/gemma4_unified/processing_gemma4_unified.py` — lines 365, 373
- `mlx_vlm/models/molmo/processing_molmo.py` — lines 649, 657
- `mlx_vlm/models/molmo2/processing.py` — lines 680, 688
- `mlx_vlm/models/fastvlm/processing.py` — lines 441, 449
- `mlx_vlm/models/phi3_v/processing_phi3_v.py` — lines 593, 600
- `mlx_vlm/models/phi4mm/processing_phi4mm.py` — lines 637, 644
- `mlx_vlm/models/kimi_vl/processing_kimi_vl.py` — lines 500, 507
- `mlx_vlm/models/locateanything/processing_locateanything.py` — lines 267, 273

In these 4 files it appears like a provided argument is never read and a hardcoded True is set:

- `mlx_vlm/models/falcon_ocr/processing_falcon_ocr.py` — 156 and 162 (two call sites)
- `mlx_vlm/models/falcon_perception/processing_falcon_perception.py` — 135 and 141 (two call sites)
- `mlx_vlm/models/kimi_k25/processing_kimi_k25.py` — 250
- `mlx_vlm/models/molmo_point/processing_molmo_point.py` — 98

With that said, I am far outside my wheelhouse in this codebase (Transformers, Tokenizers, etc...), I came about this from a code flow perspective then asked Claude to dig into it.

If this is indeed an issue please let me know and ill be happy to address it in a PR.

Thank you!

camparker:
Fixes Blaizzy/mlx-vlm#1724

13 processors ignored an explicitly-passed `trust_remote_code` — nine popped the caller's value and hardcoded `True`, four never read it at all. Passing `trust_remote_code=False` did not prevent remote code execution.

```python
# before
kwargs.pop("trust_remote_code", None)
...
    trust_remote_code=True,

# after
trust_remote_code = kwargs.pop("trust_remote_code", True)
...
    trust_remote_code=trust_remote_code,
```

This is the pattern `glm4v`, `glm4v_moe`, `glm_ocr`, `paddleocr_vl`, and `hunyuan_vl` already use — the PR brings 13 processors in line with 5 that were already correct.

Adds `TestTrustRemoteCodePassthrough` covering the `False` case on `molmo_point`. Verified on my fork that it fails without the fix and passes with it.

`internvl_chat`, `lfm2_vl`, and `phi4_siglip` also discard the argument, but in ways that differ from this issue — left out of scope, happy to file separately.

**On the default:** it stays `True`, so callers that don't pass the flag see no change in behavior. I deliberately didn't switch it to `False` — safer, but a breaking change for every existing caller, and a separate discussion from<br>  fixing the flag being ignored. Glad to open an issue for it if you'd want that.

### DIFF
--- a/mlx_vlm/models/falcon_ocr/processing_falcon_ocr.py
+++ b/mlx_vlm/models/falcon_ocr/processing_falcon_ocr.py
@@ -147,19 +147,20 @@ class FalconOCRProcessor:
     @classmethod
     def from_pretrained(cls, pretrained_model_name_or_path, **kwargs):
         kwargs.pop("use_fast", None)
+        trust_remote_code = kwargs.pop("trust_remote_code", True)
 
         model_path = Path(pretrained_model_name_or_path)
         is_local = model_path.exists() and model_path.is_dir()
 
         if is_local:
             tokenizer = AutoTokenizer.from_pretrained(
-                str(model_path), trust_remote_code=True
+                str(model_path), trust_remote_code=trust_remote_code
             )
             config_file = model_path / "config.json"
             config = json.loads(config_file.read_text()) if config_file.exists() else {}
         else:
             tokenizer = AutoTokenizer.from_pretrained(
-                pretrained_model_name_or_path, trust_remote_code=True
+                pretrained_model_name_or_path, trust_remote_code=trust_remote_code
             )
             try:
                 from huggingface_hub import hf_hub_download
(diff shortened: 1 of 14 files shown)


lucasnewman (approved):
Thanks!

Blaizzy:
That's because we patch the processor so it doesn't have any effect :)

PR #1743, opened 2026-07-27
Fix double transpose of Nemotron Omni audio conv weights on MLX checkpoints

takuyaomi:
## Description

`sanitize_audio_weights` in `nemotron_h_nano_omni` unconditionally transposes `sound_encoder` conv weights. That is correct when converting from the PyTorch checkpoint, but `load_model` runs sanitize on every load — so loading an **already-converted MLX checkpoint** transposes the weights a second time and fails:

```
ValueError: Expected shape (256, 3, 3, 1) but received shape (256, 3, 1, 3)
for parameter sound_encoder.encoder.subsampling.layers.0.weight
```

This currently affects every published MLX conversion of this architecture I tried (e.g. `mlx-community/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-4bit`).

## Fix

Add idempotence guards (same spirit as `check_array_shape` in the vision towers):

- conv2d: transpose only when the square kernel dims are in NCHW position (`shape[2] == shape[3]`)
- conv1d: transpose only when the shape matches a PyTorch layout for the configured kernel size (depthwise ends with the kernel size; pointwise carries the channel count in dim 1). The kernel size is taken from `sound_config.conv_kernel_size`.

## Verification

- Before: `mlx_vlm generate` with `mlx-community/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-4bit` fails with the shape error above.
- After: loads and answers an invoice-reading probe correctly (「880,000円」, 97 tok/s generation on M4 Max).
- Converting from the original PyTorch checkpoint still transposes as before (guards match the PyTorch layouts).

The code was written in collaboration with Claude (AI assistant); verification was run on real hardware as described above.

### DIFF
--- a/mlx_vlm/models/nemotron_h_nano_omni/audio.py
+++ b/mlx_vlm/models/nemotron_h_nano_omni/audio.py
@@ -545,38 +545,25 @@ class SoundFeatureExtractor:
         )
 
 
-def _conv1d_needs_transpose(shape, conv_kernel_size):
-    # PyTorch conv1d weights are (out, in/groups, K); MLX expects (out, K, in/groups).
-    # Guard so already-converted MLX checkpoints are not transposed twice:
-    # torch depthwise ends with the kernel size, torch pointwise carries the
-    # (large) channel count in dim 1; MLX layouts match neither pattern.
-    _, d1, d2 = shape
-    if d2 == conv_kernel_size and d1 != conv_kernel_size:
-        return True
-    if d2 == 1 and d1 > conv_kernel_size:
-        return True
-    return False
-
-
-def sanitize_audio_weights(weights, conv_kernel_size=9):
+def sanitize_audio_weights(weights):
+    # The PyTorch checkpoint ships feature_extractor buffers that are dropped
+    # below, so their presence distinguishes the PyTorch layout from an
+    # already-converted MLX checkpoint. load_model re-runs sanitize on every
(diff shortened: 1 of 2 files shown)


takuyaomi:
Reworked the fix after the CI failure: the per-shape guards conflicted with the synthetic shapes in `test_sanitize_audio_and_projection_weights`.

New approach: use the `feature_extractor` buffers as a checkpoint-format fingerprint — they exist in the PyTorch checkpoint and are dropped during conversion, so their presence tells the PyTorch layout from an already-converted MLX checkpoint. No shape heuristics needed.

- `tests/test_nemotron_h_nano_omni.py`: 6/6 pass locally
- `mlx-community/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-4bit`: loads and answers the invoice probe correctly
- Fresh conversion from the PyTorch checkpoint: unchanged behavior (fingerprint present → transposes run)

lucasnewman (approved):

lucasnewman:
@takuyaomi Can you [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits and force push so we can clear the merge requirements?

takuyaomi:
Done — rebased onto latest main and force-pushed with signed commits. Both commits now show as Verified.

lucasnewman:
Thanks!

PR #1745, opened 2026-07-27
Add Mage-VL (microsoft/Mage-VL, 4B codec-native VLM)

xocialize:
Adds `mage_vl` — [microsoft/Mage-VL](https://huggingface.co/microsoft/Mage-VL) (Apache-2.0), a 4B image/video-understanding model: a from-scratch 24-layer "Mage-ViT" tower with a 4:6:6-split 3D RoPE and 2×2 patch merger, feeding a stock Qwen3-4B-Instruct-2507 decoder.

## What's included

- `mlx_vlm/models/mage_vl/` — config, vision tower, composition. **The decoder is reused, not re-translated**: `models/qwen3`'s `LanguageModel` already accepts `inputs_embeds` and uses plain 1-D rope, which is exactly what Mage-VL's text config ("use simple 1D position_ids") calls for. `language.py` is a re-export.
- A torch-free `MageVLProcessor` (the checkpoint's remote-code processor is torch-native and fails in `prepare_inputs`), installed via `install_auto_processor_patch` per the `qwen3_vl` pattern. Image path; upstream's video path depends on a manylinux-only native extension (`codec-video-prep`) or a torch frame sampler, so torch-free video is left as a follow-up.
- `prompt_utils` entry (`LIST_WITH_IMAGE_FIRST` — the chat template is Qwen-VL-family).
- Tests: a `test_models.py` entry covering the image case and the t>1 windowed-attention case, plus `test_mage_vl_positions.py` pinning the 2×2 block-order position derivation (validated row-for-row against the reference processor's `patch_positions` output for a (1, 128, 64) grid).
- A small `convert.py` fix: processors without `save_pretrained` (Mage-VL's deliberately doesn't inherit `ProcessorMixin`) no longer crash the sidecar-copy step after the weights are written; processor files are copied verbatim instead.

## Faithfulness notes

Three upstream behaviours are reproduced deliberately and documented in `vision.py` (each looks like a bug; each is in the trained weights): interleaved `rotate_half`; a `concat([freqs, freqs])` rope table so lanes 2i/2i+1 carry different frequencies under that rotation; and the `hidden_states[-1]` tap despite an upstream comment saying second-to-last.

## Verification

Against the checkpoint's own remote code (CPU stream, fp32): vision tower rel ≤ 4.0e-5 / cos 1.0 across four grids including t=0–63; weight key set 696 ↔ 696 (0 missing / 0 unused); the vision-feature scatter hits exactly the 2048 image-token positions; **end-to-end greedy decoding is 48/48 tokens identical** to the reference on the model's own `examples/dog.jpg`.

Through the standard API: `load` → `apply_chat_template` → `generate` verified on [mlx-community/Mage-VL-8bit](https://huggingface.co/mlx-community/Mage-VL-8bit) (8-bit conversion produced by this branch's `mlx_vlm.convert`; its greedy tokens are also 48/48 identical to the PyTorch reference).

`black` run on all touched files; `python -m pytest mlx_vlm/tests/test_models.py -k mage_vl mlx_vlm/tests/test_mage_vl_positions.py` → 9 passed.

### DIFF
--- a/mlx_vlm/models/mage_vl/language.py
+++ b/mlx_vlm/models/mage_vl/language.py
@@ -1,0 +1,19 @@
+"""Mage-VL's decoder is stock Qwen3-4B-Instruct-2507 — reused verbatim, not re-translated.
+
+Path A, settled by reading both candidates rather than picking by name:
+
+* `models/qwen3` builds rope through `initialize_rope(base=rope_theta, scaling_config=rope_scaling)`
+  and advances positions by `cache.offset` — plain 1-D rope. That is exactly Mage-VL, whose text
+  config is `rope_type: "default"` and whose modeling code says, in as many words, "Use simple 1D
+  position_ids" (`modeling_mage_vl.py:1269`).
+* `models/qwen3_vl` is the closer-sounding name and the WRONG model here: its language half is
+  `MRoPERotaryEmbedding`-based and tiles `position_ids` to `(3, 1, 1)`. Mage-VL reserves 3D mRoPE
+  as future work and does not use it.
+
+`LanguageModel.__call__` already accepts `inputs_embeds`, which is the whole reason no decoder
+work is needed — the vision tower's output goes straight in.
+"""
+
+from ..qwen3.language import LanguageModel, Qwen3Model, TransformerBlock
+
+__all__ = ["LanguageModel", "Qwen3Model", "TransformerBlock"]
--- a/mlx_vlm/models/mage_vl/language.py
+++ b/mlx_vlm/models/mage_vl/language.py
@@ -1,17 +1,8 @@
-"""Mage-VL's decoder is stock Qwen3-4B-Instruct-2507 — reused verbatim, not re-translated.
+"""Mage-VL's decoder is stock Qwen3-4B-Instruct-2507, reused as-is.
 
-Path A, settled by reading both candidates rather than picking by name:
-
-* `models/qwen3` builds rope through `initialize_rope(base=rope_theta, scaling_config=rope_scaling)`
-  and advances positions by `cache.offset` — plain 1-D rope. That is exactly Mage-VL, whose text
-  config is `rope_type: "default"` and whose modeling code says, in as many words, "Use simple 1D
-  position_ids" (`modeling_mage_vl.py:1269`).
-* `models/qwen3_vl` is the closer-sounding name and the WRONG model here: its language half is
-  `MRoPERotaryEmbedding`-based and tiles `position_ids` to `(3, 1, 1)`. Mage-VL reserves 3D mRoPE
-  as future work and does not use it.
-
-`LanguageModel.__call__` already accepts `inputs_embeds`, which is the whole reason no decoder
-work is needed — the vision tower's output goes straight in.
+The text config is `rope_type: "default"` — plain 1-D rope, not mRoPE — so `models/qwen3`
+is the correct base rather than `models/qwen3_vl`. `LanguageModel.__call__` already accepts
+`inputs_embeds`, which the vision tower's output is passed through.
 """
 
 from ..qwen3.language import LanguageModel, Qwen3Model, TransformerBlock
--- a/mlx_vlm/models/mage_vl/mage_vl.py
+++ b/mlx_vlm/models/mage_vl/mage_vl.py
@@ -1,0 +1,147 @@
+"""Mage-VL — codec-native streaming multimodal model (microsoft/Mage-VL, Apache-2.0).
+
+Composition only: a Mage-ViT tower (`vision.py`, net-new) feeding a stock Qwen3-4B decoder
+(`language.py`, reused verbatim). See those modules for the reproduced-upstream quirks.
+"""
+
+from typing import List, Optional, Tuple
+
+import mlx.core as mx
+import mlx.nn as nn
+import numpy as np
+
+from .config import ModelConfig, TextConfig, VisionConfig
+from .language import LanguageModel
+from .vision import VisionModel
+
+
+def masked_scatter(
+    final_embedding: mx.array,
+    image_mask_expanded: mx.array,
+    scaled_image_features: mx.array,
+) -> mx.array:
+    shape = final_embedding.shape
+    flat_features = mx.flatten(scaled_image_features)
+    flat_embedding = mx.flatten(final_embedding)
+    flat_mask = mx.flatten(image_mask_expanded)
+
+    positions = mx.array(np.where(np.array(flat_mask))[0], mx.uint32)
+    flat_embedding[positions] = flat_features
+    return mx.reshape(flat_embedding, shape)
+
+
+class Model(nn.Module):
+    def __init__(self, config: ModelConfig):
+        super().__init__()
+        self.config = config
+        self.model_type = config.model_type
+        self.vision_tower = VisionModel(config.vision_config)
+        self.language_model = LanguageModel(config.text_config)
+
+    def get_input_embeddings(
+        self,
+        input_ids: Optional[mx.array] = None,
+        pixel_values: Optional[mx.array] = None,
+        **kwargs,
+    ) -> mx.array:
+        if pixel_values is None:
+            pixel_values = kwargs.get("pixel_values_videos", None)
+        if pixel_values is None:
+            return self.language_model.model.embed_tokens(input_ids)
+
+        patch_positions = kwargs.get("patch_positions", None)
+        grid_thw = kwargs.get("image_grid_thw", None)
+        if grid_thw is None:
+            grid_thw = kwargs.get("video_grid_thw", None)
+        grid_thw = _as_grid_list(grid_thw)
+
+        if patch_positions is None:
+            if not grid_thw:
+                raise ValueError("mage_vl needs patch_positions or a grid_thw to derive them")
+            patch_positions = _positions_from_grid(grid_thw, self.config.vision_config)
+
+        inputs_embeds = self.language_model.model.embed_tokens(input_ids)
+        hidden_states = self.vision_tower(
+            pixel_values.astype(inputs_embeds.dtype), patch_positions, grid_thw
+        )
+
+        image_token_id = self.config.image_token_id
+        video_token_id = self.config.video_token_id
+        is_image = (input_ids == image_token_id) | (input_ids == video_token_id)
+        mask_expanded = mx.expand_dims(is_image, -1)
+        mask_expanded = mx.repeat(mask_expanded, inputs_embeds.shape[-1], axis=-1)
+
+        return masked_scatter(
+            inputs_embeds, mask_expanded, hidden_states.astype(inputs_embeds.dtype)
+        )
+
+    def __call__(
+        self,
+        input_ids: mx.array,
+        pixel_values: Optional[mx.array] = None,
+        mask: Optional[mx.array] = None,
+        cache=None,
+        **kwargs,
+    ):
+        inputs_embeds = self.get_input_embeddings(input_ids, pixel_values, **kwargs)
+        return self.language_model(
+            inputs=input_ids, cache=cache, inputs_embeds=inputs_embeds, mask=mask
+        )
+
+    @property
+    def layers(self):
+        return self.language_model.model.layers
+
+    def sanitize(self, weights):
+        """Remap the hub's key namespace onto the mlx-vlm one.
+
+        hub                             -> here
+        model.language_model.*          -> language_model.model.*
+        lm_head.*                       -> language_model.lm_head.*
+        model.visual.*                  -> vision_tower.*
+        """
+        out = {}
+        for k, v in weights.items():
+            if k.startswith("model.language_model."):
+                k = "language_model.model." + k[len("model.language_model.") :]
+            elif k.startswith("model.visual."):
+                k = "vision_tower." + k[len("model.visual.") :]
+            elif k.startswith("lm_head."):
+                k = "language_model.lm_head." + k[len("lm_head.") :]
+            out[k] = v
+
+        vision = {k[len("vision_tower.") :]: v for k, v in out.items() if k.startswith("vision_tower.")}
+        vision = self.vision_tower.sanitize(vision)
+        for k, v in vision.items():
+            out["vision_tower." + k] = v
+        return out
+
+
+def _as_grid_list(grid_thw) -> List[Tuple[int, int, int]]:
+    if grid_thw is None:
+        return []
+    if isinstance(grid_thw, mx.array):
+        grid_thw = np.array(grid_thw)
+    return [tuple(int(x) for x in row) for row in np.atleast_2d(np.array(grid_thw))]
+
+
+def _positions_from_grid(
+    grid_thw: List[Tuple[int, int, int]], config: VisionConfig
+) -> mx.array:
+    """Derive (L, 3) [t, h, w] patch positions in the processor's 2x2 block order.
+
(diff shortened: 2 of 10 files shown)


ismaelvega:
### `MageVLProcessor.from_pretrained` silently routes to torch remote code on torchvision-free installs (Apple Silicon)
---
Hi @xocialize 👋🏻  ran the `mage-vl` branch end-to-end on a clean Apple-Silicon MLX setup and hit a failure that the torch-free port was specifically written to prevent. Filing here since this is the PR that introduces the port.

**Repro**

```bash
python -m venv .venv && source .venv/bin/activate
pip install git+https://github.com/xocialize/mlx-vlm.git@mage-vlm
python -c "
m, p = load('mlx-community/Mage-VL-8bit')
out = generate(m, p, apply_chat_template(p, m.config, 'Describe this image.', num_images=1),
               image=['http://images.cocodataset.org/val2017/000000039769.jpg'], max_tokens=64)
print(out.text)
"
```

**Symptom**

```
ValueError: Failed to process inputs with error:
ones_like(): argument 'input' (position 1) must be Tensor, not mlx.core.array
```

The port's own docstring in `mlx_vlm/models/mage_vl/processing_mage_vl.py` quotes this exact error as the failure it exists to prevent — but it still triggers on a standard MLX setup (no torch/torchvision installed, which is the normal Apple-Silicon case).

**Root cause**

`MageVLProcessor.from_pretrained` at `mlx_vlm/models/mage_vl/processing_mage_vl.py:61` does:

```python
image_processor = AutoImageProcessor.from_pretrained(
    pretrained_model_name_or_path, use_fast=False, trust_remote_code=False
)
```

`AutoImageProcessor.from_pretrained(use_fast=False)` resolves to upstream's `Qwen2VLImageProcessor`, which is torchvision-backed, so on a torchvision-free environment it raises:

```
ImportError: AutoImageProcessor requires the Torchvision library but it was not found in your environment.
```

That `ImportError` is then swallowed by `install_auto_processor_patch` in `mlx_vlm/models/base.py:582-584`:

```python
except Exception:
    # On any failure, fall back to previous behavior
    pass
```

So dispatch falls through to the upstream remote-code `MageVLProcessor` (configured in the checkpoint's `config.json` `auto_map.AutoProcessor`), which is torch-native — and downstream `prepare_inputs` dies with the `ones_like()` error.

The silent `except Exception: pass` in `install_auto_processor_patch` is what makes this manifest as a cryptic downstream crash instead of "torchvision missing". Worth flagging separately — even if the mage_vl port is fixed, the same trap will bite any future port that has a broken `from_pretrained`.

**Suggested fix**

Build a numpy/PIL image processor directly in `MageVLProcessor.from_pretrained`, bypassing `AutoImageProcessor`. The sibling `qwen3_vl` port already ships a torch-free `Qwen3VLImageProcessor` (`mlx_vlm/models/qwen3_vl/processing_qwen3_vl.py:142`) — same `Qwen2VLImageProcessor` schema (patch 16, merge 2, temporal_patch_size 1, CLIP statistics), identical output shapes, no torch/torchvision. The mage_vl port's docstring already says it "follows the `qwen3_vl` pattern", so reusing the image processor keeps that contract.

Sketch:

```python
@classmethod
def from_pretrained(cls, pretrained_model_name_or_path, **kwargs):
    from transformers import AutoTokenizer
    from ..base import load_chat_template
    from ..qwen3_vl.processing_qwen3_vl import Qwen3VLImageProcessor

    kwargs.pop("use_fast", None)
    kwargs.pop("trust_remote_code", None)
    tokenizer = AutoTokenizer.from_pretrained(
        pretrained_model_name_or_path, trust_remote_code=False, **kwargs
    )
    load_chat_template(tokenizer, pretrained_model_name_or_path)

    cfg = _load_preprocessor_config(pretrained_model_name_or_path)  # preprocessor_config.json
    ip_kwargs = {
        k: cfg[k] for k in (
            "patch_size", "temporal_patch_size", "merge_size",
            "min_pixels", "max_pixels", "do_rescale", "rescale_factor",
            "do_normalize", "image_mean", "image_std", "do_convert_rgb",
        ) if k in cfg
    }
    image_processor = Qwen3VLImageProcessor(**ip_kwargs)
    return cls(image_processor=image_processor, tokenizer=tokenizer)
```

I've validated this end-to-end: `AutoProcessor.from_pretrained("mlx-community/Mage-VL-8bit")` returns the torch-free MageVLProcessor, single-image QA on COCO `000000039769.jpg` produces "Two cats are sleeping on a red couch with remote controls next to them." (48/48 tokens match the bf16 reference described in the model card — I didn't re-run the full gate, just sanity-checked the obvious case).

Happy to send a PR against the `mage-vl` branch if helpful. pls let me know.

Environment: mlx-vlm 0.6.8 from `mage-vl` branch @ `xocialize/mlx-vlm`, Python 3.14.5, Apple Silicon (M-series), no torch / torchvision installed.

xocialize:
@ismaelvega Thank you for the precise diagnosis — confirmed and fixed in bf9ca87.

You had the mechanism exactly right: `AutoImageProcessor.from_pretrained` needs torchvision, the ImportError was swallowed by the AutoProcessor patch dispatcher's fallback, and dispatch fell through to the checkpoint's torch remote-code processor. My dev venv had torch installed *for the PyTorch parity oracle*, which is precisely why the torch-free case escaped — the verification environment differed from the user environment in the one dimension that mattered.

The fix takes your suggested direction: `MageVLProcessor.from_pretrained` now constructs the in-repo numpy/PIL `Qwen3VLImageProcessor` directly from `preprocessor_config.json` — no `Auto*` resolution anywhere. Gated two ways:

- old-vs-new `pixel_values` on the reference image agree to max|Δ| 4.8e-7 (float32 rounding), so nothing shifts for environments that *do* have torchvision;
- your repro now passes in a clean torch-free venv, loading `mlx_vlm.models.mage_vl.processing_mage_vl.MageVLProcessor` and producing the reference answer. Re-verified after rebasing onto the `main` merge.

Your standalone workaround is no longer needed on the current branch. One observation for a possible follow-up outside this PR: the `except Exception: pass` in `install_auto_processor_patch`'s dispatcher is what turned an ImportError into a silent wrong-processor fallback — a `logger.warning` there would have surfaced this immediately.

AlanConstantino:
I'm up at 4AM wondering why I can't get this Mage model to work on oMLX.... turns out I need the changes on this PR 😂

@xocialize You da goat fr 🐐🙏

Lazarus-931:
hey @xocialize, lint issues seems to be holding up the CI, try running: `pre-commit run --all-files` and committing!

xocialize:
CI fix pushed in `8b2555d`.

@Lazarus-931 thanks — you were right that it was the pre-commit stage, though the culprit turned out to be `autoflake` rather than `black`/`isort` (both of those passed). The hook runs with `--remove-all-unused-imports --in-place` and fails when it rewrites anything, and it was rewriting two files in this PR:

- **`mage_vl/config.py`** — `field` imported from `dataclasses`, never used. The only other occurrence of "field" in the file is the word inside a docstring ("Field-for-field identical to…"), which is why it survived review.
- **`mage_vl/mage_vl.py`** — `TextConfig` imported, never referenced.

`TextConfig` stays in the package's public surface — `__init__.py` re-exports it from `.config`, not from `mage_vl.py` — so `mlx_vlm.models.mage_vl.__all__` is unchanged. Verified all six names still resolve after the removal.

Checked locally before pushing: `autoflake --check` clean across `mlx_vlm/` (no other file, mine or upstream, is dirty), `black` clean, and `mlx_vlm/tests/test_mage_vl_positions.py` 8/8 passing.

One note on the CI state: the new run is sitting at `action_required` rather than running. Workflow runs on fork PRs need approval each time, so the checks won't re-execute until someone with write access approves that run — the red ✗ currently shown is still the old result for `bf9ca87`, not a new failure.

@ismaelvega — your torchvision-free repro is fixed in `bf9ca87` (the processor is now constructed directly instead of going through `AutoImageProcessor.from_pretrained`, so there's no torchvision import to swallow and no fallback to the checkpoint's torch remote code). Thanks again for the precise diagnosis — the reason it escaped me is that my dev venv had torch installed for the PyTorch parity oracle, so the verification environment differed from the user environment in exactly the dimension that mattered.

mlx_vlm/models/mage_vl/language.py
```diff
@@ -0,0 +1,19 @@
+"""Mage-VL's decoder is stock Qwen3-4B-Instruct-2507 — reused verbatim, not re-translated.
+
+Path A, settled by reading both candidates rather than picking by name:
```
lucasnewman: These look like implementation notes from porting, not current information about the code. Can you remove any irrelevant comments?

mlx_vlm/models/mage_vl/mage_vl.py
```diff
@@ -0,0 +1,162 @@
...
+    final_embedding: mx.array,
+    image_mask_expanded: mx.array,
+    scaled_image_features: mx.array,
+) -> mx.array:
+    shape = final_embedding.shape
+    flat_features = mx.flatten(scaled_image_features)
+    flat_embedding = mx.flatten(final_embedding)
+    flat_mask = mx.flatten(image_mask_expanded)
+
+    positions = mx.array(np.where(np.array(flat_mask))[0], mx.uint32)
```
lucasnewman: Why not just?

```py
bool_mask = image_mask_expanded.astype(mx.bool_)
final_embedding[bool_mask] = scaled_image_features
return final_embedding
```

Or at least use `mx.where`, bouncing through numpy definitely isn't necessary.

mlx_vlm/models/mage_vl/processing_mage_vl.py
```diff
@@ -0,0 +1,174 @@
...
+        if isinstance(text, str):
+            text = [text]
+        text = list(text or [])
+
+        image_inputs = {}
+        if images is not None:
+            image_inputs = dict(
+                self.image_processor(images=images, return_tensors="np")
+            )
+            grids = np.asarray(image_inputs["image_grid_thw"])
```
lucasnewman: Why not use mx.array directly here?

mlx_vlm/models/mage_vl/vision.py
```diff
@@ -0,0 +1,402 @@
...
+        # NOTE the exponent: arange(size)/size, NOT the conventional arange(0, d, 2)/d.
+        self._inv_freq_t = 1.0 / (
+            base ** (mx.arange(self.t_size, dtype=mx.float32) / self.t_size)
+        )
+        self._inv_freq_h = 1.0 / (
+            base ** (mx.arange(self.h_size, dtype=mx.float32) / self.h_size)
+        )
+        self._inv_freq_w = 1.0 / (
+            base ** (mx.arange(self.w_size, dtype=mx.float32) / self.w_size)
+        )
```
lucasnewman, line 90: These cache arrays need to be evaluated at init time so they don't crash when moving streams.

mlx_vlm/prompt_utils.py
```diff
@@ -46,6 +46,9 @@ class MessageFormat(Enum):
     "zaya1_vl": MessageFormat.LIST_WITH_IMAGE_FIRST,
     "qwen3_vl": MessageFormat.LIST_WITH_IMAGE_FIRST,
     "qwen3_vl_moe": MessageFormat.LIST_WITH_IMAGE_FIRST,
+    # Mage-VL's chat template emits <|vision_start|><|image_pad|><|vision_end|> for structured
+    # {"type": "image"} content, exactly like the Qwen-VL family it reuses.
```
lucasnewman: We don't need the comment here.

lucasnewman (changes requested):
Thanks for the submission! I added some notes inline, but please clean up the comments and fix the numpy usage / array eval at init and then we can merge.

xocialize:
Thanks for the review @lucasnewman — all five addressed in `445d59c`, rebased onto the `main` merge.

**`masked_scatter` — numpy removed.** Your literal suggestion needs the value flattened for these shapes (`final_embedding[bool_mask] = scaled_image_features` raises *"Value array shape must be broadcastable with the last 0 dimensions"*), so it's now:

```py
flat_embedding = mx.flatten(final_embedding)
flat_embedding[mx.flatten(image_mask_expanded).astype(mx.bool_)] = mx.flatten(scaled_image_features)
return mx.reshape(flat_embedding, shape)
```

You were right that the numpy trip was doing nothing but a device→host→device round trip per image. Checked equivalence rather than assuming it: bit-identical to the old implementation across float32 and bfloat16 at `(1,64,32)`, `(1,2048,2560)` and `(2,128,64)`, including the zero-image case.

**Rope caches evaluated at init.** Agreed this was a real bug rather than style — thank you, it would only have surfaced on a different stream than the one the model was constructed on, which is exactly what local single-stream runs never exercise. `mx.eval` on the three arrays at the end of `__init__`; I checked and this is the only init-time cache in the port.

**`processing_mage_vl.py` numpy.** The grid arithmetic is plain Python now (`mx.array(...).tolist()`, then integer math), and the `np` import is gone. `mage_vl` is numpy-free end to end — no module in it imports numpy. I kept the image processor call as `return_tensors="np"` since `Qwen3VLImageProcessor` is itself the torch-free numpy port and `to_mlx` converts at the boundary, matching the `qwen3_vl` sibling — happy to change that too if you'd rather it differ.

While in there I also dropped numpy from `_as_grid_list` and `_positions_from_grid`. The coercion contract is unchanged — `mx.array` already accepts mx arrays, numpy arrays and nested lists, which is what the parametrized tests pin.

**Comments.** Removed the porting-narrative docstring from `language.py`, keeping only what a reader of the code needs (why `qwen3` and not `qwen3_vl`), and dropped the `prompt_utils.py` comment. Fair criticism — that was notes-to-self, not documentation. Net −13 lines across the change.

I left one comment in `vision.py`: the `# NOTE the exponent: arange(size)/size, NOT the conventional arange(0, d, 2)/d`. It marks a genuine deviation from the usual rope formulation that upstream Mage-VL requires, and without it the line reads like a bug someone would "fix". Say the word if you'd rather it go too.

Gates: 8/8 positions tests, autoflake and black clean repo-wide, and greedy generation on the reference image is byte-identical before and after the change (48 tokens, temperature 0) — re-verified after rebasing onto the merge.

Thanks @Lazarus-931 for merging `main` in.

lucasnewman (approved):
Thanks!

lucasnewman:
@xocialize Can you [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits and force push so we can clear the merge requirements?

xocialize:
Done @lucasnewman — commits signed and force-pushed. All 7 now show as verified, and the merge requirement has cleared (the PR moved from `BLOCKED` to `MERGEABLE`).

One thing worth flagging since it changes the history you reviewed: signing meant rewriting the branch, and I could not re-sign the two `Merge branch 'main' into mage-vl` commits (from @Blaizzy and @Lazarus-931) without re-attributing them to my key. So I rebased the 7 commits onto current `main` instead and dropped those merges — their only purpose was to pull `main` in, which the rebase now does directly, and the result is a clean linear history. Nothing else changed: the `mage_vl` tree is byte-identical to `445d59c`, the commit CI passed on before the rewrite.

Re-verified after the rebase rather than just before it, since `main` had moved underneath — it brought new video/audio message formatting into `prompt_utils.py`, which is intact and untouched by my change:

- 8/8 positions tests
- autoflake and black clean repo-wide
- greedy generation on the reference image byte-identical (48 tokens, temperature 0)

The new run is at `action_required`, as fork pushes are each time — it'll need an approval whenever you next get to it. For reference the previous run on identical `mage_vl` content passed.

lucasnewman:
Thanks!

PR #1746, opened 2026-07-28
Add Kimi K3

kernelpool:


Blaizzy (approved):
LGTM, thanks a lot @kernelpool! 

I just have one comment, we can address it now or in a future PR.

mlx_vlm/prompt_utils.py
```diff
@@ -763,6 +764,25 @@ def _template_references_kw(template_processor: Any, name: str) -> bool:
             ):
                 template_processor = processor
 
+        if template_processor is None and processor is not None:
```
Blaizzy: Can we move this to the custom
processor? 

> Sorry, just noticed the comment wasn’t registered with the review.

kernelpool: yep, will do

Lazarus-931:
wow, don't know who'll be able to run this, but awesome!

Blaizzy (approved):

mlx_vlm/prompt_utils.py
Blaizzy: Thank you very much @kernelpool!

Blaizzy:
Merged ✅

PR #1749, opened 2026-07-29
Add OpenAI-compatible /v1/embeddings endpoint backed by mlx-embeddings

Lazarus-931:
For embedding support, mlx-embeddings doesn't have a http server, I think layering it on mlx-vlm to handle it for nativ seems best, this pr allows mlx-vlm to handle embedding requests from nativ.

### DIFF
--- a/mlx_vlm/server/app.py
+++ b/mlx_vlm/server/app.py
@@ -26,6 +26,7 @@ from ..vision_cache import VisionFeatureCache
 from . import request_normalization as _request_normalization
 from .anthropic import register_routes as register_anthropic_routes
 from .audio import register_routes as register_audio_routes
+from .embeddings import register_routes as register_embeddings_routes
 from .generation import (
     GenerationArguments,
     PromptTooLongError,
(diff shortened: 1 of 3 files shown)


lucasnewman:
@Lazarus-931 Hmm, relying on lazy imports to resolve the circular dependency seems very fragile for the build.

Could we pull in the embeddings models instead? It's not a huge codebase and the last commit was months ago, so it doesn't seem like a big maintenance burden.

Lazarus-931:
@lucasnewman  I was thinking we can just remove the mlx-vlm deps from mlx-embeddings by removing the vlm options from there, but yes, this can also work, I think  porting those models into here would be much simper looking forward

lucasnewman:
> @lucasnewman I was thinking we can just remove the mlx-vlm deps from mlx-embeddings by removing the vlm options from there, but yes, this can also work, I think porting those models into here would be much simper looking forward

I think it will actually be pretty small because most of the embeddings models are just LM backbones with a class token / projection / pool or whatever else at the end, but lmk if you find otherwise.

Lazarus-931:
Superseding with a PR that ports embedding models natively into mlx-vlm.

PR #1750, opened 2026-07-29
Skip full-sequence LM head during Gemma 4 prefill

### LINKED ISSUE #1732: Prefill materialises logits for every prompt position (1 MiB/token on Gemma 4) — image turns hit a hard allocation wall at ~19k tokens
## Summary

On the non-chunked prefill path, the model materialises logits for **every** prompt position — `[1, N, vocab]` in float32 — and the caller then keeps only the last row. With Gemma 4's vocabulary of 262144 that is **1 MiB of logits per prompt token**, so prompt length translates directly into a single huge allocation.

This is not an edge case for Gemma 4: its `chunked_prefill_policy` disables chunked prefill whenever visual tokens are present, so **every turn with an image takes this path**. On a 32 GB M5 the result is a hard failure once the prompt passes ~19k tokens:

```
RuntimeError: [metal::malloc] Attempting to allocate 25353916928 bytes which is
greater than the maximum allowed buffer size of 20100448256 bytes.
```

25353916928 / 262144 / 4 = 24179 — exactly the number of prompt positions.

## Environment

| | |
|---|---|
| mlx-vlm | 0.6.7 |
| mlx | 0.32.0 |
| Python | 3.14.2 |
| OS | macOS 26.5.2, Apple M5, 32 GB (Metal max buffer 18.72 GiB, recommended working set 24.96 GiB) |
| Model | `google/gemma-4-26B-A4B-it`, converted with `mlx_vlm.convert -q --q-bits 4` |

## Where it happens

`mlx_vlm/models/gemma4/language.py` — the LM head runs over the whole sequence:

```python
out = self.model(inputs, inputs_embeds=inputs_embeds, mask=mask, cache=cache, ...)
out = self.logits_from_hidden(out)        # <- [1, N, 262144] f32 for the FULL prompt
return LanguageModelOutput(logits=out, ...)
```

`mlx_vlm/generate/ar.py::_step` — and the caller immediately throws all of it away but the last row:

```python
outputs = model.language_model(y, inputs_embeds=inputs_embeds, cache=prompt_cache, **step_kwargs)
last_outputs = outputs
logits = outputs.logits[:, -1, :]         # <- only this row is ever used
```

The chunked path avoids the worst of it by construction (each chunk is at most `prefill_step_size`), but chunking is unavailable here: `models/gemma4/language.py::chunked_prefill_policy` returns `False` as soon as the prompt contains visual tokens (bidirectional attention over vision spans), so image turns always run the full prompt in one pass.

## Reproduction

Pure upstream path, one image plus a growing text prompt, `max_tokens=1`:

```python
from mlx_vlm import load, stream_generate
from mlx_lm.sample_utils import make_sampler
from PIL import Image
import mlx.core as mx

model, proc = load(MODEL_4BIT)
cfg = load_config(MODEL_4BIT)
tok, greedy = proc.tokenizer, make_sampler(0.0)
img = Image.open(SOME_IMAGE)

for reps in (200, 800, 3200):
    msgs = [{"role": "user", "content": "Note about the context subsystem. " * reps}]
    prompt = apply_chat_template(proc, cfg, msgs, num_images=1, enable_thinking=False)
    n = len(tok.encode(prompt, add_special_tokens=False))
    mx.reset_peak_memory()
    base = mx.get_active_memory() / 1024**3
    for _ in stream_generate(model=model, processor=proc, prompt=prompt, image=[img],
                             max_tokens=1, sampler=greedy, prefill_step_size=2048):
        pass
    print(n, mx.get_peak_memory() / 1024**3 - base)
```

Measured on the machine above:

| prompt tokens | peak | transient over base |
|---|---|---|
| 2 823 | 17.16 GB | +2.87 GB |
| 11 223 | 23.24 GB | +8.95 GB |
| 44 823 | — | `[metal::malloc] Attempting to allocate 65059459200 bytes` (60.6 GiB) |

The slope between the first two points is ~0.74 MiB per prompt token, consistent with a single `[1, N, 262144]` f32 tensor dominating the transient.

For reference, the same prompt lengths **without** an image stay flat (chunked prefill active): 40k tokens peaks at 18.91 GB, transient +4.62 GB.

Side observation while collecting the table: at 22 423 tokens the failure surfaced as `IndexError: list index out of range` from inside `stream_generate` rather than as an allocation error. It may be worth checking that path separately — an allocation failure that turns into an IndexError is hard to diagnose downstream.

## Suggested fix

Only the last position's logits are ever consumed during prefill, so the LM head does not need to run over the full sequence. Two options:

1. Thread a `logits_to_keep` (transformers naming) or `num_logits` kwarg from `ar.py::_step` down to the language model, and slice the hidden states before `logits_from_hidden`:

```python
if logits_to_keep:
    out = out[:, -logits_to_keep:, :]
out = self.logits_from_hidden(out)
```

Models that ignore the kwarg keep today's behaviour, so the change is opt-in per model.

2. Alternatively, expose `logits_from_hidden` on the output and let `_step` compute logits lazily from `hidden_states[:, -1:]`. That keeps the fix entirely inside `ar.py` for models that already return hidden states.

Either way the per-token cost of the prefill drops from 1 MiB to zero, and image turns stop being bounded by `vocab_size × prompt_length` instead of by actual memory.

## Impact

With this fixed, the practical context limit for image turns on 32 GB hardware moves from ~19k tokens to whatever the KV cache and activations allow — for this model that is several times larger. Until then, any application that puts retrieved context, a long user message and an image into the same turn hits a hard wall that looks like an out-of-memory error but is really a single oversized allocation.

Lazarus-931:
Gemma 4's LanguageModel projected every prompt position through the LM head before [ar.py](<http://ar.py>) sliced the final row, materialising \[1, N, vocab\] logits (\~1 MiB/token at the 262144-token vocab) during prefill.

Add an opt-in logits_to_keep path: LanguageModels expose supports_logits_to_keep, and generation asks for a single row so prefill (and each discarded chunked-prefill step) slices the hidden states before the LM head. Covers gemma4 (incl. gemma4_unified) and gemma4_text. The retained last-row logits are identical, so generation output is unchanged.

Fixes Blaizzy/mlx-vlm#1732.

All tests green!

### DIFF
--- a/mlx_vlm/generate/ar.py
+++ b/mlx_vlm/generate/ar.py
@@ -328,6 +328,8 @@ def generate_step(
         step_kwargs = kwargs
         if speculative_prefill_capture_kwargs:
             step_kwargs = {**kwargs, **speculative_prefill_capture_kwargs}
+        if getattr(model.language_model, "supports_logits_to_keep", False):
+            step_kwargs = {**step_kwargs, "logits_to_keep": 1}
 
         with mx.stream(generation_stream):
             if "decoder_input_ids" in step_kwargs:
@@ -433,12 +435,15 @@ def generate_step(
                         and processed_tokens + n_to_process > checkpoint_len
                     ):
                         n_to_process = checkpoint_len - processed_tokens
+                    chunk_kwargs = kwargs
+                    if getattr(model.language_model, "supports_logits_to_keep", False):
+                        chunk_kwargs = {**kwargs, "logits_to_keep": 1}
                     model.language_model(
                         inputs=input_ids[:, :n_to_process],
                         inputs_embeds=inputs_embeds[:, :n_to_process],
                         cache=prompt_cache,
                         n_to_process=n_to_process,
-                        **kwargs,
+                        **chunk_kwargs,
                     )
                     quantize_cache_fn(prompt_cache)
                     mx.eval([c.state for c in prompt_cache])
--- a/mlx_vlm/models/gemma4/language.py
+++ b/mlx_vlm/models/gemma4/language.py
@@ -650,6 +650,8 @@ class Gemma4TextModel(nn.Module):
 
 
 class LanguageModel(nn.Module):
+    supports_logits_to_keep = True
+
     def __init__(self, config: TextConfig):
         super().__init__()
         self.config = config
@@ -726,6 +728,7 @@ class LanguageModel(nn.Module):
         # Allow callers to pass pre-allocated sinks directly.
         hidden_sink = kwargs.pop("hidden_sink", hidden_sink)
         shared_kv_sink = kwargs.pop("shared_kv_sink", shared_kv_sink)
+        logits_to_keep = kwargs.pop("logits_to_keep", None)
 
         out = self.model(
             inputs,
@@ -738,6 +741,8 @@ class LanguageModel(nn.Module):
             shared_kv_sink=shared_kv_sink,
             **kwargs,
         )
+        if logits_to_keep:
+            out = out[:, -int(logits_to_keep) :, :]
         out = self.logits_from_hidden(out)
         return LanguageModelOutput(
             logits=out,
--- a/mlx_vlm/models/gemma4_text/language.py
+++ b/mlx_vlm/models/gemma4_text/language.py
@@ -496,6 +496,8 @@ class Gemma4TextModel(nn.Module):
 
 
 class LanguageModel(nn.Module):
+    supports_logits_to_keep = True
+
     def __init__(self, args: ModelConfig):
         super().__init__()
         self.args = args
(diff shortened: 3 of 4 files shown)


lucasnewman (approved):
Very minor nit that the naming seems a little confusing (it's sequence length to keep, right?), but otherwise lgtm.

Lazarus-931:
> Very minor nit that the naming seems a little confusing (it's sequence length to keep, right?), but otherwise lgtm.

i see the confusion, i was just following hf transformers convention, and sequences_to_keep reads as the batch dimension ( like keep N of the batched sequences)

lucasnewman:
> > Very minor nit that the naming seems a little confusing (it's sequence length to keep, right?), but otherwise lgtm.
> 
> i see the confusion, i was just following hf transformers convention, and sequences_to_keep reads as the batch dimension ( like keep N of the batched sequences)

Ok, if it's convention then it's fine.

lucasnewman (approved):

PR #1752, opened 2026-07-29
Porting embeddings

Lazarus-931:
this is a WIP to port embedding models to mlx-vlm, for easier use by nativ upstream and just because mlx-embedding was calling mlx-vlm's components, so much cleaner to have them here. 

edit: complete

### DIFF
--- a/mlx_vlm/models/qwen2_5_vl/language.py
+++ b/mlx_vlm/models/qwen2_5_vl/language.py
@@ -90,7 +90,7 @@ class Attention(nn.Module):
             kv_seq_len += cache.offset + 1
             position_ids = mx.arange(cache.offset, cache.offset + L)
             position_ids = mx.expand_dims(position_ids, axis=0)
-            position_ids = mx.tile(position_ids, (3, 1, 1))
+            position_ids = mx.tile(position_ids, (3, B, 1))
         else:
             kv_seq_len += cache.offset + 1 if cache is not None else 0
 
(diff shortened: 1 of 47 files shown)


mlx_vlm/models/qwen2_5_vl/language.py
```diff
@@ -90,7 +90,7 @@ def __call__(
             kv_seq_len += cache.offset + 1
             position_ids = mx.arange(cache.offset, cache.offset + L)
             position_ids = mx.expand_dims(position_ids, axis=0)
-            position_ids = mx.tile(position_ids, (3, 1, 1))
+            position_ids = mx.tile(position_ids, (3, B, 1))
```
lucasnewman, line 93: What’s the intention here?

lucasnewman (approved):
Looks good!

mlx_vlm/models/qwen2_5_vl/language.py
Lazarus-931: With the 1, batched m-rope broke when B > 1 and tiling (3, B, 1) gives (3, B, L) across the actual batch.

So when B == 1, it's the same if B wasn't there

mlx_vlm/utils.py
Blaizzy, line 1: Can we revert these changes? 

I would like avoid narrow changes to core files, those are generalised.

Blaizzy (commented):
LGTM, I just have one nit on the utils.py

lucasnewman (approved):
Looks good!

PR #1753, opened 2026-07-29
feat(speculative): validate Laguna DFlash contract

Thump604:
## Summary

Adds a pure Laguna S 2.1 DFlash checkpoint contract:

- parse and validate the published DFlash configuration;
- reject target-layer-count or tokenizer-vocabulary mismatches before target binding;
- validate the expected checkpoint weight keys and tensor shapes, including quantized weight auxiliaries.

## Scope

The published DFlash checkpoint has a fixed six-layer draft architecture and
requires a 48-layer Laguna target with the shared 100352-token vocabulary.
This slice makes those compatibility requirements executable before later
loader or speculative-generation wiring consumes the checkpoint.

## Tests

- `python -m pytest mlx_vlm/tests/test_laguna_dflash_drafter.py -q` (`19 passed`)
- `python -m pytest mlx_vlm/tests/test_models.py -k laguna -q` (`7 passed`)
- `pre-commit run --files ...` (Black, isort, autoflake)

## Excluded

No model loading, conversion, generation, cache handling, server wiring,
runtime configuration, or performance claim is included. Subsequent DFlash
implementation slices will build on this validation boundary.

### DIFF
--- a/mlx_vlm/speculative/drafters/laguna_dflash/config.py
+++ b/mlx_vlm/speculative/drafters/laguna_dflash/config.py
@@ -1,0 +1,292 @@
+import inspect
+from dataclasses import dataclass, field
+from typing import Any, Mapping, Optional
+
+from ....models.base import BaseModelConfig
+
+LAGUNA_DFLASH_HIDDEN_SIZE = 3072
+LAGUNA_DFLASH_INTERMEDIATE_SIZE = 12288
+LAGUNA_DFLASH_NUM_LAYERS = 6
+LAGUNA_DFLASH_NUM_ATTENTION_HEADS = 72
+LAGUNA_DFLASH_NUM_KV_HEADS = 8
+LAGUNA_DFLASH_HEAD_DIM = 128
+LAGUNA_DFLASH_SLIDING_WINDOW = 512
+LAGUNA_DFLASH_VOCAB_SIZE = 100352
+LAGUNA_DFLASH_BLOCK_SIZE = 16
+LAGUNA_DFLASH_MASK_TOKEN_ID = 12
+LAGUNA_DFLASH_TARGET_LAYERS = [1, 10, 19, 29, 38, 47]
+LAGUNA_DFLASH_TARGET_LAYER_COUNT = 48
+LAGUNA_DFLASH_AUX_LAYER_IDS = [2, 11, 20, 30, 39, 48]
+
+
+def _value(source: Any, key: str, default: Any = None) -> Any:
+    if isinstance(source, Mapping):
+        return source.get(key, default)
+    return getattr(source, key, default)
+
+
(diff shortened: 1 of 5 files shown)


mlx_vlm/speculative/drafters/laguna_dflash/config.py
```diff
@@ -0,0 +1,292 @@
...
+        default_factory=lambda: list(LAGUNA_DFLASH_AUX_LAYER_IDS)
+    )
+    causal: bool = True
+    attention_bias: bool = False
+    qkv_bias: bool = False
+    tie_word_embeddings: bool = True
+
+    def validate(self) -> None:
+        expected = {
+            "hidden_size": LAGUNA_DFLASH_HIDDEN_SIZE,
```
lucasnewman: I'm not an expert on DFlash, but shouldn't this just come from the model config instead of being hardcoded?

Thump604: Good point. The config loader already read those values, but validation duplicated them as hard-coded checkpoint facts.

Updated in a058b58e:
- required architecture and nested `dflash_config` metadata are now read from the checkpoint;
- validation checks required fields and structural compatibility rather than a copy of this revision's numeric values;
- target layer count, vocabulary compatibility, capture structure, and weight shapes remain fail-closed;
- expected weight shapes are derived from the parsed configuration.

Added coverage with a valid non-default DFlash configuration to prove the contract and tensor shapes follow checkpoint metadata, plus malformed/missing metadata cases.

Validation:
- `python -m pytest mlx_vlm/tests/test_laguna_dflash_drafter.py -q` — 10 passed
- `python -m pytest mlx_vlm/tests/test_models.py -k laguna -q` — 7 passed
- `pre-commit run --files ...` — passed

lucasnewman:
@Thump604 What's the overall intention here? The validation functions you added are only called for synthetic tests, so I'm sure they pass, but never called from the engine. What are we actually validating?

Thump604:
You were right: the prior revision only proved the validator in synthetic tests and did not establish a production call path. I have corrected the scope rather than defending that slice.

The PR now includes the actual Laguna DFlash model backend and loader registration. A checkpoint containing `dflash_config` resolves through `get_model_and_args()` to `laguna_dflash.Model`; the generic drafter compatibility gate invokes `LagunaDFlashDraftModel.validate_target_compatibility()` before serving binds the target. The focused tests now exercise both real loader selection and model-specific compatibility dispatch.

I am keeping target hidden-state capture and speculative generation/server wiring out of this PR as a separate stacked follow-up so this remains reviewable. Local focused result for this revision: 166 passed. Thank you for catching the incomplete boundary.

Lazarus-931:
@lucasnewman, verified it on actual weights. If code is good, i think good to merge

lucasnewman (approved):

PR #1754, opened 2026-07-29
fix: isolate Qwen MTP shards during sanitization

Thump604:
## Summary

Prevent Qwen 3.5 VLM sanitization from letting MTP draft-shard keys select the base model RMSNorm loading convention.

The sanitizers already discard `mtp.*` tensors. They computed `shift_norm_weights` before that discard, so an attached `language_model.mtp.*` shard caused base RMSNorm values to receive the MTP offset. This moves the decision after draft keys are filtered for both dense and MoE Qwen 3.5 VLM implementations.

## Regression coverage

Adds focused dense and MoE tests using a base norm plus an MTP-shard norm. The base value remains unchanged and no MTP key reaches the sanitized result.

## Reproduction boundary

Validated locally with standalone Qwen MTP composite artifacts whose index maps `language_model.mtp.*` tensors to `model-mtp.safetensors`:

- Qwen 3.6 35B-A3B VLM MTP
- Qwen 3.5 122B-A10B VLM MTP

Before the change, the text route emitted corrupted output even with MTP disabled. With the change, the same artifacts returned normal final text. This PR does not claim speculative-throughput improvement or change MTP decoding behavior; it fixes base-weight sanitization when a draft shard is present.

### DIFF
--- a/mlx_vlm/models/qwen3_5/qwen3_5.py
+++ b/mlx_vlm/models/qwen3_5/qwen3_5.py
@@ -142,10 +142,10 @@ class Model(Qwen3VLModel):
         return inputs_embeds, special_image_mask
 
     def sanitize(self, weights):
-        shift_norm_weights = should_shift_norm_weights(weights)
-
-        # ignore mtp weights
+        # The MTP draft shard is separate from the base model. Its presence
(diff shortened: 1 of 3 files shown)


lucasnewman (approved):
Thanks!

PR #1756, opened 2026-07-30
Inkling Small

pcuenca:
Native nvfp4 model supported:

```
mlx_vlm.generate --prompt "who are you?" --model thinkingmachines/Inkling-Small-NVFP4
```

There's also an [mxfp4 version](https://huggingface.co/mlx-community/Inkling-Small-mxfp4) I used to test conversion and inference end-to-end:

```
mlx_vlm.generate --prompt "who are you?" --model mlx-community/Inkling-Small-mxfp4
```

PR #1757, opened 2026-07-31
Inkling: batch-cache position fix, depth-8 MTP empty-cache fixes, numpy dMel input

davidtorcivia:
## Summary

Four fixes for the Inkling family, found while serving Inkling-Small (converted per #1756) behind a batch-cache engine and exercising the depth-8 MTP drafter.

## Decode positions (language.py)

The attention read `kv.offset` before `update_and_fetch` to place its relative-position band. Batch cache implementations can disagree with the plain `KVCache` by one on whether the in-flight token is counted in `.offset`, which shifts the band by one position on every decode step and derails generation within a few tokens (it shows up as immediate token loops). Deriving the query position as `S - L` after `update_and_fetch` is correct for any contiguous cache, and is exactly equal to the old value for a plain `KVCache`.

The same function now coerces the Metal kernel template args to Python ints. Engines hand offsets over as numpy or mx scalars, and `mx.fast.metal_kernel` rejects those with "Invalid template argument".

## Empty-cache handling in the MTP chain (language.py, inkling_mtp.py)

With the depth-8 Inkling-Small MTP chain, only block 0 is fed during prefill, so the later blocks are legitimately snapshotted while still empty, and `KVCache.state` raises when `keys` is None. `_snapshot_cache_state` now records empty caches with a sentinel and `_restore_cache_state` restores them to empty. `draft_eval_state` skips empty caches for the same reason on the first draft round.

## numpy input for dMel extraction (models/inkling/__init__.py)

`InklingProcessor._extract_dmel_bins` assumes torch tensors, but this pipeline runs feature extractors with numpy return types, so audio input crashes with "'numpy.ndarray' object has no attribute 'to'". The patch converts numpy to torch for the call and hands numpy back when numpy came in.

## Testing

Verified end to end on Inkling-Small: text, vision, and audio generation, plus greedy MTP speculative decode with token-identity checks against non-speculative baselines at draft block sizes 2, 3, and 4.

### DIFF
--- a/mlx_vlm/models/inkling/__init__.py
+++ b/mlx_vlm/models/inkling/__init__.py
@@ -3,3 +3,37 @@ from .inkling import Model
 from .language import LanguageModel
 
 __all__ = ["Model", "ModelConfig", "LanguageModel"]
+
+
+def _patch_hf_processor_dmel_numpy():
+    """Let InklingProcessor._extract_dmel_bins accept numpy input.
+
+    The HF processor assumes torch tensors, but this pipeline runs the
+    feature extractor with numpy return types, which crashes the dMel bin
+    extraction with `'numpy.ndarray' object has no attribute 'to'`. Coerce
+    numpy -> torch for the call and hand back numpy when numpy came in, so
+    the surrounding BatchFeature conversion stays type-consistent.
+    """
+    try:
+        import numpy as np
+        from transformers.models.inkling.processing_inkling import InklingProcessor
+    except Exception:  # transformers without inkling — nothing to patch
+        return
+    orig = InklingProcessor._extract_dmel_bins
+    if getattr(orig, "_mlx_vlm_numpy_ok", False):
+        return
+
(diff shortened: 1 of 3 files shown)


mlx_vlm/models/inkling/__init__.py
```diff
@@ -3,3 +3,37 @@
 from .language import LanguageModel
 
 __all__ = ["Model", "ModelConfig", "LanguageModel"]
+
+
+def _patch_hf_processor_dmel_numpy():
+    """Let InklingProcessor._extract_dmel_bins accept numpy input.
```
lucasnewman, line 9: @davidtorcivia There's now a native processor so this shouldn't be necessary.

lucasnewman:
@davidtorcivia Thanks for the contribution! The attention masking isn't quite right either, so instead of having you iterate on it I opened a new PR and added you as co-author in https://github.com/Blaizzy/mlx-vlm/pull/1763

PR #1758, opened 2026-07-31
Video input for models without a video_processor: frames fallback + Inkling native temporal pairs

davidtorcivia:
## Summary

`--video` is dropped silently for models whose processor has no `video_processor` component. The frames load, the generic HF processor skeleton accepts the `videos` kwarg and discards it (it only processes video when the component exists), and the model then answers the question with no visual input at all. With Inkling-Small this produced a confident hallucinated description of a video it never saw.

Three commits:

## 1. Frames-as-images fallback

When `processor.video_processor` is missing, sample the video at the requested fps and append the frames to the image list before the prompt is formatted, so the image placeholder count matches. A signature check is not enough because the skeleton always declares a `videos` parameter, so this checks for the component itself. A notice is printed so the substitution is visible instead of silent.

## 2. Frame cap (--video-max-frames, default 16)

More frames hurt past a couple dozen: at 48 near-identical frames the tested model stopped tracking the event sequence and described a static scene mixture, while 8 to 16 evenly spaced frames preserved event order at a fifth of the token cost. Long clips re-sample evenly to the cap.

## 3. Inkling native temporal pairs

Inkling's vision patches are temporal pairs `[P, T=2, H, W, C]` and the image processor duplicates a still into both slots, so the second slot can carry a different frame. For `inkling_mm_model`, video is fed that way instead: each evenly spaced anchor frame goes through the standard still pipeline as one image entity, and the adjacent frame's patches are spliced into temporal slot 1 inside `get_input_embeddings` via a `video_temporal_pixels` kwarg. Token bookkeeping stays entirely on the standard image path, and each pair carries two frames of temporal signal for the token cost of one still.

Pairing has to use adjacent frames: putting temporally distant frames in one patch reads as a double exposure (the model described one moving ball as two overlapping circles). With adjacent pairing the same clip reads correctly.

## Testing

Synthetic multi-phase 480p clips (moving shape, direction change, object appearance) on Inkling-Small at temperature 0. Before: no media segments in the prompt, hallucinated answer. Fallback: frames arrive as ordered image segments and the model describes the actual events. Native pairs: same quality as the fallback at half the token cost per unit of temporal coverage, with still images unaffected (verified against the single-image baseline). Models with a real video_processor are untouched.

### DIFF
--- a/mlx_vlm/generate/dispatch.py
+++ b/mlx_vlm/generate/dispatch.py
@@ -163,6 +163,13 @@ def parse_arguments():
         help="Frames-per-second to sample from --video.",
     )
     parser.add_argument(
+        "--video-max-frames",
+        type=int,
+        default=16,
+        help="Cap on frames sent when video falls back to ordered images "
+        "(long clips are re-sampled evenly to this count).",
+    )
+    parser.add_argument(
         "--resize-shape",
         type=int,
         nargs="+",
@@ -1382,9 +1389,22 @@ def main():
                             _np.transpose(f, (1, 2, 0)).astype("uint8")
                         )
                     )
+            sampled = len(frames)
+            max_frames = max(2, getattr(args, "video_max_frames", 16) or 16)
+            if sampled > max_frames:
+                # Long clips: keep frames evenly spaced across the full clip.
+                # Beyond a couple dozen near-identical frames, models start
+                # describing a static scene mixture instead of the sequence,
+                # so more frames costs tokens and loses the temporal thread.
+                idxs = [
+                    round(i * (sampled - 1) / (max_frames - 1))
+                    for i in range(max_frames)
+                ]
+                frames = [frames[i] for i in idxs]
             print(
                 f"{processor.__class__.__name__} has no native video support; "
-                f"sending {len(frames)} sampled frames as ordered images."
+                f"sending {len(frames)} of {sampled} sampled frames as "
+                f"ordered images."
             )
             args.image = (args.image or []) + frames
             args.video = None
--- a/mlx_vlm/generate/dispatch.py
+++ b/mlx_vlm/generate/dispatch.py
@@ -1362,6 +1362,7 @@ def main():
     # Fall back to sampling the video into frames and sending them as ordered
     # images, which any image-capable model handles.
     gen_kwargs_extra = {}
+    inkling_video_prompt = None
     if args.video:
         import inspect as _inspect
 
@@ -1382,8 +1383,10 @@ def main():
         _has_video_component = getattr(processor, "video_processor", None) is not None
         if "videos" not in _proc_params or not _has_video_component:
             frames = []
+            frame_fps = args.fps or 2.0
             for v in args.video:
-                arr, _ = _load_video(str(v), fps=args.fps or 2.0)
+                arr, _sfps = _load_video(str(v), fps=args.fps or 2.0)
+                frame_fps = _sfps or frame_fps
                 for f in arr:
                     frames.append(
                         _PILImage.fromarray(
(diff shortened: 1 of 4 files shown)


mlx_vlm/generate/dispatch.py
```diff
@@ -1349,6 +1356,141 @@ def main():
...
+        if "videos" not in _proc_params or not _has_video_component:
+            frames = []
+            frame_fps = args.fps or 2.0
+            for v in args.video:
+                arr, _sfps = _load_video(str(v), fps=args.fps or 2.0)
+                frame_fps = _sfps or frame_fps
+                for f in arr:
+                    frames.append(
+                        _PILImage.fromarray(
+                            _np.transpose(f, (1, 2, 0)).astype("uint8")
```
lucasnewman: This should already be in uint8, so no need for `.astype()` (which will copy).

mlx_vlm/generate/dispatch.py
```diff
@@ -1349,6 +1356,141 @@ def main():
...
+                if len(frames) % 2:
+                    frames.append(frames[-1])
+                n_pairs = max(2, min(max_frames, len(frames) // 2))
+                anchors = [
+                    min(round(i * (len(frames) - 2) / max(n_pairs - 1, 1)), len(frames) - 2)
+                    for i in range(n_pairs)
+                ]
+                evens = [frames[a] for a in anchors]
+                odds = [frames[a + 1] for a in anchors]
+                odd_feats = processor.image_processor.preprocess(
```
lucasnewman: This should already be an mx.array with the native processor, so no need to bounce through numpy.

mlx_vlm/generate/dispatch.py
```diff
@@ -1349,6 +1356,141 @@ def main():
...
+                ) + [{"role": "user", "content": parts}]
+                _tok = (
+                    processor.tokenizer
+                    if hasattr(processor, "tokenizer")
+                    else processor
+                )
+                inkling_video_prompt = _tok.apply_chat_template(
+                    msgs, add_generation_prompt=True, tokenize=False
+                )
+                print(
```
lucasnewman: Was this for debugging?

mlx_vlm/generate/dispatch.py
```diff
@@ -1349,6 +1356,141 @@ def main():
             prompt if isinstance(prompt, list) else [prompt]
         )
 
+    # Processors without native video support (no `videos` parameter) used to
+    # drop --video silently: the frames were loaded, the processor ignored the
+    # kwarg, and the model hallucinated an answer with no visual input at all.
+    # Fall back to sampling the video into frames and sending them as ordered
+    # images, which any image-capable model handles.
+    gen_kwargs_extra = {}
+    inkling_video_prompt = None
```
lucasnewman: In general we shouldn't have model-specific code in `dispatch.py`. We should instead have this be a subsampler utility for video -> image sequence and call it based on some model capability flag.

lucasnewman (changes requested):
@davidtorcivia This looks cool! It needs a few changes before we can merge it, please take a look at the comments.

mlx_vlm/generate/dispatch.py
davidtorcivia: Removed, load_video already gives uint8.

mlx_vlm/generate/dispatch.py
davidtorcivia: Yep its already mx, dropped the bounce.

mlx_vlm/generate/dispatch.py
davidtorcivia: Yeah, debugging leftover, removed.

mlx_vlm/generate/dispatch.py
davidtorcivia: Moved it all into generate/video.py, dispatch now just checks for a prepare_video_frame_pairs hook on the model as the capabilty flag. Also rebased on main.

davidtorcivia:
Rebased on main and fixed the failing check. The cli video test mocked a bare processor so it was tripping the new fallback and trying to actually open clip.mp4, gave the mock native video suport instead. Also added a test covering the fallback path itself.

lucasnewman (approved):
Thanks!

PR #1759, opened 2026-07-31
Inkling: decode fast path (fused sconv and MoE kernels, stacked qkvr, window-bounded attention)

davidtorcivia:
## Summary

Decode-speed work for the Inkling family, measured on Inkling-Small (q4/g64 with q8 attention and shared experts, M-series, 256 GB). Single-stream decode goes from 26.1 to 21.2 ms/step at short context, and from 70.6 to 22.0 ms/step at 4k context, so decode is now nearly flat in context length instead of collapsing as the cache grows. Prefill picks up about 5% at 8k. Everything loads existing checkpoints unchanged.

Rebased onto main after #1756 and #1763 landed; no dependencies remain and the four commits here are the whole change. The sliding-window slice and the mask variant integrate with the left-padding batch masking from #1763 (padding is re-based when keys are sliced, and applied identically by both mask kernels), and the fused sconv path defers to the masked batching path whenever a cache carries padding information.

## Where the time went

An ablation probe against the bandwidth roofline (about 9 ms/step for this model) showed the gap was spread across launch count and a few specific bad shapes rather than any one heavy op: the MoE block accounted for roughly 16 ms/step, attention 8, and the short convolutions 3.4, with each of them dominated by chains of small kernels around modest amounts of real work. Each commit takes one of those apart.

## Short convolutions (commit 1)

Each of the four sconv sites per layer ran about eight ops at decode (fp32 casts, state concat, cache slice, the 4-tap depthwise conv, residual add), 168 sites across the model. One kernel now does the conv plus residual in fp32 and writes the shifted conv state, and the decoder layer's own residual add is folded in behind the same launch. The output is numerically identical to the unfused path, including its bf16 rounding of the conv result before the residual add; the prefill and masked multimodal paths are untouched.

## MoE (commit 2)

Three changes, one per bottleneck:

- The post-gate routing chain (sigmoid, correction bias, top-k, logsigmoid softmax, route and global scaling) was a dozen small kernels per layer. It is now one simdgroup kernel, one token per simdgroup, top-k by simd_max rounds with ties resolved to the lowest lane. A first version did the selection on a single thread and was dramatically slower than the mx op chain it replaced; the lane-parallel version is what shipped.
- The two always-on shared experts went through gather_qmm with constant indices. Their weights are concatenated at load (a shape-driven remap in sanitize that handles both bf16 and quantized triplets) into plain linear layers, and the per-token expert weight is applied on the expert-major intermediate, where it distributes exactly over the concatenated down projection.
- The routed down projection was the one gather that ran far off peak: gather_qmm's vector-per-expert mode reaches about 200 GB/s on the [2048 to 4096] down shape, versus about 750 GB/s for the broadcast mode the gate/up projections use. For affine q4/g64 checkpoints a small kernel dequantizes the selected rows directly (threadgroup per output row, simdgroup per expert, one quant group per lane) and folds the top-k weighted sum over experts in, so the [N, K, out] intermediate and its mul and sum kernels disappear too. It sustains about 550 GB/s; other quantizations keep the gather path.

NVFP4 expert scales that are all ones (any non-NVFP4 checkpoint) are detected once after load and their gather+mul chains skipped.

## Stacked q/k/v/r projection (commit 3)

The four attention projections share the same input row, so their weights are stacked at load into a single `qkvr_proj`; a row concat of quantized matrices is exact since every output row keeps its own groups. Decode runs one matmul instead of four. Checkpoints from before the fusion record per-path quantization entries under the old names, so the loader predicate lets `qkvr_proj` inherit the `q_proj` entry; without that, a checkpoint with q8 attention over a q4 default would requantize the stacked tensor at the default width and fail shape checks on load.

## Banded mask and sliding-window work bounds (commit 4)

The banded-mask kernel carried S and the query offset as Metal template arguments, so every decode step and every prefill chunk shape compiled a fresh pipeline variant. The added variant reads its runtime dimensions from injected shapes and derives the offset as S minus LQ, leaving only per-layer band geometry in the template: one pipeline per layer kind, ever. Bit-identical output; callers passing any other offset keep the old kernel.

Sliding-window layers also no longer hand SDPA the full cached history. A query row can never reach keys older than window minus one positions before it, yet decode and prefill computed O(L times S) masked scores and materialized an [H, L, S] mask per layer regardless of context length. Slicing the cache views to the reachable range bounds both at window plus L for the sliding layers (35 of 42 in Inkling-Small). The dropped keys carried exactly zero softmax weight, so outputs move only at the bf16 reassociation level, and decode is exact. This change is what flattens decode in context length.

## Numbers

Bare greedy decode loop, warm, Inkling-Small q4:

| | before | after |
|---|---|---|
| decode, 100-token context | 26.1 ms/step (38.3 tok/s) | 21.2 ms/step (47.3 tok/s) |
| decode, 4k context | 70.6 ms/step (14.2 tok/s) | 22.0 ms/step (45.6 tok/s) |
| chunked prefill, 8k | 631 tok/s | 670 tok/s |

Through an OpenAI-compatible server on the same box, warm single-stream decode went from 40 to 51 tok/s.

## Numerics and testing

The kernels reproduce the unfused paths' bf16 rounding points (the conv result before the residual add, routing weights, the down projection output before the expert-weight multiply), so differences sit at the reassociation level: greedy decode over 200-token continuations flips 1 to 7 tokens per prompt, always at top-2 logit gaps of one or two bf16 ulps, and teacher-forced logits match within the same margin.

Verified: per-kernel equivalence tests against the unfused implementations (the sconv kernel is bit-exact; MoE selects identical expert sets with weights within one bf16 ulp; the mask variant is bit-identical including strided views); the tiny-model tests in test_models.py; end-to-end text, vision, and audio generation on Inkling-Small plus greedy MTP speculative decode through the depth-8 drafter, whose sanitize gets the same stacked-projection and dense-shared remaps; long-context generation at 4k compared against the unsliced path. The gate matmul into the routing kernel and a tiled prefill attention kernel were both tried and measured slower or not worth it, in case anyone is tempted.

### DIFF
--- a/mlx_vlm/utils.py
+++ b/mlx_vlm/utils.py
@@ -807,6 +807,13 @@ python -m mlx_vlm.convert --hf-path <local_dir> --mlx-path <mlx_dir>
             # Handle custom per layer quantizations
             if p in config["quantization"]:
                 return config["quantization"][p]
+            # Projections fused at load (e.g. inkling qkvr_proj) inherit the
+            # per-path entry of their first constituent from checkpoints saved
+            # before the fusion existed.
+            if p.endswith(".qkvr_proj"):
+                alias = p[: -len("qkvr_proj")] + "q_proj"
+                if alias in config["quantization"]:
+                    return config["quantization"][alias]
             if not hasattr(m, "to_quantized"):
                 return False
             # Skip layers not divisible by 64
--- a/mlx_vlm/utils.py
+++ b/mlx_vlm/utils.py
@@ -807,13 +807,6 @@ python -m mlx_vlm.convert --hf-path <local_dir> --mlx-path <mlx_dir>
             # Handle custom per layer quantizations
             if p in config["quantization"]:
                 return config["quantization"][p]
-            # Projections fused at load (e.g. inkling qkvr_proj) inherit the
-            # per-path entry of their first constituent from checkpoints saved
-            # before the fusion existed.
-            if p.endswith(".qkvr_proj"):
-                alias = p[: -len("qkvr_proj")] + "q_proj"
-                if alias in config["quantization"]:
-                    return config["quantization"][alias]
             if not hasattr(m, "to_quantized"):
                 return False
             # Skip layers not divisible by 64
--- a/mlx_vlm/models/inkling/config.py
+++ b/mlx_vlm/models/inkling/config.py
@@ -1,7 +1,25 @@
 from dataclasses import dataclass
-from typing import List, Optional, Union
+from typing import Dict, List, Optional, Union
 
 from ..base import BaseModelConfig
+
+
+def sanitize_quantization_config(quantization):
+    """Alias per-layer quantization entries onto the load-time fused ``qkvr_proj``.
+
+    Checkpoints quantized before the fusion existed key their per-layer
+    overrides by the constituent paths (``...self_attn.q_proj`` and friends).
+    The stacked projection inherits the ``q_proj`` entry: row-concat keeps each
+    output row's own quantization groups (see ``fuse_qkvr``), so the
+    constituents' settings apply unchanged to the fused matrix.
+    """
+    if not isinstance(quantization, dict):
+        return quantization
+    out = dict(quantization)
+    for key, value in quantization.items():
+        if key.endswith(".self_attn.q_proj"):
+            out.setdefault(key[: -len("q_proj")] + "qkvr_proj", value)
+    return out
 
 
 @dataclass
@@ -86,6 +104,8 @@ class ModelConfig(BaseModelConfig):
     audio_token_id: int = 200053
     vocab_size: int = 201024
     eos_token_id: Optional[List[int]] = None
+    quantization: Optional[Dict] = None
+    quantization_config: Optional[Dict] = None
 
     def __post_init__(self):
         # Coerce dict sub-configs (from config.json) into typed dataclasses, and wire the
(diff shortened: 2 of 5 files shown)


mlx_vlm/utils.py
```diff
@@ -807,6 +807,13 @@ def get_class_predicate(p, m):
             # Handle custom per layer quantizations
             if p in config["quantization"]:
                 return config["quantization"][p]
+            # Projections fused at load (e.g. inkling qkvr_proj) inherit the
```
lucasnewman: Can we instead add quantization and quantization_config to the model's ModelConfig? The you can handle this in post_init and we don't need model-specific code in utils, which is generally not allowed.

lucasnewman (changes requested):
@davidtorcivia This looks really cool! Please take a look at the comment above which we should resolve before we merge this.

mlx_vlm/utils.py
davidtorcivia: Done, moved it into ModelConfig post_init following the gpt_oss/qwen3_5 pattern and reverted utils. Verifed on a per-layer quantized checkpoint that the fused proj still picks up its entires.

lucasnewman (approved):
Awesome, thank you!

lucasnewman:
@davidtorcivia Can you run `pre-commit run --all` to clear the formatter and then we can merge?

davidtorcivia:
Done, formatter is clear now. Thanks for the qiuck review!

lucasnewman (approved):

PR #1762, opened 2026-07-31
Add image input support to /v1/embeddings endpoint

Lazarus-931:
Accept base64 data URLs and OpenAI-style image_url objects as embedding input alongside text. Image items are decoded, run through the model's processor to produce pixel_values, and embedded via the model's image path. Text input remains fully backward compatible, and models without image support return a clear error.

This is for a future upstream feature of nativ


lucasnewman (approved):
LGTM

PR #1763, opened 2026-07-31
Fix Inkling batching and caches with MTP

lucasnewman:
Based on https://github.com/Blaizzy/mlx-vlm/pull/1757, but without the processor changes and a fix for correctly masking attention when batching.


Lazarus-931 (approved):
lgtm!

PR #1764, opened 2026-07-31
Inkling: restore audio input in the bundled processor

davidtorcivia:
## Summary

Audio input for Inkling is broken on main: any `--audio` request raises `Processor InklingProcessor does not support audio parameter`. This restores it inside the bundled processor, torch-free, with dMel quantization verified bit-equal to the reference implementation and end-to-end greedy output token-identical to the last known-good state.

## What broke

The bundled `InklingProcessor` from #1756 installs itself over `AutoProcessor` for every Inkling checkpoint, and it only implements the image and text paths. `process_inputs` inspects the processor signature for an `audio` parameter, finds none, and raises before any audio touches the model. The transformers processor is not a fallback either: its `_extract_dmel_bins` calls `.to(...)` on its input, and this pipeline runs feature extractors with numpy return types, so routing audio there dies with `'numpy.ndarray' object has no attribute 'to'`.

## The fix

Audio becomes a first-class path in the bundled processor:

- `from_pretrained` resolves the feature extractor from the config nested in `processor_config.json`, by class name through the transformers namespace. An older transformers without Inkling still loads the processor for image and text work; only an actual audio input raises, with a message saying why.
- dMel quantization is implemented in plain numpy: clamp to the dMel range, then nearest bin center, computed in float64 like the reference. No torch import anywhere in the path.
- The audio placeholder emitted by the chat template expands into one audio token per valid dMel frame (mask-aware), mirroring how image placeholders expand per patch, and the processor returns `audio_input_ids` and `audio_input_ids_mask` in the shapes the model's audio tower consumes.

One deliberate oddity, commented in the code: the constructor parameter is `audio_extractor`, stored as `self.feature_extractor`. `ProcessorMixin.get_attributes` scrapes the `__init__` signature for modality keywords, and a parameter named `feature_extractor` silently becomes a required constructor argument for every checkpoint, image-only conversions included.

## Verification

- New test compares the numpy dMel binning against the transformers torch implementation on random features, exact bin centers, values just off the bin midpoints, and out-of-range values that exercise the clamp: bit-equal bin ids everywhere. Exact midpoints are excluded on purpose: numpy and torch `linspace` disagree in the last ulp of the centers, and only a feature sitting exactly on a midpoint could observe it; the clamp keeps real log-mel values off that measure-zero set. The tests skip cleanly when torch or an Inkling-aware transformers is absent.
- A second test pins the error behavior when no feature extractor is available.
- End to end on Inkling-Small (q4, M-series): greedy `--audio` generation over a 440 Hz sine fixture produces exactly the same tokens as the pre-regression transformers path did ("A brief, high-pitched electronic beep."), and the full `test_inkling.py` suite passes.

### DIFF
--- a/mlx_vlm/models/inkling/processing_inkling.py
+++ b/mlx_vlm/models/inkling/processing_inkling.py
@@ -10,6 +10,7 @@ from typing import List, Optional, Union
 
 import mlx.core as mx
 import numpy as np
+import transformers
 from PIL import Image
 from transformers import AutoTokenizer
 from transformers.feature_extraction_utils import BatchFeature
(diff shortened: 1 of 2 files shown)


mlx_vlm/models/inkling/processing_inkling.py
```diff
@@ -249,10 +351,35 @@ def from_pretrained(cls, pretrained_model_name_or_path, **kwargs):
...
+        proc_cfg_path = model_path / "processor_config.json"
+        if proc_cfg_path.exists():
+            proc_cfg = json.loads(proc_cfg_path.read_text())
+            dmel_kwargs = {
+                key: proc_cfg[key]
+                for key in ("num_dmel_bins", "dmel_min_value", "dmel_max_value")
+                if key in proc_cfg
+            }
+            fe_cfg = proc_cfg.get("feature_extractor")
+            if isinstance(fe_cfg, dict):
```
lucasnewman, line 369: No need for the transformers feature extractor here: this should just be implemented in MLX. dMel isn't very complicated (it's just logmel with scale correction and deterministic quantization) and `mlx-audio` is already a dependency.

lucasnewman:
@davidtorcivia As before I appreciate this contribution and hopefully this isn't discouraging, but I thought it would be faster to handle this myself since I'm very familiar with the gotchas around the audio input. Closing in favor of https://github.com/Blaizzy/mlx-vlm/pull/1767

Issue #1448, opened 2026-06-26
mlx-vlm convert cash


uxbug:
Using Python 3.11.14 environment 

```
mlx                       0.31.2
mlx-audio                 0.4.4
mlx-embeddings            0.1.0
mlx-lm                    0.31.3
mlx-metal                 0.31.2
mlx-vlm                   0.6.3
vmlx                      1.3.71
```

```
$ mlx_vlm.convert --hf-path huihui-ai/Huihui-Qwen3.6-35B-A3B-abliterated --mlx-path huihui-ai/Huihui-Qwen3.6-35B-A3B-abliterated-mxfp4 -q --q-mode mxfp4
[INFO] Loading
[INFO] Using dtype: bfloat16
[INFO] Quantizing
[INFO] Quantized model with 4.402 bits per weight.
libc++abi: terminating due to uncaught exception of type std::runtime_error: [METAL] Command buffer execution failed: Caused GPU Timeout Error (00000002:kIOGPUCommandBufferCallbackErrorTimeout)
[1]    2659 abort      mlx_vlm.convert --hf-path huihui-ai/Huihui-Qwen3.6-35B-A3B-abliterated   -q
```

There is no problem using mlx-lm

```
$ mlx_lm.convert --hf-path huihui-ai/Huihui-Qwen3.6-35B-A3B-abliterated --mlx-path huihui-ai/Huihui-Qwen3.6-35B-A3B-abliterated-mxfp4 -q --q-mode mxfp4
[INFO] Loading
[INFO] Using dtype: bfloat16
[INFO] Quantizing
[INFO] Quantized model with 4.253 bits per weight.
```
There is no problem in using mlx_vlm mix.

```
$ mlx_vlm.convert --hf-path huihui-ai/Huihui-Qwen3.6-35B-A3B-abliterated --mlx-path huihui-ai/Huihui-Qwen3.6-35B-A3B-abliterated-mixed_4_6 -q --quant-predicate mixed_4_6
[INFO] Loading
[INFO] Using dtype: bfloat16
[INFO] Quantizing
[INFO] Quantized model with 5.012 bits per weight.
```

PR #1765, opened 2026-07-31
Evaluate weights incrementally when saving to avoid Metal GPU timeout

axiom-of-choice:
Addresses #1448, which was closed without a fix.

## Summary

`mlx_vlm.convert` quantizes a large model fine and then dies while writing the
weights:

```
[INFO] Quantized model with 4.402 bits per weight.
RuntimeError: [METAL] Command buffer execution failed: Caused GPU Timeout Error
  (00000002:kIOGPUCommandBufferCallbackErrorTimeout)
```

`save_weights()` builds shards of up to `MAX_FILE_SIZE_GB` (5 GB) and hands each
one straight to `mx.save_safetensors`. That call forces evaluation of every lazy
tensor in the shard as a single Metal command buffer, and on a big model the
buffer runs long enough for the IOGPU watchdog to kill it. The quantization step
has already finished by then; only the write fails.

The fix evaluates each shard in small batches before saving, so no single command
buffer is big enough to trip the watchdog. Same approach used in mlx-lm for the
same class of failure (ml-explore/mlx-lm#1572).

## Changes

- `mlx_vlm/utils.py`: `save_weights()` takes `shard_gb` and `eval_every`. Before
  each `mx.save_safetensors`, the shard's tensors are evaluated in batches of
  `eval_every`. Added `DEFAULT_EVAL_EVERY = 16` and validation on both params.
- `mlx_vlm/convert.py`: both are threaded through `convert()` and exposed as
  `--shard-gb` and `--eval-every`.
- `mlx_vlm/tests/test_utils.py`: new `TestSaveWeights`, 13 tests.

Defaults preserve today's on-disk layout; `shard_gb` stays at `MAX_FILE_SIZE_GB`
and the batched eval is behaviour-preserving. Pass `eval_every=None` to get the
exact previous code path.

## Evidence

Controlled pair on the model from #1448, `huihui-ai/Huihui-Qwen3.6-35B-A3B-abliterated`
with `-q --q-mode mxfp4`, same machine, minutes apart. `eval_every=None` is the
current behaviour:

| Run | Settings | Result |
|---|---|---|
| stock | `eval_every=None`, `shard_gb=5` | crash after 138s, `kIOGPUCommandBufferCallbackErrorSubmissionsIgnored`, shard 1 at 5.35 GB and shard 2 at 0 bytes |
| fixed | `eval_every=16`, `shard_gb=1` | ok after 57s, 19 shards written |

Both runs report `4.402 bits per weight`, which matches the number in #1448
exactly, so this is the same failure the issue describes. The crash happened with
37.9 GB of disk free, and the fixed run then completed the same conversion with
25 GB free, so it is not a disk problem.

Two more data points from a 36B MoE (`qwen3_5_moe`, 102 GB bf16 source): the stock
converter died the same way while writing shard 3, and a chunked writer got
through 20 of 22 shards with no GPU error at all.

Correctness of the output:

- With default settings the emitted `model.safetensors` and
  `model.safetensors.index.json` are byte-identical to the pre-change code path,
  verified by SHA-256 against a run with `eval_every=None`.
- Multi-shard round-trip is bit-exact: 2.64 GB across 1120 quantized tensors into
  3 shards, every tensor compared with `mx.array_equal` after reload, and every
  `weight_map` entry checked to resolve to the shard that actually holds it.

## Risks

The watchdog is a timeout, so the failure is load sensitive rather than a hard
threshold. On a smaller model I could not get it to fire at all across repeated
stock runs, so shard byte size alone is not sufficient; it depends on how much
work the buffer ends up doing and what else is using the GPU. Treat the 35B
result above as the case this addresses.

The batched eval now runs on every convert, not just large models. It is
behaviour-preserving and the byte-identical check above covers that, but it is a
change to a path every conversion goes through, hence `eval_every=None` as an
escape hatch.

`eval_every=16` is a conservative default. Very wide tensors could in principle
still make one batch large; lowering `--eval-every` handles that without a code
change.

## Pre-merge

- `pytest mlx_vlm/tests/test_utils.py`, 47 passed, 13 of them new.
- `pytest mlx_vlm/tests/test_models.py -k "quant or gemma4_unified or save"`, 14
  passed. These cover the existing save then reload round-trips with quantized
  weights, which is the path being changed.
- `black` and `isort --profile=black` clean.

## Post-merge

- Worth reopening #1448 so the original reporter can confirm on their model, it
  was closed without a fix.
- Anyone hitting this on a very large model can drop `--shard-gb` to 1 as well;
  smaller shards reduce peak memory during the write.

## Note

`convert()` defaults `q_group_size=64` in its Python signature, which is invalid
for `mxfp4` and raises `[quantize] mxfp4 quantization requires group size 32 but
got 64` when the function is called directly. The CLI avoids it by passing `None`.
Unrelated to this change so I left it alone, but it looks like a real API vs CLI
inconsistency.

### DIFF
--- a/mlx_vlm/convert.py
+++ b/mlx_vlm/convert.py
@@ -9,6 +9,8 @@ import mlx.nn as nn
 from mlx.utils import tree_map_with_path
 
 from .utils import (
+    DEFAULT_EVAL_EVERY,
+    MAX_FILE_SIZE_GB,
     MODEL_CONVERSION_DTYPES,
     create_model_card,
     fetch_from_hub,
@@ -295,6 +297,8 @@ def convert(
     dequantize: bool = False,
     trust_remote_code: bool = True,
     quant_predicate: Optional[str] = None,
+    shard_gb: int = MAX_FILE_SIZE_GB,
+    eval_every: Optional[int] = DEFAULT_EVAL_EVERY,
 ):
     print("[INFO] Loading")
     model_path = get_model_path(hf_path, revision=revision)
@@ -388,7 +392,13 @@ def convert(
     if isinstance(mlx_path, str):
         mlx_path = Path(mlx_path)
 
-    save_weights(mlx_path, target, donate_weights=True)
+    save_weights(
+        mlx_path,
+        target,
+        donate_weights=True,
+        shard_gb=shard_gb,
+        eval_every=eval_every,
+    )
 
     # Copy Python and JSON files from the model path to the MLX path
     for pattern in ["*.py", "*.json"]:
@@ -534,6 +544,19 @@ def configure_parser() -> argparse.ArgumentParser:
         help="Trust remote code.",
         action="store_true",
         default=False,
+    )
+    parser.add_argument(
+        "--shard-gb",
+        help="Maximum size of each safetensors shard, in gigabytes.",
+        type=int,
+        default=MAX_FILE_SIZE_GB,
+    )
+    parser.add_argument(
+        "--eval-every",
+        help="Tensors to evaluate per batch while writing shards. Lower this if"
+        " saving a large model hits a Metal GPU timeout.",
+        type=int,
+        default=DEFAULT_EVAL_EVERY,
     )
     return parser
 
--- a/mlx_vlm/tests/test_utils.py
+++ b/mlx_vlm/tests/test_utils.py
@@ -10,10 +10,12 @@ from unittest.mock import MagicMock, patch
 import mlx.core as mx
 import mlx.nn as nn
 import pytest
+from mlx.utils import tree_flatten
 
 from mlx_vlm.convert import _preserve_existing_deepseek_v4_quantization
 from mlx_vlm.models.text_only import TextOnlyModel
 from mlx_vlm.utils import (
+    MAX_FILE_SIZE_GB,
     StoppingCriteria,
     _load_safetensors,
     apply_generation_config_defaults,
@@ -23,9 +25,11 @@ from mlx_vlm.utils import (
     load_image,
     load_model,
     load_processor,
+    make_shards,
     prepare_inputs,
     process_inputs_with_fallback,
     sanitize_weights,
+    save_weights,
     update_module_configs,
 )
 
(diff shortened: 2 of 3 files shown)


Blaizzy:
Thanks @axiom-of-choice for the contribution 

Please open a GH issue and we'll look into it 👌🏽

Blaizzy:
#1448 was fixed if you look the related PR.

Blaizzy:
What machine do you have and what version of mlx-vlm are you running? 

Please share the result of: 

> pip list | grep mlx

axiom-of-choice:
Mac mini, Apple M2 Pro (10-core), 32 GB unified memory, macOS 26.5.2 (25F84), Python 3.11.14.

```
mlx                0.32.0
mlx-audio          0.4.4
mlx-lm             0.31.3
mlx-metal          0.32.0
mlx-vlm            0.6.8
```

One detail on what I actually ran: the crash and the fix runs were both from source at `07bd830` (main, 2026-07-31), not the installed 0.6.8 wheel, so #1450 was in the tree.

Still digging into the why. I'll follow up here with a cleaner breakdown.

axiom-of-choice:
@Blaizzy Thanks for the response!

Following up on my own claim, and correcting it

I tried to build a second repro on a different architecture .

That means my 35B result is N=1, on a machine that was under load, and the crash and success runs also differed in two variables at once (`shard_gb` 5 vs 1, and `eval_every`). So I cannot claim from this that #1450 left the bug live. It may well have fixed the qwen3_5_moe path.

So I would rather present this as a reliability improvement to the write path than as a bug fix:

- `save_weights` currently lets one `mx.save_safetensors` call force evaluation of a whole shard, up to `MAX_FILE_SIZE_GB`, as a single command buffer. Bounding that work is cheap and removes a failure mode that is load sensitive and hard to reproduce on demand.
- Defaults keep the on-disk layout identical, verified by SHA-256 against the current code path, and `--shard-gb` gives people hitting a timeout a knob that does not need a code change.
- It is not architecture specific, so it does not depend on each model's sanitizer keeping graphs small.

Happy for you to treat it as low priority or close it if you do not think the write path needs the guard.

axiom-of-choice:
@Blaizzy 

To be clear about what I am proposing: a reliability improvement to the write path, not a bugfix.

  Why it is safe:

  - With defaults the emitted `model.safetensors` and `model.safetensors.index.json` are byte-identical to the current code path, verified by SHA-256 against a run with `eval_every=None`.
  - `shard_gb` defaults to `MAX_FILE_SIZE_GB`, so the on-disk layout does not change.
  - Multi-shard round-trip is bit-exact: 1120 quantized tensors across 3 shards, every tensor compared with `mx.array_equal` after reload, every `weight_map` entry checked against the shard that holds it.
  - `eval_every=None` reproduces the exact previous path if anyone wants it.
  
  Why it is an improvement:

  - Today one `mx.save_safetensors` call can force evaluation of a full 5 GB shard as a single Metal command buffer. Batching that evaluation bounds the work per buffer, which is cheap and removes a failure mode that is load sensitive and
  hard to reproduce on demand.
  - It does not depend on each model's sanitizer keeping graphs small, so it holds across architectures.
  - `--shard-gb` and `--eval-every` give anyone who hits a timeout a knob instead of needing a patch.

PR #1767, opened 2026-08-01
Add audio feature extractor for Inkling

lucasnewman:
Based on https://github.com/Blaizzy/mlx-vlm/pull/1764.

This uses an MLX-native implementation of dMel that's bit-compatible with the transformers processor which has bin centers and boundaries in fp64. It also chunks the input so max memory usage stays low (~144mb for 30 mins of audio).


Lazarus-931 (approved):
lgtm!

PR #1773, opened 2026-08-02
embedding_loader: quantize pre-quantized weights before loading

Lazarus-931:
The embedding loader built the model in full precision and called load_weights directly, so any model with a quantization block in its config (4-bit / 8-bit embedders) failed with a shape mismatch. Apply nn.quantize using the config's quantization params before load_weights, gated on a predicate that only quantizes modules that expose to_quantized and actually ship .scales in the checkpoint.

for https://github.com/Blaizzy/nativ/pull/162


lucasnewman (approved):
LGTM

Issue #1520, opened 2026-07-06
gemma4_unified fails to load in 0.6.4 (regression) — video_processor TypeError masked as "Could not import module 'Gemma4UnifiedProcessor'"


plz12345:
### Summary

`mlx-vlm==0.6.4` cannot load any `gemma4_unified` checkpoint in a torch-less MLX environment. It fails with a misleading error:

```
ModuleNotFoundError: Could not import module 'Gemma4UnifiedProcessor'. Are this object's requirements defined correctly?
```

This is a **regression in 0.6.4** — `0.6.3` loads the same model fine. The real error is a `TypeError` swallowed by a bare `except Exception: pass` in the AutoProcessor patch, so the surfaced message points the wrong way. (Note: PR #1321 previously fixed video-processor config handling for gemma4_unified and shipped in 0.6.3; the 0.6.4 changes to `gemma4/processing_gemma4.py` appear to have re-introduced the problem.)

### Environment

- mlx-vlm: **0.6.4** (broken) vs **0.6.3** (works)
- transformers: **5.12.1** (newest allowed by mlx-vlm 0.6.4's own `transformers<5.13.0,>=5.5.0` pin)
- No `torch`/`torchvision` installed (standard MLX-only setup)
- macOS / Apple Silicon, Python 3.11
- Model: any `gemma4_unified` checkpoint (repro'd with `mlx-community/gemma-4-12B-it-8bit`)

### Repro

```bash
# broken
uv run --with "mlx-vlm==0.6.4" -- \
  mlx_vlm.generate --model mlx-community/gemma-4-12B-it-8bit \
  --prompt "hi" --max-tokens 4 --no-verbose
# -> ModuleNotFoundError: Could not import module 'Gemma4UnifiedProcessor'

# works
uv run --with "mlx-vlm==0.6.3" --with "transformers<5.13" -- \
  mlx_vlm.generate --model mlx-community/gemma-4-12B-it-8bit \
  --prompt "hi" --max-tokens 4 --no-verbose
# -> generates normally
```

### Real (masked) error

Calling the processor directly, unmasked:

```python
from mlx_vlm.models.gemma4_unified.processing_gemma4_unified import Gemma4UnifiedProcessor
Gemma4UnifiedProcessor.from_pretrained("mlx-community/gemma-4-12B-it-8bit", trust_remote_code=True)
```

```
[...]

PR #1777, opened 2026-08-02
fix: tolerate null vision_config in quantization skip_vision path

### LINKED ISSUE #1776: AttributeError: 'NoneType' object has no attribute 'get' when loading quantized text-only gemma4_unified (vision_config: null)
## Summary

Loading a **quantized, text-only** fine-tune of a `gemma4_unified` model fails during the quantization step when `config.json` sets `vision_config` to `null`. The same crash would affect any quantized checkpoint that sets `vision_config: null`.

## Repro

`mlx-community/gemma-4-12b-coder-fable5-composer2.5-4bit` is a text-only coder quant — only `language_model.*` weights, vision/audio towers stripped, `config.json` has `"vision_config": null`.

```python
from pathlib import Path
from mlx_vlm.utils import load_model, load_config

d = Path("<model_dir>")  # any quantized gemma4_unified with vision_config: null
cfg = load_config(d)
model = load_model(model_path=d, config=cfg, lazy=True)
```

## Error

```
File ".../mlx_vlm/utils.py", line 799, in load_model
    skip_vision = config.get("vision_config", {}).get("skip_vision", False)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get'
```

## Root cause

The legacy quantization `skip_vision` lookup is:

```python
config.get("vision_config", {}).get("skip_vision", False)
```

`dict.get("vision_config", {})` returns the **value** when the key is present — `None` when the JSON is `"vision_config": null` — not the `{}` default. Calling `.get(...)` on `None` then crashes.

This path is only reached for **quantized** models (`if quantization is not None`), and only when `vision_config` is explicitly `null`. The `gemma4_unified` model class itself handles `vision_config is None` correctly (it skips building the vision/audio tower), so a text-only quant that sets `vision_config: null` is a valid representation — the legacy quantization flag lookup just isn't `None`-safe.

## Why text-only quants set vision_config to null

Text-only quants of unified families strip the vision/audio tower weights. To stop the model class from instantiating a tower it has no weights for, the config sets `vision_config: null` — the `gemma4_unified` `ModelConfig` otherwise defaults `vision_config` to a non-`None` `VisionConfig` via `default_factory`. Without `null`, you instead get `Missing N parameters: vision_embedder.*, embed_vision.*` on strict weight load.

## Proposed fix

Make the `skip_vision` lookup tolerate a non-dict `vision_config` (treat it as "no `skip_vision`"). Minimal, no behavior change for checkpoints that ship a real or absent `vision_config`. PR to follow.

## Related

- #1318 — proposed remapping `gemma4_unified → gemma4` + stripping vision keys (closed, not merged); the native unified model was preferred instead, which is the path this fix strengthens.
- #1382 — legacy gemma4 shared-KV weight loading (closed, not merged).
- #1520 — separate `gemma4_unified` load regression in 0.6.4 (video_processor `TypeError`).
- #1277 — `vision_embedder` layer_norm size mismatch (closed).

garnetlyx:
## What & why

Fixes #1776.

Loading a **quantized, text-only** `gemma4_unified` checkpoint (one whose `config.json` sets `"vision_config": null` because the vision/audio tower weights were stripped) crashes during the quantization step:

```
File ".../mlx_vlm/utils.py", line 799, in load_model
    skip_vision = config.get("vision_config", {}).get("skip_vision", False)
AttributeError: 'NoneType' object has no attribute 'get'
```

`dict.get("vision_config", {})` returns the **value** when the key is present — `None` for `"vision_config": null` — not the `{}` default, so the chained `.get(...)` blows up. This path only runs for quantized models.

Setting `vision_config: null` is the correct way to represent a text-only unified quant: the `gemma4_unified` model class skips the tower when `vision_config is None`, otherwise its `ModelConfig` defaults to a non-`None` `VisionConfig` (via `default_factory`) and strict weight load fails with `Missing N parameters: vision_embedder.*, embed_vision.*`. The model class already handles `None`; only this legacy quantization flag lookup did not.

## Change

Extract the `skip_vision` resolution into a small `_resolve_skip_vision(config)` helper that treats a non-dict `vision_config` as "no `skip_vision`". No behavior change for checkpoints that ship a real (`dict`) or absent `vision_config`.

## Testing

- New `mlx_vlm/tests/test_load.py` covers `vision_config` of `None`, absent, `{}`, and `{"skip_vision": True/False}`.
- `pre-commit run --all` passes (black / isort / autoflake).
- `python -m unittest mlx_vlm.tests.test_models` — 249 tests pass.
- Verified end-to-end against the model in #1776: with `vision_config: null` + this patch, `load_model` succeeds and the model builds text-only (`vision_embedder`/`embed_vision`/`embed_audio` all `None`, `language_model` present).

## Related

- #1318 — proposed remapping `gemma4_unified → gemma4` + stripping vision keys (closed, not merged); the native unified model was preferred. This PR strengthens that chosen native path rather than reviving the remap approach.
- #1382 — legacy gemma4 shared-KV weight loading (closed, not merged).
- #1520 — separate `gemma4_unified` load regression in 0.6.4 (different error).
- #1277 — `vision_embedder` size mismatch (closed).


Lazarus-931:
hi @garnetlyx, please verify your commits and we should be gtg!

garnetlyx:
> hi @garnetlyx, please verify your commits and we should be gtg!

Just signed the commit. Thanks!

mlx_vlm/utils.py
```diff
@@ -782,7 +796,7 @@ def load_model(model_path: Path, lazy: bool = False, **kwargs) -> nn.Module:
     if (quantization := config.get("quantization", None)) is not None:
         # Handle legacy models which may or may not have vision quantized
         # TODO: Re-upload the models with the new quantization config and remove this
-        skip_vision = config.get("vision_config", {}).get("skip_vision", False)
```
lucasnewman, line 785: Could this just be `(config.get("vision_config") or {}).get("skip_vision", False)`? I think the intention is clear as it is today and the separate function isn't adding much.

lucasnewman (changes requested):
@garnetlyx See my comment, if you can tweak it we can merge.

mlx_vlm/utils.py
garnetlyx: Good call — inlined as you suggested in eb7dad3 and dropped the helper + its test module. Thanks!

lucasnewman (approved):

PR #1778, opened 2026-08-02
fix(tokenizer): keep incomplete UTF-8 tail buffered at streaming flush boundary (#1771)

### LINKED ISSUE #1771: streaming detokenizer: UnicodeDecodeError on partial multi-byte UTF-8 in tokenizer_utils.add_token
**Environment:** mlx-vlm 0.6.8, mlx 0.32.0, Python 3.14, macOS 15 (M4 Max, 64GB).

While generating with qwen3_omni_moe audio inputs (degenerate byte-token sequences from the #1768/#1770 class), the stream crashes inside the detokenizer instead of surfacing the model's (bad) output:

```
File ".../mlx_vlm/generate/dispatch.py", line 1027, in stream_generate
    detokenizer.add_token(token, skip_special_token_ids=skip_special_token_ids)
File ".../mlx_vlm/tokenizer_utils.py", line 237, in add_token
    ).decode("utf-8")
UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 25-26: unexpected end of data
```

The eager `.decode("utf-8")` assumes token boundaries align with UTF-8 character boundaries — they don't for byte-level BPE, so any multi-byte character split across an internal buffer edge can raise mid-stream. The usual fix is `decode("utf-8", errors=...)` deferred until the byte buffer holds complete sequences (or `errors="replace"` at flush). I hit it via degenerate outputs, but any emoji/CJK-heavy generation that lands a split at the wrong offset should be able to trigger it.

FreazeMood:
Fixes #1771.

The BPE streaming detokenizer flushes its byte buffer whenever a token starts with a space. If the buffer ends in the middle of a multi-byte UTF-8 sequence at that moment, the current `errors="replace"` decode turns the pending bytes into replacement characters — corrupting text (CJK, Hebrew, emoji) that would have decoded cleanly once the next token's bytes arrived. The previous strict decode raised `UnicodeDecodeError` at the same spot, which is what #1771 reports.

This change decodes strictly first; on failure it distinguishes the two cases:
- **truncated sequence at the buffer edge** (`e.end == len(raw)`): flush the clean head, keep the undecoded tail buffered so it completes with the next token — lossless;
- **genuinely invalid bytes mid-buffer**: fall back to replacement, as today.

Includes a regression test that fails on current main (the head gets a replacement char appended) and passes with the fix. Verified against the reproduction in #1771 on an M-series machine.

### DIFF
--- a/mlx_vlm/tests/test_streaming_detokenizer_utf8.py
+++ b/mlx_vlm/tests/test_streaming_detokenizer_utf8.py
@@ -1,0 +1,47 @@
+"""Regression test for issue #1771: an incomplete multi-byte UTF-8
+sequence at the streaming flush boundary must neither raise nor mangle
+the already-decodable head of the buffer."""
+
+import unittest
+
+
+class FakeTokenizer:
+    """Minimal stand-in exposing what BPEStreamingDetokenizer uses."""
+
+    def __init__(self, vocab):
+        self.vocab = vocab
+
+    def get_vocab(self):
+        return self.vocab
+
+
+class TestBPEStreamingUTF8(unittest.TestCase):
+    def test_incomplete_tail_at_flush_boundary(self):
+        from mlx_vlm.tokenizer_utils import BPEStreamingDetokenizer
+
+        BPEStreamingDetokenizer.make_byte_decoder()
+        byte_decoder = BPEStreamingDetokenizer._byte_decoder  # char -> byte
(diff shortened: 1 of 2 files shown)


lucasnewman:
This was fixed in https://github.com/Blaizzy/mlx-vlm/commit/3cf2e75acd6373a5dcda00cda7779e0e6350976c

PR #1779, opened 2026-08-02
Fix APC extra_hash to use prefix-invariant semantic inputs

### LINKED ISSUE #999: Server clears Metal cache after every request, destroying KV prefix cache and forcing full re-prefill on every turn
## Description

`mlx_vlm.server` calls `mx.clear_cache()` and `gc.collect()` after every single generation request (both streaming and non-streaming). This wipes the Metal memory cache, which destroys any KV cache state from the previous request. As a result, every turn in a multi-turn conversation pays the full prefill cost from scratch, even when the conversation shares a long common prefix with the previous turn.

For large models with long conversations, this makes multi-turn latency dramatically worse than it needs to be.

## Location in Source

`mlx_vlm/server.py`, three locations:

```python
# Line ~1077 (streaming path)
mx.clear_cache()
print("Stream finished, cleared cache.")

# Line ~1105 (non-streaming path)  
mx.clear_cache()
gc.collect()
print("Generation finished, cleared cache.")

# Line ~1142 (another path)
mx.clear_cache()
```

## Impact

I serve Qwen3.5 397B A17B (4bit) via mlx_vlm on a Mac Studio M3 Ultra 512GB behind an agent framework (Hermes Agent). A typical conversation builds up to 50,000+ tokens of context over time.

With `prefill-step-size 512`, prefilling 58,000 tokens requires ~115 chunks through a 397B model. This takes several minutes. On the second turn, if the KV cache from the first turn were preserved, only the new tokens (the latest user message + assistant prefix) would need prefilling, which would take seconds instead of minutes.

Instead, every single turn pays the full multi-minute prefill cost because `mx.clear_cache()` wipes everything.

Over the course of a conversation, I observed the server log `Generation finished, cleared cache.` over 1,143 times in one session. Each of those clears threw away KV cache state that could have accelerated the next request.

## Expected Behavior

The server should preserve KV cache state between requests by default for multi-turn conversations. The `prompt_cache_state` parameter already exists in the generate call (line ~1101), suggesting the infrastructure for cache reuse is partially there.

## Suggested Fix

1. **Do not call `mx.clear_cache()` after every request.** Instead, only clear when the model is unloaded, when memory pressure requires it, or when the conversation is reset (detected by a shorter message list than the previous request, which blockops-proxy already detects as `SESSION_RESET`).

2. **Add a server flag** like `--preserve-kv-cache` or `--no-clear-cache` that skips the `mx.clear_cache()` call after generation. Users who need aggressive memory reclamation can keep the current behavior; users serving multi-turn conversations can opt into cache persistence.

3. **Implement prompt cache matching** so that when the next request shares a prefix with the previous one (which is always the case in multi-turn chat), the server reuses the cached KV state and only prefills the new tokens.

## Memory Considerations

The concern behind `mx.clear_cache()` is likely memory reclamation. On a 512GB M3 Ultra serving a single 397B model, memory is not the bottleneck since the model itself uses ~200GB and the KV cache for 60k tokens is modest in comparison. For smaller memory configurations, option 2 (a flag) would let users choose the right tradeoff for their hardware.

## Environment

- mlx_vlm version: 0.4.4
- Model: Qwen3.5 397B A17B 4bit
- macOS, Apple Silicon M3 Ultra 512GB
- Single user serving behind a proxy

f1yn:
## Summary
- Centralize APC identity in `apc.py`: hash **source** multimodal inputs (image/audio/video/tenant), not full-sequence `inputs_embeds` / `attention_mask`.
- **Store** with stable `extra_hash`; **lookup** tries stable first, then legacy (pre-fix full-tensor hash) for backward compatibility.
- Fixes cross-turn exact-cache misses on the continuous-batching server path (`exact_hits: 0` with growing context).

Fixes #999

## Test plan
- [x] `pytest mlx_vlm/tests/test_apc.py -k "extra_hash or legacy"`
- [x] `pytest mlx_vlm/tests/test_apc.py mlx_vlm/tests/test_apc_semantic_key.py`
- [x] Two-turn chat with `APC_ENABLED=1` → `/v1/cache/stats` shows `exact_hits > 0`

## Operator note
For long multi-turn sessions, set `APC_EXACT_CACHE_ENTRIES` above the default LRU of 2 (e.g. `32`).

### DIFF
--- a/mlx_vlm/apc.py
+++ b/mlx_vlm/apc.py
@@ -190,6 +190,123 @@ def model_key_dependencies(model: Any = None, processor: Any = None) -> Tuple[in
     return tuple(deps)
 
 
+def resolve_apc_image_hash(
+    prompt_kwargs: Optional[dict],
+    *,
+    image_ref: Any = None,
+) -> int:
+    """Image salt from precomputed hash, ``pixel_values``, or image path ref."""
+    kw = prompt_kwargs or {}
+    img = kw.get("_apc_image_hash")
+    if img is not None:
+        return int(img)
+    pixel_values = kw.get("pixel_values")
+    return hash_image_payload(pixel_values=pixel_values, image_ref=image_ref)
+
+
+def apc_prefix_invariant_media(
+    prompt_kwargs: Optional[dict],
+    *,
+    audio_ref: Any = None,
+    video_ref: Any = None,
+) -> Dict[str, Any]:
+    """Source multimodal inputs for APC identity — not derived full-sequence tensors."""
+    kw = prompt_kwargs or {}
+    audio = kw.get("input_features")
+    if audio is None:
+        audio = audio_ref
+    video = kw.get("pixel_values_videos")
+    if video is None:
+        video = video_ref
+    return {
+        "audio": audio,
+        "video": video,
+    }
+
+
+def apc_legacy_media(
+    prompt_kwargs: Optional[dict],
+    *,
+    audio_ref: Any = None,
+    video_ref: Any = None,
+    attention_mask: Any = None,
+) -> Dict[str, Any]:
+    """Pre-fix media dict including full-sequence embeds/masks (0.6.8–)."""
+    media = apc_prefix_invariant_media(
+        prompt_kwargs, audio_ref=audio_ref, video_ref=video_ref
+    )
+    kw = prompt_kwargs or {}
+    media["embeddings"] = kw.get("inputs_embeds")
+    mask = kw.get("attention_mask")
+    if mask is None:
+        mask = attention_mask
+    media["masks"] = mask
+    return media
+
+
+def compute_apc_extra_hash(
+    prompt_kwargs: Optional[dict],
+    *,
+    tenant: Optional[str] = None,
+    model: Any = None,
+    processor: Any = None,
+    legacy: bool = False,
+    audio_ref: Any = None,
+    video_ref: Any = None,
+    attention_mask: Any = None,
+    image_ref: Any = None,
+) -> int:
+    """APC ``extra_hash`` from prefix-invariant (default) or legacy media inputs."""
+    kw = prompt_kwargs or {}
+    if tenant is None:
+        tenant = kw.get("_apc_tenant")
+    image_hash = resolve_apc_image_hash(kw, image_ref=image_ref)
+    if legacy:
+        media = apc_legacy_media(
+            kw,
+            audio_ref=audio_ref,
+            video_ref=video_ref,
+            attention_mask=attention_mask,
+        )
+    else:
+        media = apc_prefix_invariant_media(kw, audio_ref=audio_ref, video_ref=video_ref)
+    return semantic_extra_hash(
+        tenant=tenant,
+        image_hash=image_hash,
+        media=media,
+        model=model,
+        processor=processor,
+    )
+
+
+def apc_extra_hash_candidates(
+    prompt_kwargs: Optional[dict],
+    *,
+    tenant: Optional[str] = None,
+    model: Any = None,
+    processor: Any = None,
+    audio_ref: Any = None,
+    video_ref: Any = None,
+    attention_mask: Any = None,
+    image_ref: Any = None,
+) -> Tuple[int, ...]:
+    """Stable hash first, then legacy when they differ (lookup fallback)."""
+    common = {
+        "tenant": tenant,
+        "model": model,
+        "processor": processor,
+        "audio_ref": audio_ref,
+        "video_ref": video_ref,
+        "attention_mask": attention_mask,
+        "image_ref": image_ref,
+    }
+    stable = compute_apc_extra_hash(prompt_kwargs, legacy=False, **common)
+    legacy = compute_apc_extra_hash(prompt_kwargs, legacy=True, **common)
+    if stable == legacy:
+        return (stable,)
+    return (stable, legacy)
+
+
 def semantic_extra_hash(
     *,
     tenant: Optional[str] = None,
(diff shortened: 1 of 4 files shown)


f1yn:
After patch, I can see the `matched_tokens` populated as expected.

```
~/L/dev-ai (main|✚9) $ curl -s http://127.0.0.1:10153/v1/cache/stats | python3 -m json.tool

{
    "block_size": 16,
    "num_blocks": 2048,
    "pool_used": 0,
    "lookups_hit": 15,
    "lookups_miss": 0,
    "matched_tokens": 681162,
    "served_tokens": 0,
    "token_hit_rate": 1.0,
    "evictions": 0,
    "stores": 0,
    "disk_hits": 0,
    "disk_writes": 0,
    "exact_hits": 15,
    "exact_stores": 32,
    "rejects": 0,
    "rejects_by_reason": {},
    "last_reject": null,
    "resident_bytes": 0,
    "enabled": true
}
```

gtonic:
> **Disclosure:** this investigation was done with an AI agent (Claude Code) running on
> my machine — the source tracing, the processor sweep and the first draft of this
> comment. I've reviewed it before posting, and anything that wasn't actually executed is
> marked as such below.

Thanks for this — we hit the same `exact_hits: 0` symptom on the continuous-batching
path and this PR fixes it for us. Measured on an M5 Pro (48 GB) with
`Qwen3.6-35B-A3B-UD-MLX-4bit` at ~7.5k tokens of context, with the salt fixed:

| turn | cache hit | prefill | wall |
|------|-----------|---------|------|
| 1 | 0 % | 3233 ms | 3.58 s |
| 2 | 99.5 % | 153 ms | 0.31 s |
| 3 | 99.4 % | 157 ms | 0.38 s |
| 4 | 99.3 % | 172 ms | 0.25 s |

Branching (replacing the last question instead of appending) hits at 99.4 %. A second,
architecturally different model — `mlx-community/Muse-Glimmer-30B-4bit`, which lands in
`exact` mode via `RotatingKVCache` rather than `ArraysCache` — goes 16318 ms on turn 1 to
396 ms on turn 2+ at 99.7 %.

One thing worth flagging before this lands, because it trades a performance bug for a
potential correctness one on a subset of models.

## `apc_prefix_invariant_media` is still a whitelist

```python
audio = kw.get("input_features")
video = kw.get("pixel_values_videos")
```

`inputs_embeds` — which this PR correctly removes from the default salt — was
*incidentally* covering media that this whitelist doesn't name. Several processors emit
audio under different keys, checked against the installed 0.6.12 tree:

| model | processor emits | matches whitelist? |
|-------|-----------------|--------------------|
| `gemma4_unified` | `input_features` | yes |
| `minicpmo` | `audio_features`, `audio_feature_lens` | **no** |
| `phi4mm` | `input_audio_embeds`, `audio_embed_sizes` | **no** |
| `inkling` | `audio_input_ids` | **no** |

MiniCPM-O traced end to end:

- the processor returns `audio_features` / `audio_feature_lens`
  (`processing_minicpmo.py:666-667`) and never `input_features` — the feature
  extractor's `input_features` is consumed internally at `:546`
- the model reads those keys back and merges audio into `inputs_embeds`
  (`minicpmo.py:391`)
- audio placeholder tokens are derived from raw **sample count**, not content:
  `self._get_audio_placeholder(int(len(all_waveforms[idx])))` (`:555`)

So: same text plus two different audio clips **of the same length** produce identical
`input_ids`, and `audio` in the salt is `None` for both. The `audio_ref` fallback doesn't
rescue it — `ar.py` calls `compute_apc_extra_hash(prompt_kwargs, model=..., processor=...,
legacy=False)` without passing it. `lookup_exact_cache` then matches on both `extra_hash`
and token IDs, and can serve a prefix computed from the *first* clip.

Before this PR that couldn't happen, because `inputs_embeds` made every salt unique — at
the cost of never hitting, which is exactly the bug you're fixing. The legacy candidate
doesn't help either: lookup tries stable first, so a stable-hash collision is served
before legacy is ever consulted.

**How far I took this:** traced from source, *not* reproduced. I have not run MiniCPM-O,
Phi-4-MM or Inkling through the exact-APC path, and I have not confirmed they reach the
batched prefill path at all — if they don't, this is theoretical for now. The multimodal
check I did run was image-based (two different images, byte-identical text, correct warm
hits on repeat, no collision), and images are covered separately via
`resolve_apc_image_hash` / `pixel_values`.

## Suggested direction

Invert the whitelist — fold in everything except what is provably a function of the token
sequence:

```python
_APC_SALT_EXCLUDED_KEYS = frozenset({
    "input_ids", "inputs_embeds", "attention_mask", "position_ids",
    "cache_position", "_apc_image_hash", "_apc_tenant",
})
```

That picks up new processor keys automatically instead of needing a whitelist edit per
model. The exclusion set has to name `attention_mask` explicitly — it also grows with the
prompt, so a fully generic fold over `prompt_kwargs` would reintroduce precisely the bug
this PR fixes. The error direction is asymmetric, which is what makes the inverted form
the safer default: an unnecessary key only costs cache hits, a missing one risks serving
a wrong prefix.

Happy to open a follow-up PR, or to test a revision against the models above.

## Unrelated side note

`APC_EXACT_CACHE_ENTRIES` defaults to `2` and each request stores two snapshots (the
checkpoint at `len - guard` plus the full prompt), so a single unrelated request evicts
the entire cache. Your operator note already suggests raising it — we initially misread
that eviction as the fix not working, so a higher default might save others the same
detour.

f1yn:
@Lazarus-931 why was this closed?

Lazarus-931:
apologies @f1yn for closing without commenting, https://github.com/Blaizzy/mlx-vlm/pull/1713 was already merged and addressed the same issue as this pr!

Issue #1775, opened 2026-08-02
OOM when using DFlash Speculative Decoding


anthonyalayo:
While attempting to benchmark performance improvements between llama.cpp and mlx-vlm, I wanted to use the same benchmarking script for both: https://github.com/ggml-org/llama.cpp/tree/master/tools/server/bench/speed-bench

Where my command was the following:

> (speed-bench) ~/gi/llama.cpp/to/server/bench/speed-bench > python speed_bench.py --url http://127.0.0.1:8080 --bench qualitative --category coding --osl 1024 --concurrency 1 --model mlx-community/Qwen3.6-27B-4bit

There are no logs of memory allocations coming from mlx-vlm, so I don't know for certain, but I have a guess that some sort of concurrency leak is happening which causes my mac to instantly run out of memory and crash.

<img width="3456" height="2112" alt="Image" src="https://github.com/user-attachments/assets/14439e90-e192-44d5-a47e-b36b520f126e" />

Notice the significant size of swap quickly moving up to 25GB. I had 10GB used for my system outside of the mlx-vlm speed test, and my calculations bring me to everything fitting in memory with plenty of room. The equivalent commands for llama.cpp (which show memory allocs in the logs) confirmed that everything should fit:

> HF_HUB_OFFLINE=1 mlx_vlm.server --model mlx-community/Qwen3.6-27B-4bit --draft-model anthonyya/Qwen3.6-27B-DFlash-8bit --draft-kind dflash --max-kv-size 80000 --kv-bits 8 --kv-quant-scheme uniform --enable-thinking --log-level DEBUG

But clearly that isn't happening. I looked for a max parallelization flag similar to llama.cpp and other frameworks:

<img width="691" height="63" alt="Image" src="https://github.com/user-attachments/assets/b9a78362-c8d3-4bf9-804e-b103f51a5b97" />

But I couldn't find any. I also see that there is support for continuous batching, but with no flag to turn this support off, and no clarity as to whether it is the culprit here. Can support for these be introduced? And perhaps also a guide on best practices regarding their usage?

anthonyalayo:
For some more context, this doesn't happen on derivative frameworks such as oMLX. mlx-vlm is the only mlx framework I have seen that supports all forms of kv cache quantization and dflash speculative decoding, so I would really like this to work.

Lazarus-931:
hi @anthonyalayo, thanks for the issue, yes seems strange indeed. i'd say a leak as well but will verify!

[...]

PR #1782, opened 2026-08-03
server: add --max-num-seqs to bound concurrent batch size

Lazarus-931:
## Problem

Continuous batching admits every queued request at once — the request `Queue()` is unbounded and `ResponseGenerator._collect_pending_requests` drains it fully each step — so the running batch (and peak memory) grows with client concurrency with no way to limit it. While this didn't directly cause Blaizzy/mlx-vlm#1775, it's needed because atm, mlx-vlm doesn't have any control over maximum parallel requests.

## Change

Add `--max-num-seqs` (env `MLX_VLM_MAX_NUM_SEQS`). The admission loop now takes at most `max_num_seqs - len(active)` new sequences per step and leaves the rest queued (backpressure), so the running batch never exceeds the cap; overflow is admitted as sequences finish. Defaults to **unbounded** (unchanged behavior).

## Verification

Small model, 8 concurrent requests:

* `--max-num-seqs 2`: prefills happen in **4 waves of 2** and all 8 complete from the queue.
* no cap: all **8 prefill at once**.

`test_speculative` passes (aside from an unrelated pre-existing bf16-tolerance flake).

## Files

* `mlx_vlm/server/cli.py`: `--max-num-seqs` flag → env
* `mlx_vlm/server/generation.py`: `get_max_num_seqs()` + capacity-gated admission


lucasnewman:
@Lazarus-931 Looks reasonable but tests didn't run properly.

lucasnewman (approved):
Is the name a convention used by others or is this new? I think it might be more clear as `--max-continuous-batches` or something more illustrative if it's new. Otherwise LGTM

Lazarus-931:
> Is the name a convention used by others or is this new? I think it might be more clear as `--max-continuous-batches` or something more illustrative if it's new. Otherwise LGTM

it's used in vLLM: 

https://github.com/vllm-project/vllm/blob/66b3c0e61f1e477820212201adf1ed871df7ee98/vllm/engine/arg_utils.py#L1479-L1482


llama.cpp does -np, --parallel N 

can change it, lmk

lucasnewman:
> > Is the name a convention used by others or is this new? I think it might be more clear as `--max-continuous-batches` or something more illustrative if it's new. Otherwise LGTM
> 
> it's used in vLLM:
> 
> https://github.com/vllm-project/vllm/blob/66b3c0e61f1e477820212201adf1ed871df7ee98/vllm/engine/arg_utils.py#L1479-L1482
> 
> llama.cpp does -np, --parallel N
> 
> can change it, lmk

Eh, it's fine if it's already convention, I'm good to merge.

PR #1783, opened 2026-08-03
molmo: fix image-feature scatter and prompt formatting (unusable output)

Lazarus-931:
Molmo-7B-D was either crashing or returning empty output. 

Two causes: the image-feature merge used a wrong numpy index (allocating a 15 GB tensor and corrupting a prompt embedding), and prompts were sent raw instead of wrapped as User: ... Assistant:

I changed the merge to scatter each patch to its correct token position and routed prompts through the processor's chat template. Molmo now answers correctly at ~11 GB peak instead of ~41 GB.

### DIFF
--- a/mlx_vlm/prompt_utils.py
+++ b/mlx_vlm/prompt_utils.py
@@ -102,7 +102,9 @@ MODEL_CONFIG = {
     # Prompt-only models
     "florence2": MessageFormat.PROMPT_ONLY,
     "plamo2vl": MessageFormat.PROMPT_ONLY,
-    "molmo": MessageFormat.PROMPT_ONLY,
+    # molmo is chat-formatted: its processor's template wraps the prompt as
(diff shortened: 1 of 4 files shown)


mlx_vlm/prompt_utils.py
```diff
@@ -102,7 +102,9 @@ class MessageFormat(Enum):
     # Prompt-only models
     "florence2": MessageFormat.PROMPT_ONLY,
     "plamo2vl": MessageFormat.PROMPT_ONLY,
-    "molmo": MessageFormat.PROMPT_ONLY,
+    # molmo is chat-formatted: its processor's template wraps the prompt as
+    # "User: ... Assistant:", which the raw-prompt path used to skip.
```
Blaizzy:
```suggestion
```

Blaizzy (approved):
LGTM!

PR #1791, opened 2026-08-04
fix(quant): honor per-layer quantization overrides for wrapped text models

Lazarus-931:
for LFM models, specifically  8B-A1B, fails becuase load does honor per-module quant overrides


lucasnewman (approved):
🙏

PR #1796, opened 2026-08-05
feat(server): accept structured transcription hotwords

### LINKED ISSUE #1781: Server: add backend-aware hotword and custom vocabulary support for audio transcriptions
## Summary

Add backend-aware custom vocabulary / hotword support to the OpenAI-compatible `POST /v1/audio/transcriptions` endpoint.

The endpoint currently parses `prompt` and `context`, but it does not accept a structured hotword list, and its generic prompt forwarding can silently send the value under the wrong keyword for several MLX Audio ASR backends.

## Current behavior

`AudioTranscriptionRequest` exposes:

- `prompt`
- `context`
- language and decoding fields

In `_build_stt_generate_kwargs`, `prompt` is removed from the request dump and then mapped to `context` or `text` based on `_accepts_parameter`.

Because `_accepts_parameter` considers a model with `**kwargs` to accept every parameter name, this can produce false positives and silently route the prompt incorrectly.

Examples against mlx-vlm 0.6.8 and mlx-audio 0.4.7:

| Backend | Model-level vocabulary/context hook | Current server result |
|---|---|---|
| Qwen3-ASR | `system_prompt` | `prompt` becomes `context`; Qwen receives it in `**kwargs` and discards it |
| Whisper | `initial_prompt` | Not mapped; `prompt` becomes `context` inside decoding options |
| MOSS Transcribe/Diarize | `prompt` | Direct `prompt` is removed and replaced with `context`, which is ignored |
| FunASR Nano | `hotwords` or `context` | `context` works, but the endpoint cannot accept a structured `hotwords` list |
| VibeVoice ASR | `context` | Works through the existing `context` field |
| Parakeet / Cohere ASR | No current hotword hook | Values may be accepted through `**kwargs` and then ignored |

For example:

```bash
curl http://localhost:8080/v1/audio/transcriptions \
  -F model=mlx-community/Qwen3-ASR-0.6B-8bit \
  -F prompt="Prefer the spelling: Nativ" \
  -F file=@sample.wav
```

The request succeeds, but the vocabulary hint does not reach Qwen3-ASR's `system_prompt`.

## Proposed behavior

Provide a stable transcription vocabulary API and route it explicitly by backend capability.

Possible request fields:

```
prompt: Optional[str]
context: Optional[str]
hotwords: Optional[list[str]]
```

Suggested backend mapping:

- Qwen3-ASR → build a `system_prompt` containing the vocabulary hints
- Whisper → `initial_prompt`
- MOSS Transcribe/Diarize → `prompt`
- FunASR Nano → `hotwords` directly
- VibeVoice ASR → `context`
- Unsupported backends → either clearly reject the option with HTTP 400 or expose capability metadata, rather than silently ignoring it

The OpenAI-compatible `prompt` field could remain the portable entry point, while `hotwords` is an mlx-vlm extension for callers that need a structured list.

## Implementation considerations

- Prefer explicit named-parameter detection before falling back to `**kwargs`.
- Do not treat `**kwargs` alone as evidence that a backend semantically supports `context`.
- Keep the original `prompt` available for models that expose a direct prompt parameter.
- Add endpoint tests covering the generated kwargs for Qwen3-ASR, Whisper, MOSS, FunASR Nano, and VibeVoice.
- Add a test that unsupported backends do not silently consume and discard vocabulary fields.

## Expected outcome

Clients can provide product names, people’s names, acronyms, and domain terminology once through the transcription endpoint, and supported ASR backends receive the hint using their native parameter.

LauraGPT:
Closes #1781.

## Summary

- accept repeated hotwords / hotwords[] multipart fields as a structured vocabulary list
- pass hotwords through the existing request-model and generate-signature filtering path
- preserve existing prompt and context behavior, including backends that accept multiple hint fields
- document the native Fun-ASR-Nano hotwords request

The maintainer-requested revision removes cross-backend parameter guessing, model-family routing, and artificial mutual exclusion. A backend receives only the native fields its generate signature accepts.

## Validation

- python -m pytest -q mlx_vlm/tests/test_server.py mlx_vlm/tests/test_server_audio.py: 276 passed
- focused red/green regression: a backend accepting both context and hotwords now receives both
- Ruff 0.14.10 check and format check: passed
- git diff --check against current main: passed
- merged current main at 321514d; PR remains three files and 72 added lines

All three review threads were addressed in place and resolved.


mlx_vlm/server/audio.py
```diff
@@ -587,6 +588,64 @@ def _build_stt_generate_kwargs(
...
+    accepted,
+    request: AudioTranscriptionRequest,
+) -> Dict[str, Any]:
+    hotwords = [word.strip() for word in request.hotwords or [] if word.strip()]
+    supplied = sum(bool(value) for value in (request.prompt, request.context, hotwords))
+    if supplied > 1:
+        raise HTTPException(
+            status_code=400,
+            detail="Pass only one of prompt, context, or hotwords.",
+        )
```
Blaizzy: This is interesting, could you explain why this logic is in place?

I suspect it's too brittle, and models that take two will fail unexpectedly.

mlx_vlm/server/audio.py
```diff
@@ -587,6 +588,64 @@ def _build_stt_generate_kwargs(
...
+def _unsupported_stt_vocabulary() -> HTTPException:
+    return HTTPException(
+        status_code=400,
+        detail=(
+            "Selected model does not explicitly support transcription vocabulary "
+            "through prompt, context, or hotwords."
+        ),
+    )
+
+
```
Blaizzy: Lets revert this, it's not generalizable enough.

mlx_vlm/server/audio.py
```diff
@@ -558,16 +563,12 @@ def _build_stt_generate_kwargs(
...
-    prompt = request.prompt
     accepted = _callable_parameters(model.generate)
-
-    if prompt:
-        if _accepts_parameter(accepted, "context"):
-            kwargs.setdefault("context", prompt)
-        elif _accepts_parameter(accepted, "text"):
-            kwargs.setdefault("text", prompt)
+    for name, value in _build_stt_vocabulary_kwargs(accepted, request).items():
+        kwargs.setdefault(name, value)
```
Blaizzy: Revert this

Blaizzy (commented):
Thanks @LauraGPT, I like the direction!

I left a few nits.

mlx_vlm/server/audio.py
LauraGPT: Removed the mutual-exclusion rule. A backend may now receive context and hotwords together when its generate signature accepts both; the new regression covers that combination.

mlx_vlm/server/audio.py
LauraGPT: Removed the cross-backend prompt/hotword routing helper and its model-specific parameter mapping. Hotwords now remain a native generate kwarg instead of being rewritten into another parameter.

mlx_vlm/server/audio.py
LauraGPT: Restored the original prompt handling in _build_stt_generate_kwargs. The only production additions left are the hotwords request field and repeated multipart parsing; the existing kwargs filter passes it through to backends that accept it.

LauraGPT:
Review feedback is addressed at current head e8feb0cd. The brittle cross-backend routing and mutual-exclusion behavior are removed; the PR now only adds native repeated hotwords parsing and pass-through. It is synced with current main, all three threads are resolved, and the current server suites pass 276/276 with Ruff and diff checks green. @Blaizzy, ready for re-review.

Lazarus-931:
https://github.com/Blaizzy/mlx-audio/pull/885 is up, once merged, this should be gtg!

thx for ur patience @LauraGPT!

LauraGPT:
The prerequisite Blaizzy/mlx-audio#885 has now merged as 6546e953908800219454d4483592c1db9d8871a8. This PR remains mergeable, and its test check is green at exact head 12c16b9cf668b768e57a0b43e5ff0ede23ddfe6a; no branch change is needed for the dependency. Ready for the final maintainer pass.

PR #1183, opened 2026-05-14
Add Laguna support

Blaizzy:
## Summary
- Add Laguna as a text-only language model package split into config, model, and language modules.
- Register Laguna prompt formatting and preserve tool-call metadata through prompt utilities.
- Normalize JSON-string tool-call arguments before template rendering.
- Add coverage for Laguna model loading metadata, prompt formatting, and tool-call prompt handling.

## Validation
- `uv run pytest mlx_vlm/tests/test_models.py mlx_vlm/tests/test_prompt_utils.py mlx_vlm/tests/test_utils.py -q`
- `uv run mlx_vlm.generate --model poolside/Laguna-XS.2 --prompt "Make a program to find pi" --max-tokens 1200 --temperature 0`
- Started `mlx_vlm.server` with `poolside/Laguna-XS.2` and verified one single `/v1/chat/completions` request plus two concurrent chat completion requests returned parsed `tool_calls` with HTTP 200 responses.

PR #1802, opened 2026-08-05
Add dedicated Poolside Laguna tool parser

IronBCC:
## Summary

- add a dedicated `poolside_v1` parser for Laguna's `<tool_call><arg_key>...` format
- register Poolside markers before the generic GLM parser
- add parser inference and server-response regression coverage

## Root cause

Laguna's template contains `<arg_key>` markers, so the existing registry classified it as `glm47`. The dedicated parser preserves declared string arguments and decodes non-string JSON/literal values according to the tool schema.

This follows the Laguna support path validated by PR #1183, while making the parser selection explicit.

## Validation

- direct parser assertions with the bundled Python 3.11 runtime: pass
- `mlx_vlm.server.responses_state.process_tool_calls` integration check: pass
- `compileall`: pass
- `git diff --check`: pass
- full pytest collection was unavailable locally because the bundled environment lacks `mlx_audio`

LM Studio 0.4.20's native endpoint was also tested separately; it still bypasses the Python parser and returns raw Poolside markup. That is an LM Studio response-layer limitation, not a failure of this MLX-VLM server parser.


lucasnewman (approved):
@IronBCC Thanks!

lucasnewman:
@IronBCC Can you [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits and force push so we can clear the merge requirements?

Lazarus-931:
hey @IronBCC, thanks for the pr!

tool calling is being redesigned, and a proper fix that supports glm as well is up in #2208

closing this :)

PR #1803, opened 2026-08-05
Keep family-specific STT defaults out of Whisper decode options

Lazarus-931:
The transcription endpoint always injected Parakeet-family request defaults (`frame_threshold`, `max_tokens`, `prefill_step_size`) into every STT call, and Whisper's `**decode_options` catch-all bypasses the signature filter, so every Whisper transcription crashed constructing `DecodingOptions`. Defaults now apply only when the model declares the parameter (client-provided values still pass through), fixing Blaizzy/nativ#212 — verified live with `openai/whisper-large-v3-turbo`: the reported `TypeError` on current main, a successful transcription with this change.

### DIFF
--- a/mlx_vlm/server/audio.py
+++ b/mlx_vlm/server/audio.py
@@ -393,23 +393,34 @@ async def _parse_transcription_request(
     if not timestamp_granularities:
         timestamp_granularities = _form_list(form, "timestamp_granularities[]")
 
-    payload = AudioTranscriptionRequest(
-        model=model,
-        language=_clean_form_value(form.get("language")),
-        prompt=_clean_form_value(form.get("prompt")),
-        temperature=_form_float(form.get("temperature")),
-        response_format=_clean_form_value(form.get("response_format")) or "json",
-        verbose=_form_bool(form.get("verbose"), default=False),
-        max_tokens=_form_int(form.get("max_tokens"), default=1024),
-        chunk_duration=_form_float(form.get("chunk_duration"), default=30.0),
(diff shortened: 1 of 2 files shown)


lucasnewman (approved):
This is ok for now, but seems more like a Whisper model issue -- my feeling is that unknown kwargs should be ignored at the model level in general so we don't have to modify the server for every potential downstream model's requirements.

I think it can be a follow-up though and fixing the immediate issue is more important.

Lazarus-931:
> This is ok for now, but seems more like a Whisper model issue -- my feeling is that unknown kwargs should be ignored at the model level in general so we don't have to modify the server for every potential downstream model's requirements.
> 
> I think it can be a follow-up though and fixing the immediate issue is more important.

yes, i agree, i've opened a pr upstream.

https://github.com/Blaizzy/mlx-audio/pull/873

Can we instead merge that, bump the floor, and then close this?

easier that reverting ltr

lucasnewman:
> > This is ok for now, but seems more like a Whisper model issue -- my feeling is that unknown kwargs should be ignored at the model level in general so we don't have to modify the server for every potential downstream model's requirements.
> > I think it can be a follow-up though and fixing the immediate issue is more important.
> 
> yes, i agree, i've opened a pr upstream.
> 
> [Blaizzy/mlx-audio#873](https://github.com/Blaizzy/mlx-audio/pull/873)
> 
> Can we instead merge that, bump the floor, and then close this?
> 
> easier that reverting ltr

Yes, definitely, I just merged that one.

PR #1761, opened 2026-07-31
Fix image-valued Responses function output handling

Rajeev-SG:
## Disclosure

This pull request was prepared and submitted by **OpenAI Codex** at Rajeev-SG's direction.

## Summary

Preserve image-valued function output as visual input in the Responses API adapter instead of serializing the image data as tool-message text.

Closes #1760.

## API contract

The official OpenAI Responses API reference defines `function_call_output.output` as either a string or an array of `ResponseInputText`, `ResponseInputImage`, or `ResponseInputFile` items. The image-valued tool output sent by Codex is therefore valid Responses API input; the adapter must preserve the image as visual input rather than stringify it.

Reference: https://developers.openai.com/api/reference/resources/responses/methods/create

## Root cause

`_append_response_item_to_prompt()` converted every non-string function output with `json.dumps()`. When a tool such as `view_image` returned an `input_image` content block, the complete base64 data URL became ordinary text in the next prompt.

The change now:

- extracts `input_image` and `image_url` blocks into the visual input list;
- preserves accompanying text blocks in the tool message;
- keeps unknown content blocks as JSON for backward compatibility;
- uses `[Image output attached]` when an output contains only images.

## Failure evidence

This was reproduced in a real Codex + Qwen3.6 + MLX-VLM conversation:

- Codex-side conversation count before the follow-up: **56,861 tokens**
- image data URL returned by the tool: **391,202 characters**
- MLX-VLM prompt count after JSON serialization: **305,961 tokens**
- request including output allowance: **310,057 tokens**
- configured KV limit: **163,840 tokens**
- result before the fix: HTTP 400

The initial image request succeeded. The failure occurred on the automatic follow-up after the image-valued function output was appended.

### Regression coverage

- Added converter tests for image-only and mixed text/image function outputs.
- Added a public `/responses` endpoint test proving the image reaches the generation call as visual input and the tool message does not contain base64 data.
- Full server suite: **253 passed**
- Responses-state tests: **2 passed**
- Pre-commit checks: **black, isort, and autoflake passed**

### Real user-flow proof

An equivalent local routing fix was exercised against the installed, otherwise unmodified MLX-VLM service:

1. Started a fresh Codex task using local Qwen3.6.
2. Attached the original failure screenshot.
3. Required the model to call `view_image`.
4. Confirmed the returned image was converted to visual input.
5. Confirmed the follow-up `/v1/responses` request returned HTTP 200.
6. Confirmed the model completed with `IMAGE_TOOL_LOOP_OK`.

[...]

PR #1806, opened 2026-08-05
Fix /v1/responses image reference output

lucasnewman:
Based on https://github.com/Blaizzy/mlx-vlm/pull/1761, but expanded to cover properly inserting outputs associated with specific turns.


Blaizzy (approved):
LGTM, thanks!

PR #1808, opened 2026-08-05
Add support for Minimax-H3

lucasnewman:
This also adds support for `--output-modality video` / `mlx_vlm.generate_video` since this is the first video output model we support.


Lazarus-931 (approved):
lgtm!

PR #1811, opened 2026-08-07
Split unterminated reasoning correctly when the prompt opens the thinking block 

Lazarus-931:
`_split_thinking` gains a `starts_in_thinking` parameter, and the OpenAI non-streaming call site now passes the same prompt_has_open_thinking(formatted_prompt, …) signal that streaming already uses. When a model's chat template opens the thinking block in the prompt (so no <think> ever appears in generated text) and the block never closes, the output is now correctly returned as reasoning instead of being mislabeled as the answer.

solves https://github.com/Blaizzy/nativ/issues/223


lucasnewman (approved):
LGTM

PR #1813, opened 2026-08-08
moe-offloading

Lazarus-931:
Pages routed MoE experts from an on-disk store instead of holding every expert resident in RAM, so a checkpoint bigger than a Mac's memory can run at all — only the experts a token actually routes to get loaded. `mlx_vlm.load()` detects a repacked checkpoint and swaps in the offloaded layers automatically, no caller-visible API change.

## Small-scale: fits models that otherwise wouldn't load at all

Measured on an M4 mini (16GB RAM), 80-token generation:

| Model | Peak mem | tok/s |
|---|---|---|
| OLMoE-1B-7B-0125-Instruct-4bit | 3.5 GB | 41.2 |
| DeepSeek-V2-Lite-Chat-4bit-mlx | 8.4 GB | 18.9 |
| gpt-oss-20b-MXFP4-Q8 | 9.8 GB | 20.3 |
| Qwen1.5-MoE-A2.7B-Chat-4bit | 8.4 GB | 25.5 |
| ERNIE-4.5-21B-A3B-PT-4bit | 11.1 GB | 17.2 |
| Qwen3-30B-A3B-4bit (128 experts) | 10.6 GB | 14.1 |

ERNIE-21B and Qwen3-30B are the point: both fit under 16GB offloaded, neither would load resident on this machine. Output bit-identical to resident (270/270 tests).

Decode tok/s across context length (512–32K tokens, same M4 mini):

| Context | OLMoE (64 experts) | Qwen3-30B-A3B (128 experts) |
|---|---|---|
| 512 | 66.6 | 23.4 |
| 1024 | 63.6 | 23.4 |
| 2048 | 60.0 | 20.5 |
| 4096 | 46.7 | 20.2 |
| 8192 | 38.8 | OOM |
| 16384 | 27.8 | — |
| 32768 | 18.0 | — |

Qwen3-30B-A3B OOMs between 4K and 8K tokens on this 16GB machine — a real ceiling, not something this PR regresses (unmodified per-layer-eval code hits the same wall at the same length).

## Extreme scale: beats a dedicated competing engine on the exact case offloading exists for

DeepSeek-V4-Flash-4bit (141GB) doesn't fit in 128GB RAM either — the same situation OLMoE/Qwen3-30B hit on a 16GB mini, just an order of magnitude bigger. Compared against DwarfStar (a from-scratch C/Metal engine built specifically for this, with its own SSD-streaming expert cache), matched to DwarfStar's own reported budget (71.06 GiB), fresh process per data point, same prompts, same M5 Max:

| Context | mlx-vlm prefill | DwarfStar prefill | mlx-vlm decode | DwarfStar decode |
|---|---|---|---|---|
| 512 | 100.3 tok/s | 63.8 | 10.6 tok/s | 6.5 |
| 2048 | 280.9 tok/s | 155.2 | 10.0 tok/s | 6.2 |
| 4096 | 286.0 tok/s | 149.8 | 10.8 tok/s | 5.8 |
| 8192 | 304.2 tok/s | 261.9 | 10.0 tok/s | 5.5 |

At ctx=8192, peak memory is 65.9 GB vs DwarfStar's own 79.7 GB planned figure. mlx-vlm wins prefill, decode, and memory at every tested context length under this matched-budget comparison.

Two approaches were tried and rejected with real measurements before landing on the two fixes below: a `gather_qmm`-fused batched matmul (~9x slower than the per-expert loop even with zero tensor-assembly cost — the primitive itself is slower at this shape) and OS-level background layer prefetch (no measurable benefit once genuinely cold-tested).

## Reliability and observability (this PR, prior to the perf fixes below)

- Byte-budgeted LRU eviction in `ExpertStore` — DeepSeek V4's 256-experts/layer scale otherwise grows resident set unboundedly and OOMs.
- Refuses an unchunked prefill against an offloaded model instead of silently pinning every expert across every layer in one graph until eval.
- `--expert-cache-gb` CLI flag / `expert_cache_gb` kwarg to size the resident-expert budget explicitly.
- `/v1/moe-offload/stats` endpoint (hits/misses/evictions/hit rate) for observability and tuning.
- KV-cache headroom reservation: the auto budget now measures actual per-token KV growth (via two real tiny forward passes through the model's own cache) and reserves for it, so a large `max_kv_size` doesn't starve the expert budget.
- Also fixes two bugs found in review: fused switch layers (Laguna/MiniMax-M3-VL) failed to load through the real `mlx_vlm.load()` path, and a missing/corrupted expert file silently produced garbage for one layer instead of erroring. Both covered by regression tests.

## Two performance fixes that produced the extreme-scale numbers above

1. **Dropped `gc.collect()` from eviction.** It was called on every eviction batch and was 63.6% of wall-clock time in a real run (~10ms × 3991 calls) — a full Python GC cycle over a heap of thousands of live large arrays, not real work. Popping an expert's keys already drops the last reference, so CPython's own refcounting frees it immediately; `gc.collect()` only matters for reference cycles, which don't exist in this structure. `mx.clear_cache()` (MLX's own allocator cache) stays.
2. **Bulk-load a layer's experts when most of it is touched anyway.** A real prefill chunk routes over virtually all of `num_experts` per layer regardless of top-k (confirmed at every one of 7 real model configs tested, from Llama4's 16 experts/top-1 to Kimi-K2's 384/top-8) — so per-expert LRU caching just thrashes there, evicting/reloading layers competing for a shared budget. `ExpertStore.get_all()` loads the whole layer file once and bypasses per-expert bookkeeping when a call's unique touched experts exceed half of `num_experts` — a routing-behavior threshold, not a model-specific one; decode's small, selective touches are unaffected.

Both are pure `ExpertStore`/`OffloadedSwitchGLU` bookkeeping — no DeepSeek-specific logic, verified bit-identical to the unoptimized path via a synthetic harness swept across all 7 real model configs above, and the full test suite passes unchanged.

## Cross-model coverage fix

`repack()` classifies expert tensors via a generic naming-convention regex (`plan()`), which matches 5 of 8 scoped models' real upstream checkpoints directly (DeepSeek, Kimi-K2, GLM-4.5-Air, Qwen3-30B-A3B, ERNIE-4.5-21B) but not 3 that use different naming (MiniMax-M2's Mixtral-style `w1/w2/w3`, gpt-oss's native fused MXFP4 layout, Llama4's headerless fused tensor). Rather than adding per-vendor regexes to `plan()`, `repack()` now falls back to that model's own `sanitize()` — the same normalization `mlx_vlm.load()` already applies for regular loading — only when the fast path finds zero experts, so all 5 already-working models take the exact same code path as before. Verified end-to-end against a real tiny `minimax.Model` round-tripped through raw Mixtral-style naming → `repack()` → `load_model()` (0.09% relative logit diff, new regression test), and mechanically for gpt-oss/Llama4's real naming conventions. Caught two real bugs along the way: a `sanitize()` that gates its whole MoE-restructuring block on a "layer 0 present" global probe (would have silently skipped every other layer), and `load_model()`'s actual two-step sanitize chain (`Model.sanitize()` then `LanguageModel.sanitize()`) not being fully replicated.

### Reproduction

```bash
python3 -c "from mlx_vlm.moe_offload import repack; repack('<checkpoint>', '<offload-dir>')"
python3 -c "
model, tok = load('<offload-dir>')
print(generate(model, tok, 'Explain how photosynthesis works in three sentences.', max_tokens=80))
"
```


lucasnewman:
@Lazarus-931 Overall this looks really cool. Should we add a cli entrypoint for repacking a checkpoint? Also it would be great to update the docs to cover usage.

Lazarus-931:
> @Lazarus-931 Overall this looks really cool. Should we add a cli entrypoint for repacking a checkpoint? Also it would be great to update the docs to cover usage.

will do

lucasnewman:
@Lazarus-931 Failing tests, otherwise looks good

lucasnewman:
@Lazarus-931 Needs signed commits to merge

lucasnewman (approved):

PR #1814, opened 2026-08-08
Keep serving when an optional model preload fails

Lazarus-931:
A failed optional model preload (image-generation, TTS, speech-to-text, or embedding) currently raises out of the ASGI lifespan and aborts server startup entirely, so one unloadable speech model takes down a server whose language model had already loaded, this guards those four preloads to log and continue, records each failure on runtime.preload_failures and exposes it in the runtime snapshot, and leaves the language-model preload fatal since that is the core capability. 

for https://github.com/Blaizzy/nativ/issues/226


lucasnewman (approved):
LGTM

PR #1817, opened 2026-08-08
Add Nemotron-VoiceChat-11B full-duplex speech model

lucasnewman:
[Original Model Card](https://huggingface.co/nvidia/NVIDIA-NemotronLabs-VoiceChat-11B)

This also adds support for the `/v1/realtime` endpoint in the server to allow a streaming full-duplex session. I added an example because it's not completely trivial to setup with live mic/speaker integration correctly.

The 8-bit and 4-bit model variants can run in realtime on my M5 Max, so this is right at the edge of usable but it's still early days for full duplex models :)

Note this requires a new release of `mlx-audio` which implements/updates some of the codec and ASR model components needed.


Lazarus-931 (approved):
nice, lgtm!

small CI issue!

lucasnewman:
@Lazarus-931 Yeah, we need to do an `mlx-audio` release and then it will be clear.

mlx_vlm/models/nemotron_voicechat/convert.py
```diff
@@ -0,0 +1,300 @@
...
+
+import argparse
+import json
+import os
+import shutil
+from pathlib import Path
+
+from huggingface_hub import snapshot_download
+
+OFFICIAL_REVISION = "c5d3b70183b6bb9d7553590e111b05685049751c"
```
Blaizzy: Do we need the revision default pinned?

mlx_vlm/models/nemotron_voicechat/convert.py
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    tts_source = speech["tts_config"]
+    tts_backbone = tts_source["backbone_config"]
+    character = tts_source["cas_config"]["backbone_config"]["encoder"]
+    mog = tts_source["mog_head_config"]
+    codec = speech["codec_config"]
+
+    config = {
+        "mlx_runtime_config_version": RUNTIME_CONFIG_VERSION,
```
Blaizzy: Do we need this?

mlx_vlm/server/app.py
```diff
@@ -59,7 +61,7 @@ def _server_api_key() -> Optional[str]:
     return key if key else None
 
 
-def _require_management_api_key(request: Request) -> None:
+def _require_management_api_key(request: HTTPConnection) -> None:
```
Blaizzy, line 68: This seems to be a false positive change and a potential regression

Blaizzy (commented):
LGTM,

Just got a couple nits👌🏽

Blaizzy:
Will make a new mlx-audio release tomorrow.

Let's land all the important pending changes on that side before it

mlx_vlm/server/app.py
lucasnewman: This is actually a valid change since we can either have a WebSocket or a Request come through now. Both implement `.headers` so it won't regress anything.

mlx_vlm/models/nemotron_voicechat/convert.py
lucasnewman: Nope, I removed this.

mlx_vlm/models/nemotron_voicechat/convert.py
lucasnewman: Removed.

Lazarus-931 (approved):
lgtm!

PR #1821, opened 2026-08-09
Assert sanitize() idempotency in tests and document the invariant

danmackinlay:
Follow-up to #1718 ([comment](https://github.com/Blaizzy/mlx-vlm/issues/1718#issuecomment-5232128770)).

Since #1498, `load_model` calls `sanitize()` on every checkpoint, including ones already converted to MLX layout. So every `sanitize()` now has to detect for itself whether its input still needs converting. That requirement isn't written down anywhere and isn't tested, and the bugs it produces have been fixed one architecture at a time — #1521, then the sanitize fixes in 0.6.7–0.6.9.

This doesn't touch library code. It just makes the invariant explicit and checkable.

### What's here

- **`mlx_vlm/tests/sanitize_invariants.py`** (new) — `assert_sanitize_idempotent(model, hf_weights)`. Runs `sanitize()`, runs it again on its own output, and asserts nothing moved. Reports shape *and* value changes, since the value-changing case (a norm shift, a scale rewrite) is the one that loads fine and generates garbage.
- **`mlx_vlm/tests/test_models.py`** — three tests already assert this per-architecture (`:8415`, `:8832`, `:8867`); refactored onto the helper. No new fixtures, no new model configs.
- **`CONTRIBUTING.md`** — states the invariant next to the existing "add a model test" instruction, with a table of the conversion classes `check_array_shape` doesn't cover (norm-weight shifts, expert/qkv/gate_up splits, quant-scale rewrites, MTP shard filtering) and how to detect each. Also corrects the test paths in that section, which still pointed at `src/tests/`.

### Why the helper rather than per-architecture fixes

I couldn't find a currently-broken `sanitize()` at HEAD. What I did find is that several are idempotent only incidentally — an earlier step consumes the key a later guard tests:

- `qwen3_next/language.py:410` — early return keys on `model.layers.0.mlp.experts.0.up_proj.weight`, which line 422 `pop`s while stacking experts. That's what stops the unguarded `v + 1.0` norm shift at line 439 being applied twice.
- `gpt2/language.py` — transposes keyed on unprefixed `h.{i}.…`, and the `model.` prefixing runs afterwards.
- `hunyuan/language.py` — gated on `gate_and_up_proj`, which the first pass consumes.

All correct today, but correct as a side effect of unrelated restructuring rather than by assertion — so reordering breaks them, and in the norm-shift case it breaks them silently.

### Verification

Tests pass under both runners. Mutation-tested to confirm the helper has teeth — removing the `qwen3_omni_moe` audio guard reproduces the 0.6.9 failure signature:

```
AudioModel: sanitize() changed 'thinker.audio_tower.conv2d1.weight' on a second
pass -- shape (4, 3, 3, 1) became shape (4, 3, 1, 3). sanitize() must detect
already-converted weights and leave them alone.
```

Guard restored, green again. `black` clean.

Unrelated: `TestAWQ::test_fold_into_linear_is_identity` and `test_fold_into_norm_is_identity` fail for me on a clean checkout of `main` too, so they're not from this branch.

A parametrised sweep over every architecture would be better than three call sites, but each one needs a hand-built config and a plausible torch-layout weight dict, so I haven't attempted it. Happy to extend the helper to more architectures if you'd like it in this PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,13 +31,45 @@ To determine the model layer names, we suggest either:
 - Look at the names of the weights by inspecting `model.safetensors.index.json`
   in the Hugging Face repo.
 
-Additionally, add a test for the new model type to the [model
-tests](https://github.com/Blaizzy/mlx-vlm/tree/main/src/tests/test_models.py).
+If the model needs a `sanitize()` to convert weights from the PyTorch layout,
+note that `load_model` calls it on **every** checkpoint, including ones already
+converted to MLX layout. `sanitize()` must therefore be idempotent: running it
+on its own output has to be a no-op.
 
-From the `src/` directory, you can run the tests with:
+For conv weight transposes, `check_array_shape` is the established guard. It
+does not cover the other conversions a `sanitize()` may perform, and each needs
+its own "has this already been done?" check:
+
+| conversion | detect it by |
+|---|---|
+| conv transpose | target layout (`check_array_shape`, or compare against `self.<conv>.weight.shape`) |
+| norm-weight shift (`v + 1.0`) | a marker that only an unconverted checkpoint carries; `v.ndim == 1` is **not** one, since it holds before and after |
+| expert / qkv / gate_up splits and merges | absence of the source key |
+| quantization-scale rewrites | the scale dtype or grouping already being the MLX one |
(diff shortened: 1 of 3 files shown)


lucasnewman (approved):
Thanks!

PR #1822, opened 2026-08-09
Fix TurboQuant batch decode, MTP KV quantization, and attention sinks

### LINKED ISSUE #1016: prefill_attention always returns None after #909 — silent fallback to full dequantize
## Summary

After [PR #909](https://github.com/Blaizzy/mlx-vlm/pull/909) dropped `_TurboQuantProdCodec` in favor of MSE-only, `prefill_attention()` **always returns `None`** during prefill (L>1). The fallback in `models/base.py:212-221` then calls `cache.dequantize()`, decompressing the **entire** KV cache to fp16 on every attention call during prefill. This negates TurboQuant's memory savings during prefill and causes activation memory to scale linearly with context length.

This doesn't surface in mlx-vlm's own generate flow (which uses post-conversion: prefill with fp16 KVCache, convert to TQ after), but it affects any downstream project that creates TQ caches *before* prefill (on-the-fly quantization).

## Root Cause

`prefill_attention()` at [turboquant.py:5087-5092](https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/turboquant.py) checks:

```python
if not (
    isinstance(self.key_codec, _TurboQuantProdCodec)
    and isinstance(self.value_codec, _TurboQuantMSECodec)
    ...
):
    return None
```

But `_ensure_codecs()` (line 4829) builds both codecs with `mode="mse"` → creates `_TurboQuantMSECodec`. Since ProdCodec was intentionally removed in #909, this check **always fails** and `prefill_attention` is dead code.

The fallback in `models/base.py`:
```python
# lines 212-221
dequantized_keys, dequantized_values = cache.dequantize()
return mx.fast.scaled_dot_product_attention(
    queries, dequantized_keys.astype(queries.dtype),
    dequantized_values.astype(queries.dtype), scale=scale, mask=mask,
)
```

At 128K context on gemma-4-31B, this materializes ~31 GB of activation memory per attention call during prefill (per [PR #939 benchmarks](https://github.com/Blaizzy/mlx-vlm/pull/939)).

## How We Discovered This

We maintain [omlx](https://github.com/jundot/omlx), which uses a continuous-batching scheduler with on-the-fly TQ quantization — empty `TurboQuantKVCache` is created *before* prefill so tokens are quantized during `update_and_fetch()`. This means TQ caches are active during prefill, triggering the dequantize fallback on every chunk.

We traced the call chain while investigating why peak memory savings were near zero ([omlx PR #717](https://github.com/jundot/omlx/pull/717), [omlx issue #661](https://github.com/jundot/omlx/issues/661)).

For reference, llama.cpp and vLLM both avoid this by design — their attention kernels read quantized cache directly via fused kernels, and KV caches are allocated in quantized format from the start.

### Option A: Wire `quantized_attention()` as fallback (bridge fix)

`quantized_attention()` ([turboquant.py:4976](https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/turboquant.py)) already exists, works with MSECodec, and processes KV in 2048-token chunks via codebook lookups. It's just never called from the SDPA path. A 3-line change in `models/base.py`:

```python
# Replace dequantize fallback with chunked quantized attention
result = cache.prefill_attention(queries, scale=scale, mask=mask)
if result is not None:
    return result
return cache.quantized_attention(queries, scale=scale, mask=mask)
```

Pros: Minimal change, immediately bounds activation memory to chunk-sized intermediates.
Cons: Python-level loops, slower than fused Metal for short-medium prompts.

### Option B: Land PR #939 (proper fix)

[PR #939](https://github.com/Blaizzy/mlx-vlm/pull/939) rewrites `prefill_attention` with flash-style tiled online softmax using existing Metal kernels. Bounds activation at ~512 MB constant regardless of context length.

Pros: Proper solution, near-optimal performance.
Cons: Larger change (453+/1816-), still open.

These aren't mutually exclusive — Option A could land as a bridge while Option B is finalized.

## Related Issues & PRs

**mlx-vlm:**
- [#909](https://github.com/Blaizzy/mlx-vlm/pull/909) (merged) — Dropped ProdCodec, MSE-only. Caused `prefill_attention` to become dead code.
- [#939](https://github.com/Blaizzy/mlx-vlm/pull/939) (open) — Flash-style tiled TBQ prefill. The proper fix.
- [#973](https://github.com/Blaizzy/mlx-vlm/pull/973) (open) — Rotation-basis mismatch on power-of-two head dims.
- [#967](https://github.com/Blaizzy/mlx-vlm/pull/967) (merged) — Race condition in fused fast-quantize Metal kernels.
- [#858](https://github.com/Blaizzy/mlx-vlm/pull/858) (merged) — Original TurboQuant implementation.

**mlx-lm:**
- [#1060](https://github.com/ml-explore/mlx-lm/issues/1060) — TurboQuant discussion. TheTom confirmed MSE-only better. Asymmetric K/V mandatory for dense models.

**omlx (downstream, where we found this):**
- [#717](https://github.com/jundot/omlx/pull/717) — Wire up TQ + skip_last + on-the-fly quantization
- [#661](https://github.com/jundot/omlx/issues/661) — "TurboQuant has zero effect on memory" (root cause: dead code + this fallback)
- [#705](https://github.com/jundot/omlx/issues/705) — Sub-8-bit TQ ~8x slower (possibly related)

YuhuaBillChen:
Three defects in the quantized-KV path, all only reachable with `--kv-bits`
set. Each is independent; they surfaced in sequence because fixing the first
exposed the next.

## 1. The batch cache had no attention path of its own

`BatchTurboQuantKVCache` was served by a single branch in `models/base.py`
that dequantized the whole cache and handed it to
`mx.fast.scaled_dot_product_attention`:

```python
if isinstance(cache, BatchTurboQuantKVCache):
    dequantized_keys, dequantized_values = cache.dequantize(keys, values)
    return mx.fast.scaled_dot_product_attention(...)
```

Every decode step therefore allocated a transient proportional to the context
length. Since the server's continuous-batching path always builds
`BatchTurboQuantKVCache` (via `_make_cache` in `generate/ar.py`), `--kv-bits`
on a served model stored the KV quantized but paid to materialize it in full
on every step.

The single-sequence `TurboQuantKVCache` has `decode_attention`,
`prefill_attention` and `quantized_attention`; the batch class had none of
them. The comment on `BatchTurboQuantKVCache.group_size` even reads *"not used
for TurboQuant (it has its own attention path)"*, which reads like the
omission was accidental.

Those kernels never depended on the single-sequence storage layout. They read
the codecs plus the `keys_state`/`values_state` the caller passes in, and they
already index a leading batch dimension (`BQH = B * n_q_heads`, grid
`B * n_repeats`). So they move to a `_TurboQuantAttentionMixin` shared by both
caches — no kernel code is duplicated — and the batch cache dispatches through
it when it holds a single row with no left padding:

```python
def _turboquant_attention_applies(cache) -> bool:
    if not isinstance(cache, BatchTurboQuantKVCache):
        return True
    if not cache.is_single_row():
        return False
    return bool((cache.left_padding == 0).all())
```

Wider or left-padded batches keep the dequantizing fallback; the kernels would
otherwise attend to padded positions as if they were real tokens. The
chunk-size constants move to the mixin along with the kernels, since reading
`decode_key_chunk_size` off the batch cache would otherwise raise
`AttributeError` mid-decode.

## 2. Single-row MTP bypassed `--kv-bits` entirely

`make_speculative_prompt_cache` special-cased one configuration:

```python
if draft_kind == "mtp" and batch_size == 1:
    return cache.make_prompt_cache(lm)
return make_cache(lm, left_padding)
```

`make_cache` is what applies `--kv-bits`. Taking `cache.make_prompt_cache`
instead left the KV unquantized however the server was configured. Inspecting
the 16 full-attention layers of Qwen3.6-27B:

| configuration | KV cache type |
| --- | --- |
| MTP, `batch_size=1` | `KVCache` ×16 — unquantized |
| MTP, `batch_size=2` | `BatchTurboQuantKVCache` |
| no drafter, `batch_size=1` | `BatchTurboQuantKVCache` |

Only that one combination diverged, and it is the configuration a
single-stream server runs in.

Routing it through `make_cache` exposed the third defect. Target verification
narrows the cache one draft token at a time
(`keys[:, :, : prefix_len + i + 1, :]`), which raised `TypeError:
'_QuantizedStateProxy' object is not subscriptable`. That is the real reason
the special case existed — the rollback contract it also has to satisfy
(`is_trimmable`/`trim`/`zero_row_tail`) was already met. The proxy now serves
contiguous slices along the time axis by slicing the quantized state, and
rejects indexing it cannot answer without dequantizing.

## 3. Attention sinks were dropped on the batch path

The batch branch called `mx.fast.scaled_dot_product_attention` without
forwarding `sinks`, so a model that uses them silently received the no-sink
result. Sharing the fused kernels made this visible, because the
single-sequence branch raises on sinks instead.

Neither behaviour is correct. The fused kernels have no sink term, so a
request carrying sinks now skips them and takes the dequantizing path, which
forwards `sinks` to MLX. This is a fix to existing behaviour rather than a
consequence of the other two changes.

## Results

Measured on a 24GB M-series machine serving Qwen3.6-27B behind an external
watchdog set at 21GB, one request at a time, a fresh process per data point.

**32K prompt tokens including a 2048px image, mxfp4 weights, 4-bit
TurboQuant KV, MTP drafter:** decode previously grew from 20GB to 21.5GB
within five seconds of prefill finishing and was killed before emitting a
token. It now completes 5,472 tokens at an 18.34GB peak.

**64K prompt tokens including an image, 3-bit weights, 3-bit TurboQuant KV,
MTP drafter:** 64,268 prompt tokens in, 6,468 generated, 18.64GB peak.

Both runs were re-measured on this branch after rebasing onto `ffd7aeff`, so
the numbers reflect the code as submitted.

## Scope

The gain is confined to decode. `prefill_attention` returns `None`
unconditionally because of #1016: `_build_codec` is only ever called with
`mode="mse"`, so `_TurboQuantProdCodec` is never constructed and the codec
type check at the top of `prefill_attention` always fails. Confirmed for both
cache classes:

```
TurboQuantKVCache      prefill_attention -> None
BatchTurboQuantKVCache prefill_attention -> None
TurboQuantKVCache      decode_attention  -> array(1, 4, 1, 64)
BatchTurboQuantKVCache decode_attention  -> array(1, 4, 1, 64)
```

Prefill therefore still takes the dequantizing path on both caches and its
throughput is unchanged (~107 tok/s before and after). This PR does not
attempt to fix #1016; it only lets the batch cache reach the same paths the
single cache already reaches. Once #1016 is resolved, the batch cache picks up
the prefill fix through the shared mixin.

## Tests

Adds `test_batch_turboquant_attention.py` and
`test_speculative_quantized_kv.py`, covering the shared surface, the
single-row/padding guard, numerical equivalence against the dequantizing
fallback, sinks being applied rather than dropped or rejected, head_dim
64/128/256 against 1, 4 and 6 query heads per KV head, and a regression test
asserting per-step peak memory does not scale with context length.

I checked the suite catches the regressions it is meant to catch: reverting
just the chunk-size constants makes 6 tests fail with the same
`AttributeError` seen in production, and restoring the raise-on-sinks
behaviour fails both sinks tests.

Two server tests that pinned the old single-row MTP branch are replaced by one
parametrised test asserting the new contract: every speculative configuration
goes through the supplied `make_cache`.

Full suite on this branch: 1816 passed, 1 failed. The failure is
`test_speculative.py::test_qwen_target_verify_quantized_linear_matches_singleton_batch_path`,
which fails identically on an unmodified checkout of `ffd7aeff` — pre-existing
and unrelated.

## What I could not verify

End-to-end model runs are Qwen3.6-27B only (GQA 6:1, head_dim 256, with and
without a vision input). Coverage for other head geometries and for attention
sinks is numerical-equivalence testing against the dequantizing path, not real
model output — I do not have a sinks model to run. Multi-row batches and
left-padded rows deliberately keep the existing fallback, so they should be
unaffected either way.


YuhuaBillChen:
Rebased onto current main and revalidated locally.

The conflict with #1842 was in make_speculative_prompt_cache. That PR kept the batch_size == 1 → cache.make_prompt_cache shortcut, which is the line this PR removes: that path bypasses make_cache, so --kv-bits is silently ignored for single-row speculation. At 32K a fp16 cache costs ~8.4GB where 4-bit TurboQuant needs ~2.1GB.

I resolved it in favour of this PR's behaviour, which now also covers DFlash — single-row DFlash from #1842 inherits the same bypass. This does not weaken the new drafter: for batch_size == 1, rollback_speculative_cache takes is_batch=False and only calls c.trim(), which the batch caches implement (is_trimmable/trim/zero_row_tail) — that is why the multi-row path already used them. The parametrised test covers mtp/dflash/eagle3 at batch 1 and 2, keeping the eagle3 case added in #1842.

The latest workflow is still showing action_required, so the checks have not run on this PR yet. Could a maintainer approve the workflow run when convenient?

On scope: the first two commits form one change — the fused batch attention path and the quantized speculative cache. Splitting them would give MTP quantized KV while still dequantizing the whole cache each step, which is the problem the pair is meant to fix. The third commit (attention sinks fallback, ~10 lines of non-test code) is genuinely separable, and I'm happy to move it into its own PR if a narrower review scope is preferred.

YuhuaBillChen:
@Lazarus-931 could you approve the pending workflow run when you get a chance? Thanks!

YuhuaBillChen:
Thanks for approving the workflow and merging main, @Lazarus-931  checks are green and the branch is clean now.
One snag: because that merge commit makes you the last pusher, branch protection now needs the approving review to come from someone else. @lucasnewman would you mind taking a look when you have a moment?

YuhuaBillChen:
Friendly ping — this one is green and mergeable, just needs an approving review from someone other than the last pusher. @lucasnewman Many thanks~

lucasnewman:
@YuhuaBillChen Can you run some benchmarks that show the relative performance and memory usage before and after your PR? I believe your memory savings claim but it's not clear to me how this impacts overall tokens/sec for prefill & decode.

mlx_vlm/models/base.py
```diff
@@ -289,6 +289,21 @@ def quantized_scaled_dot_product_attention(
...
+    They read the quantized state directly, so they need one contiguous
+    sequence starting at index 0. That always holds for the single-sequence
+    cache; the batch cache qualifies only while it carries a single row with
+    no left padding (padded rows would be attended to as real tokens).
+    """
+    if not isinstance(cache, BatchTurboQuantKVCache):
+        return True
+    if not cache.is_single_row():
+        return False
+    return bool((cache.left_padding == 0).all().item())
```
lucasnewman: This forces a new gpu sync due to the `.item()` call, which I suspect may be a performance issue.

YuhuaBillChen: Good catch. I removed the .item() call from the attention hot path. BatchTurboQuantKVCache now caches fused-attention eligibility as a Python boolean and refreshes it only at batch lifecycle boundaries (filter, extend, and state restore), where synchronization is already unavoidable. The per-token attention dispatch only reads that boolean.

I also added lifecycle regression coverage. The focused attention suite passes 36/36.

An isolated ABBA test measured the old guard at 100.49 us/call versus 0.017 us/call for the cached boolean. End-to-end, removing the sync improved no-MTP decode from 14.27 to 15.28 tok/s (+7.1%) and MTP decode from 18.73 to 19.74 tok/s (+5.4%). Prefill and peak memory were effectively unchanged.

YuhuaBillChen:
Thanks — I ran fresh-process ABBA comparisons with a Qwen3.6-27B MXFP4 target, Q8 vision tower, fixed 2048x1542 image, MTP block 3 where applicable, APC disabled, a 20.5 GiB watchdog, and 250 ms external footprint sampling. Each equal-work row below is the median of two valid runs with identical prompt and output lengths.

| Workload | Prefill, before -> after | Decode, before -> after | OS peak footprint, before -> after |
|---|---:|---:|---:|
| 4K + 512, fused TQ attention only | 125.4 -> 125.5 tok/s (+0.1%) | 10.7 -> 15.3 tok/s (+42.8%) | 17,908 -> 17,460 MiB (-448 MiB) |
| 4K + 512, full MTP path | 127.0 -> 125.2 tok/s (-1.4%) | 22.0 -> 19.8 tok/s (-10.2%) | 17,872 -> 17,692 MiB (-180 MiB) |
| 16K + 512, full MTP path | 119.9 -> 117.1 tok/s (-2.3%) | 21.58 -> 14.59 tok/s (-32.4%) | 19,019 -> 17,968 MiB (-1,051 MiB / 5.5%) |

The 16K row is the clearest equal-work memory comparison. The decode cost is real, but it is the cost of actually enabling the requested TurboQuant Q4 KV cache: before this PR, the single-row speculative shortcut silently used unquantized KV despite `--kv-bits 4`.

To separate generic 4-bit cost from TurboQuant overhead, I also ran all three KV backends under the same patched runtime and 16K + 512 protocol:

| KV backend | Prefill | Decode | OS peak footprint | MLX peak |
|---|---:|---:|---:|---:|
| Unquantized 16-bit | 119.94 tok/s | 21.13 tok/s | 19,011 MiB | 17.61 GB |
| MLX native uniform Q4 | 120.43 tok/s | 18.92 tok/s | 17,687 MiB | 16.87 GB |
| TurboQuant Q4 | 116.32 tok/s | 14.57 tok/s | 17,968 MiB | 17.41 GB |

For a capacity boundary, the upstream no-MTP path tripped the 21 GiB watchdog twice at 8,086 prompt tokens (0/2 completed, 21,081-21,084 MiB), while the patched path completed 2/2 at 17,609-17,616 MiB. With MTP enabled at the same prompt length both arms completed, but their generated lengths differed (1,024 versus 127 tokens), so I am not using that pair for throughput or peak-memory claims.

The native uniform Q4 experiment exposed a separate pre-existing Qwen MTP batch rollback bug. I isolated that fix in #1986 rather than expanding this PR.

Finally, I removed the `.item()` synchronization noted in the inline review. The isolated guard cost fell from 100.49 us/call to 0.017 us/call; full-MTP decode improved from 18.73 to 19.74 tok/s (+5.4%), while prefill, peak memory, and MTP draft acceptance (62.3%) were unchanged. The focused attention suite passes 36/36.

YuhuaBillChen:
One follow-up after updating the branch: the `.item()` synchronization you flagged was removed in `94ba841` (now included in head `4627daa`). Fused-attention eligibility is cached and refreshed only during cache lifecycle mutations, so the per-token decode path no longer synchronizes. The benchmark results are in the comment above.

One related compatibility issue surfaced during that validation: on current upstream, native uniform Q4 KV plus Qwen MTP fails in target verification with `'tuple' object has no attribute 'shape'`. It reproduces without this PR at `batch_size=2`, so the underlying bug is pre-existing; #1822 makes the same path reachable at `batch_size=1` by correctly honoring `--kv-bits 4`.

I kept that fix separate in #1986 rather than expanding this PR. Ideally #1986 should land first, then #1822 can be rebased/updated on top of it.

@lucasnewman both #1822 and #1986 are currently waiting for maintainer approval before their workflows can start. Could you approve the runs when you get a chance? Thanks!

YuhuaBillChen:
@lucasnewman Hi, can you help kick off the workflow? Also addressed the .item() sync and added benchmarks above. Thanks 😄

YuhuaBillChen:
@lucasnewman The latest signed push (Black formatting only) has put the workflow back into action_required. Could you approve the workflow run when you get a chance? Thanks!

YuhuaBillChen:
Rebased onto main (7acac8a) to pick up the MRoPE split fix from #2020 was the only red in the last run. @lucasnewman  mind approving the workflow so it can go green?

mlx_vlm/speculative/utils.py
```diff
@@ -122,8 +121,14 @@ def make_speculative_prompt_cache(
     left_padding,
     make_cache: Callable,
 ):
-    if batch_size == 1:
-        return cache.make_prompt_cache(lm)
+    del draft_kind, batch_size
+    # `make_cache` is what honours --kv-bits. Single-row speculation used to
```
lucasnewman: This is the only part of the change that gives me pause since it affects all speculative decoders. Have you tested anything outside of MTP (DFlash, EAGLE, etc)?

Minor nit: why do we need the variable deletion here? Also, we should tighten the comment -- it doesn't need to reflect the implementation details of your patch or what existed before, just what the code will do once the change is landed.

YuhuaBillChen: Thanks — both points are fixed now.

I removed the unused del and shortened the comment to describe the current behavior.

I added DFlash coverage to the cache-construction contract test, including
both single-row and batched inputs. Since dspark, laguna, and
muse_glimmer_assistant all come to dflash, they share that path.

I also added an end-to-end test using a small Muse Glimmer target and
assistant. It runs through `generate_step` and confirms that the target cache
remains a 4-bit `BatchTurboQuantKVCache` after decoding. Existing MTP and
EAGLE3 coverage remains unchanged.

Full suite on MLX 0.32.2: 2668 passed, 7 skipped.

YuhuaBillChen:
@lucasnewman sorry to ping again but do you have time for another look, thanks. CI all green and dflash addressed.

YuhuaBillChen:
Rebased and squashed onto current main@6102cb4.

The resulting diff is unchanged in scope, while incorporating the latest
speculative-verifier, native-Q4, APC, and Qwen3.8 updates from main.

Validation on MLX 0.32.2:
- focused TurboQuant/speculative coverage: 74 passed
- full suite: 2969 passed, 7 skipped, 75 subtests passed
- Black, isort, autoflake, compilation, and diff checks pass

YuhuaBillChen:
Hi @lucasnewman  would you mind having another look? Green since Aug 28 with the dflash coverage you asked for.

Anything still blocking, or just needs a slot?

Thanks a ton!

lucasnewman (approved):
Thanks!

YuhuaBillChen:
Many thanks!

PR #1823, opened 2026-08-09
Add server-aware generate and chat CLIs

Lazarus-931:
## Summary

  - Add optional server reuse to `mlx_vlm generate` and `mlx_vlm chat`.
  - Reuse a running mlx-vlm server through its OpenAI-compatible
    `/v1/chat/completions` API instead of cold-loading the model again.
  - Add `--base-url`, `--api-key`, `--server`, and `--local`.
  - Preserve the existing local path for unsupported or local-only options.

  ## Behavior

  For plain text and vision generation, the CLI probes a running server and
  streams its response when available. `--server` requires remote execution;
  `--local` explicitly bypasses the server.

  Options that cannot be represented faithfully by the request API remain local.
  A reachable server that returns an authentication or HTTP error is surfaced,
  rather than silently falling back to a separate local model load.

  ## Validation

  - `uv run --with pytest python -m pytest mlx_vlm/tests/test_cli.py mlx_vlm/tests/test_server.py -q`
    - 283 passed
  - Formatted with Black; pre-commit Black, isort, and autoflake passed.
  - End-to-end verified on an M4 Mac mini against a real mlx-vlm server using
    `mlx-community/Qwen3-1.7B-4bit`.

  ## Related

  This provides the engine-side server-aware CLI behavior needed by the Nativ CLI:
  https://github.com/Blaizzy/nativ/pull/228

### DIFF
--- a/mlx_vlm/chat.py
+++ b/mlx_vlm/chat.py
@@ -23,6 +23,14 @@ from mlx_vlm.generate import (
     PromptCacheState,
     stream_generate,
 )
+from mlx_vlm.generate.remote import (
+    RemoteServerError,
+    _image_url,
+    resolve_base_url,
+    sampling_params,
+    server_available,
+    stream_chat,
+)
 from mlx_vlm.prompt_utils import apply_chat_template
 from mlx_vlm.utils import load_image
 from mlx_vlm.vision_cache import VisionFeatureCache
@@ -35,9 +43,12 @@ class MLXVisionChat:
         temperature: float = 0.7,
         max_tokens: int = 1000,
         verbose: bool = False,
+        remote_args=None,
         **kwargs,
     ):
         self.console = Console()
+        self.model_path = model_path
+        self.remote_args = remote_args
         self.verbose = verbose
         self.temperature = temperature
         self.max_tokens = max_tokens
@@ -56,10 +67,16 @@ class MLXVisionChat:
             self.chat_template_kwargs["thinking_mode"] = thinking_mode
         self.stream_kwargs = kwargs
 
-        with self.console.status("[bold green]Loading model..."):
-            self.model, self.processor = load(model_path)
-
-        rprint("[bold green]Model loaded successfully![/bold green]")
+        if remote_args is None:
+            with self.console.status("[bold green]Loading model..."):
+                self.model, self.processor = load(model_path)
+            rprint("[bold green]Model loaded successfully![/bold green]")
+        else:
+            self.model = None
+            self.processor = None
+            rprint(
+                f"[bold green]Using server at {resolve_base_url(remote_args)}.[/bold green]"
+            )
         self.print_help()
 
     def print_help(self) -> None:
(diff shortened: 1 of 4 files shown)


mlx_vlm/chat.py
```diff
@@ -194,6 +232,28 @@ def main():
...
+        default=None,
+        help="API key for the server (default: $MLX_VLM_API_KEY).",
+    )
+    server_mode = parser.add_mutually_exclusive_group()
+    server_mode.add_argument(
+        "--server",
+        action="store_true",
+        help="Require a running server instead of loading the model locally.",
+    )
+    server_mode.add_argument(
```
Blaizzy, line 252: We don't need this flag 

If server url is not present, it gets skipped

mlx_vlm/generate/dispatch.py
```diff
@@ -73,6 +73,32 @@ def parse_arguments():
...
+        help="API key for the server (default: $MLX_VLM_API_KEY).",
+    )
+    server_mode = parser.add_mutually_exclusive_group()
+    server_mode.add_argument(
+        "--server",
+        action="store_true",
+        help="Require the server: send the request over HTTP instead of loading "
+        "the model locally (errors if the server is unreachable).",
+    )
+    server_mode.add_argument(
```
Blaizzy, line 97: Here too

mlx_vlm/generate/remote.py
```diff
@@ -0,0 +1,259 @@
...
+import json
+import mimetypes
+import os
+import sys
+import urllib.error
+import urllib.parse
+import urllib.request
+
+DEFAULT_BASE_URL = "http://127.0.0.1:8080"
+DEFAULT_PREFILL_STEP_SIZE = 2048
```
Blaizzy, line 19: Inherits from generate or server 

This is a duplicate

mlx_vlm/generate/remote.py
Blaizzy, line 1: This file doesn't seem well thought out.

Let's not let agents handle everything. Careful system planning and thinking should have precedence.

mlx_vlm/generate/remote.py
```diff
@@ -0,0 +1,259 @@
...
+            return True
+    except urllib.error.HTTPError as exc:
+        if exc.code == 404:
+            return False
+        raise RemoteServerError(_http_error_message(exc)) from exc
+    except OSError:
+        return False
+
+
+def eligible_for_server(args):
```
Blaizzy, line 71: This is wrong, the server handles most of it

mlx_vlm/generate/remote.py
```diff
@@ -0,0 +1,259 @@
...
+    )
+    if any(disqualifying):
+        return False
+    if getattr(args, "revision", "main") != "main":
+        return False
+    prefill_step_size = getattr(args, "prefill_step_size", None)
+    return prefill_step_size in (None, DEFAULT_PREFILL_STEP_SIZE)
+
+
+def _image_url(path):
```
Blaizzy, line 116: Server already handles it

Blaizzy (commented):
I like the direction, let's just increase the quality of the thinking and manual review.

PR #1824, opened 2026-08-09
Preserve packed TurboQuant states in exact APC snapshots [New]

YuhuaBillChen:
Stacked on #2072. Until that PR lands, the GitHub diff also includes its exact-APC ownership-transfer changes; the intended delta here is the packed TurboQuant snapshot support.

## Summary

- keep exact-APC TurboQuant snapshots in their packed representation in memory and on disk
- preserve the exact TurboQuant state type, key/value bit widths, seed, offset, and codec configuration
- restore packed rows directly into `BatchTurboQuantKVCache` without a BF16 round trip
- keep non-consuming merge semantics defensive: only the explicit consuming path borrows a single restored row

## Why

After #2072, exact APC releases restored sources layer by layer and avoids redundant ownership copies. TurboQuant snapshots still took an avoidable representation round trip, however:

```text
packed live TQ KV -> BF16 exact snapshot -> disk -> BF16 restore -> packed batch TQ KV
```

That expanded both the shard and the restore working set, then spent time requantizing data that was already quantized. This PR adds a distinct `turboquant_kv` exact-snapshot schema and keeps the state packed end to end.

Existing float exact snapshots remain readable through the existing `kv` schema. Other cache adapters keep their current behavior.

## Isolated results

I compared current #2072 with this PR in fresh processes on a 24 GB M-series machine. The synthetic cache has the 16 accumulating full-attention layers of Qwen3.8-27B (4 KV heads, head dimension 256) and TQ4 KV. The 16K results are medians of two runs in ABBA order.

| 16K exact snapshot metric | #2072 | This PR | Change |
|---|---:|---:|---:|
| On-disk shard | 2,048.1 MiB | 260.1 MiB | -87.3% |
| Snapshot MLX peak | 3,597.5 MiB | 780.5 MiB | -78.3% |
| Snapshot construction | 184.7 ms | 14.0 ms | -92.4% |
| Restore lookup MLX peak | 3,510.0 MiB | 260.5 MiB | -92.6% |
| Restore lookup | 261.9 ms | 82.2 ms | -68.6% |
| Merge into batch TQ cache | 265.8 ms | 0.44 ms | -99.8% |

A 64K capacity point showed the same scaling:

| 64K exact snapshot metric | #2072 | This PR |
|---|---:|---:|
| On-disk shard | 8,192.4 MiB | 1,040.4 MiB |
| Restore lookup MLX peak | 11,051.0 MiB | 1,040.5 MiB |
| Restore lookup | 2.135 s | 0.171 s |
| Merge into batch TQ cache | 0.836 s | 0.00052 s |

These are cache-transformation measurements, not full-server process-footprint claims.

## Real-model canary

I also ran a cold/warm server canary with Qwen3.8-27B MXFP4, its Q8 vision tower, and TQ4 KV. The cold request had 5,012 prompt tokens. A second identical request restored 5,011 tokens from the packed exact snapshot, returned HTTP 200, and reduced wall time from 40.32 s to 0.351 s. The restored cache continued as `BatchTurboQuantKVCache`; no traceback or watchdog trip occurred.

## Tests

- Q4, Q3.5, and asymmetric K3/V8 packed disk round trips
- detached snapshot ownership and consuming/non-consuming merge behavior
- single-row and padded two-row merge, followed by cache growth
- full suite on MLX 0.32.2: `2916 passed, 7 skipped, 75 subtests passed`
- Black, isort, autoflake, Python compilation, and `git diff --check` pass

### DIFF
--- a/mlx_vlm/apc.py
+++ b/mlx_vlm/apc.py
@@ -815,6 +815,79 @@ def _decode_checkpoint_tree(structure: dict, load_array) -> Any:
     raise ValueError(f"unsupported checkpoint tree node: {kind!r}")
 
 
+def _encode_turboquant_state(
+    value: Any,
+    tensor_prefix: str,
+    arrays: Dict[str, mx.array],
+    counter: List[int],
+) -> Optional[dict]:
+    """Encode a packed TurboQuant state while preserving NamedTuple types."""
+    if isinstance(value, mx.array):
+        name = f"{tensor_prefix}_t{counter[0]}"
+        counter[0] += 1
+        arrays[name] = value
+        return {"kind": "array", "name": name}
+    if value is None:
+        return {"kind": "none"}
+    if isinstance(value, tuple):
+        from . import turboquant as tq
+
+        state_types = {
+            cls.__name__
+            for cls in (
+                tq.TurboQuantMSEState,
+                tq.TurboQuantProdState,
+                tq.TurboQuantPolarState,
+                tq.TurboQuantPolarProdState,
+                tq.TurboQuantSplitState,
+            )
+        }
+        items = [
+            _encode_turboquant_state(item, tensor_prefix, arrays, counter)
+            for item in value
+        ]
+        if any(item is None for item in items):
+            return None
+        type_name = type(value).__name__
+        if type_name not in state_types:
+            return None
+        return {"kind": "state", "type": type_name, "items": items}
+    return None
+
+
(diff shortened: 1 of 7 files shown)


Blaizzy:
Hey @YuhuaBillChen 

Thank you very much for this optimization, the results look great!

However, I'm going to have to close this PR because right now we want to have a stable APC stability before optimize.

Please an issue and we'll revisit this optimization in a couple weeks.

PR #1827, opened 2026-08-09
Fall back to globbing when the safetensors index is stale

eptan:
load_model() filters the shards named in model.safetensors.index.json by existence. When the index is stale (checkpoints re-quantized or renamed without regenerating the index, as some hub repos have shipped), the filtered list is non-empty but incomplete, so the *.safetensors glob fallback never fires and loading proceeds with a partial weight set that only fails much later in load_weights with a missing-parameter error that names tensors, not the real cause.

Extract shard resolution into _resolve_weight_files(): use the index only when every shard it names exists; otherwise warn that the index is stale and discover shards by globbing. Behavior is unchanged for complete indexes, absent indexes, and unreadable indexes.

### DIFF
--- a/mlx_vlm/tests/test_utils.py
+++ b/mlx_vlm/tests/test_utils.py
@@ -1,5 +1,6 @@
 import base64
 import json
+import logging
 import struct
 from io import BytesIO
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


lucasnewman:
We'd prefer to fix the repos instead of patching over it here. Closing.

PR #1829, opened 2026-08-09
Warn when resize_shape is ignored by a custom image processor

eptan:
process_image() silently discards a caller-supplied resize_shape whenever the model ships its own BaseImageProcessor subclass, so the --resize-shape CLI flag and the resize_shape generate kwarg are no-ops for those models with no indication to the user.

Emit a logging.warning naming the processor class and pointing at its own sizing options instead, and add tests covering both processor kinds plus the no-warning path.


lucasnewman:
@eptan Can you [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits and force push so we can clear the merge requirements?

eptan:
@lucasnewman Done. I signed the commits and force pushed

lucasnewman (approved):
@eptan Thanks!

PR #1832, opened 2026-08-10
Server aware chat

Lazarus-931:
Adds --base-url/--api-key to mlx_vlm chat so the REPL runs against a running mlx-vlm server instead of loading the model in-process — each turn posts the conversation to the server's /v1/chat/completions and streams the reply, reusing the already-loaded weights. With no base URL it behaves exactly as before, loading and running the model locally.

> **Stacked on #1831.** Until #1831 merges, this diff also shows its `generate`/`remote.py` changes; it reduces to the `chat.py` delta once #1831 lands. Review #1831 first.

### DIFF
--- a/mlx_vlm/chat.py
+++ b/mlx_vlm/chat.py
@@ -2,6 +2,7 @@ import argparse
 import codecs
 import os
 import sys
+import urllib.error
 from typing import Dict, List
 
 from rich import print as rprint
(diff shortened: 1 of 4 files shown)


lucasnewman:
@Lazarus-931 Is this so you can exercise the server path from chat for testing or something else?

Lazarus-931:
> @Lazarus-931 Is this so you can exercise the server path from chat for testing or something else?

we wanted to work on nativ's cli support, and the idea was, instead of wrapping it in Nativ's logic, we have mlx-vlm itself sense the server the user has up

Lazarus-931:
though I paused on it since I wanted this to be well made and wasn't sure

Issue #1833, opened 2026-08-10
DFlash batch verification can desynchronize shared KV positions and miss EOS


Thump604:
## Problem

Direct batched DFlash can commit different accepted depths per row even though the target KV cache has one shared physical sequence axis. Zeroing a rejected tail does not move a shorter row back from the maximum committed RoPE/cache position. A later round can therefore verify a row at the wrong logical position.

The direct DFlash helpers also do not receive model EOS IDs. They can continue after a first-bonus or emitted EOS until max_tokens. If only some first-bonus rows are EOS, the reduced verifier batch must filter target-cache membership before its first forward call.

## Root cause

The batch walk returns independent per-row acceptance counts, but cache rollback is performed once on a shared sequence axis. EOS handling exists at higher generation layers but was not propagated into the direct DFlash round helpers.

## Proposed fix

- Reconcile the remaining divergent-position concern with merged #2113 before proposing another fix. The original uniform-depth clamp was withdrawn after the Gemma benchmark in #1834 failed output parity; the historical Gemma case has not been requalified on current main.
- Propagate model EOS IDs into singleton and batch DFlash rounds.
- Filter initially finished rows from target caches before reduced-batch verification.
- Add rollback, unequal-budget, first-bonus EOS, emitted EOS, and mixed-row regressions.

#1834 now covers EOS handling and cache membership only. Its August test and benchmark results are historical; the current-main composition still needs conflict resolution and validation.

PR #1834, opened 2026-08-10
Fix DFlash EOS handling in direct generation

Thump604:
Refs #1833.

## Scope
- Pass model EOS IDs into direct DFlash singleton/batch generation; stop on first-bonus or emitted EOS.
- Remove initially finished rows from target caches before reduced-batch verification.
- Merge main `5c9b5f52` without rewriting public ancestry; preserve #2113 uniform-acceptance behavior.

## Validation
At `070041ff`: all 227 tests in `mlx_vlm/tests/test_speculative.py` passed (no skips), including 10 EOS boundary cases. Black, AST parsing, and main-relative diff checks passed.

The earlier uniform-clamp proposal was withdrawn after the historical Gemma benchmark failed output parity. This EOS-only refresh does not claim new Gemma numerical parity or performance results. Any further divergent-position fix needs a fresh reproduction on main with #2113; it is not included here.


Blaizzy:
This is really cool, would love to see the benchmark results

Thump604:
Thanks — I have refreshed the draft on current `main` at `d3c1e45c` and changed the issue linkage to `Refs #1833` while qualification is still incomplete. The four-file DFlash correctness scope is unchanged.

Focused validation on the rebased head: 22 DFlash tests passed (138 unrelated speculative tests deselected); compile and diff checks are clean.

I agree benchmark evidence is the remaining requirement. I will post a paired current-main vs fix run using a matching target+DFlash artifact, with B=2 homogeneous and deliberately divergent rows. It will report output/EOS equivalence, drafted/accepted counts, warmed latency/throughput, and peak memory. I am keeping the PR in draft and am not claiming performance results before that run.

Thump604:
I ran the paired benchmark requested, and it surfaced a correctness blocker in the current uniform-clamp implementation, so I am keeping this draft and not presenting it as qualified.

Hardware/artifacts:
- Apple M2 Ultra, 128 GB
- target: Gemma 4 31B BF16 (`842da3794eaa0b77d5f08bae87a17459d91ff475` artifact path revision)
- matching Gemma 4 31B DFlash BF16 drafter (model SHA256 `b3f28580103d6e8a0a1d9389646957b9309df5982558afd754b9810357f425b7`)
- base `7a741ebf`; PR `d3c1e45c`
- greedy decoding, one warmup + 3 measured trials per arm

Median results:

| workload | base tok/s | PR tok/s | delta | output | accepted/drafted (base → PR) |
|---|---:|---:|---:|---|---|
| B=1 | 7.340 | 7.354 | +0.18% | identical | 30/82 → 30/82 |
| B=2 homogeneous | 26.171 | 26.139 | -0.12% | identical | 98/192 → 98/192 |
| B=2 deliberately divergent, budgets 12/64 | 7.624 | 7.351 | -3.58% | **different** | 38/112 → 37/115 |

Peak memory was approximately 66 GB in the divergent case for both arms.

I checked the divergent long row against two independent references: B=1 DFlash and dense target-only generation. Both match current main/base exactly. The PR diverges (`…fast data retrieval, typically offering…` versus the reference `…fast data retrieval—ideally providing…`).

The uniform clamp is token-valid at the immediate verifier boundary, but it changes when a longer-accepted row is re-forwarded. On this real checkpoint that changes later greedy output, so the current patch does not establish the claimed target-equivalent batch correction and also regresses the divergent workload by about 3.6%.

My recommendation is to split/retain the independently valid EOS handling and withdraw the uniform-clamp claim until request-local positions or a rowwise fallback can be implemented and benchmarked. I will revise the draft accordingly rather than weaken the oracle.

Lazarus-931:
hi @Thump604, do you want to fold the other half of the fix onto here?

or open a sep pr?

Both ways are ok :)

Thump604:
Thanks, I'd keep this PR focused on EOS handling. I withdrew the earlier clamp because it changed the Gemma 4 long-row output against both independent references. #2113 has since added related uniform-acceptance handling, so I'll reconcile that overlap before proposing any separate follow-up. Its reported Qwen3.5 validation does not establish parity for the historical Gemma case. This PR also needs a current-main conflict refresh and validation; the August results do not qualify that composition. I'll correct #1833's stale clamp recommendation as well.

Thump604:
I've refreshed the EOS-only patch against `5c9b5f52`, preserving #2113 and the existing branch history. At `070041ff`, all 227 speculative tests pass, including the 10 EOS boundary cases. This is source regression coverage, not a rerun of the historical Gemma parity benchmark.

Thump604:
I fixed the compatibility regression for low-level DFlash targets that do not expose an outer `config`, without changing the existing DFlash2/DSpark fixtures or parity checks. At `46fb58eb`, the eight previously failing CI cases and twelve EOS boundary cases pass locally (20 total). [CI is running](https://github.com/Blaizzy/mlx-vlm/actions/runs/33887251962); this is source-test coverage, not production Gemma qualification.

Issue #1715, opened 2026-07-25
Prompt-cache prefix reuse corrupts rotating (sliding-window) KV caches — silent wrong output, crash with speculative decoding


justSunny12:
## Summary

The per-turn prompt-cache prefix reuse in `mlx_vlm/generate/dispatch.py` trims the KV cache with a raw slice that is only valid for the flat `KVCache`. Applied to a rotating (sliding-window) cache it slices a **ring buffer** by a **logical** length and overwrites `offset`, while leaving the ring's own state (`_idx`, `start_position`) untouched.

Two failure modes, both reproduced:

* **`RotatingKVCache` → silent context corruption.** No exception; the window's token order is scrambled. Greedy decoding of the same prompt diverges from a cold run.
* **`BufferedRotatingKVCache` → hard crash.** This is the class speculative decoding installs (`speculative/mtp.py::_buffer_mtp_target_cache`), and it stays in the `PromptCacheState` for subsequent turns:

```
ValueError: [broadcast_shapes] Shapes (1,8,4,256) and (1,8,1029,256) cannot be broadcast
  BufferedRotatingKVCache._ensure_capacity (mlx_vlm/models/cache.py)
```

This affects any model with sliding-window attention (found on Gemma 4, `sliding_window: 1024`, where half the layers are rotating).

## Environment

| | |
|---|---|
| mlx-vlm | 0.6.7 |
| mlx | 0.32.0 |
| Python | 3.14.2 |
| OS | macOS 26.5.2, Apple M5, 32 GB |
| Model | `google/gemma-4-26B-A4B-it`, converted with `mlx_vlm.convert -q --q-bits 4` |
| Drafter | `gemma-4-26B-A4B` MTP, `draft_block_size=2` |

## Root cause

`mlx_vlm/generate/dispatch.py`, in the `prompt_cache_state` reuse branch:

```python
# Trim cache to prefix_len in case it includes generated tokens
for c in kv_cache:
    if hasattr(c, "keys") and c.keys is not None:
        cached_len = c.keys.shape[2]
        if cached_len > prefix_len:
            c.keys = c.keys[:, :, :prefix_len, :]
            c.values = c.values[:, :, :prefix_len, :]
            if hasattr(c, "offset"):
                c.offset = prefix_len
```

For a rotating cache:

* `c.keys.shape[2]` is the **physical** buffer width (window, plus speculative slack), not the logical length;
[...]

PR #1596, opened 2026-07-14
fix(apc): exact hybrid + TurboQuant warm match live kv layout 

weklund:
## Summary

Fixes continuous-batching crashes under **APC + `--kv-bits`** when a second request joins mid-flight with an **exact-mode** (hybrid) warm row, and makes warm restore use the same **quant backend** as live gen (uniform vs TurboQuant).

**Root cause (#1579):** live `_make_cache` leaves full-attn layers quantized (last layer float when `n > 2`); exact multi warm always built float `BatchKVCache` → `_extend_cache` type mismatch.

**Also on this branch (#1562 / supersedes open #1580):** the same last-layer policy for **block** warm paths via shared `should_quantize_kv_layer`, so block APC+kv staggered join is fixed too.

**TurboQuant:** warm fill no longer forces `BatchQuantizedKVCache(bits=int(3.5))`. When `turboquant_enabled(bits, scheme)`, restore builds `BatchTurboQuantKVCache` / stream `TurboQuantKVCache`.

This is a **correctness / crash** fix (join layout contract). Not a performance feature. No on-disk APC format change (float store; requant on restore).

Closes / addresses **#1579**. Includes **#1562** (also tracked by #1580 — this branch is #1580 tip + #1579).

## Changes

- `should_quantize_kv_layer` in `models/cache.py` (shared policy)
- Block warm: stream/batch multi use policy (#1562)
- Exact multi: `kv_quant_config` + post-merge align to live types (#1579)
- `_fill_batch_layer_cache` / stream fill: TurboQuant vs uniform by scheme
- `ar.py` / `dispatch.py`: pass `scheme` in quant config
- Tests in existing modules (`test_apc_quantized.py`, `test_batch_quantized_cache.py`)

## Test plan

- [x] Unit: `pytest mlx_vlm/tests/test_apc_quantized.py mlx_vlm/tests/test_batch_quantized_cache.py` → 80 passed, 1 skipped
- [x] Live matrix (11 cached models): staggered cold/warm APC+kv8, co-insert, TQ (`kv=3.5/turboquant`), warm pipeline → **113/114**
- [x] Exact hybrids (Qwen3.5 0.8/2/4B, gemma-4-e2b): uniform cold stagger + **TQ staggered** green
- [x] Server concurrent long prefix: APC+kv8 / APC off+kv8 / float+APC → HTTP 200 (incl. Qwen3.5-0.8B, gemma-4-e2b)
- [ ] CI

### Known residual (out of scope)

[...]

PR #1835, opened 2026-08-10
Decline prefix-cache reuse for non-trimmable recurrent caches

kylesyx:
## Problem

Multi-turn generation crashes from the **second turn onward** on hybrid linear-attention models (Qwen3.5 / Qwen3.6 MoE):

```
File "mlx_vlm/generate/dispatch.py", line 928, in stream_generate
    c.trim(n_drop)
AttributeError: 'ArraysCache' object has no attribute 'trim'
```

Reproduced on `mlx-vlm==0.6.10` (also present on `main` @ ffd7aef) with `Qwen3.6-35B-A3B-MLX-4bit` (`model_type: qwen3_5_moe`) via `mlx_vlm.chat_ui`, and by any caller that passes a `prompt_cache_state`.

## Cause

`_prefix_cache_trim_amount` decides whether a prompt cache can be rolled back to a shared prefix. Its only guard is `_cache_fully_retained`, which answers a *different* question — "has this cache already evicted part of the prefix?" — and deliberately does not use `is_trimmable()` (see its docstring re: `BufferedRotatingKVCache`).

Nothing checks whether the cache supports `trim()` **at all**.

`ArraysCache` backs the linear-attention (GatedDeltaNet) layers of hybrid models. It holds a fixed-size recurrent state with no per-token history, so:

- `is_trimmable()` → `False`
- it defines no `trim` method
- it exposes none of `caches` / `start_position` / `max_size`

That last point is what makes it silent: it falls straight through `_cache_fully_retained` to the final `return True`, reuse is accepted, and `stream_generate` then calls `c.trim(n_drop)` on it.

Qwen3.5/3.6 interleave 3 linear-attention layers per full-attention layer (40 layers, `full_attention_interval=4`), so the cache list always contains `ArraysCache` entries and every conversation with a shared prefix hits this.

The `is_trimmable()` + `trim()` pairing was what #1715 originally recommended (mirroring `mlx_lm.models.cache.trim_prompt_cache`); the merged fix replaced it with the retention check and the trimmability half was dropped.

## Fix

Add `_cache_supports_trim`, mirroring `mlx_lm.models.cache.can_trim_prompt_cache`:

```python
def can_trim_prompt_cache(cache: List[Any]) -> bool:
    return all(c.is_trimmable() for c in cache)
```

with two additions over the mlx-lm version: recursion into `CacheList.caches`, and treating a raising/absent `is_trimmable` as not trimmable. It is consulted before `_cache_fully_retained`, so reuse is declined and the caller cold-prefills — the fallback the docstring already documents for `None`.

`n_drop == 0` still takes the warm path: the cache already *is* the shared prefix, so a recurrent state carries over as-is with no rollback. Append-only follow-up turns keep their cache hit.

## Cost

When a hybrid model has a partially-matching prefix, the turn now re-prefills instead of crashing. This does not make prefix reuse work for these architectures — recurrent state genuinely cannot be rolled back to an arbitrary token boundary (see ml-explore/mlx-lm#980; lmstudio-ai/mlx-engine#287 disables cross-turn caching outright for the same reason). This PR only makes the existing "decline and cold-prefill" path handle them instead of raising.

## Tests

Three cases added to `TestPrefixCacheReuseTrim`:

- `test_recurrent_cache_is_not_reusable` — bare `ArraysCache` declines
- `test_mixed_flat_and_recurrent_is_not_reusable` — one non-trimmable entry declines the whole cache, so flat layers can't roll back while recurrent ones don't and desync
- `test_recurrent_cache_with_exact_prefix_still_reusable` — `n_drop == 0` stays on the warm path

The first two fail on `main` (`assert 40 is None`) and pass with the fix.

```
$ pytest mlx_vlm/tests/test_generate.py -k TestPrefixCacheReuseTrim
11 passed
```

`test_generate.py`, `test_apc.py`, `test_speculative.py` → 304 passed. One unrelated pre-existing failure, `test_qwen_target_verify_quantized_linear_matches_singleton_batch_path` (bfloat16 exact-equality on a quantized matmul), which fails identically on unmodified `main`.

Also verified end-to-end: a 3-turn conversation on `Qwen3.6-35B-A3B-MLX-4bit` through `chat_ui` crashes on turn 2 before the fix and completes with correct cross-turn context after.

Related: #1715 (introduced the guard), ml-explore/mlx-lm#980, ml-explore/mlx-swift-lm#157.

### DIFF
--- a/mlx_vlm/generate/dispatch.py
+++ b/mlx_vlm/generate/dispatch.py
@@ -669,6 +669,77 @@ def _cache_fully_retained(c: Any) -> bool:
     return True  # KVCache / QuantizedKVCache retain the whole sequence
 
 
+def _cache_supports_trim(c: Any) -> bool:
+    """Whether ``c`` can be rolled back at all, i.e. whether ``trim()`` is usable.
+
+    Complements :func:`_cache_fully_retained`, which asks whether the prefix is
+    still *present*. A recurrent entry has no per-token history to roll back to:
+    ``ArraysCache`` -- used by the linear-attention (GatedDeltaNet) layers of
+    hybrid models such as Qwen3.5/Qwen3.6 MoE -- holds a fixed-size state, reports
+    ``is_trimmable() == False`` and defines no ``trim`` at all. It also exposes
+    none of the attributes ``_cache_fully_retained`` keys off, so it falls through
+    that check as reusable and the caller then dies on ``c.trim(n_drop)`` with
+    ``AttributeError: 'ArraysCache' object has no attribute 'trim'``.
+
+    This mirrors ``mlx_lm.models.cache.can_trim_prompt_cache``.
+    """
+    children = getattr(c, "caches", None)
+    if children is not None:  # CacheList
+        return all(_cache_supports_trim(x) for x in children)
+    if not hasattr(c, "trim"):
+        return False
+    is_trimmable = getattr(c, "is_trimmable", None)
+    if callable(is_trimmable):
+        try:
+            return bool(is_trimmable())
+        except Exception:
+            return False
+    return True
+
+
+def _cache_logical_len(c: Any) -> Optional[int]:
+    """How many tokens ``c`` logically holds, or ``None`` when that is unknowable.
+
+    ``_prefix_cache_trim_amount`` used to read this off a top-level ``offset``.
+    Two shapes have no such attribute and silently collapsed to ``0``:
+
+    * ``CacheList``, which keeps its length in its children -- used per layer by
+      hybrid models (inkling / zaya1_vl / falcon_h1).
+    * ``ArraysCache``, whose fixed-size recurrent state has no token axis at all
+      -- the whole cache for mamba / mamba2 / rwkv7, and the linear-attention
+      layers of Qwen3.5 / Qwen3.6 MoE.
+
+    A ``0`` there makes ``n_drop == 0``, which reports the cache as fully
+    reusable and skips both rollback guards below. For a recurrent cache that is
+    not a conservative answer but a wrong one: the prompt is sliced down to the
+    shared prefix while the state still represents the old full sequence.
+
+    ``offset`` is still preferred wherever it exists, so flat and rotating caches
+    keep their exact current semantics (note ``RotatingKVCache.size()`` clamps to
+    ``max_size`` and is deliberately *not* used here). Otherwise recurse into
+    ``caches``, and fall back to ``0`` only for a provably empty cache. A
+    non-empty cache with no way to state its length returns ``None`` -- fail
+    closed, because the alternative is reusing stale state.
+    """
+    offset = getattr(c, "offset", None)
+    if offset is not None:
+        return int(offset or 0)
+    children = getattr(c, "caches", None)
+    if children is not None:  # CacheList: length lives in the children
+        lens = [_cache_logical_len(x) for x in children]
+        if any(n is None for n in lens):
+            return None
+        return max(lens, default=0)
+    empty = getattr(c, "empty", None)
+    if callable(empty):
+        try:
+            if empty():
+                return 0  # nothing cached yet, so nothing to roll back
+        except Exception:
+            return None
+    return None
+
+
 def _prefix_cache_trim_amount(kv_cache: List[Any], prefix_len: int) -> Optional[int]:
     """Trailing tokens to drop so ``kv_cache`` keeps only its first ``prefix_len``.
 
@@ -678,11 +749,18 @@ def _prefix_cache_trim_amount(kv_cache: List[Any], prefix_len: int) -> Optional[
     taking an arbitrary rotation of the window and leaving the ring index stale:
     silent output corruption, or a broadcast crash once speculative decoding wraps
     the cache in ``BufferedRotatingKVCache``. Returns the number of tokens to drop
-    (``0`` when the whole cache is reusable), or ``None`` when an entry has already
-    evicted part of the prefix and the caller must cold-prefill instead.
+    (``0`` when the whole cache is reusable), or ``None`` when the caller must
+    cold-prefill instead -- because an entry's cached length cannot be
+    established, because it does not support ``trim()`` at all, or because it has
+    already evicted part of the prefix.
     """
-    cached_len = max((int(getattr(c, "offset", 0) or 0) for c in kv_cache), default=0)
+    lens = [_cache_logical_len(c) for c in kv_cache]
+    if any(n is None for n in lens):
+        return None
+    cached_len = max(lens, default=0)
     n_drop = max(0, cached_len - prefix_len)
+    if n_drop and not all(_cache_supports_trim(c) for c in kv_cache):
+        return None
     if n_drop and not all(_cache_fully_retained(c) for c in kv_cache):
         return None
     return n_drop
--- a/mlx_vlm/tests/test_generate.py
+++ b/mlx_vlm/tests/test_generate.py
@@ -28,6 +28,7 @@ from mlx_vlm.generate import ar as ar_module
 from mlx_vlm.generate import dispatch as dispatch_module
 from mlx_vlm.generate import normalize_resize_shape
 from mlx_vlm.models.cache import (
+    ArraysCache,
     BatchKVCache,
     BatchPoolingCache,
     BatchRotatingKVCache,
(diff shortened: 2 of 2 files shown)


lucasnewman:
@kylesyx Can you [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits and force push so we can clear the merge requirements?

pierre427:
Same defect class was fixed in mlx-lm — you may find the write-up useful, since the diagnosis
matches yours exactly (`ArraysCache` has no `trim`, falls through the retained-prefix guard
because it exposes none of `caches` / `start_position` / `max_size`, and hybrid models always
have those entries in the cache list).

mlx-lm **#1596** takes the other branch of the fork: instead of declining reuse, it makes these
caches trimmable by recording bounded state checkpoints at prefill chunk boundaries, so the
prompt cache can roll back to a shared prefix and the hit is preserved. That matters most for
exactly the shapes that trigger this crash — regenerate and edited-message turns on
Qwen3.5/3.6, where declining reuse means re-prefilling the whole shared prefix every turn.

Declining reuse is the right stop-gap and strictly better than crashing. If you want the cache
hit back afterwards, the checkpoint approach ports over cleanly — the interleave pattern you
describe (3 linear-attention layers per full-attention layer) is the same one it was built for.

kylesyx:
@lucasnewman Signed and force-pushed — `daa342d` now shows as verified.

While re-pushing I found the branch had picked up a `Merge branch 'main'` commit from a third party, so I rebased onto current `main` instead and the branch is a single signed commit again. Rebased cleanly; no conflicts with #1713, which landed in the same area.

Re-ran after the rebase: `TestPrefixCacheReuseTrim` 11 passed, and `test_generate.py` + `test_apc.py` + `test_speculative.py` 311 passed. The one failure, `test_qwen_target_verify_quantized_linear_matches_singleton_batch_path`, is pre-existing and fails identically on unmodified `main` (bfloat16 exact-equality on a quantized matmul).

@pierre427 Thanks — agreed the checkpoint approach is the better end state, and it's the one worth having for regenerate/edited-turn flows. I'd rather keep this PR as the minimal stop-gap since it's scoped to the guard and doesn't touch the cache classes; happy to see ml-explore/mlx-lm#1596 land and follow up separately once there's an upstream contract to mirror.

pierre427:
Makes sense — minimal guard now, checkpoint contract when there's an upstream one to mirror. Thanks for checking the #1713 interaction.

diffen:
Independent reproduction and validation on the newer Qwen3.8 family:

- Released environment: mlx-vlm 0.6.13, mlx-lm 0.31.3, mlx 0.32.0, Gradio 6.24.0.
- Model/UI: mlx-community/Qwen3.8-27B-nvfp4 through mlx_vlm.chat_ui.
- Before this patch: turn 1 completed, then turn 2 and its retry both failed in stream_generate at dispatch.py:859 with AttributeError: ArraysCache has no attribute trim. The Python/Gradio server remained alive, confirming this was the recurrent-cache rollback path rather than an OOM.
- On this PR branch: TestPrefixCacheReuseTrim passes 11/11.
- Exact-model end-to-end check: a three-turn chat completed with correct retained context on turns 2 and 3 and no server traceback. Reported peak MLX memory was 16.34-16.35 GB.

This confirms that Qwen3.8-27B-nvfp4 is also affected and that declining the unsafe partial-prefix reuse fixes its multi-turn chat path.

lucasnewman:
@kylesyx Thanks for reporting the issue. There are still come cases that need to be covered here which you can see from these tests which will fail even with your changes. Let me know if you'd like to make the changes or have us handle it.

```py
def test_nested_non_trimmable_cache_is_not_reusable(self):
    # Hybrid recurrent models can wrap an attention cache and recurrent state
    # in a CacheList. CacheList has no top-level offset, so the cached length
    # must be obtained from its children before deciding whether trimming is
    # required.
    c = CacheList(self._fill(KVCache(), 20), ArraysCache(2))
    assert c.size() == 20
    assert dispatch_module._prefix_cache_trim_amount([c], 8) is None

def test_nonempty_recurrent_cache_without_length_is_not_reusable(self):
    # ArraysCache state does not expose a logical token length or offset.
    # Assigning a synthetic offset in a test hides that production behavior;
    # without a trustworthy length, divergent-prefix reuse must be declined.
    c = ArraysCache(1)
    c[0] = mx.zeros((1, 4))
    assert not c.empty()
    assert c.size() == 0
    assert dispatch_module._prefix_cache_trim_amount([c], 8) is None
    
def test_nested_trimmable_cache_uses_child_length(self):
    # A composite made entirely from trimmable caches should still reuse the
    # shared prefix; this guards against conservatively rejecting every
    # CacheList while fixing the recurrent-cache case above.
    c = CacheList(self._fill(KVCache(), 20), self._fill(KVCache(), 20))
    assert c.size() == 20
    assert dispatch_module._prefix_cache_trim_amount([c], 8) == 12
```

Lazarus-931:
@kylesyx, also wouldn't this not run for mamba/mamba2/rwkv7 (pure-ArraysCache) and inkling/zaya1_vl/falcon_h1 (CacheList-per-layer) models?

pierre427:
Yes. There is a second issue here: _prefix_cache_trim_amount derives cached_len only from the top-level .offset attribute. Bare ArraysCache and CacheList expose no top-level offset, so both collapse to cached_len == 0; that makes n_drop == 0 and bypasses both _cache_supports_trim and _cache_fully_retained.

For pure recurrent and per-layer CacheList models, the likely failure is therefore silent stale-state reuse rather than the original AttributeError: input_ids is sliced to the shared prefix while the cache still represents the old full sequence. The mixed Qwen case happens to work because a sibling KVCache contributes a nonzero offset.

I think exact append and rollback need separate decisions: use PromptCacheState.token_ids to prove that no rollback is required, recursively obtain length from CacheList children where possible, and fail closed for a nonempty cache whose logical length is unknown. A production-shaped ArraysCache exact-prefix test should also avoid assigning synthetic .offset, since that attribute does not exist in the real cache.

lucasnewman (changes requested):
Needs fixes for failing test cases identified.

kylesyx:
@lucasnewman I'll take it — thanks for writing the cases out, they land on something my patch missed.

They fail *above* the guard I added rather than inside it. `_prefix_cache_trim_amount` derives `cached_len` from a top-level `offset`, which neither `CacheList` nor a bare `ArraysCache` exposes, so both collapse to `cached_len == 0` → `n_drop == 0` → `_cache_supports_trim` and `_cache_fully_retained` are *both* short-circuited and reuse is accepted. Qwen3.5/3.6 only got fixed by my patch because a sibling `KVCache` in the same list happens to contribute a nonzero offset.

Which means @Lazarus-931 is right about mamba/mamba2/rwkv7 and the per-layer-`CacheList` models (inkling/zaya1_vl/falcon_h1), and @pierre427's reading of it is right too: there the failure was never the `AttributeError`, it's silent stale-state reuse — the prompt is sliced to the shared prefix while the cache still holds the old full sequence.

Your second case is also a fair hit on my tests. `c.offset = 100` invents an attribute the real `ArraysCache` doesn't have, so those assertions described a shape that can't occur in production. Rewritten without it.

**Pushed:** a recursive `_cache_logical_len()` that

- prefers `offset` wherever it exists, so flat and rotating caches keep their exact current semantics — note `RotatingKVCache.size()` clamps to `max_size`, so `size()` is deliberately *not* used as the length source (there's a test pinning this);
- recurses into `CacheList` children for the length;
- falls back to `0` only for a provably empty cache, and returns `None` for a non-empty cache whose logical length can't be established — fail closed, since the alternative is reusing stale state.

Still confined to the guard; no cache classes touched.

**Tests** — your three are in verbatim, plus coverage for the shapes above:

```
TestPrefixCacheReuseTrim                        17 passed
full suite, main vs. this branch          2148 → 2157 passed, 0 failed
```

8 of the 9 added tests fail on unmodified `main` (the 9th is the all-`KVCache` `CacheList` warm-path case, which must keep returning `0` either way).

**One consequence worth naming rather than burying:** pure-recurrent models now cold-prefill every turn, since their length isn't knowable from the cache alone. That's what your second case asks for and it's the right default, but it does give up a cache hit they currently get. @pierre427's `PromptCacheState.token_ids` suggestion is how to win it back — prove the turn is a pure append and the rollback question never arises. I'd rather land that separately than widen this one.

Also rebased onto current `main` and squashed back to a single signed commit — the branch had picked up two `Merge branch 'main'` commits. The last CI run here was from Aug 20 against an older base; this push should get a fresh one.

songofhawk:
Independent confirmation on the current stable release and a different Qwen3.8 quant/UI path:

- Hardware: M2 Pro, 32 GB unified memory
- Runtime: mlx-vlm 0.6.17, mlx-lm 0.31.3, mlx 0.32.2
- Model: orcarouter/Qwen3.8-27B-Uncensored-MLX, affine 4-bit/g64
- Entry point: mlx_vlm.chat with a local snapshot path
- Result: multi-turn chat eventually reports `AttributeError: ArraysCache object has no attribute trim`; the process remains alive, so this is the prompt-cache rollback path rather than OOM or a corrupt checkpoint.

I also checked out the latest PR head (`08b0c9e`) and ran the focused suite locally:

```
python -m pytest mlx_vlm/tests/test_generate.py -k TestPrefixCacheReuseTrim -q
17 passed, 120 deselected
```

This extends the earlier Qwen3.8 NVFP4/chat_ui confirmation to the affine 4-bit checkpoint and the CLI chat entry point on v0.6.17. The additional nested/non-length-bearing cache cases requested in review are present and passing at the current head. A fresh maintainer review would be helpful because the PR is still blocked by the earlier changes-requested state.

kylesyx:
Thanks @songofhawk — useful to have the affine 4-bit checkpoint and the `mlx_vlm.chat` CLI path covered. The earlier confirmation was NVFP4 through `chat_ui`, so that's a different quant *and* a different entry point landing on the same rollback.

One thing that isn't visible from the PR page: **CI has not actually run on the current head.** The `Test PRs` workflow on `08b0c9e` came back `action_required` with zero jobs started — it's waiting on a maintainer to approve workflows for this fork PR. So the red X still showing here is the Aug 20 run, which predates both the rebase and these changes.

@lucasnewman whenever you have a moment: your three cases are in verbatim, plus the pure-recurrent (mamba/mamba2/rwkv7) and per-layer-`CacheList` shapes @Lazarus-931 raised. Locally on `08b0c9e`:

```
TestPrefixCacheReuseTrim                 17 passed
full suite, main vs. this branch   2148 -> 2157 passed, 0 failed
```

`main` has moved 84 commits since this branch's base; GitHub still reports it `MERGEABLE`, but say the word and I'll rebase onto current `main`.

Blaizzy:
Thanks @kylesyx!

Closing this PR because it's fixed in main, but I encourage you to look into the many other open issues.

Issue #1766, opened 2026-07-31
Mage-VL: dense/frame-sampled video input support


josiah-nelson:
The built-in `mage_vl` model supports image inference but raises `NotImplementedError` for video input. I'd like to contribute dense/frame-sampled video support matching upstream `microsoft/Mage-VL` processor semantics: the upstream frame-sampling policy, per-frame timestamps (`tokens_per_second=1`, `temporal_patch_size=1`), explicit `patch_positions` in `(t, row, col)`, 2×2 spatial merge, and 3D RoPE.

I have a PyTorch-oracle parity harness (goldens for preprocessing tensors, vision features, final logits, and deterministic generations against the pinned Microsoft checkpoint) and would validate the implementation against it.

Before opening the PR, any preferences on:

- where video frame-sampling policy should live (processor kwargs vs. video preprocessor config)?
- how you'd like video inputs shaped in the processor API?

A follow-up proposal for codec-sparse patch selection (sparse `patch_positions`) would come separately.

Lazarus-931:
hi @josiah-nelson, thanks for wanting to contribute.

i think we can reuse  `load_video` from utils.py, no need to re-make for Mage-VL specific, as for the processor api, it should match existing video models where `video` arg accepts list of decoded clips and returns `video_grid_thw`, plus expanded tokens like qwen2_5_vl!

feel free to open a sep pr for codec-sparse patch_positions as well!

josiah-nelson:
Thanks! agreed on both points. I'll match the qwen2_5_vl processor and use `load_video` as the single loader. 

Mage-VL's prompt embeds per-frame timestamps from the source video, and its reference sampling uses `int(duration × fps)` capped at `max_frames` without `frame_factor` rounding. I'm thinking the best bet would be a backward-compatible tweak to `load_video` that optionally returns the sampled frame indices + native fps, and honors `frame_factor=1`. That keeps one loader while letting timestamp-aware models (not just mage) reconstruct exact source timing. 

Will open the dense-video PR on that basis, and codec-sparse `patch_positions` as a follow-up PR!

PR #1836, opened 2026-08-10
feat(server): video_url falls back to frames when processor lacks video support

x7peeps:
# feat(server): video_url falls back to frames when processor lacks video support

## The problem

The OpenAI-compatible `/v1/chat/completions` endpoint accepts `video_url`
content parts but routes them straight to the model's processor. Processors
that have no native video path raise `NotImplementedError` when asked to
ingest the `videos` kwarg — e.g. Microsoft's **Mage-VL** (`mage_vl` model
class), where `processing_mage_vl.py` explicitly raises:

```python
raise NotImplementedError(
    "mage_vl video input needs the upstream codec pipeline "
    "(not available as a torch-free path yet); image input only for now"
)
```

The end result for clients (Hermes Agent, Open WebUI, anything OpenAI-compatible)
is a hard 500 Internal Server Error and a silently-dropped video.

The **CLI already handles this** via `generate/dispatch.py` (line 1298 onward):
when `processor_handles_video(processor) == False`, it samples frames, subsamples
to ≤16 evenly-spaced frames, and re-attaches them as ordered images. The server
just never gained the same guard.

## What changed

**`mlx_vlm/server/openai.py` — `chat_completions_endpoint` (~50 lines)**

After `get_cached_model()` resolves the model/processor and before
`apply_chat_template()` is called, when `videos` are present and the processor
does not support videos natively, the server now:

1. Imports the same helpers the CLI uses (`processor_handles_video`,
   `sample_video_frames`, `subsample_evenly`) from `mlx_vlm.generate.video`.
2. Samples evenly-spaced frames at 2 fps.
3. Subsamples to ≤16 frames.
4. Appends the PIL frames to `images` and clears `videos` so `prepare_inputs`
   never sees the kwarg the processor rejects.

Processors that natively support videos (Qwen3-VL, Gemma4-VL, etc.) skip the
fallback entirely. Frame extraction failures log a warning and degrade to the
original behavior so the caller still sees a clear error.

**`mlx_vlm/tests/test_server.py` — two new tests**

| Test | Asserts |
|---|---|
| `test_chat_completions_video_url_falls_back_to_frames_when_processor_rejects_video` | Request returns 200; sampled frames appended to `num_images`; `videos` kwarg cleared |
| `test_chat_completions_video_url_keeps_video_kwarg_when_processor_supports_video` | Native-video path is not perturbed; `sample_video_frames` not called |

## Who should enable this

Anyone running a local `mlx-vlm` server with a model whose processor rejects
the `videos` kwarg. Today that includes Mage-VL (and any future codec-native
streaming model that ships an image-only processor). Upgrading the server is
enough — clients keep sending `video_url` content parts unchanged.

## Platform compatibility

| Platform | Status |
|---|---|
| macOS / Apple Silicon | ✅ primary verification target |
| Linux | ✅ works wherever ffmpeg + Python are available |
| Windows | ✅ same |
| Cloud-hosted mlx-vlm (rare) | ✅ no platform-specific code |

No new dependencies — `sample_video_frames` already calls `ffmpeg` via
`mlx_vlm.utils.load_video`, same as the CLI fallback.

## Quick-start

```bash
# Start the server with Mage-VL (or any image-only processor)
python -m mlx_vlm.server --model mlx-community/Mage-VL-8bit --port 8123

# Send a video_url content part — frames are sampled automatically
curl http://127.0.0.1:8123/v1/chat/completions \
  -H 'Content-Type: application/json' -d '{
    "model": "Mage-VL-8bit",
    "messages": [{"role": "user", "content": [
      {"type": "video_url", "video_url": {"url": "/path/clip.mp4"}},
      {"type": "text", "text": "What is happening?"}
    ]}]
  }'
```

Before this fix: `500 Internal Server Error`. After: a normal completion.

## Verification (self-tested, real output)

M5 Max 128GB, `mlx-vlm server` + `mlx-community/Mage-VL-8bit`:

```
$ curl … /v1/chat/completions  (video_url)
→ 200 OK  prompt_tokens=8219 images=16 audio=0 videos=0
→ "The video shows a sports broadcast, likely from a football match,
    featuring a panel of commentators. … BBC Sport logo … live broadcast
    from the BBC."   [12.5s end-to-end, prefill 1007 tok/s, decode 60 tok/s]
```

Server log confirms the fallback activation:

```
INFO — Processor MageVLProcessor has no native video support;
       sampling frames for 1 submitted video(s)
INFO — Generation queued: request=… prompt_tokens=8219 max_tokens=200
       images=16 audio=0 videos=0
```

Test suite: `mlx_vlm/tests/test_server.py` (+2 new tests, 267 total) passes.
`black` / `isort` / `autoflake` (the repo's pre-commit hooks) clean.

## Related

- Hermes Agent PR [#82996](https://github.com/NousResearch/hermes-agent/pull/82996) —
  complementary client-side guard: Hermes `video_analyze` also extracts frames
  for local VLM endpoints. Either fix alone solves the user-visible breakage;
  together they make every OpenAI-compatible client able to send video to any
  mlx-vlm-served processor.
- Upstream tracker [mlx-vlm issue #1766](https://github.com/Blaizzy/mlx-vlm/issues/1766) —
  the Mage-VL native video path that will eventually subsume this fallback
  when the codec pipeline becomes torch-free. Until then, this guard keeps
  every codec-native model usable through the OpenAI server.

## Footprint

- 2 files, +210/-1
- Zero new dependencies
- All existing tests pass unchanged (267 passed)
- Mimics an established CLI path (generate/dispatch.py) rather than inventing
  a new abstraction; behavior matches ecosystem expectations


lucasnewman:
@x7peeps Thanks for calling this out! I'd prefer not to inline a bunch of the logic into the server-side endpoints, so I put up a little bit of a different approach in https://github.com/Blaizzy/mlx-vlm/pull/1924 which should solve the issue for you.

PR #1837, opened 2026-08-10
fix(prompt_utils): render tool-call turns that carry no text (#1785)

### LINKED ISSUE #1785: /v1/messages: assistant tool-call turn with no text crashes chat template with 'NoneType' object is not iterable
## Summary

A multi-turn tool-use conversation through the Anthropic-compatible `/v1/messages` endpoint (`mlx_vlm/server/anthropic.py`) crashes with a 500 the moment the model's tool call has no accompanying text — an ordinary, common shape for any agentic tool-use loop, not an edge case. Root cause confirmed via the real server-side traceback (below), not a guess.

This looks like the same underlying defect class as #1034 (open) and PR #1024 (closed, unmerged) — both about `mlx_vlm.server` sending `content: null` for text-free tool-call turns — but manifesting on the newer `/v1/messages` endpoint (added in #1196, closing #1060) with a harder failure mode: a hard crash instead of #1034's infinite loop.

## Environment

- `mlx-vlm` 0.6.8 (confirmed on PyPI: 78 commits between 0.6.8→0.6.9 touch new model support / embeddings / KV-cache fixes — nothing in the Anthropic endpoint or chat-template/tool-call path, so 0.6.9 does not appear to fix this)
- Model: `mlx-community/Qwen3-VL-32B-Instruct-4bit`
- Server: `mlx_vlm.server --model mlx-community/Qwen3-VL-32B-Instruct-4bit --port 8082`
- Client: real `anthropic` Python SDK (`AsyncAnthropic`, `base_url` pointed at the local server), non-streaming `messages.create()`
- macOS, Apple Silicon

## Reproduction

1. Send a `/v1/messages` request with one user turn + a tool schema, where the model's first response is a tool call with no accompanying text (`finish_reason=tool_use`, e.g. a simple "navigate to this page" style tool with nothing to narrate).
2. Send the follow-up request that continues the conversation (assistant's tool-call turn + the tool result appended) — this is a completely ordinary second turn in any tool-use loop.
3. The server's own log shows iteration 1 complete normally (`finish_reason: tool_use`), then the second `POST /v1/messages` returns 500.

## Real traceback (captured from the server's own terminal)

```
2026-08-03 20:56:10,063 - INFO - Request completed: endpoint=/v1/messages model=mlx-community/Qwen3-VL-32B-Instruct-4bit stream=False backend=continuous_batching prompt_tokens=3437 generated_tokens=28 elapsed=15.121s prefill=243.5 tok/s decode=29.9 tok/s finish_reason=tool_use in_flight=0
INFO:     127.0.0.1:57149 - "POST /v1/messages HTTP/1.1" 200 OK
  File ".../mlx_vlm/server/anthropic.py", line 495, in anthropic_messages_endpoint
                       ^^^^^^^^^^^^^^^^^^^^
  File ".../mlx_vlm/server/app.py", line 716, in <lambda>
    apply_chat_template=lambda *args, **kwargs: _server_package_attr(
                                                ^^^^^^^^^^^^^^^^^^^^^
  File ".../mlx_vlm/prompt_utils.py", line 941, in apply_chat_template
    return get_chat_template(processor, messages, add_generation_prompt, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../mlx_vlm/prompt_utils.py", line 782, in get_chat_template
    return template_processor.apply_chat_template(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../transformers/processing_utils.py", line 2165, in apply_chat_template
    prompt, generation_indices = render_jinja_template(
                                 ^^^^^^^^^^^^^^^^^^^^^^
  File ".../transformers/utils/chat_template_utils.py", line 581, in render_jinja_template
    rendered_chat = compiled_template.render(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../jinja2/environment.py", line 1295, in render
    self.environment.handle_exception()
  File ".../jinja2/environment.py", line 942, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "<template>", line 64, in top-level template code
TypeError: 'NoneType' object is not iterable
INFO:     127.0.0.1:57149 - "POST /v1/messages HTTP/1.1" 500 Internal Server Error
```

## Root cause

1. `mlx_vlm/server/anthropic.py`'s `_anthropic_messages_to_internal` (lines 358-363) builds the assistant message dict for a tool-call turn with no accompanying text as:
   ```python
   msg: Dict[str, Any] = {"role": message.role, "content": content_text}
   if tool_calls:
       msg["tool_calls"] = tool_calls
       if not content_text:
           msg["content"] = None
   ```
   `content: null` alongside `tool_calls` is a valid, OpenAI-spec-compliant shape for a tool-call-only turn.

2. Qwen3-VL's own chat template (`chat_template.jinja`, cached model snapshot) handles the assistant role as:
   ```jinja
   {%- elif message.role == "assistant" %}
       {{- '<|im_start|>' + message.role + '\n' }}
       {%- if message.content is string %}
           {{- message.content }}
       {%- else %}
           {%- for content_item in message.content %}
               {%- if 'text' in content_item %}
                   {{- content_item.text }}
               {%- endif %}
           {%- endfor %}
       {%- endif %}
   ```
   It only branches on "is this a string" vs. "assume iterable" — there's no `None` branch, so `content: None` falls into `for content_item in message.content` and Jinja raises `'NoneType' object is not iterable` (confirmed at the template's own line 64 in the traceback above).

The fix belongs on the server side (step 1), not the template — `content` should be coerced to `""` (or the key omitted) rather than `None` before calling `apply_chat_template`, matching one of the two fixes proposed in #1034 ("Server side — in the OpenAI-compat request handler, substitute a single space (or any non-empty placeholder) for empty/null `content` on assistant tool_call messages before template rendering. Defensive and template-agnostic.") — except this needs the identical treatment in `anthropic.py`'s own message-building path, not just `server.py`'s.

## Impact

This isn't cosmetic — for anyone driving a real multi-turn agentic tool-use loop against `/v1/messages` (e.g. a browser-automation agent), any tool call with no accompanying text kills the whole conversation with a 500 on the very next turn. In our case the underlying task the model was performing had actually already succeeded by the time the crash hit (real side effects had already landed), so the crash turned a real success into an opaque, uninformative failure for the calling application.

## Suggested fix

In `_anthropic_messages_to_internal` (`mlx_vlm/server/anthropic.py`), don't set `msg["content"] = None` for a tool-call-only turn — use `""` (or omit the key) instead, so it satisfies `message.content is string` in every chat template that follows this pattern, not just Qwen3-VL's. Happy to open a PR with this one-line change plus a regression test if that's useful.

Anai-Guo:
## Problem

A multi-turn tool-use conversation dies with a 500 as soon as the model emits a tool call with no accompanying text — an ordinary shape in any agentic loop, not an edge case:

```
File ".../mlx_vlm/prompt_utils.py", line 782, in get_chat_template
    return template_processor.apply_chat_template(...)
  ...
  File "<template>", line 64, in top-level template code
TypeError: 'NoneType' object is not iterable
```

## Root cause

Both compatibility layers build a text-free tool-call turn as `content: None`, which is the spec-compliant OpenAI shape:

- `mlx_vlm/server/anthropic.py:358-363` sets `msg["content"] = None` when `tool_calls` are present and `content_text` is empty.
- `mlx_vlm/server/openai.py:1650` forwards `message.content` verbatim, so a client sending `"content": null` lands the same way.

Chat templates commonly branch on *"is this a string"* and otherwise assume the value is iterable — Qwen3-VL's `chat_template.jinja` is representative:

```jinja
{%- if message.content is string %}
    {{- message.content }}
{%- else %}
    {%- for content_item in message.content %}   {# <- None reaches here #}
```

There is no `None` branch, so rendering raises `TypeError: 'NoneType' object is not iterable`.

## Fix

Normalize `content: None` to `""` on assistant turns that carry `tool_calls`, immediately before the template is rendered.

`get_chat_template` is the single choke point shared by `/v1/chat/completions`, `/v1/messages`, `/v1/responses` and the library `generate()` path, so one normalization covers all of them and stays template-agnostic rather than patching each endpoint separately. An empty string takes the `is string` branch and renders identically to a turn whose text genuinely is empty, so no template that already handles this shape changes behaviour. The caller's message list is never mutated — a shallow copy is made only when a message actually needs patching.

## Verification

Reproduced against a Jinja template reduced from Qwen3-VL's assistant branch, on this commit's parent vs. this commit:

| | before | after |
|---|---|---|
| `content: None` + `tool_calls` | `TypeError: 'NoneType' object is not iterable` | renders, tool call preserved, turn closed, generation prompt emitted |

Four regression tests added to `mlx_vlm/tests/test_prompt_utils.py`:

- `test_null_content_tool_call_turn_renders` — the crash case
- `test_null_content_renders_like_empty_string` — `None` and `""` produce byte-identical prompts
- `test_caller_messages_are_not_mutated` — no write-back into the caller's list
- `test_null_content_without_tool_calls_is_left_alone` — only tool-call turns are touched

`mlx_vlm/tests/test_prompt_utils.py` passes in full (40 tests). `black` 26.3.1 and `isort --profile=black` 5.13.2 (the pinned pre-commit versions) leave both files unchanged.

## Scope

Closes #1785.

Partially addresses #1034: that report covers both `content: null` **and** `content: ""` for Gemma 4. The `null` half is fixed here. The `""` half is not — Gemma 4's template gates turn closure on `not message.get('content')`, which is already true for an empty string, so it needs the template-side fix (option 1 in that issue) or a non-empty placeholder. Substituting a non-empty placeholder globally would inject stray text into every other model's prompt, so I have deliberately left it out of this change rather than trade a crash for a silent prompt regression.

Credit to @agent-smith-ddp for the precise server-side traceback in #1785 and to @reza-yousefi for identifying the empty-content/tool-call class in #1034. @agent-smith-ddp offered a PR on the `anthropic.py` one-liner — happy to close this in favour of that if preferred; I went to the shared choke point because the same shape reaches the template from three endpoints.

### DIFF
--- a/mlx_vlm/prompt_utils.py
+++ b/mlx_vlm/prompt_utils.py
@@ -582,6 +582,39 @@ def get_message_json(
     )
 
 
+def _normalize_tool_call_content(
+    messages: List[Dict[str, Any]],
+) -> List[Dict[str, Any]]:
+    """Coerce ``content: None`` to ``""`` on assistant tool-call turns.
+
+    A tool-call turn carrying no accompanying text is an ordinary shape in an
+    agentic loop, and both the OpenAI and Anthropic compatibility layers
+    forward it as ``{"content": None, "tool_calls": [...]}``. Chat templates
+    commonly branch on ``message.content is string`` and otherwise iterate the
+    value, so ``None`` raises ``'NoneType' object is not iterable`` during
+    rendering. An empty string takes the string branch and renders identically
+    to a turn whose text really is empty.
+
+    The input list is left untouched; a shallow copy is made only when a
+    message actually needs patching.
+    """
+    normalized = None
+
+    for index, message in enumerate(messages):
+        if (
+            isinstance(message, dict)
+            and message.get("content") is None
+            and message.get("tool_calls")
+        ):
+            if normalized is None:
+                normalized = list(messages)
+            patched = dict(message)
+            patched["content"] = ""
+            normalized[index] = patched
+
+    return messages if normalized is None else normalized
+
+
 def get_chat_template(
     processor,
     messages: List[Dict[str, Any]],
@@ -590,6 +623,8 @@ def get_chat_template(
     **kwargs,
 ) -> Any:
     """Apply chat template using processor's tokenizer."""
+
+    messages = _normalize_tool_call_content(messages)
 
     def _get_image_token() -> str:
         if processor is None:
--- a/mlx_vlm/tests/test_prompt_utils.py
+++ b/mlx_vlm/tests/test_prompt_utils.py
@@ -762,3 +762,126 @@ class TestModelSpecificPromptContracts:
             "text",
         ]
         assert result[0]["content"][-1]["text"] == "OCR:"
+
+
+class TestToolCallOnlyAssistantTurn:
+    """A tool-call turn carrying no text must still render (#1785, #1034)."""
+
+    # Faithful reduction of the Qwen3-VL assistant branch: it checks whether
+    # content is a string and otherwise assumes the value is iterable.
+    ASSISTANT_BRANCH_TEMPLATE = (
+        "{%- for message in messages %}"
+        '{%- if message.role == "assistant" %}'
+        "{{- '<|im_start|>assistant\n' }}"
+        "{%- if message.content is string %}{{- message.content }}"
+        "{%- else %}"
+        "{%- for content_item in message.content %}"
+        "{%- if 'text' in content_item %}{{- content_item.text }}{%- endif %}"
+        "{%- endfor %}"
+        "{%- endif %}"
+        "{%- if message.tool_calls %}"
+        "{%- for tc in message.tool_calls %}"
+        "{{- '<tool_call>' + tc.function.name + '</tool_call>' }}"
+        "{%- endfor %}"
+        "{%- endif %}"
+        "{{- '<|im_end|>\n' }}"
+        "{%- else %}"
+        "{{- '<|im_start|>' + message.role + '\n'"
+        " + (message.content or '') + '<|im_end|>\n' }}"
+        "{%- endif %}"
+        "{%- endfor %}"
+        "{%- if add_generation_prompt %}"
+        "{{- '<|im_start|>assistant\n' }}{%- endif %}"
+    )
+
+    @staticmethod
+    def _processor(template):
+        from jinja2 import Environment
+
+        class _Processor:
+            chat_template = template
+
+            def apply_chat_template(
+                self,
+                messages,
+                tokenize=False,
+                add_generation_prompt=False,
+                **kwargs,
+            ):
+                compiled = Environment().from_string(template)
+                return compiled.render(
+                    messages=messages,
+                    add_generation_prompt=add_generation_prompt,
+                    **kwargs,
+                )
+
+        return _Processor()
+
+    @staticmethod
+    def _messages(content):
+        return [
+            {"role": "user", "content": "Open the docs page."},
+            {
+                "role": "assistant",
+                "content": content,
+                "tool_calls": [
+                    {
+                        "id": "call_1",
+                        "type": "function",
+                        "function": {
+                            "name": "navigate",
+                            "arguments": {"url": "/docs"},
+                        },
+                    }
+                ],
+            },
+            {"role": "tool", "tool_call_id": "call_1", "content": "ok"},
+        ]
+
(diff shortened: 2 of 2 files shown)


lucasnewman:
Fixed in https://github.com/Blaizzy/mlx-vlm/pull/1864

PR #1838, opened 2026-08-10
Add Muse Glimmer model support

Blaizzy:
## Summary

- add a native MLX-VLM port for the `muse_glimmer` architecture
- implement the 52-layer hybrid sliding/full-attention language model, gated GQA, centered RMS normalization, output scaling, and checkpoint key sanitization
- implement the ViT-G/14 vision tower, window/full attention schedule, positional interpolation, pixel shuffle, adapter, and language projection
- add the Muse Glimmer image processor, multimodal token expansion, AutoProcessor registration, and prompt formatting
- optimize BF16 execution without quantization or projection fusion: compile exact centered RMSNorm and vision GELU/rotary pointwise work, keep masking on-device, reuse host grid metadata, and skip identity reorders
- add focused configuration, normalization, masking, processor, cache, quantization, multimodal-forward, and optimization exactness tests

## Why

`meta-models/Muse-Glimmer-30B` was not recognized or loadable by MLX-VLM. This adds the architecture and preprocessing needed to load its original checkpoint and run text and image generation through both the CLI and OpenAI-compatible server.

The canonical model path is:

```text
meta-models/Muse-Glimmer-30B
```

## BF16 numerical fix

Transformers computes centered RMSNorm by normalizing and applying `(1 + weight)` in FP32, then casting once to BF16. The generic BF16 fast path rounded before applying the centered weight. Across 52 cached decode layers, that error made the two leading candidates tie at generated token 189 and changed the greedy token.

The exact operation order is compiled to recover most of its execution overhead. Model weights and surrounding activations remain BF16.

## BF16 performance

Measured with the full checkpoint and a 200-token generation:

| Metric | Initial port | Current exact path |
| --- | ---: | ---: |
| Prompt, fresh process | 102.411 tok/s | 96.028 tok/s |
| Generation | 12.067 tok/s | 11.897 tok/s |
| Peak memory | 59.655 GB | 59.701 GB |
| Vision encode (16x16 patch grid) | 72.555 ms | 68.104 ms |

Vision latency remains 6.1% lower. Compilation improves the exact centered-norm generation result from 11.515 tok/s uncompiled to 11.897 tok/s. The remaining prompt cost is first-use compilation.

## Numerical validation

Compared with the Transformers BF16 implementation:

- all 200 greedy generated token IDs are identical
- text logits: 0.4979% relative L2, cosine 0.999988, same top-1 token
- vision features: 8.3509% relative L2, cosine 0.996512
- multimodal logits: 3.8328% relative L2, cosine 0.999397, same top-1 token

The previous fast centered-norm path had 0.8558% text-logit relative L2 and diverged at generated token 189. The remaining logits are not bit-identical because MLX and MPS use different BF16 linear and attention kernels.

The deliberate FP32 vision rotary calculation is retained; compilation only groups its elementwise operations.

## Validation

- `133 passed, 2 skipped` across `test_muse_glimmer.py` and `test_processors.py`
- Ruff E/F checks passed
- `git diff --check` passed
- strict loading passed against the full 59.6 GB checkpoint
- generate CLI passed using `--model meta-models/Muse-Glimmer-30B` and returned `MODEL_NAME_OK`
- server startup passed using `--model meta-models/Muse-Glimmer-30B`
- an OpenAI-compatible request with `"model": "meta-models/Muse-Glimmer-30B"` returned HTTP 200 and `SERVER_MODEL_NAME_OK`
- image generation correctly identified two cats in the sample image


johntdavies:
Thanks for putting this together — I converted `meta-models/Muse-Glimmer-30B` with this branch and hit one bug worth flagging before it merges. Opened #1839 against `pc/muse-glimmer-30b` with a fix and a regression test.

**Quantized conversions come out broken.** `NormedEmbedding` subclasses `nn.Embedding`, so `nn.quantize` calls the inherited `to_quantized()`, which returns a plain `QuantizedEmbedding` and drops the weightless `embed_norm`. Unnormalized embeddings then enter the residual stream.

The nasty part is that it also bites at **load** time, not just convert time — `nn.quantize` runs again when loading a quantized checkpoint, so correct weights on disk don't save you. Any 4-bit conversion published from this branch would be broken for everyone downloading it, and it fails silently: full speed, fluent-looking output, wrong content.

4-bit / `group_size=64`:

```
'The capital of France is' -> [('\n', 11.5), (' ', 11.31), ('\n\n', 11.31)]
```

`--quant-predicate mixed_4_6` fails the same way — it protects `lm_head`/`v_proj`/`down_proj` but not `embed_tokens`.

Your architecture port itself is fine, which is what took me a while to establish. bf16 through this same code gives:

```
'The capital of France is' -> [(' Paris', 17.0), (':', 13.5), (' **', 13.44)]
```

I also diffed the port against the reference `MuseGlimmer` implementation in `transformers` 5.15.0 — attention, centered RMSNorm, NoPE on the global layers, `output_multiplier`, softcapping and SwiGLU all match. Swapping the bf16 embedding table back into the 4-bit model restored it exactly, and a manually re-wrapped *4-bit* embedding worked too, so bit width was never the problem.

The fix follows the `FusedLinear.to_quantized()` precedent in `mlx_lm/models/afm7.py`. No checkpoint keys change, since `embed_norm` is weightless.

After it, 4-bit matches bf16 exactly, and end-to-end on an M4 Pro 64 GB (4-bit `group_size=64`, 5.216 bits/weight, 18 GB, 19.55 GB peak):

- Text: `Hello` -> `Hello! How can I help you today?` at 16.4 tok/s
- Vision: correctly described three shapes, colours and positions in a synthetic test image

Happy to rework it if you'd rather solve it a different way — e.g. handling this generically in the quantization path so any norm-carrying embedding is covered, rather than per-model.

https://claude.ai/code/session_01BRUXXJQrVfBDSu8LfzKzt9

Blaizzy:
Thanks a lot @johntdavies, I was still iterating on it :)
 
 There is a simpler fix I've applied here.

lucasnewman (approved):

PR #1839, opened 2026-08-10
Preserve embed_norm when quantizing Muse Glimmer embeddings

johntdavies:
Targets `pc/muse-glimmer-30b` (the branch behind #1838) rather than `main`, so it can be merged straight into that PR.

## Problem

Quantized conversions of `meta-models/Muse-Glimmer-30B` produce a broken model. `NormedEmbedding` subclasses `nn.Embedding`, so `nn.quantize` calls the inherited `to_quantized()`, which returns a plain `QuantizedEmbedding` and silently drops the weightless `embed_norm`. Unnormalized embeddings then flow into the residual stream.

Two things make this easy to miss:

- **It fails silently.** The model loads, runs at full speed, and emits fluent-looking text. Only the content is wrong.
- **It bites at load time, not just convert time.** `nn.quantize` runs again when loading a quantized checkpoint, so the norm is dropped a second time. Correct weights on disk do not help — anyone downloading a 4-bit conversion gets the broken model.

## Repro

4-bit, `group_size=64`, on the released checkpoint:

```
'The capital of France is' -> [('\n', 11.5), (' ', 11.31), ('\n\n', 11.31)]
```

`--quant-predicate mixed_4_6` fails identically: it protects `lm_head`/`v_proj`/`down_proj`, but not `embed_tokens`.

The same port at bf16 is fine, which is what localized this to quantization rather than the architecture:

```
'The capital of France is' -> [(' Paris', 17.0), (':', 13.5), (' **', 13.44)]
```

Swapping the bf16 embedding table back into the 4-bit model restored it exactly, and a manually re-wrapped *4-bit* embedding also worked — so bit width was never the issue, the missing norm was.

## Fix

Add `QuantizedNormedEmbedding` and a `to_quantized()` override that carries the norm across. This follows the existing precedent in `mlx_lm/models/afm7.py`, where `FusedLinear.to_quantized()` returns a matching quantized subclass; it also threads `mode`/`**kwargs` for current MLX quantization modes.

No checkpoint keys change — `embed_norm` is weightless — so existing conversions keep loading, and they start behaving correctly once this is in.

## Verification

After the fix, 4-bit logits match bf16 exactly:

```
'The capital of France is' -> [(' Paris', 17.0), (':', 13.5), (' **', 13.4)]
```

End-to-end on `meta-models/Muse-Glimmer-30B`, 4-bit `group_size=64` (5.216 bits/weight, 18 GB), M4 Pro 64 GB:

- Text: `Hello` -> `Hello! How can I help you today?` (58 prompt tokens @ 93 tok/s, 16.4 tok/s generation, 19.55 GB peak)
- Vision: correctly identified three shapes, their colours and positions in a synthetic test image

Added `test_quantization_preserves_embedding_norm`, which asserts behaviourally rather than on the class name. Without the fix it fails with `AssertionError: quantization dropped embed_norm: QuantizedEmbedding has no weightless embedding norm`; with it, all 7 tests in the file pass.

https://claude.ai/code/session_01BRUXXJQrVfBDSu8LfzKzt9

### DIFF
--- a/mlx_vlm/models/muse_glimmer/language.py
+++ b/mlx_vlm/models/muse_glimmer/language.py
@@ -36,6 +36,18 @@ class CenteredRMSNorm(nn.Module):
         return normalized * (1.0 + self.weight)
 
 
+class QuantizedNormedEmbedding(nn.QuantizedEmbedding):
+    """Quantized NormedEmbedding that keeps the weightless embedding norm.
+
+    ``nn.Embedding.to_quantized`` returns a plain ``QuantizedEmbedding``, which
+    would silently drop ``embed_norm`` and feed unnormalized embeddings into the
+    residual stream.
+    """
+
+    def __call__(self, inputs: mx.array) -> mx.array:
+        return self.embed_norm(super().__call__(inputs))
+
+
 class NormedEmbedding(nn.Embedding):
     def __init__(self, vocab_size: int, hidden_size: int, eps: float):
         super().__init__(vocab_size, hidden_size)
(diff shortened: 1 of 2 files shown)


lucasnewman:
@johntdavies Thanks, this is covered in https://github.com/Blaizzy/mlx-vlm/pull/1838 now.

PR #1842, opened 2026-08-11
Add Muse Glimmer DFlash assistant support

Blaizzy:
## Summary

- add a dedicated MLX backend for `meta-models/Muse-Glimmer-30B-assistant`
- auto-detect `muse_glimmer_assistant` checkpoints as DFlash drafters
- capture the target layers expected by the published assistant and support speculative KV-cache rollback
- match the Transformers assistant architecture, raw target embedding lookup, bidirectional sliding attention, and checkpoint schema
- optimize Muse DFlash with compiled residual/norm transitions, pipelined context projection, adaptive startup depth, and a dedicated batch-1 server loop
- avoid per-round allocator-cache flushes
- document generate CLI and server usage

## Impact

Muse Glimmer 30B can use its published BF16 5-layer, 16-token DFlash assistant through both the generation CLI and the OpenAI-compatible server. The optimization does not change weights, quantize the assistant, add fused model kernels, or bypass the target model's output scaling and softcap.

Throughput remains prompt-dependent because draft acceptance changes with the tokenized prefix. The chat server keeps its original first verifier shape to preserve its better acceptance path.

## Performance

Target: local 4-bit Muse model at `/tmp/muse-glimmer-publish.kLaN71/4bit-quantized-embed`  
Drafter: `meta-models/Muse-Glimmer-30B-assistant`  
Greedy decoding, up to 200 tokens, batch size 1

| Path | Before | After | Acceptance after |
|---|---:|---:|---:|
| Raw generate path, cold | 52.14 tok/s | 55.33 tok/s | 3.10 tokens/round |
| Raw generate path, warm | 52.23 tok/s | 55.39 tok/s | 3.10 tokens/round |
| Streaming chat server, warm | 43.60 tok/s | 45.20 tok/s | 2.44 tokens/round |

- raw prompt: 15 tokens; generation stopped at EOS after 196 tokens; 24.64 GB peak memory
- server chat prompt: 71 tokens; 200 generated tokens
- the remaining raw-versus-chat throughput difference is acceptance-driven: the full chat prefix changes the sequence proposed by the assistant

## Validation

- `24 passed` across the focused Muse Glimmer and DFlash tests
- `155 passed` in the broader speculative suite with the known unrelated Qwen quantized-linear identity test excluded
- compiled target and assistant transitions are bit-identical to their uncompiled operation order in focused BF16 tests
- preprojected assistant context produces identical greedy draft tokens
- singleton server dispatch preserves stop handling
- pre-commit hooks pass for every changed file
- server smoke and two repeated streaming benchmarks returned HTTP 200

## Numerical note

The tiny deterministic target matches autoregressive greedy generation exactly. On the real 4-bit Muse target, greedy output can still diverge from tokenwise decoding because target verification processes a block at once; the observed boundary was traced to BF16/4-bit target matmul shape numerics rather than assistant or cache corruption. The previously tested exact affine path remains reverted as requested.


pierre427:
Ran this branch through an independent bench on an **8-bit target** (the PR body has 4-bit numbers, so this complements them). Checked out `pr-1842`'s head, loaded the real `meta-models/Muse-Glimmer-30B-assistant` via `load_drafter`; target = Muse-Glimmer-30B quantized bf16→8-bit (9.31 bpw). Single-user B=1, cold-interleaved (3 prompts × 3 rounds, Apple Silicon):

| arm | decode t/s | accept/round | vs plain |
|---|--:|--:|--:|
| plain (no draft) | 17.8 | — | 1.00× |
| DFlash (q8 assistant) | 40.2 | 2.25 | 2.26× |
| DFlash (q4 assistant) | 41.7 | 2.28 | **2.34×** |

So ~2.3× on the 8-bit target — a bigger relative win than on 4-bit (the slower target leaves more room for the draft to amortize). Loads and runs clean on the real checkpoint; output matches the no-draft greedy on the code-heavy prompt and tie-flips on a couple (the usual block-verify-vs-single-token divergence on a quantized target). Works end-to-end on real weights + 8-bit — nice work on the batched rollback.

(Context: I'd opened a duplicate — #1849 — before spotting this; closed it in favor of yours. There's a downstream consumer waiting on this in Rapid-MLX raullenchai/Rapid-MLX#1844, whose DFlash bridges into mlx-vlm's `load_drafter`.)

Blaizzy:
Thanks @pierre427, I'm happy it's working well for you!
I'm still fixing some stuff before its ready

mlx_vlm/models/muse_glimmer/language.py
```diff
@@ -24,6 +23,47 @@ def _centered_rms_norm(x: mx.array, weight: mx.array, eps: float) -> mx.array:
...
+@mx.compile
+def _prepare_mlp_input(
+    residual: mx.array,
+    attention_output: mx.array,
+    post_attention_weight: mx.array,
+    pre_feedforward_weight: mx.array,
+    post_attention_eps: float,
+    pre_feedforward_eps: float,
+) -> tuple[mx.array, mx.array]:
+    hidden_states = residual + _centered_rms_norm_impl(
```
lucasnewman: Is this supposed to be the compiled `_centered_rms_norm` or are you intending to compile this block separately? If not, do we actually need `_centered_rms_norm`?

lucasnewman (approved):
Looks reasonable to me although I'm not an expert here. One question inline above.

mlx_vlm/models/muse_glimmer/language.py
Blaizzy: An artifact,  refactorting!

PR #1843, opened 2026-08-11
Wire Muse Glimmer's reasoning channel through the server

fredchu:
## Problem

#1841 added Muse Glimmer's harmony markers to the model config:

```python
# mlx_vlm/models/muse_glimmer/config.py
thinking_start_token: str = "to=self<|message|>"
thinking_end_token: str = "<|eom|>"
```

but nothing in the server ever reads them. `_build_gen_args()` resolves the tokens as
request field → `MLX_VLM_THINKING_*_TOKEN` env → `None`, so on a default server run the
model's whole chain of thought comes back inside `message.content`:

```
to=self<|message|>Say exactly: hello world

We need to say exactly: hello world ...

No extra.<|eom|><|start|>assistant to=user<|message|>hello world
```

with `message.reasoning` / `message.reasoning_content` both `null`.

Because the reasoning sits inside `content` *and* inside the token budget, any answer
longer than a sentence tends to exhaust `max_tokens` before the final channel is reached,
so `content` ends up being pure chain of thought with no answer in it at all.

Two smaller things fall out of the same area:

* Even after the thinking split, the final `<|start|>assistant to=user<|message|>` header
  is left in `content` for every consumer to strip.
* `to_template_kwargs()` forwards `reasoning_effort` to the chat template, but Muse
  Glimmer's template does not read that name — it renders `Reasoning strength: <value>.`
  from a `reasoning_strength` variable (unconditionally, defaulting to `high`). So the
  OpenAI-style `reasoning_effort` field silently does nothing for this model.

## Changes

1. **`request_normalization.py`** — `_build_gen_args()` takes an optional `config` and falls
   back to the model config's `thinking_start_token` / `thinking_end_token` when neither the
   request nor the env var supplies them. Resolution order is unchanged otherwise
   (request → env/CLI → model config → `None`). `config` is threaded from the existing call
   sites in `app.py`, `openai.py` (3) and `anthropic.py` (2); all of them already have it in
   scope from `get_cached_model(...)`.

2. **`generation.py`** — `to_template_kwargs()` emits `reasoning_strength` alongside
   `reasoning_effort` with the same value. Templates that don't reference it ignore the extra
   kwarg, so this is a no-op for every other model.

3. **`responses_state.py`** — `_strip_content_markers()` also removes harmony channel headers
   (`<|start|>…<|message|>`) and stray `<|eom|>` / `<|eot|>` markers. The streaming path holds
   back a partially-received header so a header split across chunks isn't emitted as content.

## Result

Default run, no flags:

```bash
python -m mlx_vlm.server --model /path/to/Muse-Glimmer-30B-4bit --port 8088
```

| | before | after |
|---|---|---|
| `content` | `to=self<\|message\|>…<\|eom\|><\|start\|>assistant to=user<\|message\|>hello world` | `hello world` |
| `reasoning` | `null` | the chain of thought |

`reasoning_effort` now actually steers the model (same prompt, `mlx-community/Muse-Glimmer-30B-4bit`):

| `reasoning_effort` | reasoning length | completion tokens |
|---|---|---|
| `low` | 186 chars | 60 |
| `xhigh` | 523 chars | 168 |

## Testing

New `mlx_vlm/tests/test_muse_glimmer_reasoning.py` (5 tests, no weights/network/server):
non-streaming split, streaming split with the channel header broken across chunks, marker
stripping, config/env/request precedence in `_build_gen_args`, and the
`reasoning_strength` alias.

```
mlx_vlm/tests/test_muse_glimmer_reasoning.py    5 passed
mlx_vlm/tests/test_server.py + test_atem_tool_parser.py + test_gemma4_tool_parser.py
                                              285 passed
```

Manually verified on an M1 Max (32 GB) against `mlx-community/Muse-Glimmer-30B-4bit`:
non-streaming and streaming both return clean `content`, ATEM tool calls from #1841 still
parse correctly, and `lmstudio-community/Qwen3-VL-8B-Instruct-MLX-4bit` (non-harmony) is
unaffected.

## Notes

* Two behaviours I deliberately left alone, happy to fold either in if you'd prefer:
  * Streamed `content` keeps a leading space (`" hello world"`) where the non-streaming path
    `.strip()`s it. That difference predates this change.
  * `reasoning_effort: "none"` is passed through to `reasoning_strength` as-is. Muse Glimmer
    documents only `low / medium / high / xhigh`, so `none` lands as a value it wasn't trained
    on and yields minimal-but-nonzero reasoning rather than none. Mapping disabling efforts to
    `low` would be a judgement call I didn't want to make unilaterally.

### DIFF
--- a/mlx_vlm/server/anthropic.py
+++ b/mlx_vlm/server/anthropic.py
@@ -481,7 +481,10 @@ async def anthropic_messages_endpoint(http_request: Request):
 
         try:
             gen_args = _build_gen_args(
-                request, processor, tenant_id=_read_tenant_id(http_request)
+                request,
+                processor,
+                config=config,
+                tenant_id=_read_tenant_id(http_request),
             )
         except Exception as e:
             return _anthropic_error_response(400, str(e))
@@ -1055,7 +1058,10 @@ async def anthropic_count_tokens_endpoint(http_request: Request):
             _anthropic_messages_to_internal(request)
         )
         gen_args = _build_gen_args(
-            request, processor, tenant_id=_read_tenant_id(http_request)
+            request,
+            processor,
+            config=config,
+            tenant_id=_read_tenant_id(http_request),
         )
         template_kwargs = gen_args.to_template_kwargs()
         if tool_choice is not None:
--- a/mlx_vlm/server/app.py
+++ b/mlx_vlm/server/app.py
@@ -163,12 +163,13 @@ def _server_runtime_snapshot() -> dict:
 
 
 def _build_gen_args(
-    request, processor=None, tenant_id: Optional[str] = None
+    request, processor=None, config=None, tenant_id: Optional[str] = None
 ) -> GenerationArguments:
     """Build GenerationArguments from a compatible API request."""
     return _request_normalization._build_gen_args(
         request,
         processor=processor,
+        config=config,
         tenant_id=tenant_id,
         structured_logits_processor_builder=_build_structured_logits_processors,
     )
--- a/mlx_vlm/server/generation.py
+++ b/mlx_vlm/server/generation.py
@@ -810,6 +810,8 @@ class GenerationArguments:
             kw["reasoning"] = self.reasoning
         if self.reasoning_effort is not None:
             kw["reasoning_effort"] = self.reasoning_effort
+            # Muse Glimmer's chat template reads the reasoning_strength alias.
+            kw["reasoning_strength"] = self.reasoning_effort
         if self.thinking_budget is not None:
             kw["thinking_budget"] = self.thinking_budget
         if self.thinking_start_token is not None:
--- a/mlx_vlm/server/openai.py
+++ b/mlx_vlm/server/openai.py
@@ -820,7 +820,10 @@ async def responses_input_tokens_endpoint(request: Request):
         del model
         chat_tools, _ = _response_tool_registry(openai_request.tools)
         gen_args = _build_gen_args(
-            openai_request, processor, tenant_id=_read_tenant_id(request)
+            openai_request,
+            processor,
+            config=config,
+            tenant_id=_read_tenant_id(request),
         )
         template_kwargs = gen_args.to_template_kwargs()
         if openai_request.tool_choice is not None:
@@ -986,7 +989,10 @@ async def responses_endpoint(request: Request):
 
         try:
             gen_args = _build_gen_args(
-                openai_request, processor, tenant_id=_read_tenant_id(request)
+                openai_request,
+                processor,
+                config=config,
+                tenant_id=_read_tenant_id(request),
             )
         except Exception as e:
             raise HTTPException(status_code=400, detail=str(e))
@@ -1696,7 +1702,10 @@ async def chat_completions_endpoint(request: ChatRequest, http_request: Request)
 
         try:
             gen_args = _build_gen_args(
-                request, processor, tenant_id=_read_tenant_id(http_request)
+                request,
+                processor,
+                config=config,
+                tenant_id=_read_tenant_id(http_request),
             )
         except Exception as e:
             raise HTTPException(status_code=400, detail=str(e))
--- a/mlx_vlm/server/request_normalization.py
+++ b/mlx_vlm/server/request_normalization.py
@@ -134,6 +134,7 @@ def _build_structured_logits_processors(
 def _build_gen_args(
     request,
     processor=None,
+    config=None,
     tenant_id: Optional[str] = None,
     structured_logits_processor_builder=_build_structured_logits_processors,
 ) -> GenerationArguments:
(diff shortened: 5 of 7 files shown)


Aryagarg23:
Oh yeah. This PR fixed my issues! <3

Aryagarg23:
> Oh yeah. This PR fixed my issues! <3

Following up with an actual test run, plus one case that still leaks.

For context, I'd hit the same bug and written pretty much the same fix before I found this PR. I only ran into the case below because I was testing my own version against an agent loop, and mine had the same hole.

Tested on an M5 Max against `mlx-community/Muse-Glimmer-30B-8bit`, so a different chip and quant to your M1 Max run. Streaming, non-streaming, ATEM tool calls and `reasoning_effort` all behave the way you describe.

The leak is when the model answers without reasoning first. `add_generation_prompt` has already written `<|start|>assistant`, so the header the model emits is bare and `_HARMONY_CHANNEL_HEADER` doesn't match it:

```python
from mlx_vlm.server.responses_state import _strip_content_markers

_strip_content_markers("<|start|>assistant to=user<|message|>391")   # '391'
_strip_content_markers("to=user<|message|>391")                      # 'to=user<|message|>391'
```

So `content` comes back as `"to=user<|message|>17 × 23 = 391"`. Streaming does the same thing once the deltas are reassembled.

I couldn't work out what makes the model skip the reasoning block. It isn't tool count or prompt length, 40 dummy tools still reasons and 5 real ones doesn't. The case I have is five G-Drive tool schemas with no system prompt, which does it every time, and adding a system prompt makes it stop.

I had a crack at fixing it. Non-streaming is a one-liner, matching a bare `to=…<|message|>` in `_strip_content_markers` covers it and leaves the thinking split alone. Streaming is the harder bit, since the header arrives over several deltas and the existing hold-back is keyed on `<|start|>`.

Happy to keep going on it if that's useful.

Lazarus-931:
hi @fredchu thanks for the pr!

I tested this against main and works for me, but note when reasoning is skipped, it can emit a bare Harmony header, so the API returns `to=user<|message|>391` instead of `391`.

Streaming has the same issue because the header arrives across tokens (to, =user, <|message|>) and is emitted before it can be recognized. Could we buffer and strip both bare and full Harmony headers before the first visible content, and maybe a test or two for these cases?

fredchu:
Thanks both — fixed.

@Aryagarg23 you were right on both counts, and the reproduction was exactly what I needed.

**Non-streaming**: `_strip_content_markers` now also strips a leading bare `to=...<|message|>` header.

**Streaming**: `ThinkingStreamState` holds back any text that could still grow into a bare header until the first visible content is emitted. One thing that bit me while writing it — the thinking open marker `to=self<|message|>` *also* matches a bare-header pattern, so my first version swallowed the entire reasoning block and the existing tests went red immediately. The check now stands down while the buffer still looks like an open marker.

There is a subtler case in the other direction: once content has started, a chunk that merely *begins* with `to=...<|message|>` must be left alone — a user can legitimately ask what that marker means. Streaming emits therefore pass `strip_bare_header=False` and rely on the state machine having already consumed the real header.

**Tests**: both headers are driven through *every* split point of the payload and one character at a time, rather than a few hand-picked boundaries — token-boundary luck is exactly the failure mode this fix exists to prevent. Plus the two "leave it alone" cases.

**One caveat I want to be upfront about**: I could not reproduce the no-reasoning condition on my hardware. On an M1 Max with the 4bit conversion the model reasons every single time, even with `reasoning_effort: "none"`, so I have no live-server proof of the bare-header path itself. What I did instead was capture verbatim raw output from a live server on the PR base commit and drive the state machine with it across all 397 split points, and separately confirm that live-server behaviour is unchanged by this commit (non-streaming with tools, streaming without tools). If either of you can run the bare-header case against a real model, that would close the gap.

Separately, while testing this I hit what looks like a pre-existing problem: with `tools` supplied, streaming never emits the final assistant content at all — every delta is reasoning. It reproduces on the PR base as well and this PR doesn't touch that path, so I opened #1850 for it rather than widening this one.

The workflow run for the new commit is sitting at `action_required`, so it needs an approval before CI will report.

fredchu:
Heads-up on an overlap I noticed while looking around: #1845 touches the same spot in `ThinkingStreamState.feed` as the commit I just pushed here.

@Aryagarg23's `last` flush inserts just before the `return`, and e108e95 also adds a couple of lines there (setting `content_started`). They will conflict textually. Happy to be the one who rebases, whichever lands first.

There is also a small semantic seam worth flagging for whoever merges second. The flush path calls `_strip_content_markers(held)` with the default `strip_bare_header=True`, while this PR routes the streaming emits through `strip_bare_header=False` — by that point the state machine has already consumed the real header, and a chunk that merely *starts* with `to=...<|message|>` mid-content has to survive. After a merge those two paths would disagree with each other. Passing `strip_bare_header=False` in the flush too is probably right, since by definition it is flushing leftovers rather than a header, but I would rather raise it than quietly pick one.

Separately, and not a conflict: if #1848 lands, the config threading in this PR (passing `config=` through the three `openai.py` call sites) becomes redundant — the server can read `processor.config` instead. Happy to switch to that if you would rather keep config attachment out of the server.

nazerim:
Reconciliation note: **superseded by https://github.com/Blaizzy/mlx-vlm/pull/1884** (draft).

Thanks for the reasoning-channel work — after reviewing against current main:

- **Kept from #1843:** the `reasoning_strength` alias for Glimmer's chat template (now in #1884).
- **Already on main:** the config-token fallback — merged via #1848 ("Use model reasoning config and response templates", `_request_field_or_default`). Note the Glimmer model config itself has no `thinking_start_token`/`thinking_end_token` (both `None`), so that fallback is a no-op for Glimmer anyway.
- **Dropped:** the `responses_state.py` harmony-marker changes (target the old `ThinkingStreamState` path; the unanchored header strip contradicts the anchored handling in #1878) and the old test file (superseded by #1878's `test_glimmer_server_handling.py` + #1884's tests).

Happy to close this in favor of #1884 + #1878 once they land.

PR #1845, opened 2026-08-11
Flush the stream buffer on the last token

Aryagarg23:
## Summary

```python
s = ThinkingStreamState()
s.feed("hello <").content              # 'hello '        the '<' never arrives

s = ThinkingStreamState()
s.feed("<think>")
s.feed("cut off </thi").reasoning      # 'cut off '      the '</thi' never arrives
```

`_split_partial` keeps those trailing characters back in case `<think>` or `</think>` follows, and nothing sends them once the stream ends. Hits reasoning and content alike.

- `feed()` takes a `last` flag and empties whatever is left when it's set
- wired up at the five streaming call sites in `openai.py` and `anthropic.py`
- no change for callers that don't pass it

## Why

Any response ending in `<`, or a partial like `<thi`, quietly loses those characters. Asking for a response of exactly `<` on main:

```
                            non-streaming   streaming
Qwen2.5-VL-3B-Instruct-4bit      '<'            ''
Qwen2-VL-2B-Instruct-4bit        '<'            ''
gemma-3-4b-it-4bit               '<'            ''
```

## Validation

Three tests in `TestThinkingStreamState` covering content, reasoning and a normal stream. Failure set unchanged from main. `pre-commit run --all` clean.

### DIFF
--- a/mlx_vlm/server/anthropic.py
+++ b/mlx_vlm/server/anthropic.py
@@ -671,7 +671,9 @@ async def anthropic_messages_endpoint(http_request: Request):
                         for event in start_message_event():
                             yield event
 
-                        thinking_delta = thinking_state.feed(delta)
+                        thinking_delta = thinking_state.feed(
+                            delta, last=bool(getattr(token, "finish_reason", None))
+                        )
                         delta_reasoning = thinking_delta.reasoning
                         delta_content = thinking_delta.content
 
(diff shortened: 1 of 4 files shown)


Lazarus-931 (approved):

Blaizzy (approved):
LGTM, thanks!

PR #1847, opened 2026-08-11
Add Apertus 1.5 (image + audio + text)

Gusanidas:
Adds [swiss-ai/Apertus-v1.5-8B](https://huggingface.co/swiss-ai/Apertus-v1.5-8B) and [-70B](https://huggingface.co/swiss-ai/Apertus-v1.5-70B).

Apertus 1.5 turns media into discrete tokens in the text vocabulary instead
of projected embeddings. Images go through a VQGAN-style tokenizer (131k
codes, 16×16 px per code), audio through a WavTokenizer codec (4k codes,
40 codes/s at 24 kHz). The decoder reuses the existing `ApertusModel`
(xIELU, QK-norm, llama3 rope) with a split vocabulary: 266,752 input
tokens, 131,072-logit output head. First slice is single prompt (batch
size 1).

Port notes:

- Both codec towers stay fp32 end to end, also under `convert -q`: code
  assignment is an argmax over the codebook and half precision flips codes.
  These are the same modules the reference pins via
  `_keep_in_fp32_modules_strict`.
- The processor is torch-free (numpy/PIL). The uint8 bicubic resize is
  bit-exact with torchvision's integer CPU path, since resize drift also
  flips codes.
- Only the audio encode path is ported (the Vocos decoder weights are
  dropped); weight-norm and torch-LSTM parametrizations are fused into
  plain MLX weights at load. The tokenizer loads with
  `trust_remote_code=False`, like kimi_k3 and mage_vl.
- The chat template emits `<s>` itself, so apertus1p5 joins the
  template-owns-markers set in `should_add_special_tokens`. The server had
  a stale gemma-only copy of that check; it now calls the shared helper.
- The SFT/ORPO losses now mask labels beyond the output head. MLX
  cross-entropy accepts out-of-range targets silently, so with a split
  vocabulary any media position in a training batch was contributing a
  corrupt gradient. The regression test uses the released id ranges.

## Parity with the reference

Real 8B checkpoint vs the transformers fork the model card prescribes,
same inputs:

- Tokenized prompts, pixel values (1e-7) and all discrete codes match
  exactly: 744/744 vision codes, 418/418 audio codes, plus 8 synthetic
  clips down to sub-hop lengths.
- Logits: same argmax, KL ≤ 3e-4. Teacher-forced argmax agreement over 24
  tokens: 24/24 on audio, 23/24 on image (the miss is a bf16 near-tie).
- On the real model: correct image descriptions, correct full speech
  transcription, system prompts and multi-turn honored, 73% on an MMLU
  sample.

## Performance

M2 Max 96 GB, bf16, greedy, 64 new tokens, warm:

| Scenario | Prompt tokens | Prefill tok/s | Decode tok/s | Peak mem |
|---|---|---|---|---|
| Text | 76 | 305 | 22.6 | 18.4 GB |
| Image (586×640) | 1,597 | 467 | 22.1 | 24.8 GB |
| Audio (10.4 s) | 492 | 509 | 23.1 | 18.8 GB |

Decode sits at the bf16 bandwidth ceiling for an 8B, so the multimodal
wrapper costs nothing there — after encoding, media are just tokens.
Qwen2.5-VL-7B decodes at 25 tok/s on the same machine; the torch reference
on MPS gets 11–16 tok/s, and in my runs its MPS output was wrong (the
fork's xIELU targets CUDA; on CPU it is correct at ~1 tok/s). The real
cost of the discrete-token design is the image budget, ~3× the prompt
tokens of ViT-patch models.

## Limitations

- Real-checkpoint validation covered the 8B. The 70B config parses into
  the expected shape but I have not run it.
- Audio must be mono 24 kHz; anything else errors instead of downmixing.
- The head returns the physical 131,072 logits; the reference pads
  inference logits to the logical 266,752 vocab with dtype-min. Argmax and
  sampling are unaffected.
- Shared-framework behavior applies: media positions are fixed by the
  message formatter, one audio clip per prompt.

## Tests

`TestApertus1p5` (27 tests) covers config parsing against the vendored
real `config.json` (which caught `eos_token_id = [2, 68, 72]` hiding
inside `text_config`), sanitize/conversion including the weight-norm and
LSTM fusions, fp32 codebook preservation under quantization, the
torchvision-reference resize, placeholder expansion and scatter for both
modalities, and a save/reload round trip. `test_trainer.py` adds the
out-of-vocab label-masking regression test.

### DIFF
--- a/mlx_vlm/models/apertus1p5/__init__.py
+++ b/mlx_vlm/models/apertus1p5/__init__.py
@@ -1,0 +1,16 @@
+from . import processing_apertus1p5  # noqa: F401 (installs processor patch)
+from .apertus1p5 import Model
+from .config import ModelConfig, TextConfig, VisionTokenizerConfig
+from .language import LanguageModel
+from .processing_apertus1p5 import Apertus1p5Processor as Processor
+from .vision import VisionTokenizer
+
+__all__ = [
+    "Model",
+    "ModelConfig",
+    "TextConfig",
+    "VisionTokenizerConfig",
+    "LanguageModel",
+    "VisionTokenizer",
+    "Processor",
+]
--- a/mlx_vlm/models/apertus1p5/apertus1p5.py
+++ b/mlx_vlm/models/apertus1p5/apertus1p5.py
@@ -1,0 +1,393 @@
+import re
+from typing import Optional
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from ..base import InputEmbeddingsFeatures, LanguageModelOutput
+from .audio import AudioTokenizer
+from .config import ModelConfig
+from .language import LanguageModel
+from .vision import VisionTokenizer, ensure_channels_last
+
+_AUDIO_LSTM_KEY = re.compile(r"^(.*\.lstm)\.(weight|bias)_(ih|hh)_l(\d+)$")
+
+
+def masked_scatter(input_tensor: mx.array, mask: mx.array, source: mx.array):
+    mask_flat = mask.flatten().astype(mx.int32)
+    indices = mx.cumsum(mask_flat) - 1
+    aligned = source.flatten()[indices % source.size]
+    return mx.where(mask_flat, aligned, input_tensor.flatten()).reshape(
+        input_tensor.shape
+    )
+
+
+class Model(nn.Module):
+    """Image/audio/text Apertus 1.5 inference.
+
+    The initial port intentionally supports one prompt at a time. Custom
+    attention masks are accepted for API compatibility but ignored; the
+    Apertus backbone builds its own causal mask.
+    """
+
+    def __init__(self, config: ModelConfig):
+        super().__init__()
+        self.config = config
+        self.model_type = config.model_type
+        if config.tie_word_embeddings and not config.text_config.tie_word_embeddings:
+            # utils.load_model rebuilds text_config from the raw nested dict
+            # after ModelConfig.from_dict, discarding the hoisted top-level
+            # flag; reapply it before the language model reads it.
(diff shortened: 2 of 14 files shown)


RaphaelKreft:
Hello @Gusanidas , Maintainers, 

glad to see such a nice community contributions for Apertus 1.5 : )

**Just as a heads-up:** we are currently still in the process of upstreaming Apertus 1.5 to transformers, so the checkpoint layout and reference implementation are not final.

So I would suggest to wait with the final merge until Apertus 1.5 is upstreamed and thus the reference implementation and the checkpoint are fixed (to cross check, validate and potentially port improvements and bug-fixes).

The transformers upstream PR: https://github.com/huggingface/transformers/pull/47662

Lazarus-931:
> Hello @Gusanidas , Maintainers,
> 
> glad to see such a nice community contributions for Apertus 1.5 : )
> 
> **Just as a heads-up:** we are currently still in the process of upstreaming Apertus 1.5 to transformers, so the checkpoint layout and reference implementation are not final.
> 
> So I would suggest to wait with the final merge until Apertus 1.5 is upstreamed and thus the reference implementation and the checkpoint are fixed (to cross check, validate and potentially port improvements and bug-fixes).
> 
> The transformers upstream PR: [huggingface/transformers#47662](https://github.com/huggingface/transformers/pull/47662)

hi @RaphaelKreft, thanks for the heads up!

appreciate open-sourcing the model, kudos to u guys!

as for the pr, we'll wait on finished changes to transformers before merging!

Gusanidas:
Thanks @RaphaelKreft, makes sense. I will revalidate when https://github.com/huggingface/transformers/pull/47662 lands and update whatever has changed.

PR #1851, opened 2026-08-11
Add Ministral3 embedder: reuse backbone with bidirectional mask

Aryagarg23:
`nvidia/Nemotron-3-Embed` ships as `model_type: ministral3` with `is_causal: false`. No embedding variant is registered for that type. The loader builds the causal class and fails:

```
Unsupported embedding model: Received 146 parameters not in model:
embed_tokens.weight, layers.0.input_layernorm.weight, ...
```

- adds `ministral3_embedding`, wrapping `Ministral3Decoder` the way `lfm2_embedding` wraps its backbone
- padding mask instead of a causal one
- llama-4 attention scale at offset 0, then pool and normalise
- one line in `EMBEDDING_MODEL_REMAPPING`, no change to the existing `ministral3` module

## Validation

Against `sentence-transformers` on `nvidia/Nemotron-3-Embed-1B-BF16`, normalised:

```
                   mlx       reference
cos(query, doc1)   0.4700    0.4729
cos(query, doc2)   0.0596    0.0601
cos(query, doc3)   0.0090    0.0094
```

Seven tests in `test_models.py`:

- forward pass
- bidirectional backbone
- masked positions ignored
- invariance to padding and batch order
- `sanitize`
- remapping

Removing the bidirectional mask fails one. Removing the padding mask fails two. `pre-commit run --all-files` clean.


Aryagarg23:
Serving over `/v1/embeddings` needs #1854. `load_embedding_model` works without it.

Aryagarg23:
cheeky rebase :)

Lazarus-931:
lgtm!

cc: @lucasnewman

lucasnewman (approved):
Thanks!

Issue #1830, opened 2026-08-09
fix: Idefics3/SmolVLM LoRA training crashes with 3 bugs (v0.6.10)


pupa3066:
## Summary

LoRA fine-tuning of SmolVLM (Idefics3 architecture) via `mlx-vlm` trainer crashes with 3 cascading bugs. All are in the training path — inference works fine.

**Environment:** mlx-vlm 0.6.10, Apple M1 8GB, macOS, Python 3.13

## Bug 1: Connector output not reshaped (3D → 2D)

**File:** `mlx_vlm/models/idefics3/idefics3.py`  
**Error:** `ValueError: Image features and image tokens do not match: tokens: 1377, features 17`

The connector (perceiver resampler) outputs `(n_sub_images, 81, 2048)` but `_prepare_inputs_for_multimodal()` expects 2D `(total_features, hidden_dim)`. `image_features.shape[0]` returns 17 instead of 17×81=1377.

**Fix:** After `image_features = self.connector(image_features)`, add:
```python
image_features = image_features.reshape(-1, image_features.shape[-1])
```

## Bug 2: Sequence truncation breaks image token alignment

**File:** `mlx_vlm/trainer/sft_trainer.py`  
**Error:** `ValueError: Image features and image tokens do not match: tokens: 447, features 1053`

`max_seq_length` truncation removes image tokens from `input_ids`, but `pixel_values` still contains ALL sub-images. A 3024×4032 photo creates 13 sub-images (1053 tokens), but truncation to 512 leaves only 447 tokens.

**Fix:** Use `max_seq_length=4096` or trim `pixel_values` after truncation.

## Bug 3: Cache passed as array during grad_checkpoint

**File:** `mlx_vlm/models/idefics3/language.py`  
**Error:** `AttributeError: 'mlx.core.array' object has no attribute 'offset'`

During training with `grad_checkpoint=True`, the cache parameter becomes an mlx array instead of None/KVCache. The attention layer then crashes accessing `cache.offset`.

**Fix:** Guard with `if cache is not None and hasattr(cache, 'update_and_fetch'):`

## Reproduction

```python
from mlx_vlm import load
from mlx_vlm.trainer import train, TrainingArgs, VisionDataset, get_peft_model, find_all_linear_names
from PIL import Image
import mlx.optimizers as optim
import json
from huggingface_hub import hf_hub_download

model, processor = load("mlx-community/SmolVLM-Instruct-4bit")
[...]

PR #1852, opened 2026-08-11
Fix Idefics3/SmolVLM LoRA training crashes (#1830)

Lazarus-931:
Idefics3's forward omitted the `mask` argument that every other VLM accepts, so the trainer's positional call bound the attention mask to `cache` and crashed in create_attention_mask. Accept and forward `mask`, which also fixes SmolVLM (an Idefics3 subclass).

Sequence truncation also cut image placeholder tokens while pixel_values kept every sub-image, breaking the one-feature-per-image-token alignment the model asserts on. iterate_batches now skips examples whose image tokens would not survive truncation, warns once per example with the max_seq_length to raise, and fails only when no example is trainable.

### DIFF
--- a/mlx_vlm/models/base.py
+++ b/mlx_vlm/models/base.py
@@ -210,7 +210,11 @@ def create_attention_mask(
     h, cache=None, window_size: Optional[int] = None, return_array: bool = False
 ):
     N = h.shape[1]
-    if cache and hasattr(cache, "make_mask"):
+    if (
+        cache is not None
+        and not isinstance(cache, mx.array)
+        and hasattr(cache, "make_mask")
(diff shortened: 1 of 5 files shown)


lucasnewman (approved):
LGTM. Also take a look at the changes in https://github.com/Blaizzy/mlx-vlm/pull/1858 (which will conflict if we merge this).

lucasnewman (approved):

Lazarus-931:
> LGTM. Also take a look at the changes in #1858 (which will conflict if we merge this).

yeah, had 2 things I included here from there, pr closed 

thx @pupa3066

Issue #1820, opened 2026-08-09
todo: create `/v1/rerank` to work alongside /v1/embeddings


Lazarus-931:
/v1/embeddings was made here since we didn't use mlx-embeddings upstream for nativ, but `/v1/rerank` was never included.

It would only receive a query and document go through the model together, and it emits a single relevance score.

PR #1853, opened 2026-08-12
Add text and multimodal reranking endpoint

Lazarus-931:
handles #1820, reranker was missing and should be supported, will add endpoint support in nativ as well

this for now can allow us to setup much better features in Nativ using reranker, such as improving smart search in artifacts or for general endpoint support

verified tests and gtg!


lucasnewman (approved):

TheAngryPit:
Tested this PR directly on the Nativ v0.3.1 bundled Python runtime, fully offline, using an existing cached `mlx-community/Qwen3-Reranker-0.6B-mxfp8` checkpoint.

Observed:

- `POST /v1/rerank`: HTTP 200
- relevant document ranked first (`relevance_score: 0.96875`)
- `/health`: reported the model under `loaded_models.reranker` with `model_kind: reranker`
- `POST /unload`: success
- `POST /rerank`: HTTP 404

Could this PR register `/rerank` as an alias for the same handler? LiteLLM/vLLM-compatible rerank integrations commonly target `/rerank`, while other clients use `/v1/rerank`. Supporting both would avoid a downstream adapter and matches the dual-route pattern already used by several server endpoints.

I opened Blaizzy/nativ#301 for the Nativ discovery/UI/API integration that depends on this PR.

PR #1854, opened 2026-08-12
Eager-eval cached RoPE arrays in the remaining variants

Aryagarg23:
`SuScaledRoPE`, `Llama3RoPE` and `YarnRoPE` build `_freqs` lazily. Load on one thread, run on another:

```
RuntimeError: There is no Stream(gpu, 1) in current thread.
```

That is what the server does. `get_cached_model` and the `/v1/embeddings` worker are different threads, so `POST /v1/embeddings` 500s for any model on one of these three. Once `_freqs` is built lazily on the loading thread nothing can evaluate it afterwards, so it has to happen in `__init__`.

#1272 fixed this for `MRoPERotaryEmbedding`, `ProportionalRoPE` followed. This gives the remaining three the same `eager_eval_arrays` / `eval_cached_arrays` pair.

## Validation

Six tests in `test_rope_utils.py`. Three follow `test_proportional_rope_evals_private_helper_arrays_on_init`; three build each variant on one thread and run it on another. Against `rope_utils.py` from main:

```
                on main           with this
SuScaledRoPE    Stream(gpu, 0)    ok
Llama3RoPE      Stream(gpu, 2)    ok
YarnRoPE        Stream(gpu, 4)    ok
```

29 passed. `pre-commit run --all-files` clean.

End to end with `nvidia/Nemotron-3-Embed-1B-BF16`, loaded via #1851:

```
before   POST /v1/embeddings -> 500
after    2048 dim, "a cat sat on the mat" scores 0.4698 against
         "a kitten rested on the rug", 0.0632 against "quarterly earnings rose"
```

### DIFF
--- a/mlx_vlm/models/rope_utils.py
+++ b/mlx_vlm/models/rope_utils.py
@@ -53,6 +53,7 @@ class SuScaledRoPE(nn.Module):
 
         freqs = base ** (mx.arange(0, dims, 2, dtype=mx.float32) / dims)
         self._freqs = mx.array(long_factor, dtype=mx.float32) * freqs
+        self.eval_cached_arrays()
 
         def default_scale(factor):
             return math.sqrt(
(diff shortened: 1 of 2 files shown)


lucasnewman (approved):
@Aryagarg23 Thank you!

Issue #1812, opened 2026-08-07
Unable to load model with stripped vision tower


noseheavy:
Attempting to load a model with only text layers will lead to load failure, despite `mlx_lm` capable of successfully loading and performing inference. Reproducible with model `leonsarmiento/ThinkingCap-Qwen3.6-27B-3bit-mlx`:

```
% mlx_vlm.chat --model ~/.omlx/models/leonsarmiento/ThinkingCap-Qwen3.6-27B-3bit-mlx
Fatal error: Missing 393 parameters: 
vision_tower.blocks.0.attn.proj.bias,
...
vision_tower.blocks.9.norm2.weight,
vision_tower.merger.linear_fc1.bias,
vision_tower.merger.linear_fc1.weight,
vision_tower.merger.linear_fc2.bias,
vision_tower.merger.linear_fc2.weight,
vision_tower.merger.norm.bias,
vision_tower.merger.norm.weight,
vision_tower.patch_embed.proj.bias,
vision_tower.patch_embed.proj.weight,
vision_tower.pos_embed.weight.
```

This occurs despite of trying `"language_model_only": true` through `config.json`. Behavior with other inference engines reveal model files are not the cause, with oMLX also successful in inference, with notable option to override model type (LLM/VLM) detection previously used to resolve similar issues.

PR #1855, opened 2026-08-12
Support text-only Qwen3.5 checkpoints (#1812)

Lazarus-931:
Treat an absent or null vision_config, or an explicit language_model_only flag, as a text-only configuration. Skip constructing the vision tower in that case and reject image or video inputs with a clear error.

### DIFF
--- a/mlx_vlm/models/qwen3_5/qwen3_5.py
+++ b/mlx_vlm/models/qwen3_5/qwen3_5.py
@@ -52,7 +52,10 @@ class Model(Qwen3VLModel):
         # only initialize nn.Module, skip the initialization of vision_tower and language_model in the parent class
         nn.Module.__init__(self)
         self.config = config
-        self.vision_tower = VisionModel(config.vision_config)
+        if config.vision_config is None or config.language_model_only:
+            self.vision_tower = None
+        else:
+            self.vision_tower = VisionModel(config.vision_config)
         self.language_model = LanguageModel(config.text_config, config)
 
     def get_input_embeddings(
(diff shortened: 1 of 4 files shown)


Blaizzy:
Could you share the model you are using to build this? All qwen3_5 are VLMs, if they don't have vision odds are the model type should be different as well or it's non standard in which case requires careful consideration.

mlx_vlm/utils.py
```diff
@@ -691,6 +692,8 @@ def load_model(model_path: Path, lazy: bool = False, **kwargs) -> nn.Module:
     modules = ["text", "vision", "perceiver", "projector", "audio"]
     model_config = update_module_configs(model_config, model_class, config, modules)
     model_config = apply_generation_config_defaults(model_config, config)
+    if text_only_config and hasattr(model_config, "language_model_only"):
+        model_config.language_model_only = True
```
lucasnewman, line 696: `language_model_only` is already set in the `model_config` in this case, right? Or are you wanting to override the value if it's set to false?

mlx_vlm/models/qwen3_5/qwen3_5.py
```diff
@@ -79,6 +82,9 @@ def get_input_embeddings(
                 rope_deltas=rope_deltas,
             )
 
+        if self.vision_tower is None:
+            raise ValueError("this Qwen3.5 model was loaded without a vision tower")
```
lucasnewman, line 86: How about "This model doesn't support image input." for clarity?

Lazarus-931:
> Could you share the model you are using to build this? All qwen3_5 are VLMs, if they don't have vision odds are the model type should be different as well or it's non standard in which case requires careful consideration.

i used leonsarmiento/ThinkingCap-Qwen3.6-27B-3bit-mlx, which declares a VLM architecture and sets language_model_only= false, while shipping no vision tower.

but this issue effects several models so i'm closing for a more generalized fix

PR #1856, opened 2026-08-12
Add exact-mode APC prefix reuse for the DFlash speculative path

pierre427:
## Summary

The DFlash speculative path never consults the APC subsystem — `_run_speculative` always prefills the whole prompt and reports `cached_tokens=0`. For agent traffic (a stable system prompt + tools + a growing conversation) that re-prefills a large shared prefix on every turn.

This PR adds **exact-mode cross-request prefix reuse for the DFlash path**, plus a set of hardening fixes to the shared APC core that came out of a deep review of the feature. Built on the pluggable-adapter registry from #1638 / #1629.

### What's new

**`speculative/apc_dflash.py`** — per-row exact-prefix lookup/store for the DFlash batch loop:
- `lookup_batch` merges per-row exact-prefix snapshots into a batched warm cache (`make_warm_batch_exact_cache_multi`); miss rows join as empty caches; only the (right-padded) per-row suffixes are prefilled, with `prepare()`/`finalize()` rolling the right-pad KV out.
- **Mid-prefill static-prefix checkpointing** — the piece that makes reuse work on sliding-window models at all. A post-hoc trimmed checkpoint is only lossless while a `RotatingKVCache` hasn't wrapped, so on any real prompt every checkpoint gets refused and only the full-length entry survives — which exact-lookup then rejects (an entry whose length equals the query can't produce a suffix). Capturing the checkpoint *during* prefill, at the moment the cache offset equals the boundary, sidesteps the wrap problem: the sliding layers physically hold the true window ending there. The boundary is the end of the first rendered message (min 2048 tokens), so a *different question* on the same system prompt hits.
- Store policy: checkpoints stored before the full-length entry (a count-bounded LRU evicts oldest-inserted — full-first let a row evict its own full entry), a store-size floor (`APC_DFLASH_STORE_MIN_TOKENS`, default 1024), and text-only/media guards symmetric with lookup.
- Server wiring: fail-safe throughout (any APC fault degrades to a cold prefill), per-row `cached_tokens` reporting, and prefill rate computed from tokens actually prefilled rather than the full prompt.

### Hardening of the shared core

- **Window compaction in `RotatingKVCacheCloneAdapter.clone`**: a one-shot prefill retains *all* tokens in a rotating cache, so a snapshot stored O(prompt) per sliding layer instead of O(window) — ~1.4 GB waste per 36k-token entry on a 39-sliding-layer 30B. When the buffer is linear retained-all with `keep == 0`, the clone now keeps only the last `max_size` positions (lossless for batch merge, disk round-trip, and direct decode continuation — covered by tests). This also narrows the entry-size spread that motivates #1576; the two compose (floor + compaction + ceiling).
- **`store_exact_cache(clone=..., disk=...)`**: `clone=False` lets callers hand over freshly built snapshots without a second multi-GB deep copy; an identical re-store (every warm hit re-stores its full prefix) is now a dedup fast path that refreshes LRU recency without cloning (new stat `exact_store_dedups`); `disk=False` keeps an entry memory-only (a full-length entry is only ever hit by the next turn of the same in-process conversation — persisting it wrote ~870 MB/session that nothing could read).
- **Attention-sink guard**: the batch cache classes are `keep`-blind (`merge` takes the last window tokens, `extract` drops `keep`), so a sink model would be silently corrupted on warm restore. `dflash_apc_mode` and the store path now refuse APC for any cache carrying `keep > 0`, recursively through `CacheList`/tuple composites.

### Validation

- `mlx_vlm/tests/test_apc_dflash_exact.py` — model-free port of a 91-check battery: store order/LRU survival, floor, compaction (+negatives), warm round-trip **bit-equality** vs a cold prefill through a compacted snapshot, sink gates incl. nested composites, re-store dedup, clone/disk ownership, boundary rules, trimmed-checkpoint exactness, second-clone and decode-continuation round-trips.
- Live A/B on an M5, 30B sliding-window model (window 2048, 39 sliding + 13 full layers), 7.7k-token prompt: cold prefill 9.13 s @ 848 tok/s → warm request with a **different question** on the same system prompt restored 7,728 cached tokens and prefilled 15 tokens in **0.13 s**.
- The feature + fixes went through five rounds of adversarial review (independent-reviewer loop) before this PR; the findings that survived are encoded as tests.

### Relationship to open APC work

- #1629 / #1638 — built directly on the capability-driven adapter registry.
- #1623 — complementary: that fixes degenerate one-token entries in checkpoint-length *selection*; this adds a store-side floor on the new DFlash path.
- #1576 — same underlying concern (count-bounded LRU, size-unbounded entries); window compaction removes the sliding-layer share of the spread, and a ceiling still composes on top.
- #1835 — same hybrid-model motivation: exact snapshots are the safe reuse unit where block concat isn't.

Not included (deliberately): model-specific parsers/drafters, and two speculative-path sampling fixes we found during review (single shared sampler across coalesced rows; logits processors ignored during verify) — happy to file those separately if wanted.

@Blaizzy could you review when you get a chance? Happy to split this into core-fixes vs dflash-wiring if you'd prefer smaller pieces.

### DIFF
--- a/mlx_vlm/apc.py
+++ b/mlx_vlm/apc.py
@@ -598,6 +598,7 @@ class APCStats:
     disk_writes: int = 0
     exact_hits: int = 0
     exact_stores: int = 0
+    exact_store_dedups: int = 0
     rejects: int = 0
     rejects_by_reason: Dict[str, int] = field(default_factory=dict)
     last_reject: Optional[Dict[str, Any]] = None
@@ -626,6 +627,7 @@ class APCStats:
             "disk_writes": self.disk_writes,
             "exact_hits": self.exact_hits,
             "exact_stores": self.exact_stores,
+            "exact_store_dedups": self.exact_store_dedups,
             "rejects": self.rejects,
             "rejects_by_reason": dict(self.rejects_by_reason),
             "last_reject": (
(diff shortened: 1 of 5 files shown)


pierre427:
Rebased onto current main `8683ec19` (over the merged #1842/#1848 Muse
work); head `22e9e432` → `ccb7e093`. PR-authored changes are otherwise
unchanged — the only range-diff delta is the inherited #1842 guard change
in `server/generation.py` (`shape[1] > 1` → `> prefill_step_size`),
verified compatible: checkpointed prompts at or under the step size route
through this PR's one-shot-split branch, which handles exactly that case.
Also verified #1842's `make_speculative_prompt_cache` change (plain caches
for all kinds at B==1) composes with `snapshot_prompt_cache_row`, which
handles both plain and batch-shaped caches.

On the rebased head: `test_apc_dflash_exact.py` 17/17;
`test_speculative.py` 155/156, with the one failure
(`test_qwen_target_verify_quantized_linear_matches_singleton_batch_path`)
reproduced identically on a pristine-main worktree — pre-existing on this
machine, not introduced by the rebase.

lucasnewman:
@pierre427 I see the intention but this PR and #1857 add a net +3k LoC for this very specific case, which ends up duplicating a lot of the APC logic and is very difficult to review and maintain. Can you see if you can find a simpler solution and fold it into a single PR? Also, if the hashing for APC is not working correctly for multimodal inputs we should break that out into a separate PR instead of patching over it here. Thanks!

PR #1857, opened 2026-08-12
Give the drafter real context after an APC warm hit (stacked on #1856)

pierre427:
## Summary

**Stacked on #1856** (same branch lineage — the diff shows both until that merges; only the top commit `4881886` is this PR). Happy to fold it into #1856 instead if you'd rather review once.

#1856 ships the APC warm path handing the drafter a **single** hidden-state position after a warm hit (the per-row last-real-position gather — the simple correct-for-verification choice at ragged right-padding). That's a measured acceptance tax. This follow-up gives the drafter its real context back.

### The fix

- `assemble_warm_drafter_context` (speculative/dflash.py): after a warm hit, each row gets its **full un-padded suffix** hidden states (ragged per-row — drafting is row-wise in the batch loop), plus `context_offsets[i] = prefix_len[i]` seeding an absolute RoPE position shift so suffix keys sit at positions `cached..cached+S−1`, exactly where a cold prefill would put them.
- `_drafter_pos_shift` / `pos_shift`: an **additive attribute** on the drafter cache rather than a pre-seeded `cache.offset` — the rotating cache's offset doubles as storage bookkeeping, and pre-seeding it while the buffer isn't exactly full corrupts `_update_in_place`. The qwen3/laguna DFlash attentions now read `offset = cache.offset + pos_shift` (default 0 ⇒ byte-identical old behavior; gemma4 inherits qwen3's).
- `_apply_round1_window`: companion round-1 context truncation to the drafter's training window (window−1), applied **after** the warm offsets so a warm suffix longer than the window composes (`pos_shift = cached + skipped`). This also bounds the transient 5-layer × prompt-length KV spike the old full-length round-1 context stored on long prompts.

### Measured (30B sliding-window target, 17.3k-token prompt, greedy, same-question continuations, n=3/arm)

| drafter context on warm hit | accept |
|---|---:|
| full prompt (cold reference) | 4.41 |
| suffix + pos-shift (**this PR**) | **4.47** |
| single position (#1856 as filed) | 3.88 |

The −12% warm-hit acceptance dip is eliminated; warm hits now decode at reference acceptance. We also measured the obvious next step — persisting ~136 MB/entry of *prefix* hidden states so warm hits start with the full window — and it bought nothing over the suffix (+0.06, noise): drafter acceptance is driven by local context. So this small version is the complete fix, not a first installment.

### Notes for review

- `MUSE_WARMCTX=0` and `MUSE_ROUND1_TRUNC=0` are call-time A/B kill switches restoring the old behaviors (that's how the table above was produced).
- One behavior caveat, documented in-code: a hypothetical drafter config with `sliding_window` set but **mixed** sliding/full layer types would now have its full-attention layers' round-1 context trimmed too (the old in-attention trim was sliding-layers-only). No in-tree checkpoint has that shape; `MUSE_ROUND1_TRUNC=0` restores the old behavior.
- Cold-path RoPE equivalence with the previous in-attention trim (`cache.offset += skip`) is preserved: after the rounds-loop truncation, S arrives at ≤ window−1, the internal trim no-ops, and `pos_shift(skipped)` reproduces the exact offsets it used to produce. Covered by `mlx_vlm/tests/test_dflash_warm_context.py` (numeric equality of the `MUSE_WARMCTX=0` path against the old gather, tail-keep/pos-shift math, offset+truncation composition) plus the existing dflash/laguna/APC packs (92 + 12 + 16 + 38 green).

### DIFF
--- a/mlx_vlm/apc.py
+++ b/mlx_vlm/apc.py
@@ -598,6 +598,7 @@ class APCStats:
     disk_writes: int = 0
     exact_hits: int = 0
     exact_stores: int = 0
+    exact_store_dedups: int = 0
     rejects: int = 0
     rejects_by_reason: Dict[str, int] = field(default_factory=dict)
     last_reject: Optional[Dict[str, Any]] = None
@@ -626,6 +627,7 @@ class APCStats:
             "disk_writes": self.disk_writes,
             "exact_hits": self.exact_hits,
             "exact_stores": self.exact_stores,
+            "exact_store_dedups": self.exact_store_dedups,
             "rejects": self.rejects,
             "rejects_by_reason": dict(self.rejects_by_reason),
             "last_reject": (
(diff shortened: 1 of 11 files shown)


pierre427:
Rebased on the updated #1856 — three commits total: the updated base
(`ccb7e093`), the warm-context fix (`0ad80f49`), and a new commit
(`1ac3c001`) extending the same `pos_shift` treatment to the
`muse_glimmer_assistant` drafter merged in #1842, which this PR couldn't
cover originally because that drafter wasn't upstream yet. The extension is
7 insertions / 1 deletion in one file: queries and keys both derive from
the single shifted `cache_offset`
(`cache.offset + int(getattr(cache, "pos_shift", 0))`), and `pos_shift=0`
preserves the previous cold-path offset (all 11 existing muse drafter
tests pass unchanged).

One interaction surfaced during the rebase, fixed in the amended
warm-context commit: #1842's new B==1 singleton fast path would have
bypassed warm-hit drafter context. Dispatch is now
`batch_size == 1 and not context_offsets` — cold B==1 keeps upstream's
singleton path (the upstream dispatch test still passes); warm APC hits
keep this PR's batched flow — pinned by a new test,
`test_warm_hit_batch_size_one_dispatches_batch_rounds`, mirroring the
upstream cold-dispatch test.

On the branch tip: `test_dflash_warm_context.py` 21/21,
`test_muse_glimmer_dflash_drafter.py` 11/11, `test_speculative.py` 155/156
(same pre-existing failure as noted on #1856).

lucasnewman:
@pierre427 Please see the comment in #1856 -- let's come up with a single PR that is specifically crafted to solve it.

PR #1859, opened 2026-08-12
Publish the model capabilities to clients via the OpenAI /v1/models

dcartman:
This pull request adds a categories array to the model definitions in the /v1/models response.

The model definition:
    {
      "id": "Runpod/FLUX.2-klein-4B-mflux-4bit",
      "object": "model",
      "created": 1786496646,
      "capabilities": [
        "image_editing",
        "image_generation"
      ]
    }

The capabilities supported are:
IMAGE_EDITING
TEXT_GENERATION
VISION
AUDIO 
EMBEDDINGS 
TOOLS 
SPEECH_TO_TEXT 
TEXT_TO_SPEECH 
REASONING 
VIDEO 
DRAFTER 

The primary changes and logic can be found in: capabilities.py


Lazarus-931:
hi @dcartman, thanks for the pr.

sorry i'm a bit confused, wouldn't this just be needed for image models? why would audio/embeddings/text, etc be touched?

Lazarus-931:
hey @dcartman, thanks for the pr!

i think this was a bit over-the-top, we just need the guard that raises the clear "text-to-image model doesn't support editing" error when a txt-image model is selected and user tries to send img for editing

dcartman:
HAHA! Yeah, I guess it was. 

I figured if I was adding capability checks in mlx-vlm then it should manage the truth for all model capabilities and report them out. Letting the clients be a little thinner. Currently the nativ client reads through the locally cached models to determine what capabilities each has. Seemed like redundancy to have capability checks in both places.

I'll wind it back. Though I don't like idea of the user needing to send a message only to receive what boils down to an error message in response. I would rather have that indicated up front in the UI (hence sending the capabilities in the /models api).

PR #1860, opened 2026-08-12
fix(server): return 400 for over-budget /v1/messages streams

pierre427:
### What

The streaming branch of `anthropic_messages_endpoint` (`/v1/messages` with `stream: true`) returns **HTTP 500 `api_error`** with a `"400: ..."` message when the prompt exceeds the context budget. It should return **HTTP 400 `invalid_request_error`**.

### Why

`_preflight_stream_context_budget` (in `server/app.py`) raises `HTTPException(400)` for an over-budget prompt. The streaming branch awaited it with no `except HTTPException` handler, so the exception fell through to the endpoint's outer `except Exception`, which wraps everything as `_anthropic_error_response(500, str(e), "api_error")` — producing `"400: <detail>"` under HTTP 500.

This matters to clients: a 500 signals a retryable server fault, so SDKs (including the Anthropic SDK) back off and retry on what is actually a deterministic client error. The non-streaming path and the OpenAI routers already return a proper 400 here — this brings the Anthropic streaming path in line.

### Fix

Wrap the preflight call in `except HTTPException -> _anthropic_error_response(e.status_code, str(e.detail))`, the same conversion the OpenAI routers use. The preflight already records the failure metric, so the handler does not double-record.

### Test

`mlx_vlm/tests/test_server_anthropic_stream_preflight.py` — model-free, `TestClient`-based:
- `test_stream_over_budget_returns_anthropic_400_not_500` — forces the preflight to reject; asserts 400 + `invalid_request_error` + no `"400:"` prefix leak. **Fails on `main` with `assert 500 == 400`.**
- `test_stream_within_budget_starts_stream` — preflight passes -> streaming 200.

`black --check` clean; the full `test_server.py` suite (278) plus these 2 all pass (280).


lucasnewman:
Can you [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits and force push so we can clear the merge requirements?

pierre427:
Signed and force-pushed — `2b72c9d4` now verifies.

I also restacked #1861 (`e0e61193`) onto the new signed base. There are no file changes in either rewrite.

GitHub currently shows no check runs for the new SHA; please let me know if the workflow needs approval or a rerun.

PR #1862, opened 2026-08-12
Add configurable local model registry

Lazarus-931:
currently in Nativ, model's are discovered in it's code, I think that this belongs in mlx-vlm because model discovery and resolution are engine-level responsibilities that every client should share, while implementing them in Nativ would duplicate logic, and tightly couple the app to MLX-VLM internals.

a separate pr will be opened in Nativ to allow it to handle the user-facing workflow: letting users add folders and aliases, persisting those settings, passing them to mlx-vlm, and displaying/selecting the models mlx-vlm discovers.


Blaizzy:
Closing this PR as not planed

PR #1870, opened 2026-08-12
Allow CacheList extraction with plain KVCache children

### LINKED ISSUE #1869: CacheList.extract() fails when an extracted child is a plain KVCache
## Description

`CacheList.extract()` delegates row extraction to every child cache:

```python
return CacheList(*(cache.extract(idx) for cache in self.caches))
```

This works while a child is a batch-aware cache such as `BatchKVCache`. However, `BatchKVCache.extract()` returns a plain `KVCache`, and `KVCache` does not currently implement `extract()`.

Consequently, a `CacheList` extracted from a batch cannot itself be extracted again:

1. Merge single-row `CacheList(KVCache())` instances into a batch.
2. Extract one row; the child becomes a plain `KVCache`.
3. Extract that already-single-row cache again.
4. `CacheList.extract()` raises `AttributeError`.

This can affect cache handoff and continuous-batching paths that collapse or re-extract caches after transitioning from a batched cache to a single-row cache. Several model cache factories also place plain `KVCache` children in a `CacheList`, including DeepSeek V3.2, GLM MoE DSA, Inkling, and Zaya1-VL.

## Reproduction

Run this script from the repository root:

```python
import mlx.core as mx

from mlx_vlm.models.cache import CacheList, KVCache


def make_row(value: float) -> CacheList:
    child = KVCache()
    keys = mx.full((1, 1, 3, 4), value)
    values = mx.full((1, 1, 3, 4), value + 10)
    child.update_and_fetch(keys, values)
    return CacheList(child)


batched = CacheList.merge([make_row(1.0), make_row(2.0)])

# BatchKVCache.extract() returns a plain KVCache child.
first = batched.extract(0)
print(type(first[0]).__name__, first[0].keys.shape, first[0].offset)

# This currently fails because KVCache has no extract() method.
first_again = first.extract(0)
print(type(first_again[0]).__name__, first_again[0].keys.shape, first_again[0].offset)
```

## Actual behavior

The second extraction fails:

```text
KVCache (1, 1, 3, 4) 3
AttributeError: 'KVCache' object has no attribute 'extract'
```

## Expected behavior

A plain `KVCache` should support extracting a row while preserving only its active prefix and retaining the single-row batch dimension:

```text
KVCache (1, 1, 3, 4) 3
KVCache (1, 1, 3, 4) 3
```

The extracted key and value arrays should be contiguous, have a batch size of one, and exclude unused allocated capacity beyond `offset`.

byoan:
Fixes #1869

## Summary

`CacheList.extract()` delegates extraction to its children, but extracting a `BatchKVCache` produces a plain `KVCache`. Because `KVCache` did not implement `extract()`, the resulting composite cache could not be extracted again.

This change adds row extraction support directly to `KVCache` so that the child cache continues to satisfy the interface expected by `CacheList`.

The implementation:

- extracts one batch row while preserving the batch dimension;
- copies only the active key/value prefix up to `offset`;
- makes the extracted arrays contiguous;
- preserves empty-cache behavior;
- supports Python-style negative row indices;
- raises `IndexError` for out-of-range row indices.

Regression coverage includes populated multi-row caches, repeated `CacheList` extraction, negative and invalid indices, and empty caches.

## Tests

```text
python -m pytest mlx_vlm/tests/test_cache.py -q
```

Results:

- 4 tests passed.

### DIFF
--- a/mlx_vlm/models/cache.py
+++ b/mlx_vlm/models/cache.py
@@ -343,6 +343,26 @@ class KVCache(_BaseCache):
         self.offset -= n
         return n
 
+    def extract(self, idx):
+        cache = KVCache()
+        if self.keys is None:
+            if idx not in (0, -1):
+                raise IndexError("KVCache row index out of range")
+            return cache
+
+        batch_size = int(self.keys.shape[0])
+        if idx < 0:
+            idx += batch_size
+        if idx < 0 or idx >= batch_size:
+            raise IndexError(
+                f"KVCache row index {idx} out of range for batch size {batch_size}"
+            )
+
+        cache.keys = mx.contiguous(self.keys[idx : idx + 1, :, : self.offset, :])
+        cache.values = mx.contiguous(self.values[idx : idx + 1, :, : self.offset, :])
+        cache.offset = self.offset
+        return cache
+
     def to_quantized(self, group_size: int = 64, bits: int = 4) -> QuantizedKVCache:
         quant_cache = QuantizedKVCache(group_size=group_size, bits=bits)
         quant_cache.offset = self.offset
--- a/mlx_vlm/tests/test_cache.py
+++ b/mlx_vlm/tests/test_cache.py
@@ -1,0 +1,55 @@
+import mlx.core as mx
+import pytest
+
+from mlx_vlm.models.cache import CacheList, KVCache
+
+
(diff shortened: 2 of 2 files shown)


lucasnewman (approved):
Thanks!

PR #1874, opened 2026-08-13
feat(server): expose process and speculative runtime identity

pierre427:
# feat(server): expose process and speculative runtime identity

## Summary

- add a per-process `server_instance_id`;
- expose whether speculative decoding is enabled and its effective kind;
- return both fields from `/health` and the internal runtime snapshot;
- add focused endpoint coverage.

## Motivation

Operational gates need to distinguish a real process restart from a successful
probe against the old process. They also need to verify that the expected
speculative backend is active instead of trusting an operator label or launch
command.

The instance ID is generated once at module load and is intentionally opaque.
The speculative snapshot reports `enabled` plus `kind`, using the server's
effective runtime state.

## Verification

- focused health test — 1 passed, 287 deselected;
- Black 26.3.1 — pass for changed files.

This draft is independent of the cancellation and resumable-prefill work.

### DIFF
--- a/mlx_vlm/server/app.py
+++ b/mlx_vlm/server/app.py
@@ -5,6 +5,7 @@ import os
 import secrets
 import sys
 import time
+import uuid
 from contextlib import asynccontextmanager
 from threading import Lock
 from types import SimpleNamespace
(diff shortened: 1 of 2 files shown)


lucasnewman:
@pierre427 I don't know what "operational gates" means in your description. What exactly are you trying to do with the server instance id? The speculative decoding metrics reporting looks reasonable, though.

pierre427:
Fair point — “operational gates” was vague. The intended use is a restart/readiness check:

1. Read the current `server_instance_id`.
2. Restart the server.
3. Poll `/health`.
4. Continue only once it is healthy and the ID has changed.

This prevents a failed restart from being mistaken for a successful one when the old process is still answering health probes. The ID is only an ephemeral per-process nonce—not a durable host identity, client-affinity key, or metric label.

The intended scope is the current single-process server invocation; it should not be considered stable across a multi-worker or load-balanced deployment. I can rename it to `process_instance_id` and clarify the PR description if that makes the semantics clearer. Glad the speculative decoding reporting looks reasonable.

Issue #1871, opened 2026-08-12
SAM 3.1: shape issue during the weight load at detector_model.vision_encoder.backbone.embeddings.patch_embeddings.projection.weight


ryohajika:
Same situation as SAM3 issues #1550 #1545 .
Assuming the pytorch model conversion or transpose of array shape did go wrong.

```bash
python -m mlx_vlm.models.sam3_1.generate --task track --video pingpong.mp4 --prompt "a person" --model mlx-community/sam3.1-bf16

(skip)

ValueError: Expected shape (1024, 14, 14, 3) but received shape (1024, 14, 3, 14) for parameter detector_model.vision_encoder.backbone.embeddings.patch_embeddings.projection.weight
```

mlx-vlm Version: 0.6.12
Python 3.12.13

PR #1877, opened 2026-08-13
Skip conv transposes for pre-converted SAM 3.1 checkpoints (#1871)

Lazarus-931:
sanitize() transposed every 4D weight unconditionally, so loading a checkpoint already stored in MLX layout transposed it a second time and produced (1024, 14, 3, 14) instead of (1024, 14, 14, 3). Detect the layout from the patch-embed conv, whose in_channels is unambiguously 3, and skip the transposes when the checkpoint is already converted, matching the fix SAM 3 received in #1545.

for https://github.com/Blaizzy/mlx-vlm/issues/1871 as well


lucasnewman (approved):
LGTM

PR #1878, opened 2026-08-13
Handle Muse Glimmer recipient-routing output in the server

nazerim:
## Summary

Muse Glimmer (`mlx_vlm/models/muse_glimmer`) uses recipient-routing in its
output: it emits `to=self<|message|>…<|eom|>` (internal reasoning channel),
`to=user<|message|>` (answer header), `to=<tool><|message|>` (before an ATEM
tool call), and `<|eom|>`/`<|eot|>` continuation markers (token 200007 is
**not** an eos token, so it survives decode). The server previously leaked
all of these into `message.content`, breaking agent loops (opencode) that
rely on clean content plus parsed tool calls.

This PR handles Glimmer's output with a **dedicated, self-contained state
machine**, model-gated so other models are unaffected.

## Changes

- **New `mlx_vlm/server/glimmer_stream.py`** — `MuseGlimmerStreamState`: a
  single state machine (NORMAL / THINKING / TOOL) owning all region logic in
  one buffer. Handles thinking markers, ATEM tool blocks, recipient headers
  (bare `to=…<|message|>` and full `<|start|>assistant to=…<|message|>`
  forms), quoted-marker preservation, partial-marker holding across chunk
  boundaries, `<|eom|>`/`<|eot|>` boundaries, and truncated-flush cleanup.
  `make_glimmer_stream_state(processor)` selects it for Glimmer processors.
- **`mlx_vlm/server/openai.py` + `anthropic.py`** — branch on
  `make_glimmer_stream_state`: Glimmer uses the new state machine (which
  handles thinking + tool suppression internally); all other models keep the
  existing `ThinkingStreamState` + `suppress_tool_call_content` path
  byte-identical to main.
- **`mlx_vlm/server/responses_state.py`** — final-parse fixes, ATEM-gated so
  other parsers are unchanged:
  - `process_tool_calls`: anchor on `tool_call_block_start` (parses bare
    tool calls without the reasoning envelope, and multiple blocks);
    turn-header regex; envelope cleanup bounded at the next header;
    `<|eom|>`/`<|eot|>` → header stripping.
  - `_split_thinking`: Glimmer branch collects reasoning blocks only at
    routing boundaries (quoted `to=self<|message|>` mid-answer survives);
    answer-only output stays in content; leading-marker stripping (headers
    at start or after continuation markers; trailing partials).
  - `_response_output_items_from_text`: pass `processor=` so the responses
    endpoint's final content is stripped consistently.
- **`mlx_vlm/tool_parsers/atem.py`** — expose `tool_call_block_start`.

## Why a dedicated state machine

Earlier iterations extended `ThinkingStreamState` +
`suppress_tool_call_content` — two interacting stateful functions that
disagreed about stream state and repeatedly regressed (tool markup leaking,
answers dropped, quoted markers swallowed, end-marker tails leaking). A
single state machine with one buffer eliminates that class of bugs by
construction; non-Glimmer models are untouched.

## Relationship to #1843

This overlaps with #1843 ("Wire Muse Glimmer's reasoning channel through the
server", @fredchu) on the shared goal of keeping Glimmer's reasoning out of
`content`. Differences:

- #1843 wires the model config's thinking tokens through
  `_build_gen_args()` (config/env/request precedence) and adds
  `reasoning_effort` → `reasoning_strength`.
- This PR instead detects the Glimmer processor and applies model-gated
  stripping via the dedicated state machine, additionally handling: bare
  ATEM tool calls (no reasoning envelope), multiple tool blocks,
  chunk-boundary header leaks, `<|eot|>`, and quoted-marker preservation in
  both streamed and final output.

**Reconciliation recommendation:** land this PR first; rebase #1843 keeping
its `request_normalization.py` (config-token fallback — note: its `config`
parameter is currently shadowed and inert) and `generation.py`
(`reasoning_strength` alias) changes, and dropping its `responses_state.py`
changes (they target the old path and their unanchored header strip
contradicts quoted-marker preservation). The two compose cleanly —
verified: 2150 passed with that combination.

## Testing

- **`mlx_vlm/tests/test_glimmer_stream.py`** (45 tests, no weights/network):
  stream-loop scenarios — thinking routing, answer/tool header stripping,
  quoted markers mid-answer (incl. split across chunks), full-form headers,
  `<|eom|>`/`<|eot|>` boundaries, multi-block tool calls, chunk-boundary
  splits at multiple sizes, truncation flush, prose integrity (spaces and
  indentation preserved), non-Glimmer passthrough.
- **`mlx_vlm/tests/test_glimmer_server_handling.py`** (11 tests): final-parse
  routing, bare/envelope/multi-block `process_tool_calls`, truncated
  envelopes, quoted `to=self` survival, `<|eot|>`.
- **Full suite: `2150 passed, 5 skipped`** — the single failure
  (`test_speculative.py::test_qwen_target_verify_quantized_linear_matches_singleton_batch_path`)
  is pre-existing and unrelated (numerical mlx-op test, imports no server
  code; fails identically on clean `main`).
- **Fuzz**: ~60k inputs × chunk sizes 1–33 (931k feeds) — zero hangs,
  zero exceptions, bounded buffer.
- **Non-Glimmer safety**: `ThinkingStreamState`,
  `suppress_tool_call_content`, `make_response_stream_state`,
  `prompt_has_open_thinking` are byte-identical to `main`; Mistral
  `[TOOL_CALLS]` and other parsers behave identically.

### DIFF
--- a/mlx_vlm/server/anthropic.py
+++ b/mlx_vlm/server/anthropic.py
@@ -22,7 +22,6 @@ from .generation import (
     _count_prompt_tokens,
 )
 from .glimmer_stream import make_glimmer_stream_state
-
 from .responses_state import (
     make_response_stream_state,
     process_tool_calls,
--- a/mlx_vlm/server/anthropic.py
+++ b/mlx_vlm/server/anthropic.py
@@ -21,6 +21,8 @@ from .generation import (
     _build_metrics_envelope,
     _count_prompt_tokens,
 )
+from .glimmer_stream import make_glimmer_stream_state
+
 from .responses_state import (
     make_response_stream_state,
     process_tool_calls,
@@ -536,16 +538,21 @@ async def anthropic_messages_endpoint(http_request: Request):
                 open_block_type = None
                 full_output = ""
                 text_output = ""
-                thinking_state = make_response_stream_state(
-                    processor,
-                    prompt_has_open_thinking(
-                        formatted_prompt,
-                        gen_args.enable_thinking,
+                glimmer_state = make_glimmer_stream_state(processor)
+                thinking_state = (
+                    glimmer_state
+                    if glimmer_state is not None
+                    else make_response_stream_state(
+                        processor,
+                        prompt_has_open_thinking(
+                            formatted_prompt,
+                            gen_args.enable_thinking,
+                            gen_args.thinking_start_token,
+                            gen_args.thinking_end_token,
+                        ),
                         gen_args.thinking_start_token,
                         gen_args.thinking_end_token,
-                    ),
-                    gen_args.thinking_start_token,
-                    gen_args.thinking_end_token,
+                    )
                 )
                 in_tool_call = False
                 tc_start = tool_module.tool_call_start if tool_module else None
@@ -674,9 +681,13 @@ async def anthropic_messages_endpoint(http_request: Request):
                         delta_reasoning = thinking_delta.reasoning
                         delta_content = thinking_delta.content
 
-                        in_tool_call, delta_content = suppress_tool_call_content(
-                            full_output, in_tool_call, tc_start, delta_content
-                        )
+                        if glimmer_state is None:
+                            in_tool_call, delta_content = suppress_tool_call_content(
+                                full_output,
+                                in_tool_call,
+                                tc_start,
+                                delta_content,
+                            )
 
                         if delta_reasoning is not None and gen_args.enable_thinking:
                             yield open_block("thinking")
--- a/mlx_vlm/server/glimmer_stream.py
+++ b/mlx_vlm/server/glimmer_stream.py
@@ -101,9 +101,7 @@ class MuseGlimmerStreamState:
                 idx = self.buffer.find(_GLIMMER_TOOL_END)
                 if idx < 0:
                     # Hold partial end-marker tails so they never leak.
-                    hold = self._partial_marker_len(
-                        self.buffer, (_GLIMMER_TOOL_END,)
-                    )
+                    hold = self._partial_marker_len(self.buffer, (_GLIMMER_TOOL_END,))
                     if hold:
                         break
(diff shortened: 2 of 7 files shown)


nazerim:
Hi @lucasnewman — CI flagged a pre-commit failure on this PR (black/isort/autoflake reformatted the server files). Fixed: ran `pre-commit run --all` locally, committed the formatting (3f7ab99, signed), and pushed. The new 'Test PRs' run needs workflow approval again. Thanks!

nazerim:
Closing this — not because the problem is wrong, but because we no longer serve Glimmer through `mlx_vlm.server`.

Our Glimmer deployment moved to oMLX, and oMLX shipped its own channel/recipient parser for Muse Glimmer (`omlx/adapter/muse_glimmer.py`, from Blaizzy/mlx-vlm-adjacent work jundot landed as #2586): a `_MuseChannelSplitter` that classifies each header's recipient (`self` → thinking, `user`/absent → visible answer, anything else → tool channel) plus ATEM `<atem:invoke>` extraction with schema coercion. That is the same job as this PR, done in the adapter layer where oMLX keeps per-model output handling.

Meanwhile this branch has fallen 267 commits behind main and now conflicts in `server/anthropic.py`, `server/openai.py` and `server/responses_state.py` — the server was since restructured into the `server/` package (and tool-calling extracted into `tools/` in #2208), so a rebase would really be a re-implementation against the new layout.

To be explicit about what is *not* resolved: `mlx_vlm.server` still has no Muse Glimmer recipient handling (`muse_glimmer` does not appear anywhere under `mlx_vlm/server/` on current main), so anyone running Glimmer on the mlx-vlm server will still see tool calls and reasoning leak into visible text. If someone wants to pick this up, the logic to port is in the two commits here (`6e0c2351`, formatting `3f7ab99`), and the reasoning-effort alias half of this work already landed via #1884.

PR #1879, opened 2026-08-13
Load text-only checkpoints of vision-language models

### LINKED ISSUE #1812: Unable to load model with stripped vision tower

Lazarus-931:
Checkpoints that declare a VLM `model_type` but ship no vision weights fail to load, because an absent `vision_config` is defaulted to an empty one and the model then builds a full vision tower that the weights cannot fill. When the config declares no vision or audio, drop the submodules the checkpoint has no weights for before the strict load.

fixes #1812 

also prevents same issue for vlm models without vision towers such as gemma4, muse_glimmer, smolvlm, molmo, pixtral, cohere_compass


lucasnewman (approved):
Looks good, although I wonder if it should log the missing module name so it's not a silent failure if there's a real weight mismatch?

mlx_vlm/utils.py
```diff
@@ -568,12 +568,22 @@ def _is_text_only_config(config: dict) -> bool:
...
-        if not isinstance(child, nn.Module):
+        if name == "language_model" or not isinstance(child, nn.Module):
             continue
-        if any(key.startswith(f"{name}.") for key in weights):
+        if not tree_flatten(child.parameters()) or name in weighted_modules:
             continue
         setattr(model, name, None)
+        dropped_modules.append(name)
+
+    if dropped_modules:
```
Lazarus-931: added here

Lazarus-931:
> Looks good, although I wonder if it should log the missing module name so it's not a silent failure if there's a real weight mismatch?

yes, i didn't think that actually, added logging for those models

lucasnewman (approved):

PR #1883, opened 2026-08-13
fix(rope): support per-batch array offsets in eager RoPE

nazerim:
## Summary

Fix the eager RoPE implementation for **per-batch array offsets** (the offset type passed by `BatchRotatingKVCache` / `BatchKVCache`). Without this, multi-request continuous batching crashes on the first merged decode step.

## The bug

`_eager_rope_angles` computed positions as:

```python
positions = mx.arange(x.shape[-2], dtype=mx.float32) + offset
```

For a batch KV cache, `offset` is a per-batch array of shape `(B,)` (each row's absolute position). Broadcasting `arange(S)` (shape `(S,)`) with `offset` (shape `(B,)`) silently grows the seq dim to `B`:

- A decode step enters with `S=1` and a 2-row batch: `positions` becomes `(2,)`, cos/sin end up `(1, 1, 2, D)`, and the rope output grows from `(2, 32, 1, 128)` to `(2, 32, 2, 128)`.
- `cache.update_and_fetch` then receives `S=2` keys, while the attention mask was built with `N=1` (`make_mask(N)` uses the real query length).
- `mx.fast.scaled_dot_product_attention` raises:

```
ValueError: [broadcast_shapes] Shapes (2,1,1,2048) and (2,32,2,2049) cannot be broadcast.
```

This is the "batched path" 500 seen when the server batches two concurrent streaming requests (e.g. an agent sending parallel turns), one of which has prefilled a full sliding window. It only manifests with **B ≥ 2** (the offset array size must differ from `S=1`), which is why single-request tests never caught it. It is not specific to any model — any model using `implementation="eager"` RoPE with batch caches is affected.

## The fix

- `_eager_rope_angles` now builds explicit `(B, S)` positions for array offsets (`arange(S)[None, :] + offset[:, None]`) and `(B, S, F)` angles; the scalar path is unchanged.
- `_compiled_eager_rope` expands cos/sin to `(B, 1, S, D)` for the array case (`[:, None]`) vs `(1, 1, S, D)` for the scalar case (`[None, None]`).

The scalar path is bit-identical to before (verified with `mx.array_equal` against the previous implementation). Array offsets produce results identical to applying the rope per-row with each row's scalar offset.

## Verification

- **Regression test**: `test_rope_utils.py::test_eager_rope_per_batch_offset_does_not_expand_seq` — checks the seq dim is preserved for `(B, S=1)` array offsets and that array offsets equal per-row scalar application, for both traditional and non-traditional layouts.
- **Full suite**: `pytest mlx_vlm/tests/` → 2116 passed, 1 failed (pre-existing unrelated Qwen-speculative failure), 5 skipped.
- **Live**: with 2 concurrent streaming requests (68 + 34,258 tokens) the server previously crashed at ~45 s with the broadcast error; with this fix both requests complete (`finish_reason=stop`, zero errors). Minimal repro in the test; a server-level repro used the exact error above.

## Repro (pre-fix)

```python
from mlx_vlm.models.rope_utils import EagerRoPE
import mlx.core as mx

rope = EagerRoPE(128, base=500000.0, scale=1.0)
x = mx.ones((2, 32, 1, 128))
print(rope(x, offset=mx.array([86, 34258])).shape)
# pre-fix: (2, 32, 2, 128)  <- seq dim expanded to the batch size
# post-fix: (2, 32, 1, 128)
```

### DIFF
--- a/mlx_vlm/models/rope_utils.py
+++ b/mlx_vlm/models/rope_utils.py
@@ -13,21 +13,15 @@ _FULL_COS = "full"
 
 
 def _eager_rope_angles(x, offset, frequencies, scale, traditional):
-    # offset may be a scalar (single-row cache) or a per-batch array
-    # (BatchRotatingKVCache / BatchKVCache offsets). Build (B, S) positions
-    # for the array case so the seq dim is not accidentally expanded to B.
-    S = x.shape[-2]
-    if isinstance(offset, mx.array) and offset.ndim > 0:
-        positions = (
-            mx.arange(S, dtype=mx.float32)[None, :] + offset[:, None]
-        )  # (B, S)
-    else:
-        positions = mx.arange(S, dtype=mx.float32) + offset  # (S,)
+    # offset is always an mx.array here (EagerRoPE.__call__ wraps it): 0-d for
+    # a scalar offset or (B,) for a per-batch offset (BatchRotatingKVCache /
+    # BatchKVCache). Give it a trailing axis so broadcasting keeps the seq dim
+    # S instead of accidentally expanding it to B.
+    positions = (
+        mx.arange(x.shape[-2], dtype=mx.float32)[None, :] + offset[..., None]
+    )  # (1, S) or (B, S)
     positions = positions * scale
-    if positions.ndim == 2:
-        angles = positions[:, :, None] * frequencies[None, None]  # (B, S, F)
-    else:
-        angles = positions[:, None] * frequencies[None]  # (S, F)
+    angles = positions[:, :, None] * frequencies[None, None]  # (B, S, F)
     if traditional:
         return mx.repeat(angles, 2, axis=-1)
     return mx.concatenate([angles, angles], axis=-1)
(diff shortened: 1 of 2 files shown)


Lazarus-931:
thx for the pr @nazerim, quick q

could this could be simplified if we just give offset a trailing axis and let broadcasting do the work, since `EagerRoPE` already  converts the offset with mx.array(offset, ...)?

nazerim:
Great catch — thanks for the suggestion, it's valid and now applied in `b9fdc7b`.

One nuance found while verifying: `offset[:, None]` on a 0-d array raises in MLX (`Too many indices for array with 0 dimensions`), so the single path uses `offset[..., None]` — trailing-axis broadcast works for both the 0-d scalar (`(1,)`) and per-batch `(B,)` offsets. `EagerRoPE.__call__` already wraps `offset` in `mx.array(offset, dtype=mx.float32)`, so no `isinstance` check is needed.

Verification:
- Bit-identical to the previous two-branch implementation for: scalar int, 0-d array, `(B,)` array, S=1 and S=3, traditional and non-traditional, fp32 and int32 offsets — through the compiled (`@mx.compile`) path that `__call__` actually uses.
- Regression test extended with an explicit 0-d array offset case; full suite: 2116 passed, 1 pre-existing unrelated failure (Qwen speculative).

Also, the failing CI check was our own commit: `pre-commit run --all` (black) modified the test file from the first commit, tripping the `git diff --quiet` guard. That's fixed by this commit too.

nazerim:
Note for context: with #1848 (merged) Muse Glimmer's `language.py` now calls `initialize_rope(..., implementation="eager")` — i.e. upstream's default Glimmer serving path runs through `EagerRoPE`, exactly the code this fix hardens for per-batch offsets. That makes the batched-path crash reachable on the plain server path (any 2 concurrent streaming requests), not just a hypothetical batch-cache scenario.

lucasnewman:
@nazerim Can you [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits and force push so we can clear the merge requirements?

nazerim:
@lucasnewman Commits are now signed and rebased on main

lucasnewman (approved):
Thanks!

PR #1885, opened 2026-08-13
lfm2_vl: implement official image splitting (tiling)

ivanfioravanti:
## Problem

The LFM2-VL compat patch in `mlx_vlm/models/lfm2_vl/processing_lfm2_vl.py` disables image splitting, so every image — including large screenshots — is downscaled to a single view capped at `max_image_tokens` (256 tokens ≈ 512×512 px). The official LFM2.5-VL pipeline (`Lfm2VlImageProcessorFast` in transformers) splits large images into up to 10 tiles of 512×512 plus a thumbnail, with each tile introduced by `<|img_row_R_col_C|>` marker tokens and the thumbnail by `<|img_thumbnail|>`.

The accuracy impact is large. Measured on ScreenSpot-v2 with `LiquidAI/LFM2.5-VL-3B-MLX-8bit`, identical prompt/scoring/decoding:

| web items | untiled (current) | tiled (this PR) |
|---|---|---|
| 15-item subset | 33.3% | **80.0%** |
| 40-item subset | 37.5% | **75.0%** |

(Liquid's published ScreenSpot-v2 web number for this model is 82.2% on vLLM with the official tiled preprocessing.)

## Changes

- `Lfm2VlNumpyImageProcessor`: transformers-parity tiling — `_is_image_too_large` (area > `max_image_tokens × patch² × downsample² × max_pixels_tolerance`), aspect-preserving grid layout within `[min_tiles, max_tiles]` (`_find_closest_aspect_ratio`), resize → row-major 512² tiles + downsampled thumbnail last, each emitted as its own `pixel_values` row / `spatial_shapes` entry with its own attention mask.
- `_patched_call`: official text expansion — `<|image_start|>`, `<|img_row_r_col_c|>` before each tile's `<image>`×256, `<|img_thumbnail|>` before the thumbnail tokens, `<|image_end|>`, mirroring `Lfm2VlProcessor._build_image_tokens` in transformers (including ceil-per-dimension downsampling padding via `_num_image_tokens_from_patch_grid`).
- Small images (below the too-large threshold) keep the exact previous single-view behavior; multi-image interleaving unchanged.

No shared code changes: `merge_input_ids_with_image_features` already scatters concatenated per-row embeddings into all `<image>` positions with an exact-count check, so tiles-as-rows aligns natively.

## Config policy (point for discussion)

The official defaults (`do_image_splitting=True, min_tiles=2, max_tiles=10, use_thumbnail=True`) are re-applied on load because the LiquidAI `*-MLX-*` repos currently ship `do_image_splitting: false`, which reproduces the handicap rather than reflecting a deliberate user choice. Overrides are honored via `Lfm2VlProcessor.from_pretrained(model, do_image_splitting=False)` or per-call processor kwargs (popped before kwarg routing). Happy to invert this if you prefer strict config honoring.

## Validation

- `LiquidAI/LFM2.5-VL-3B-MLX-8bit`, torch-free path:
  - 2560×1440 → grid 4×2, 8 tiles (32,32) + thumbnail (24,42), expanded prompt contains the full marker sequence and exactly 2300 `<image>` tokens (8×256 + 252); per-row embedding count matches token count.
  - 960×540 → single view, 252 tokens, no markers, `input_ids` byte-identical to the previous path.
  - Multi-image (tiled + small) alignment verified; marker tokens resolve as single tokenizer ids (124908–124921, 125008).
- E2E grounding generation on ScreenSpot web items parses cleanly with no embedding/token misalignment.
- Tests: 3 new processor tests + 1 updated (it asserted the old single-view behavior), all 129 `test_processors` tests and 13 `test_models` lfm2_vl tests pass; the 2 `TestAWQ` failures are pre-existing on `main`.

References implemented against: `transformers` `models/lfm2_vl/image_processing_lfm2_vl.py` / `processing_lfm2_vl.py` (authoritative), and llama.cpp ≥ b10107 `tools/mtmd/mtmd-image.cpp` `mtmd_image_preprocessor_lfm2` (which shipped the same tiling for GGUF).

### DIFF
--- a/mlx_vlm/models/lfm2_vl/processing_lfm2_vl.py
+++ b/mlx_vlm/models/lfm2_vl/processing_lfm2_vl.py
@@ -34,7 +34,6 @@ from transformers.models.lfm2_vl.processing_lfm2_vl import (
 
 from ..base import install_auto_processor_patch, load_chat_template
 
-
 # Official transformers preprocessing defaults for LFM2-VL: images larger than
 # `max_pixels_tolerance` * `max_image_tokens` * patch^2 * factor^2 pixels are
 # split into a grid of `tile_size` tiles (grid within [min_tiles, max_tiles],
@@ -707,9 +706,7 @@ def _patched_call(self, images=None, text=None, **kwargs):
     # max_tiles, use_thumbnail) regardless of how the processor's kwarg
     # routing classifies them.
     splitting_overrides = {
-        key: kwargs.pop(key)
-        for key in _OFFICIAL_SPLITTING_DEFAULTS
-        if key in kwargs
+        key: kwargs.pop(key) for key in _OFFICIAL_SPLITTING_DEFAULTS if key in kwargs
     }
 
     # Ensure we're using the slow processor (fast requires PyTorch tensors)
--- a/mlx_vlm/models/lfm2_vl/processing_lfm2_vl.py
+++ b/mlx_vlm/models/lfm2_vl/processing_lfm2_vl.py
@@ -453,9 +453,13 @@ class Lfm2VlNumpyImageProcessor(ImageProcessingMixin):
         return self.preprocess(images, return_tensors=return_tensors, **kwargs)
 
 
-# Try to import the slow image processor to force its use. Some Transformers
-# versions import torch from the SigLIP2 processor module, so fall back to a
-# local PIL/NumPy implementation when torch/torchvision are absent.
+# Try to import the slow image processor. Some Transformers versions import
+# torch from the SigLIP2 processor module. Historically the real Siglip2
+# processor was used when importable, but it has no LFM2-VL tiling and its
+# (B, C, H, W) output never matched the packed-patch contract the MLX model
+# consumes — so behavior silently differed between torch and torch-free
+# environments. The NumPy processor below is now used unconditionally; this
+# import only feeds the _SLOW_PROCESSOR_AVAILABLE fallback below.
 try:
     from transformers.models.siglip2.image_processing_siglip2 import (
         Siglip2ImageProcessor,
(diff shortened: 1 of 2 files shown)


Lazarus-931:
hi @ivanfioravanti, thanks for the pr!

quick q

`Lfm2VlNumpyImageProcessor` runs if `Siglip2ImageProcessor` raises an import error(when user does not have `torch`), if import succeeds (`torch` present), the real Siglip2 processor runs → no tiling, no thumbnail?

ivanfioravanti:
Great catch @Lazarus-931 I'm on it. 
And then I'll do a final round of the full benchmark (2 hours) to see final results vs GGUF.

ivanfioravanti:
Good catch — exactly right. As the PR stood, `_patched_from_pretrained` built the real `Siglip2ImageProcessor` whenever its import succeeded (torch present), and that path has neither tiling nor the thumbnail: it returns `(B, C, H, W)` tensors, so `_patched_call` fell back to the single-view square-guess grid info. The tiling only reached the torch-free NumPy path.

Fixed in 7cbfec9: the NumPy image processor is now used **unconditionally** — in `_patched_from_pretrained`, the patched `__init__`, and the call-time `_ensure_slow_processor`. It's the only implementation with the official tiling + thumbnail and the packed-patch/`spatial_shapes` output the MLX model consumes (the real Siglip2 output never actually matched that contract — the `_compute_image_grid_info` square-guess fallback existed precisely for it). Behavior is now identical with or without torch installed.

Regression tests added:
- `test_non_numpy_image_processor_swapped_with_tiling` — a stand-in for the torch-present Siglip2 gets swapped out before any image is processed (and raises if it's ever called), with full tile/marker/token-count assertions on a 2560×1440 input
- `test_from_pretrained_uses_numpy_image_processor_despite_siglip2` — `from_pretrained` ignores an importable Siglip2 and re-applies the official splitting defaults

All 130 processor tests pass.

ivanfioravanti:
Still testing this in a full benchmark suite using mlx-vlm and GGUF on LFM 2.5 VL 3B and comparing results. 🤞🏻

ivanfioravanti:
Before (but temp 0) and After (temp 0.2). Now this crushes GGUF. Prompts still to be tuned on both sides but it's a great start.
Screenshot 2026-08-13 at 07 10 28

Screenshot 2026-08-13 at 18 42 42

ivanfioravanti:
I think I've found an issue on llamacpp side, in case I'll open a detailed issue there, for the time being I'm focused on running this on MLX 💪

lucasnewman:
@ivanfioravanti Thanks for implementing this, it looks good to me!

Would you mind [signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits and force pushing so we can clear the merge requirements?

ivanfioravanti:
Thanks @lucasnewman done. Everything should be updated and ok.

ivanfioravanti:
Final benchmarks after prompts have been fixed is this.
HPoZ2rRWcAAp66G

lucasnewman (approved):
🚀

PR #1887, opened 2026-08-13
fix(molmo_point): honor point_color in draw_points_on_image

Anai-Guo:
## What

`draw_points_on_image` accepts and documents a `point_color` parameter, but never reads it:

```python
colors = ["red", "blue", "green", "yellow", "purple", "orange", "cyan", "magenta"]

for obj_id, img_num, x, y in points:
    color = colors[obj_id % len(colors)]   # point_color is never consulted
```

So every marker is colored from the hardcoded per-object palette and an explicit
`point_color="blue"` silently does nothing. The declared default (`"red"`) also
misdescribes what actually happens — with the palette in place, object 0 is red,
object 1 is blue, object 2 is green, and so on.

## Fix

`color = point_color or colors[obj_id % len(colors)]`, with the default changed to
`None`. The per-object palette is clearly deliberate (it is what keeps several
predicted objects distinguishable in one image), so it stays the default; the
parameter now works when a caller actually passes one.

**No behavior change on the default path** — callers that don't pass `point_color`
get exactly the same image as before.

## Verification

`point_utils.py` imports only `numpy` and `PIL`, so it can be exercised standalone.
Three points with object ids 0/1/2 drawn on a blank canvas, sampling the fill pixel
inside each marker:

| call | before | after |
|---|---|---|
| `draw_points_on_image(img, points)` | `(255,0,0) (0,0,255) (0,128,0)` | `(255,0,0) (0,0,255) (0,128,0)` — unchanged |
| `draw_points_on_image(img, points, point_color="blue")` | `(255,0,0) (0,0,255) (0,128,0)` — ignored | `(0,0,255) (0,0,255) (0,0,255)` — honored |

### DIFF
--- a/mlx_vlm/models/molmo_point/point_utils.py
+++ b/mlx_vlm/models/molmo_point/point_utils.py
@@ -79,7 +79,7 @@ def draw_points_on_image(
     points: List[Tuple[int, int, float, float]],
     output_path: Optional[str] = None,
     point_radius: int = 8,
-    point_color: str = "red",
+    point_color: Optional[str] = None,
     label: bool = True,
 ) -> Image.Image:
     """Draw point predictions on an image.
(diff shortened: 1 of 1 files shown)


lucasnewman (approved):
Good catch, thanks!

lucasnewman:
@Anai-Guo Can you [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits and force push so we can clear the merge requirements?

Anai-Guo:
@lucasnewman Done — the commit is now GPG-signed (verified) and force-pushed. The reopen was just to refresh the branch head after re-signing. Thanks!

PR #1890, opened 2026-08-13
Keep pending padding metadata aligned after filtering

### LINKED ISSUE #1889: Batch cache filtering leaves pending padding metadata out of sync
## Description

Batch cache `filter()` methods remove or reorder rows in their key/value tensors, offsets, and left-padding state. When right-padded prefill is active, however, some caches also hold temporary per-row metadata that is not filtered with the rest of the batch:

- `BatchKVCache._right_padding`
- `BatchRotatingKVCache._lengths`
- `BatchQuantizedKVCache._right_padding`

After filtering, the cache tensors and the pending padding metadata can therefore describe different batches.

For `BatchRotatingKVCache`, the next prefill update combines a one-row offset with stale two-row lengths. Broadcasting expands the cache operation back to two rows, and concatenating the next one-row chunk fails with a shape mismatch.

For `BatchKVCache` and `BatchQuantizedKVCache`, the mismatch is applied during `finalize()`. Instead of raising an error, broadcasting can silently expand a filtered one-row cache back to two rows.

This is exposed whenever rows are filtered after `prepare()` and before `finalize()`, such as cancellation or admission changes at a chunked-prefill boundary.

## Reproduction

Run this script from the repository root:

```python
import mlx.core as mx

from mlx_vlm.models.cache import (
    BatchKVCache,
    BatchQuantizedKVCache,
    BatchRotatingKVCache,
)


def random_kv(batch, length, dim):
    return (
        mx.random.normal((batch, 2, length, dim)),
        mx.random.normal((batch, 2, length, dim)),
    )


for cache, dim in (
    (BatchKVCache([0, 0]), 32),
    (BatchQuantizedKVCache([0, 0], group_size=32, bits=8), 32),
):
    cache.prepare(right_padding=[2, 0], lengths=[4, 6])
    cache.update_and_fetch(*random_kv(2, 3, dim))
    cache.filter(mx.array([1], dtype=mx.int32))
    print(type(cache).__name__, "before finalize:", cache.offset.shape)
    cache.finalize()
    print(type(cache).__name__, "after finalize:", cache.offset.shape)


rotating = BatchRotatingKVCache(32, [0, 0])
rotating.prepare(right_padding=[2, 0], lengths=[4, 6])
rotating.update_and_fetch(*random_kv(2, 3, 32))
rotating.filter(mx.array([1], dtype=mx.int32))
rotating.update_and_fetch(*random_kv(1, 2, 32))
```

## Actual behavior

The standard and quantized caches return to a two-row shape during `finalize()`:

```text
BatchKVCache before finalize: (1,)
BatchKVCache after finalize: (2,)
BatchQuantizedKVCache before finalize: (1,)
BatchQuantizedKVCache after finalize: (2,)
```

The rotating cache fails on the next update:

```text
ValueError: [concatenate] All the input array dimensions must match exactly except for the concatenation axis
```

## Expected behavior

Filtering a batch cache should apply the same row selection and ordering to all per-row state, including metadata pending until `finalize()`. Subsequent prefill updates and finalization should preserve the filtered batch size and operate on the correct rows.

byoan:
Fixes #1889

## Summary

Batch cache filtering already applies row indices to key/value tensors, offsets, and left padding, but it left temporary right-padding metadata unchanged. Filtering during a prepared prefill could therefore leave one-row cache tensors paired with metadata for the original batch.

For rotating caches, that mismatch caused the next chunk update to fail when stale `_lengths` broadcast the cache back to the old batch size. For standard and quantized batch caches, `finalize()` could silently expand a filtered cache back to the old number of rows.

This change keeps the pending metadata aligned in all affected cache implementations:

- filters `BatchKVCache._right_padding`;
- filters `BatchRotatingKVCache._lengths`;
- filters `BatchQuantizedKVCache._right_padding`.

Regression tests reorder rows before removing one, then continue or finalize prefill. This verifies both batch-size alignment and correct metadata association after filtering.

## Test plan

```text
python -m pytest mlx_vlm/tests/test_apc_adapters.py -k "BatchRightPadPrefill" -q
```

### DIFF
--- a/mlx_vlm/models/cache.py
+++ b/mlx_vlm/models/cache.py
@@ -985,6 +985,8 @@ class BatchKVCache(_BaseCache):
             self.values = self.values[batch_indices]
         self.offset = self.offset[batch_indices]
         self.left_padding = self.left_padding[batch_indices]
+        if self._right_padding is not None:
+            self._right_padding = self._right_padding[batch_indices]
 
         # Shift left to reduce padding
         min_left_pad = self.left_padding.min().item()
@@ -1341,6 +1343,8 @@ class BatchRotatingKVCache(_BaseCache):
             self.values = self.values[batch_indices]
         self.offset = self.offset[batch_indices]
         self.left_padding = self.left_padding[batch_indices]
+        if self._lengths is not None:
+            self._lengths = self._lengths[batch_indices]
 
     def extend(self, other):
         """
@@ -1789,6 +1793,8 @@ class BatchQuantizedKVCache(_BaseCache):
             self.values = tuple(v[batch_indices] for v in self.values)
         self.offset = self.offset[batch_indices]
         self.left_padding = self.left_padding[batch_indices]
+        if self._right_padding is not None:
+            self._right_padding = self._right_padding[batch_indices]
 
         min_lp = self.left_padding.min().item()
         if min_lp > 0:
--- a/mlx_vlm/tests/test_apc_adapters.py
+++ b/mlx_vlm/tests/test_apc_adapters.py
@@ -472,7 +472,40 @@ def _fill_batch_turbo(left_padding, seq_len, bits=4.0):
     return cache, k, v
 
 
-class TestBatchRotatingRightPadPrefill:
+class TestBatchRightPadPrefill:
+    @staticmethod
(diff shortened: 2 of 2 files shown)


lucasnewman (approved):
Thanks!

PR #1893, opened 2026-08-14
Add native sequence classifier rerankers

Lazarus-931:
reranker support was currently very limited to qwen, this adds native BERT, XLM-RoBERTa, and ModernBERT sequence-classifier scoring, and also replaces the endpoint’s mlx-lm causal-mask import with our own native implementation.

all has been verified + tested


lucasnewman (approved):
Looks good!

PR #1895, opened 2026-08-14
Support per-generator prefill step size overrides

### LINKED ISSUE #1894: ResponseGenerator cannot configure prefill step size per instance
## Description

The server currently resolves its prefill step size exclusively from the process-wide `PREFILL_STEP_SIZE` environment variable. `ResponseGenerator` does not accept a `prefill_step_size` argument, and its continuous batching, diffusion, and speculative generation paths each read the global value directly.

This works for the command-line server, which sets the environment variable before starting, but it prevents applications embedding mlx-vlm from assigning different prefill step sizes to different generator instances in the same process. Changing the environment is not a safe substitute because it affects every generator and can race with active generation threads.

## Reproduction

Run this script from the repository root. The no-op thread prevents model loading, so no model or network access is required.

```python
import os

import mlx_vlm.server.generation as generation


class DormantThread:
    def __init__(self, *args, **kwargs):
        pass

    def start(self):
        pass


generation.Thread = DormantThread
os.environ["PREFILL_STEP_SIZE"] = "2048"

small_generator = generation.ResponseGenerator(
    model_path="unused-small",
    prefill_step_size=1024,
)
large_generator = generation.ResponseGenerator(
    model_path="unused-large",
    prefill_step_size=4096,
)

print(small_generator.prefill_step_size)
print(large_generator.prefill_step_size)
```

## Actual behavior

The first generator cannot be constructed:

```text
TypeError: ResponseGenerator.__init__() got an unexpected keyword argument 'prefill_step_size'
```

The only available configuration is `PREFILL_STEP_SIZE`, which is shared by the entire process.

## Expected behavior

`ResponseGenerator` should accept an optional `prefill_step_size` argument. An explicit value should belong to that generator and be used by all of its prefill paths, while omitting the argument should preserve the existing environment-variable fallback.

The reproduction should print:

```text
1024
4096
```

This would let embedding applications configure generators independently without changing the existing CLI behavior.

byoan:
Fixes #1894

## Summary

`ResponseGenerator` previously read `PREFILL_STEP_SIZE` directly from the process environment in each generation backend. Embedding applications could not assign different prefill step sizes to generator instances without mutating global process state.

This change adds an optional `prefill_step_size` argument to `ResponseGenerator`. Explicit values are converted to integers and stored on the generator, while omitting the argument preserves the existing `PREFILL_STEP_SIZE` environment fallback.

The effective per-generator value is now used consistently by:

- continuous `BatchGenerator` prefill;
- diffusion generation;
- speculative prefill.

The internal resolver retains an environment fallback for subclasses that bypass `ResponseGenerator.__init__`, preserving compatibility with existing embedding integrations.

Tests cover environment fallback and explicit precedence, then verify that the configured value reaches all three generation paths.

## Test plan

```text
python -m pytest mlx_vlm/tests/test_server.py -q
```

```text
318 passed
```

### DIFF
--- a/mlx_vlm/server/generation.py
+++ b/mlx_vlm/server/generation.py
@@ -1083,6 +1083,7 @@ class ResponseGenerator:
         quantized_kv_start=DEFAULT_QUANTIZED_KV_START,
         top_logprobs_k=0,
         apc_manager: Optional["_apc.APCManager"] = None,
+        prefill_step_size: Optional[int] = None,
     ):
         self.model_path = model_path
         self.adapter_path = adapter_path
@@ -1102,6 +1103,11 @@ class ResponseGenerator:
         self.quantized_kv_start = quantized_kv_start
         self.top_logprobs_k = top_logprobs_k
         self.apc_manager = apc_manager
+        self.prefill_step_size = (
+            get_prefill_step_size()
+            if prefill_step_size is None
+            else int(prefill_step_size)
+        )
         self.apc_mode = None
         self.tokenizer = None
         self.requests: Queue = Queue()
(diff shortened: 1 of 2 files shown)


lucasnewman (approved):
Thanks!

PR #1897, opened 2026-08-14
Accept list rollback state for DeepSeek V4 MTP

### LINKED ISSUE #1896: DeepSeek-V4 MTP speculative decoding fails during cache rollback
## Description

The batched MTP generation path tracks the number of accepted speculative tokens for each active row as a Python list. When a speculative block is not fully accepted, `_mtp_rounds` passes that list to the language model's `rollback_speculative_cache` method.

DeepSeek-V4 currently expects the acceptance values to be an MLX array. It calls `.max()`, `.size`, and `.tolist()` directly on the value, so rollback fails as soon as MTP supplies the Python list produced by the shared speculative decoding code.

This prevents DeepSeek-V4 MTP generation from recovering from a partially rejected speculative block.

## Reproduction

Run the following script from the repository root. It exercises the same Python-list input passed by the MTP generation loop without requiring model weights.

```python

from mlx_vlm.models.deepseek_v4.language import LanguageModel


class DummyCache:
    keys = None

    def __init__(self):
        self.trims = []

    def trim(self, amount):
        self.trims.append(amount)


cache = DummyCache()
accepted = LanguageModel.rollback_speculative_cache(
    SimpleNamespace(),
    [cache],
    None,
    [0],
    block_size=2,
)

print(accepted)
print(cache.trims)
```

## Actual behavior

The rollback fails before updating the cache:

```text
AttributeError: 'list' object has no attribute 'max'
```

## Expected behavior

DeepSeek-V4 rollback should accept the Python list supplied by the MTP generation path, while preserving support for scalar integers and MLX arrays.

The reproduction should complete and print:

```text
0
[1]
```

The same normalized acceptance values should also be used by the GDN cache restoration path and the per-row tail cleanup used for batched generation.

byoan:
Fixes #1896

## Summary

The shared MTP generation loop passes per-row acceptance lengths to `rollback_speculative_cache` as a Python list. DeepSeek-V4 previously treated that value as an MLX array and called array-specific methods on it, causing rollback to fail after a partially rejected speculative block.

This change normalizes scalar integers, MLX arrays, and iterable acceptance values to a Python list at the start of DeepSeek-V4 rollback. The normalized values are then used consistently to:

- determine the maximum accepted length and uniform cache trim;
- restore and replay GDN-backed cache state;
- detect batched rollback;
- clear rejected per-row cache tails.

An empty acceptance sequence now raises an explicit error instead of failing later while computing the maximum.

Tests cover Python-list input for both the regular cache-trimming path and the GDN restore-and-replay path. The existing batched TurboQuant rollback coverage continues to exercise MLX-array input and per-row tail cleanup.

## Test plan

```text
python -m pytest mlx_vlm/tests/test_speculative.py -q
```

```text
158 passed, 2 warnings
```


byoan:
@lucasnewman Could you have a look at this when you have the chance? Thanks!

PR #1898, opened 2026-08-14
Report why an APC lookup reused a prefix or did not

Lazarus-931:
A deployment can see an aggregate hit rate but not why a request re-prefilled, so a cold miss looks the same as a usable prefix that a media constraint rejected. The lookup now returns its outcome alongside the plan and counts outcomes in the stats snapshot.


lucasnewman (approved):

Blaizzy:
Closing it in favour of a issue with reproducible example and expected output

PR #1899, opened 2026-08-14
Add Qwen3.8-27B support

Lazarus-931:
Adds native Qwen3.8-27B support with released-config coverage, blockwise FP8 loading, ModelOpt NVFP4 adaptation, and MTP splitting.

WIP, experimenting with sep mixed precision for decode and prefill

So I was experimenting with different percisons for the overall qwen3_5 backbone, and for qwen3.8 specifically, 

image

To summarize, mxfp4 is the best precision for this model considering perf + quality of output


mlx_vlm/generate/ar.py
Blaizzy, line 1: Please revert this

mlx_vlm/models/qwen3_5/weights.py
Blaizzy, line 1: Please move this into language.py

mlx_vlm/utils.py
Blaizzy, line 1: Please follow the DeepSeek V4 pattern here instead of adding the model-specific FP8 transformation helpers to `utils.py`.

Keep the Qwen FP8 conversion and `make_quantization_config` in `models/qwen3_5/language.py`; `utils.py` should only dispatch to the model-owned configuration:

```python
elif quant_method == "fp8" and config.get("model_type") == "qwen3_5":
    from .models.qwen3_5.language import make_quantization_config

    quantization = make_quantization_config(model)
```

This mirrors the existing DSV4 branch directly and keeps the format-specific implementation in the model folder.

Lazarus-931: yes, should have done so. will change!

Lazarus-931:
holding off, want to make fp8->bf16 conversion into mxfp8 instead with specific quantized matmul

Blaizzy:
We already have mxfp8 support by default.

mlx_vlm/generate/ar.py
```diff
@@ -2726,7 +2726,6 @@ def _next(self, **kwargs):
                 elapsed = time.perf_counter() - tic
                 self._prompt_time_counter += elapsed
                 self._record_prompt_batch_time(self._prompt_batch, elapsed)
-                self._prompt_tokens_counter += n
```
lucasnewman, line 2737: This doesn't seem related / correct?

lucasnewman (changes requested):
@Lazarus-931 Looks good other than the `ar.py` change, can you fix that first?

lucasnewman:
@Lazarus-931 Tests are failing

lucasnewman (approved):

jrp2014:
Which specific models are supported?

PR #1903, opened 2026-08-14
Complete PoolingCache continuous batching support

### LINKED ISSUE #1902: PoolingCache does not fully support continuous batching
## Description

DeepSeek-V4 uses hybrid cache entries composed of a rotating KV cache and one or more pooling caches. The continuous batching path currently has two gaps when handling this layout.

First, `_make_cache` knows how to convert regular KV and rotating caches into their batch-aware counterparts, but it has no conversion for `PoolingCache`. Because conversion recurses through `CacheList`, attempting to create a batched DeepSeek-V4 cache eventually reaches the pooling entry and raises a `ValueError`.

Second, `_extend_cache` uses the presence of a `left_padding` attribute to decide whether a cache is already batched. A `CacheList` does not expose that attribute even when all of its children are already batch-aware. When a new request joins an active DeepSeek-V4 generation batch, the nested `CacheList` is therefore passed through `merge` again. This treats `BatchRotatingKVCache` and `BatchPoolingCache` instances as single-row cache objects and breaks their batch metadata.

Together, these issues prevent PoolingCache-based models from using the full continuous batching lifecycle: initial cache construction followed by admission of concurrent requests.

## Reproduction

Run the following script from the repository root. It uses the same cache types as DeepSeek-V4 but does not require model weights.

```python
import mlx.core as mx

from mlx_vlm.generate import ar
from mlx_vlm.models.cache import (
    BatchPoolingCache,
    BatchRotatingKVCache,
    CacheList,
    PoolingCache,
)


class PoolingModel:
    def make_cache(self):
        return [PoolingCache(ratio=4)]


def make_row(prompt_length):
    rotating = BatchRotatingKVCache(max_size=16, left_padding=[0])
    keys = mx.random.normal((1, 1, prompt_length, 4))
    values = mx.random.normal((1, 1, prompt_length, 4))
    rotating.update_and_fetch(keys, values)
    rotating.finalize()

    pooling = BatchPoolingCache(ratio=4, left_padding=[0])
    pooling.prepare(lengths=[prompt_length])
    kv = mx.random.normal((1, prompt_length, 3))
    gate = mx.random.normal((1, prompt_length, 2))
    ready_kv, _, _ = pooling.accumulate_windows(kv, gate, offset=0)
    pooled = mx.random.normal((1, ready_kv.shape[1] // 4, 3))
    pooling.update_and_fetch(pooled)
    pooling.finalize()

    return [CacheList(rotating, pooling)]


try:
    caches = ar._make_cache(PoolingModel(), [0, 0])
    print(type(caches[0]).__name__, caches[0].remainder)
except Exception as error:
    print("cache creation failed:", repr(error))

try:
    caches = ar._extend_cache(make_row(6), make_row(5))
    rotating, pooling = caches[0].caches
    print(rotating.keys.shape[0], pooling.remainder)
except Exception as error:
    print("cache admission failed:", repr(error))
```

## Actual behavior

Initial conversion rejects `PoolingCache` because the type is not supported by `_make_cache`.

If batch-aware nested caches are supplied directly, admitting the second request attempts to merge them again and fails while treating the list-valued `BatchPoolingCache.remainder` as a scalar remainder.

## Expected behavior

`PoolingCache` should be converted to `BatchPoolingCache` while preserving its compression ratio and initializing state for every batch row.

Cache admission should recognize a nested `CacheList` as already batched when all of its children implement the batch `filter` and `extend` operations. The reproduction should then print:

```text
BatchPoolingCache [0, 0]
2 [2, 1]
```

byoan:
Fixes #1902

## Summary

DeepSeek-V4 combines rotating KV caches with pooling caches inside `CacheList` entries. Continuous batching could neither construct this cache layout nor safely extend an active batch when another request joined.

This change completes support for that lifecycle in two places:

- `_make_cache` now converts `PoolingCache` into `BatchPoolingCache`, preserving the pooling ratio and initializing per-row state from the batch padding information.
- `_extend_cache` now determines whether an entry is already batch-aware from its `filter` and `extend` capabilities. For nested `CacheList` entries, it checks every child recursively. This avoids passing existing batch caches through their single-cache `merge` path a second time.

The first regression test verifies direct PoolingCache conversion for continuous batching. The second builds two realistic DeepSeek-V4-style cache rows, admits the joining row into the active batch, and checks that both rotating and pooling cache state remain aligned across the resulting two-row batch.

## Test plan

```text
python -m pytest mlx_vlm/tests/test_generate.py -q
```

```text
119 passed, 2 warnings
```


lucasnewman:
@byoan Thanks for the submission. We need to handle left-padding support properly to make this actually work. I opened https://github.com/Blaizzy/mlx-vlm/pull/1974 to cover the remaining support needed.

PR #1915, opened 2026-08-15
Page a hybrid's attention layers and checkpoint only the rest

Lazarus-931:
A hybrid cannot slice recurrent or convolution state to a position the way attention KV can, so a single such layer sends the whole model to exact mode and every snapshot then copies attention KV the block pool already handles; on LFM2.5-1.2B that is 143 MB where 0.39 MB would do. Page the layers that can page and checkpoint only the ones that cannot, taking checkpoints at the boundaries the prefill already stops at so enabling the cache does not change how a prompt is split. A request that repeats a prompt reuses what it did before; one that shares an opening and then diverges resumes from the last shared boundary instead of re-prefilling from zero. Off by default behind APC_COMPOSITE.

Batched requests fall back to exact, and checkpoints are memory only.

| model | mode | pass | reuse (tok) | ttft vs APC-off | ttft vs exact |
| LFM2.5-1.2B-Instruct-4bit | composite | 17/17 | 8192 | 114x | 1.01x |
| LFM2.5-1.2B-Instruct-8bit | composite | 14/14 | 4096 | 100x | 1.07x |
| Qwen3.5-9B-MLX-4bit | composite | 14/14 | 4080 | 90x | 0.97x |
| LFM2-VL-1.6B-4bit | composite | 11/11 | 4096 | 71x | 1.04x |
| Qwen3.5-0.8B-4bit | composite | 14/17 | 4080 | 66x | 1.11x |
| Qwen3.5-2B-4bit | composite | 14/17 | 4080 | 60x | 1.01x |
| OLMo-1B-hf-4bit-mlx | block | 11/11 | 4096 | 49x | 1.02x |
| Qwen3.5-0.8B-8bit | composite | 14/14 | 4080 | 44x | 1.04x |
| gemma-3-4b-it-4bit | composite | 5/11 | 2064 | 39x | 1.00x |
| LFM2.5-8B-A1B-MLX-4bit | composite | 14/14 | 4096 | 33x | 0.78x |
| Qwen3.5-0.8B-bf16 | composite | 11/11 | 4080 | 31x | 0.96x |
| plamo-2-1b | None | 0/11 | - | - | - |

WIP: I'm working on batched support
apc_reuse

An 8,000-token prompt is cached, then a second request shares a decreasing part of it and diverges. Exact mode keeps one snapshot near the end, so it matches nothing until the whole prompt repeats. Composite resumes from the last shared boundary, in stride-sized steps. Below the first boundary neither has anything to offer.


Lazarus-931:
Review findings addressed in four commits.

Blocks matched by a composite hit were never returned to the pool. Every plan until composite carried either a warm cache with its blocks already released, or blocks with no warm cache, and the caller took ownership only in the second case. Composite returns both, so a run of six requests left 163 blocks pinned from the first hit onward. Ownership is now taken unconditionally, which is a smaller condition than the one it replaces and a no-op for every earlier plan.

The whole-prompt snapshot was taken on the first generated token, by which point the generation loop has already stepped the cache, so the entry held one token more than its key claimed and cloning does not trim. Exact mode has done this since before composite existed. Both now snapshot when prefill finishes, which is the last moment the cache holds the prompt and nothing else; a position that does not match the prefill length is dropped rather than stored.

Composite's state checkpoints were living in the exact-prefix cache, which keeps two entries by default, assumes its entries are whole prompt caches, and writes through to disk. The first meant an eight-thousand-token prompt kept only its last two boundaries, so the divergence case this work exists for did not work at stock settings at all; every measurement so far had the harness raise that limit. They now have their own store, unreachable from an exact lookup, bounded by the same byte budget that picks the stride, and never written to disk. At stock settings an eight-thousand-token prompt keeps five boundaries in 97.7 MB and a request sharing its first three thousand tokens resumes from an early one.

Past fifty thousand tokens the pool stores one layer-major snapshot instead of blocks. A composite harvest hands it only the pageable layers, so that snapshot was a partial cache in the exact store and left the prefix with no blocks to match. That tier is now taken only for a whole harvest.

Numbers in the description predate these commits and are being re-measured.

Lazarus-931:
closing in favor of #1960

PR #1921, opened 2026-08-16
Do not start KV quantization in the middle of prefill

Lazarus-931:
> ⚠️ **Experimental / draft** — still validating.

`quantized_kv_start` is a fixed 5000, so any prompt longer than that crosses the threshold while the prompt is still being ingested: the cache converts mid-prefill and every remaining chunk then attends through the slower quantized path, paying for quantization during the phase that is compute-bound and before the cache is large enough for the bandwidth saving to arrive. It also inflates peak memory, because the conversion overlaps the float cache it replaces — enough that `--kv-bits` could push peak above running unquantized at all (13.0 GB against 7.9 GB on Qwen3-4B at 32K). Hold the default until the prompt is in, so the conversion happens once at the prefill/decode boundary; decode still runs fully quantized, which is where the saving was always coming from. An explicit `--quantized-kv-start` is honored untouched, including one that fires during prefill.

Prompts over 5000 tokens produce different tokens, since the prompt's KV is now quantized once rather than incrementally; shorter prompts are byte-identical.

| model | ctx | prefill tok/s | peak GB |
|---|---:|---:|---:|
| Qwen3-0.6B-4bit | 8K | 13,398 → 22,297 | 2.52 → 1.94 |
| Qwen3-0.6B-4bit | 32K | 2,578 → 8,097 | 8.30 → 4.81 |
| Qwen3-4B-4bit | 8K | 2,610 → 3,432 | 5.27 → 4.16 |
| Qwen3-4B-4bit | 32K | 802 → 1,978 | 13.01 → 7.92 |
| Qwen3-8B-4bit | 8K | 1,709 → 2,003 | 7.62 → 6.54 |
| Qwen3-8B-4bit | 32K | 687 → 1,421 | 15.44 → 10.40 |

`--kv-bits 4` on defaults, M5 Max. Each cell ran main/branch/branch/main so drift cancels; peak memory was identical across repeats. 54 checkpoints across 16 architectures, 3 trees × 15 cells, zero new failures.

WIP: 64K and 128K on Qwen3.6/3.8-27B next.

image

Prefill and peak memory at 8K and 32K, before and after. The gain grows with context because that is where the fixed threshold lands mid-prompt.

### DIFF
--- a/mlx_vlm/generate/ar.py
+++ b/mlx_vlm/generate/ar.py
@@ -36,6 +36,7 @@ from .common import (
     _chunked_prefill_enabled,
     generation_stream,
     maybe_quantize_kv_cache,
+    resolve_quantized_kv_start,
     wired_limit,
 )
 from .types import GenerateKwargs, ProcessorLike, Unpack
@@ -247,6 +248,12 @@ def generate_step(
         Generator[Tuple[mx.array, mx.array], None, None]: A generator producing
           one token and a vector of log probabilities.
     """
+
+    quantized_kv_start = resolve_quantized_kv_start(
+        kv_bits,
+        quantized_kv_start,
+        input_ids.shape[-1] if input_ids is not None else None,
+    )
 
     quantize_cache_fn = functools.partial(
         _generate_module_override("maybe_quantize_kv_cache", maybe_quantize_kv_cache),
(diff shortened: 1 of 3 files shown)


Lazarus-931:
Closing — the gain is concentrated on small models at long context and largely disappears at 27B scale, so it isn't worth the behaviour change for prompts over the threshold. Parking the branch rather than pursuing it.

PR #1922, opened 2026-08-16
Fix SmolVLM split-image token expansion

### LINKED ISSUE #1919: token/feature count mismatch after #1905 removed the SmolVLM multimodal-merge override
**SmolVLM2 inference still broken after #1905: `_prepare_inputs_for_multimodal` now rejects the token/feature count mismatch the old override was written to reconcile.**

Thanks for #1905 (`bd5f926e`, "Fix SmolVLM inference with flattened image features") — it correctly resolves the `not enough values to unpack (expected 3, got 2)` crash that #1852's feature-flattening introduced for SmolVLM. But on `main` at `60fcc7ff`, `mlx-community/SmolVLM2-2.2B-Instruct-mlx` (revision `8445160`) now fails one step later with `ValueError: Image features and image tokens do not match: tokens: 81, features 1053` (from `mlx_vlm/models/idefics3/idefics3.py:175`). The cause is that #1905 deleted SmolVLM's `_prepare_inputs_for_multimodal` override entirely and fell back to the Idefics3 base method, but the two models tokenise images differently: the SmolVLM processor emits a small number of `<image>` placeholder tokens (81 here for a single image) that each stand for a *group* of vision features, whereas the base Idefics3 scatter requires exactly one `<image>` token per feature row (`n_image_mask_elements == image_features.size`). The removed override contained precisely the logic that handled this — the `num_image_tokens % S != 0` group check and the per-image expansion — so the fix removed the accommodation along with the shape bug. Any single-image prompt reproduces it: `python -m mlx_vlm.generate --model mlx-community/SmolVLM2-2.2B-Instruct-mlx --revision 844516024a1c4400d34489b89ee067d794e432ed --image any.jpg --prompt "Describe this image." --max-tokens 50` — worked on `ad98c8b4^`, unpack-crashed from `ad98c8b4`, and count-mismatch-crashes from `bd5f926e`.

The likely resolution is to keep the base class's flattened `(N·S, D)` feature contract but restore a SmolVLM-specific `_prepare_inputs_for_multimodal` that maps groups of features onto the sparser placeholder tokens (i.e. re-implement the deleted override against 2-D features rather than dropping it), or alternatively have the SmolVLM processor expand each `<image>` placeholder to `S` tokens so the base scatter's one-to-one invariant holds. It would also be worth adding a single-image *inference* case for SmolVLM to `test_models.py` alongside the LoRA/shape tests #1905 added — both regressions in this chain would have been caught by one end-to-end generate call. Environment: macOS 26.6.1, M5 Max, mlx `0.32.1.dev20260815+9ab977b56`, mlx-vlm editable at `60fcc7ff`, Python 3.13.14;

atirna:
SmolVLMProcessor never asked the image processor for row/col info, so a split image kept the single-image 81-token prompt while the vision encoder emitted 13x81 features. After #1905 flattened onto the Idefics3 1:1 scatter, that shows up as `ValueError: tokens: 81, features 1053`.

I default `return_row_col_info` to True, same as Idefics3 here and HuggingFace's SmolVLM processor, so the prompt expands one image-token block per tile plus global. The flattened merge stays.

Fixes #1919

## Verification
- before, on current `main`: processor `__call__` did not pass `return_row_col_info`; a 3x4 split stayed at 81 `<image>` tokens
- after: 3x4 + global expands to 1053 `<image>` tokens; flattened 1:1 tests still pass; 81 vs 1053 still raises

### DIFF
--- a/mlx_vlm/models/smolvlm/processing_smolvlm.py
+++ b/mlx_vlm/models/smolvlm/processing_smolvlm.py
@@ -174,8 +174,8 @@ class SmolVLMProcessor(ProcessorMixin):
             images = self.image_processor.fetch_images(images)
             images = make_nested_list_of_images(images)
 
-            # Separate image kwargs
-            images_kwargs = {}
+            # Separate image kwargs — always request row/col info for tiling
+            images_kwargs = {"return_row_col_info": True}
             for k in list(kwargs.keys()):
                 if k in ("return_row_col_info",):
                     images_kwargs[k] = kwargs.pop(k)
--- a/mlx_vlm/tests/test_models.py
+++ b/mlx_vlm/tests/test_models.py
@@ -1789,6 +1789,27 @@ class TestSmolVLMFlattenedImageFeatures(unittest.TestCase):
         self.assertEqual(output.shape, inputs_embeds.shape)
         self.assertTrue(mx.array_equal(output, expected).item())
 
+    def test_prepare_inputs_rejects_split_image_token_mismatch(self):
+        """#1919 production shape: 81 placeholders vs 13 flattened tiles.
+
+        The Idefics3 1:1 scatter must keep rejecting this. The processor is
+        what expands the prompt to 13 x 81 tokens; do not paper over the
+        mismatch by dropping extra tiles.
+        """
+        from mlx_vlm.models import smolvlm
+
+        model = SimpleNamespace(config=SimpleNamespace(image_token_index=9))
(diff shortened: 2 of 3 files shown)


Lazarus-931 (approved):
lgtm, thx @atirna!

Lazarus-931:
@atirna, can you sign your commits just so we can merge?

atirna:
signed it, same patch. should be verified now

PR #1923, opened 2026-08-16
feat(server): add conservative DFlash APC prefix reuse

pierre427:
## Summary

This is the smaller single-PR replacement requested in the review of #1856
and #1857.

- add opt-in exact-prefix APC reuse to the text-only B=1 DFlash server path;
- reuse the request's existing tenant-scoped semantic key;
- restore one exact prompt cache directly and prefill only its uncached suffix;
- report cached tokens through the existing streaming metrics;
- preserve absolute drafter RoPE positions with an additive offset that does
  not alter cache storage bookkeeping; and
- leave batched and multimodal requests on the existing cold path.

The implementation calls the shared `apc_lookup_plan()` and
`store_exact_cache()` APIs directly. It does not add a parallel APC module or
change the shared APC manager/adapters.

## Scope

Deliberately excluded from this PR:

- multimodal APC or multimodal hashing changes;
- mixed warm/cold batches and B>=2 reuse;
- static system-prefix checkpoints;
- disk/LRU policy changes; and
- resumable prefill scheduling.

Those are independent concerns and are not required for the conservative
growing-conversation path here.

## Validation

- 13 focused DFlash APC/offset regressions passed;
- full server suite: 352 passed;
- generate/APC suites: 189 passed;
- speculative, Laguna, Muse, and APC suites: 208 passed, with one unrelated
  quantized-linear exact-equality failure reproduced on pristine `main`.

The change is default-off unless an APC manager is configured. A live
model-backed cold/warm parity run remains a follow-up gate before this draft is
marked ready for review.

### DIFF
--- a/mlx_vlm/server/generation.py
+++ b/mlx_vlm/server/generation.py
@@ -159,6 +159,47 @@ def _drop_prefill_kwargs(prompt_kwargs: dict, keys: List[str], n: int) -> dict:
     return out
 
 
+def _dflash_apc_text_only(request) -> bool:
+    """Return whether a queued request has no multimodal prompt state."""
+    if request.images or request.videos or request.audio:
+        return False
+    raw_inputs = request.raw_inputs or {}
+    return not any(
+        key in raw_inputs
+        for key in ("pixel_values", "pixel_values_videos", "input_features")
+    )
+
+
+def _dflash_apc_lookup(
+    request,
+    input_ids: List[int],
+    *,
+    batch_size: int,
+    draft_kind: str,
+    apc_manager,
+    apc_mode: Optional[str],
+):
+    """Look up a conservative B=1, text-only exact-cache DFlash prefix."""
+    if (
+        batch_size != 1
+        or draft_kind != "dflash"
+        or apc_manager is None
+        or apc_mode != "exact"
+        or request.apc_semantic_hash is None
+        or not _dflash_apc_text_only(request)
+    ):
+        return None
+    return _apc.apc_lookup_plan(
+        apc_manager,
+        input_ids,
+        extra_hash=int(request.apc_semantic_hash),
+        apc_mode="exact",
+        safe_lookup_min=0,
+        suffix_is_text_only=lambda _prefix_len: True,
+        prefix_has_media=lambda _prefix_len: False,
+    )
+
+
 def _run_chunked_speculative_prefill(
     lm,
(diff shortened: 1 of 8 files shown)


gtonic:
Independent confirmation of the problem this PR solves, plus a data point and a possible smaller alternative — I ran into the same thing from the DFlash 2 side and only found this PR afterwards.

**Symptom.** With `--draft-kind dflash`, every request reports `cached_tokens=0`, including the second turn of a conversation whose prompt is a strict prefix extension. `APC_TRACE=1` logs no lookup and no store event at all. The same conversation with `--draft-kind mtp` on the same build reports `cached_tokens=5772/5788`.

**Where it diverges.** `ResponseGenerator._run` routes every drafter except MTP into `_run_speculative`, which builds its own prompt cache via `make_speculative_prompt_cache` and never receives `apc_manager` — it only computes `apc_semantic_hash` for logging. Visible in the logs as `backend=speculative_dflash` versus `backend=continuous_batching`. Worth ruling out one plausible red herring: it is not the chunked-prefill difference. MTP with `--prefill-step-size 65536` (i.e. effectively unchunked) still hits the cache.

**Possible smaller alternative.** The batching generator looked to me like it can already drive these drafters: it is generic over `draft_kind` (`speculative_prefill_kwargs`, `speculative_hidden_state`, `SpeculativeGenerationBatch`) and is handed `apc_manager`, `draft_kind` and `draft_block_size` in the same call — only the routing condition keeps non-MTP drafters away from it. Gating that branch behind an env var was 25 lines and gave, on an M5 Pro with Qwen3.8-27B-4bit and a DFlash 2 drafter (5.8k-token conversation, second turn):

| | `cached_tokens` | decode, 64 / 66 / 76 tok | decode, 5767 tok |
|---|---|---|---|
| `_run_speculative` | 0 | 40.8 / 38.2 / 45.6 t/s | 38.4 t/s |
| batching path | 5748 / 5788 | 38.7 / 40.7 / 43.0 t/s | 40.9 t/s |

Greedy output stayed bit-identical, `--draft-block-size` kept taking effect, `--max-num-seqs 2` was fine, and MTP was unaffected.

I am not proposing that as a competing change — this PR is further along and handles cases (RoPE offsets, the other drafter families' contracts) that my quick experiment did not. I closed my own PR (#1966) in favour of this one; posting the trace evidence and the numbers in case they are useful for review.

PR #1924, opened 2026-08-16
Add video subsampling to images fallback in the server

lucasnewman:
Same idea as https://github.com/Blaizzy/mlx-vlm/pull/1836 but using our existing helper file for video (used by the cli for e.g. Inkling) so we can keep the server-side minimal.


Lazarus-931 (approved):
lgtm!

PR #1926, opened 2026-08-16
Load a custom model class from a model_file declared in config.json

noahzelezny:
## What

Ports `mlx_lm`'s `model_file` mechanism to `load_model`: when a checkpoint's `config.json` declares `model_file: <name>.py`, the model class is imported from that file inside the checkpoint instead of the built-in registry.

**No behavior change for any existing model** — the new branch only triggers when the key is present, and the registry path is byte-for-byte what it was (there's a test asserting exactly that).

## Why

Checkpoints with custom weight formats can then run on a **stock mlx-vlm install** — the runtime travels with the weights, no fork or patch required. `mlx_lm` users already rely on this mechanism for text-only use of such checkpoints (it resolves `model_file` the same way, via `importlib.util.spec_from_file_location`); this brings vision-language parity.

Concrete motivation: we ship vector-quantized builds of Qwen3.5-397B-A17B (custom expert format decoded by JIT `mx.fast.metal_kernel` kernels bundled in the checkpoint's `model.py`). They load under stock `mlx_lm` today; with this hook the same checkpoints gain vision under stock `mlx-vlm`.

## Tests

`mlx_vlm/tests/test_model_file_loading.py`:
- loads a synthetic checkpoint end-to-end through the real `load_model` and asserts the class came from the checkpoint's `model.py`
- asserts a clear `FileNotFoundError` when the declared file is missing
- asserts the registry path is untouched (unknown `model_type` without `model_file` still raises `ValueError`)

All existing `test_utils.py` tests pass.

### DIFF
--- a/mlx_vlm/utils.py
+++ b/mlx_vlm/utils.py
@@ -599,9 +599,7 @@ def get_model_and_args(config: dict, model_path: Optional[Path] = None):
                 f"config.json declares model_file={model_file!r} but "
                 f"{model_file_path} does not exist"
             )
-        spec = importlib.util.spec_from_file_location(
-            "custom_model", model_file_path
-        )
+        spec = importlib.util.spec_from_file_location("custom_model", model_file_path)
         arch = importlib.util.module_from_spec(spec)
         spec.loader.exec_module(arch)
         return arch, "custom"
--- a/mlx_vlm/utils.py
+++ b/mlx_vlm/utils.py
@@ -585,8 +585,27 @@ def get_class_predicate(skip_vision=False, weights=None, quantization_config=Non
     return predicate
 
 
-def get_model_and_args(config: dict):
-    """Resolve a model package and its normalized model type."""
+def get_model_and_args(config: dict, model_path: Optional[Path] = None):
+    """Resolve a model package and its normalized model type.
+
+    If the config declares ``model_file`` and ``model_path`` is provided, the
+    model module is imported from that file inside the checkpoint (the same
+    mechanism mlx_lm supports) instead of the built-in registry.
+    """
+    if model_path is not None and (model_file := config.get("model_file")):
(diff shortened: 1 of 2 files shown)


mlx_vlm/utils.py
```diff
@@ -692,7 +693,24 @@ def load_model(model_path: Path, lazy: bool = False, **kwargs) -> nn.Module:
     for wf in weight_files:
         weights.update(_load_safetensors(wf))
 
-    model_class, _ = get_model_and_args(config=config)
+    if (model_file := config.get("model_file")) is not None:
```
lucasnewman: @noahzelezny Can we inline this into `get_model_and_args()` instead? That function is already responsible for model class resolution so it makes more sense to me to just extend it directly.

noahzelezny:
Done — inlined into `get_model_and_args()` as suggested. One signature note: the function previously only received `config`, which can't locate a file inside the checkpoint, so it gains an optional `model_path: Optional[Path] = None` parameter (default preserves every existing call site). `load_model` passes its path; when `model_path` is absent or the config has no `model_file`, resolution is exactly the registry path as before. Tests unchanged and green (the end-to-end test exercises the new path through `load_model`).

lucasnewman:
Thanks! Looks reasonable to me but @Blaizzy should weigh in since this touches a core file.

lucasnewman:
@noahzelezny Can you [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits and force push so we can clear the merge requirements?

noahzelezny:
@lucasnewman Done — commits are signed and force-pushed. I also rebased onto main while I was at it, so the merge commit is gone and the branch is current. Thank you!

lucasnewman (approved):
Thanks!

PR #1929, opened 2026-08-16
Fix PoolingCache APC restoration

### LINKED ISSUE #1928: PoolingCache cannot be restored or merged by exact APC
## Description

`PoolingCache` satisfies the state and metadata contract used by exact Automatic Prefix Caching, but two parts of that contract do not currently round-trip through the generic APC adapter.

First, the inherited `_BaseCache.from_state` implementation creates an uninitialized object and assigns its state before assigning its metadata. Restoring a PoolingCache with buffered remainder state calls `accumulate_windows` from the state setter, and that method needs the compression `ratio` stored in metadata. Because the ratio has not been restored yet, snapshot cloning fails.

Second, the exact-cache adapter passes each row's prefix length to custom cache merge methods as `merge(caches, prefix_lens)`. `PoolingCache.merge` only accepts the cache list, so combining warm and cold exact-cache rows fails before `BatchPoolingCache.merge` can reconstruct their pooled lengths and remainder state.

These failures prevent PoolingCache-based models from reliably storing and restoring exact APC snapshots or assembling mixed warm/cold APC batches.

## Reproduction

Run the following script from the repository root. It exercises both failures without loading model weights.

```python
import mlx.core as mx

from mlx_vlm.apc_adapters import merge_cache_entries
from mlx_vlm.models.cache import PoolingCache


buffered = PoolingCache(ratio=4)
kv = mx.arange(6, dtype=mx.float32).reshape(1, 3, 2)
gate = mx.arange(3, dtype=mx.float32).reshape(1, 3, 1)
buffered.accumulate_windows(kv, gate, offset=0)

try:
    restored = PoolingCache.from_state(buffered.state, buffered.meta_state)
    print("restored:", restored.ratio, restored.remainder)
except Exception as error:
    print("restore failed:", repr(error))

warm = PoolingCache(ratio=4)
kv = mx.arange(18, dtype=mx.float32).reshape(1, 6, 3)
gate = mx.ones((1, 6, 2), dtype=mx.float32)
ready_kv, _, _ = warm.accumulate_windows(kv, gate, offset=0)
warm.update_and_fetch(mx.ones((1, ready_kv.shape[1] // 4, 3)))
cold = PoolingCache(ratio=4)

try:
    merged = merge_cache_entries([warm, cold], [6, 0])
    print("merged:", merged.remainder, merged._pool_lengths, merged._processed)
except Exception as error:
    print("merge failed:", repr(error))
```

## Actual behavior

Buffered restoration fails because `ratio` is not available when the state setter calls `accumulate_windows`:

```text
restore failed: AttributeError("'PoolingCache' object has no attribute 'ratio'")
```

Exact row merging fails because the adapter supplies the additional prefix-length argument:

```text
merge failed: TypeError('PoolingCache.merge() takes 2 positional arguments but 3 were given')
```

## Expected behavior

PoolingCache should initialize itself from its metadata before assigning buffered state. Its merge method should accept the exact-cache adapter's prefix lengths while continuing to derive batch progress from each row's own pooled length and remainder.

The reproduction should print:

```text
restored: 4 3
merged: [2, 0] [1, 0] [6, 0]
```

This preserves the existing PoolingCache state model while making it compatible with the generic exact APC clone and merge contracts.

byoan:
Fixes #1928

## Summary

Exact APC could identify `PoolingCache` as snapshot-compatible, but restoring and merging pooling caches failed across several generic APC boundaries.

`PoolingCache.from_state()` now constructs the cache with its compression ratio before assigning buffered state. This ensures that restoring a remainder buffer can safely call `accumulate_windows()`.

APC batch-shape detection now inspects nested `CacheList` children recursively. A live single-row layout such as `CacheList(RotatingKVCache, PoolingCache)` is therefore cloned as a single-row cache instead of being incorrectly passed through `CacheList.extract()`. Genuinely batched composite layouts continue to use row extraction.

`PoolingCache.merge()` and `BatchPoolingCache.merge()` now both accept the optional `prefix_lens` argument supplied by the exact-cache adapter. The prefix lengths do not need to be applied directly because the batched pooling cache reconstructs each row’s progress from its pooled length, remainder, and compression ratio.

Regression tests cover:

- snapshotting a realistic single-row `CacheList(RotatingKVCache, PoolingCache)` while preserving the compression ratio and buffered KV/gate remainder;
- extracting the corresponding batched `CacheList(BatchRotatingKVCache, BatchPoolingCache)` layout;
- accepting and forwarding `prefix_lens` through the pooling-cache merge contract;
- rebuilding warm and cold composite rows end to end through `make_warm_batch_exact_cache_multi()`, including their rotating-cache offsets and pooling-cache progress metadata.

## Test plan

```text
python -m pytest mlx_vlm/tests/test_apc_adapters.py -q
```

```text
44 passed, 2 warnings
```

### DIFF
--- a/mlx_vlm/apc.py
+++ b/mlx_vlm/apc.py
@@ -3983,9 +3983,18 @@ def extract_prompt_cache_from_batch(
 
 def _prompt_cache_is_batch_shaped(caches: Sequence[Any]) -> bool:
     """True when every entry can row-extract (Batch* / ArraysCache layout)."""
+
+    def can_extract_row(cache: Any) -> bool:
+        if not callable(getattr(cache, "extract", None)):
+            return False
+        children = getattr(cache, "caches", None)
+        if children is None:
+            return True
+        return all(can_extract_row(child) for child in children)
+
     if not caches:
         return False
-    return all(callable(getattr(c, "extract", None)) for c in caches)
+    return all(can_extract_row(cache) for cache in caches)
 
 
 def snapshot_prompt_cache_row(
--- a/mlx_vlm/models/cache.py
+++ b/mlx_vlm/models/cache.py
@@ -2151,12 +2151,7 @@ class PoolingCache(_BaseCache):
 
     @classmethod
     def merge(cls, caches, prefix_lens=None):
-        # APC's generic exact-cache adapter passes prefix lengths to custom
-        # merge implementations. Pooling caches derive their batched progress
-        # from each row's pooled length and remainder, so no extra alignment is
-        # required here.
-        del prefix_lens
-        return BatchPoolingCache.merge(caches)
+        return BatchPoolingCache.merge(caches, prefix_lens)
 
 
 class BatchPoolingCache(_BaseCache):
@@ -2497,8 +2492,11 @@ class BatchPoolingCache(_BaseCache):
         return cache
 
     @classmethod
-    def merge(cls, caches):
+    def merge(cls, caches, prefix_lens=None):
         """Merge a list of PoolingCache instances into a BatchPoolingCache."""
+        # APC passes prefix lengths to custom merge implementations. Pooling
+        # caches derive progress from each row's pooled length and remainder.
+        del prefix_lens
         B = len(caches)
         if not all(c.ratio == caches[0].ratio for c in caches):
             raise ValueError(
--- a/mlx_vlm/models/cache.py
+++ b/mlx_vlm/models/cache.py
@@ -2118,6 +2118,14 @@ class PoolingCache(_BaseCache):
     def meta_state(self, v):
         self.ratio = v
 
+    @classmethod
+    def from_state(cls, state, meta_state):
+        # Restoring buffered state calls ``accumulate_windows``, which needs
+        # the compression ratio before the generic state setter can run.
+        obj = cls(meta_state)
+        obj.state = state
+        return obj
+
     def is_trimmable(self):
         return self.pooled is None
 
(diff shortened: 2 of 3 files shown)


Lazarus-931:
hi @byoan, thanks for the pr!

quick q:

 `_prompt_cache_is_batch_shaped` only inspects the top-level entry, and `CacheList` defines extract, so a single-row `CacheList(RotatingKVCache, PoolingCache)` gets classified as batch-shaped.

and btw the test passes because it wraps a batch child `(CacheList(BatchPoolingCache))`, which a single-row live cache never is, therefore i don't think #1928 get fixed e2e ?

also `BatchPoolingCache.merge`  still has the 2-arg signature, needs the third `prefix_lens=None`

lmk if you have any questions :)

Lazarus-931 (changes requested):
needs some changes :)

byoan:
Thank you for the review 

I updated `_prompt_cache_is_batch_shaped` to inspect `CacheList` children recursively, so a live single-row `CacheList(RotatingKVCache, PoolingCache)` is cloned rather than incorrectly treated as batch-shaped. I also kept coverage for the genuinely batched layout.
`BatchPoolingCache.merge` now accepts `prefix_lens=None`, with `PoolingCache.merge` forwarding the adapter argument, that was indeed my bad

I also added an end-to-end regression that snapshots realistic warm and cold composite rows and rebuilds them through make_warm_batch_exact_cache_multi, checking the resulting rotating and pooling cache state, so that we cover the whole scenario

Let me know if this works for you!

Lazarus-931 (approved):
lgtm, thx

Issue #1932, opened 2026-08-17
Idefics3/SmolVLM: generation runs past <end_of_utterance> and leaks it into the output — the template's turn terminator is not a stop token


jrp2014:
**Summary**

`mlx-community/Idefics3-8B-Llama3-bf16` (revision `8c2a30c488…`) ends every answer with a literal `<end_of_utterance>` in the returned text, then keeps generating until it happens to hit a Llama EOS. The chat template teaches the model to end each turn with `<end_of_utterance>` (token id 128258), but that id is never in the stopping criteria, so `generate()` neither stops on it nor strips it. This is deterministic and image-independent.

**Evidence — the checkpoint's own metadata**

- `chat_template.json`: `…{{ '<image>' }}{% endif %}{% endfor %}<end_of_utterance>\n{% endfor %}` — every turn is terminated with `<end_of_utterance>`.
- `special_tokens_map.json`: `<end_of_utterance>` is in `additional_special_tokens`; `eos_token` is `<|end_of_text|>` (128001).
- `config.json` `text_config.eos_token_id`: `[128001, 128008, 128009]` — the three Llama-3 EOS ids. **128258 (`<end_of_utterance>`) is not among them.**
- No `generation_config.json` in the snapshot.

**Where it goes wrong**

`mlx_vlm/utils.py` builds `StoppingCriteria` from `config.get("eos_token_id")` / the tokenizer's `eos_token_ids` (lines ~1024, 1136–1143). For this checkpoint that yields only the Llama ids, so the token the model was actually trained to stop on is invisible to the stopping criteria. Observed tail of a real answer:

```
…c column style, inscriptions, staircase.<end_of_utterance>
```

The model then continues past its own turn boundary. (In the direct-API path there's no post-hoc splitter to hide this.

**Reproduction**

```bash
python -m mlx_vlm.generate \
  --model mlx-community/Idefics3-8B-Llama3-bf16 \
  --revision 8c2a30c488 \
  --image any.jpg \
  --prompt "Describe this image in one sentence." \
  --max-tokens 100 --temperature 0.0
```

Expected: the answer ends cleanly at the model's turn boundary. Actual: `<end_of_utterance>` appears verbatim at the end of the text, and generation continues.

**Suggested fix**

[...]

PR #1774, opened 2026-08-02
Stop Idefics3 generation at end of utterance

jrp2014:
## Summary

- expose Idefics3's declared `<end_of_utterance>` token as a processor-specific EOS ID
- preserve processor-specific EOS IDs when generation resets stopping criteria from the model config
- preserve order and avoid duplicate EOS IDs

## Root cause

`Idefics3Processor` adds `<end_of_utterance>` as a special token, but it was not part of mlx-vlm's stopping criteria. `generate()` also resets the criteria from `model.config.eos_token_id`, so adding the token only during processor loading would not survive generation startup.

With `mlx-community/Idefics3-8B-Llama3-bf16` at revision `8c2a30c48864f3251701b7bde40f601d25535098`, native generation ended with:

```text
Keywords: the cut, halesworth, suffolk, england, uk, brick building, red brick, windows, white sign, brick wall, gravel, blue sky, clouds.<end_of_utterance>
```

Passing `--eos-tokens '<end_of_utterance>'` removed the delimiter without changing the answer, confirming the missing stop ID.

## Validation

- `python -m pytest mlx_vlm/tests/test_processors.py::TestIdefics3Processor mlx_vlm/tests/test_utils.py -k "Idefics3Processor or stopping_criteria" -q` — 6 passed
- cached native generation with the same model revision and no `eos_tokens` override — `HAS_END_OF_UTTERANCE False`, effective stop IDs `[128001, 128258]`
- Black check passes for the two edited production files; autoflake passes

Environment: mlx-vlm base `f47a3cedbc709f2c1d25eded3cc39405c7a88742`, MLX `0.32.1.dev20260802+fb5133e10`, Transformers `5.14.1`, tokenizers `0.22.2`, Python `3.13.13`, Apple M5 Max.

The repository's combined pre-commit invocation does not currently converge under Python 3.13: Black 26.3.1 and isort rewrite existing imports in `processing_idefics3.py` and `test_processors.py` in opposite directions. Those unrelated rewrites are excluded from this PR.

PR #1936, opened 2026-08-17
Fix EOS token for Idefics3

lucasnewman:
Same approach as https://github.com/Blaizzy/mlx-vlm/pull/1774 but also fixes the server path.

### DIFF
--- a/mlx_vlm/models/idefics3/processing_idefics3.py
+++ b/mlx_vlm/models/idefics3/processing_idefics3.py
@@ -100,6 +100,11 @@ class Idefics3Processor(ProcessorMixin):
     image_processor_class = "AutoImageProcessor"
     tokenizer_class = "AutoTokenizer"
 
+    @property
+    def additional_eos_token_ids(self):
+        token_id = self.tokenizer.convert_tokens_to_ids(self.end_of_utterance_token)
+        return [token_id] if token_id is not None else []
+
     def __init__(
         self,
         image_processor,
--- a/mlx_vlm/server/generation.py
+++ b/mlx_vlm/server/generation.py
@@ -1156,6 +1156,7 @@ class ResponseGenerator:
                 stop_tokens.update(config.eos_token_id)
             elif config.eos_token_id is not None:
                 stop_tokens.add(config.eos_token_id)
+        stop_tokens.update(getattr(processor, "additional_eos_token_ids", ()))
 
         draft_model = None
         draft_kind = self.draft_kind_override or os.environ.get("MLX_VLM_DRAFT_KIND")
--- a/mlx_vlm/tests/test_processors.py
+++ b/mlx_vlm/tests/test_processors.py
@@ -1571,6 +1571,14 @@ class TestIdefics3Processor(_ProcessorTestBase, unittest.TestCase):
         self.assertIn("<row_1_col_1>", result)
         self.assertIn("<row_2_col_2>", result)
 
+    def test_end_of_utterance_is_an_additional_eos_token(self):
+        processor = self._make_processor()
+        processor.tokenizer.convert_tokens_to_ids = lambda token: (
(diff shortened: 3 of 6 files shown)


lucasnewman:
Resolves https://github.com/Blaizzy/mlx-vlm/issues/1932

jrp2014:
https://github.com/Blaizzy/mlx-vlm/issues/1932 is a fuller account, also mentioning some additional models in a follow-up comment there.

Lazarus-931 (approved):
lgtm!

jrp2014:
**Follow-up on #1936: `additional_eos_token_ids` fixes Idefics3 (verified) — two related leaks remain, with different mechanisms**

Verified today at `625f71fa` on `mlx-community/Idefics3-8B-Llama3-bf16`: `<end_of_utterance>` no longer leaks and `finish_reason` is `stop`. 

Thanks — that's exactly the fix. While confirming it I re-checked the other models that leak control tokens into final text on the direct `generate()` path. Two remain, and they're worth distinguishing because neither is the Idefics3 mechanism.

**1. `mlx-community/Kimi-VL-A3B-Thinking-2506-bf16` (rev `fb254434d4…`) — leaks `<|im_assistant|>` because `generation_config.json` overrides a correct `config.json`**

The answer ends `…under a clear blue sky.<|im_assistant|>` (then stops one token later on `[EOS]`, so `finish_reason` is `stop`). The checkpoint's files disagree about EOS:

- `config.json`: `eos_token_id: [163584, 163586]` = `<|im_end|>`, `<|im_assistant|>` — and the chat template ends assistant turns with `<|im_assistant|>`, so this list is the correct one.
- `text_config.eos_token_id`: `163585`
- `generation_config.json`: `eos_token_id: [163585]`

`load_config()` merges `generation_config.json` over `config.json`, so the live `model.config.eos_token_id` is `[163585]` and `StoppingCriteria` gets only that — 163586 never reaches the stop set, and the turn terminator leaks. `additional_eos_token_ids` from #1936 doesn't help here because it's per-processor and only Idefics3 defines it. Two possible fixes: (a) when `config.json` and `generation_config.json` both specify `eos_token_id`, take the **union** rather than letting the latter win — for stop tokens, more is strictly safer; or (b) give the Kimi-VL processor an `additional_eos_token_ids` returning `<|im_assistant|>`/`<|im_end|>`, like Idefics3. (a) fixes the class; (b) fixes the instance.

**2. Empty thinking / answer wrappers that the server strips but the direct path returns**

- `mlx-community/diffusiongemma-26B-A4B-it-8bit` (rev `7b95e38870…`) and `…-mxfp8` (rev `ded389e478…`): every answer is prefixed `<|channel>thought\n<channel|>…`. mlx-vlm already knows this exact pair — `server/responses_state.py:44` lists `("<|channel>thought", "<channel|>")` and `server/app.py:262-263` special-cases it — so the server hides it, but direct `generate()` returns it verbatim. Zero references to these markers under `mlx_vlm/generate/`, `utils.py`, or the `diffusion_gemma` module.
- `mlx-community/GLM-4.6V-nvfp4` (rev `2da6855d4e…`, `glm4v_moe`): the whole reply is wrapped `<|begin_of_box|>…<|end_of_box|>`. Control: `GLM-4.6V-Flash-mxfp4` (`glm4v`, same eos, same rendered prompt) does *not* leak these, so it's specific to the MoE variant's checkpoint — weakest of the three, listed for completeness.

These two are a consistency gap between the server and direct entry points rather than a stop-token bug: a direct-API caller has no `responses_state` splitter and gets a worse result than the server for the same model. Applying the same marker handling in the direct path (or documenting that direct callers must strip them) would close it.

Repro for any of them: `apply_chat_template` + `generate()` with `--max-tokens 200`, any image.

lucasnewman:
Yep, thanks, we generally target one PR per fix so we can track the others separately (I didn't close the issue).

PR #1938, opened 2026-08-17
Fix Qwen speculative decoding with quantized batch cache

Ptico:
## Summary

This fixes two Qwen3.5-family speculative-decoding crashes when exact APC
reuse, native MTP, and uniform quantized KV are enabled together:

- Treat the quantized key/value tuple trees returned by
  `BatchQuantizedKVCache` correctly during target verification.
- Give `BatchQuantizedKVCache` the bounded trim contract needed to roll back
  rejected MTP drafts while keeping its active length and per-row offsets in
  sync.

There is no CLI or API change. This is related to the quantized-cache coverage
tracked in #1559, but it does not close that broader issue.

## Failure mechanics

On an exact warm-cache hit, Qwen target verification receives quantized keys
and values as `(packed, scales, biases)` tuples. The current path reads
`keys.shape` and slices both values as plain arrays, causing:

```text
```

Using the existing `kv_sequence_length()` helper and `tree_map()` makes the
same target-verification loop work for array and quantized tuple storage.

After that first failure is fixed, a partially accepted MTP draft exposes a
second problem. `BatchQuantizedKVCache.is_trimmable()` returns `False`, so
Qwen's rollback code classifies the full-attention KV cache as recurrent GDN
state and eventually fails with:

```text
TypeError: 'BatchQuantizedKVCache' object does not support item assignment
```

The cache can be rolled back without reallocating its packed storage: clamp
the trim to `_idx`, decrement `_idx`, and decrement every row's logical
`offset`, matching the contract of the sibling batch cache.

This mirrors the merged TurboQuant rollback fix in #1453. Both batch
quantized caches were introduced in #1030 with non-trimmable stubs, but only
`BatchTurboQuantKVCache` was subsequently updated; the uniform
`BatchQuantizedKVCache` implementation retained the original unsupported
stub. Neither #1030's test plan nor its discussion covered trim or speculative
rollback.

## Live Qwen3.8 validation

The failures and fixes were exercised with deterministic synthetic prompts on:

- mlx-vlm `main` at `edf7b77f7e3e2277a50e5e47ebd29aac2a064483`
- MLX 0.32.0, mlx-lm 0.31.3, Python 3.12.13
- `mlx-community/Qwen3.8-27B-4bit@3e6447f082e89cc7f0bc6e5441afd38dfce760ff`
- `mlx-community/Qwen3.8-27B-MTP-4bit@b643c01b6d3b094e325edb6ebd832e16c486c575`
- exact APC, native MTP block size 3, uniform Q8 KV from token zero, one
  sequence, and a 16,384-token KV limit

Current upstream base `625f71fae24f0d5c5ee7f1ec747094e815393405`
still contains both failure implementations; the new regression tests fail on
that unpatched base and pass with this commit.

| Arm | Warm cached path | Result |
| --- | ---: | --- |
| Stock `main` | 8,872 matched tokens | HTTP 500: quantized tuple has no `shape` |
| Tuple-aware target verification only | 3,414 cached tokens | HTTP 500 during rejected-draft rollback |
| Both changes | five warm turns reusing 3,414–4,542 tokens | 6/6 HTTP 200 |

The passing arm deliberately varied each continuation so MTP rejected draft
tokens and rollback ran repeatedly:

| Turn | Prompt tokens | Cached | Drafted | Accepted | mlx-vlm `peak_memory` | Elapsed |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 3,414 | 0 | 266 | 123 | 18.035 | 38.902 s |
| 2 | 3,696 | 3,414 | 260 | 125 | 18.978 | 25.024 s |
| 3 | 3,980 | 3,696 | 210 | 150 | 19.216 | 22.967 s |
| 4 | 4,263 | 3,980 | 226 | 142 | 19.216 | 29.366 s |
| 5 | 4,542 | 4,263 | 242 | 134 | 19.300 | 35.280 s |
| 6 | 4,824 | 4,542 | 96 | 50 | 19.357 | 16.136 s |

`peak_memory` is mlx-vlm's reported MLX allocator metric, not whole-machine
resident memory.

## Regression coverage

- Direct bounded-trim and per-row-offset behavior for
  `BatchQuantizedKVCache`.
- Qwen target verification with uniform quantized batch KV, compared with
  sequential singleton decoding.
- Qwen speculative rollback recognizing and trimming uniform quantized batch
  KV rather than treating it as GDN state.

## Tests

- `pre-commit run --files` for all four changed files: passed.
- Repository suite excluding the command-line smoke harness and one confirmed
  upstream baseline assertion: `2314 passed, 6 skipped, 1 deselected`.
- The deselected exact-bit quantized-linear assertion fails identically on an
  untouched worktree at base `625f71f` in this MLX environment; it is
  independent of the cache and target-verification changes here.


Lazarus-931:
hi @Ptico, thx for the pr!

quick q: 

wouldn't turboquant hits this same `tree_map` line and raises `'_QuantizedStateProxy' object is not subscriptable`  reachable, could we also fold in a fix for that?

also small nit for tests, could you add a ragged multi-row version? the new one is single-row only.

yujunkui-web:
## Reproduction confirmed on v0.6.15 + an additional trigger: concurrent batched requests

Confirming that the bug from #1947 is **still present in v0.6.15** (released 2026-08-18, which does not include this PR) — and I found an additional reproduction path that may be useful as a test case for this fix.

### New data point: crash triggers under concurrency even with *short* prompts

#1947 reproduced with a single long-context request (~13K+ prompt tokens). On v0.6.15, single requests work fine (short *and* long prompts), but **4 concurrent `/v1/chat/completions` requests with ~20-token prompts crash the batch immediately** — every request returns HTTP 500. So on the latest release the remaining failure mode appears to be specifically the **batched (B>1) MTP verify path with a quantized KV cache**, which matches exactly what this PR targets.

### Server command

```bash
python -m mlx_vlm.server \
  --model mlx-community/Qwen3.8-27B-8bit \
  --draft-model mlx-community/Qwen3.8-27B-MTP-8bit --draft-kind mtp \
  --kv-bits 8 \
  --host 127.0.0.1 --port 8080
```

### Reproduction (4 concurrent requests)

```python
import json, urllib.request
from concurrent.futures import ThreadPoolExecutor

def chat(i):
    payload = {"model": "mlx-community/Qwen3.8-27B-8bit",
               "messages": [{"role": "user", "content": f"Count from {i} to {i+10}."}],
               "max_tokens": 100}
    req = urllib.request.Request("http://127.0.0.1:8080/v1/chat/completions",
            data=json.dumps(payload).encode(),
            headers={"Content-Type": "application/json"})
    return json.load(urllib.request.urlopen(req))

with ThreadPoolExecutor(max_workers=4) as ex:
    list(ex.map(chat, range(4)))  # all 4 raise HTTPError 500
```

### Traceback (identical to #1947, via the continuous-batching path)

```
File ".../mlx_vlm/server/generation.py", line 2388, in _step
  prompt_responses, responses = batch_gen.next(**kwargs)
File ".../mlx_vlm/generate/ar.py", line 1540, in next
  tok_list, round_meta = next(self._rounds_iter)
File ".../mlx_vlm/speculative/utils.py", line 181, in run_speculative_server_rounds
  yield from _mtp_rounds_batch(
File ".../mlx_vlm/speculative/mtp.py", line 919, in _mtp_rounds_batch
  verify = _mtp_verify_target(
File ".../mlx_vlm/speculative/mtp.py", line 159, in _mtp_verify_target
  result = _mtp_verify_without_logits(lm, verify_input, prompt_cache)
File ".../mlx_vlm/speculative/mtp.py", line 64, in _mtp_verify_without_logits
  result = verify_hidden(verify_input, prompt_cache)
File ".../mlx_vlm/models/qwen3_5/language.py", line 2871, in speculative_verify_hidden
  out = self(...)
File ".../mlx_vlm/models/qwen3_5/language.py", line 1641, in __call__
  prefix_len = keys.shape[-2] - L
```

### Environment

- mlx-vlm **0.6.15** (pip), mlx 0.32.1, mlx-lm 0.31.3
- macOS 26.5.2, Apple M5 Max 128 GB, Python 3.12.13
- Model: `mlx-community/Qwen3.8-27B-8bit` (qwen3_5 arch)
- Drafter: `mlx-community/Qwen3.8-27B-MTP-8bit`

### Workaround

Dropping `--kv-bits 8` (keeping MTP) restores full stability under concurrency.

Happy to test this PR's branch on my setup if that helps — just let me know.

Ptico:
Hi @Lazarus-931, sorry for taking so long, both requests should be addressed 

  - TurboQuant target verification now uses representation-aware KV length/slicing helpers. _QuantizedStateProxy remains quantized while exposing bounded token slicing.
  - The regression test is now ragged B=2 and parametrized across uniform Q8 and TurboQuant 3.5-bit caches.
  - The fix was ported to the new Qwen3_5ExactSpeculativeVerifier location after syncing with current main.

Validation:

  - 218 passed across the speculative and batch-quantized test modules.
  - Formatting checks passed.
  - Uniform Q8 concurrent B=4 reproduction: 4/4 HTTP 200.
  - Uniform Q8 warm exact-APC rollback: 6/6 HTTP 200, with rejected drafts every turn.
  - TurboQuant B=4 reproduction: 4/4 HTTP 200.
  - TurboQuant ragged B=4 with 28/73/241/553-token prompts: 4/4 HTTP 200, with rejected drafts on every row.

One note: I have pretty humble machine and can't run 8 bit. The live runs used the pinned Qwen3.8 27B 4-bit target and MTP sidecar. @yujunkui-web, an additional run with your exact 8-bit checkpoints would be welcome, but the failing batch KV verification path is now covered both by unit tests and live concurrent runs

PR #1939, opened 2026-08-17
Reformat trainer for all modalities

Goekdeniz-Guelmez:
The whole goal is to revamp the trainer backend to support all (current/future) modalities and algorithms. This has to be scalable, efficient and future ready. Once this is merged, the next PR's will slowly integrate and implement the actual source codes. This is a big table to turn but not impossible.


mlx_vlm/trainer/stt/cross-entorpy/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""Supervised Fine-Tuning (SFT) for Vision-Language Models.
```
lucasnewman, line 1: fyi there's a typo in the file name: "entorpy"

Goekdeniz-Guelmez: ahh thanks :D!

PR #1937, opened 2026-08-17
fix(deps): declare sentencepiece explicitly (fixes ernie4_5_moe_vl import / red CI)

Anai-Guo:
## Problem

`main` CI is red on every PR opened after ~17:50 UTC today, and a fresh `pip install mlx-vlm` can no longer load ERNIE 4.5 VL:

```
mlx_vlm/models/ernie4_5_moe_vl/processing_ernie4_5_moe_vl.py:10: in <module>
    import sentencepiece as spm
E   ModuleNotFoundError: No module named 'sentencepiece'
```

8 tests fail (`test_models.py::TestModels::test_ernie4_5_moe_vl`, `TestChunkedPrefillRoPE::test_ernie_chunked_prefill_rope`, and all 5 `TestErnie4_5VLProcessor` / `TestErnie4_5VLPatch` cases).

## Root cause

`processing_ernie4_5_moe_vl.py` imports `sentencepiece` at module scope (it backs `Ernie4_5_VLTokenizer.sp_model`), but `requirements.txt` never declared it. Until today it happened to be installed transitively:

* **mlx-audio 0.4.8** (2026-08-10) → `mlx-lm>=0.31.1` as a **core** dependency → `sentencepiece`
* **mlx-audio 0.5.0** (2026-08-17 17:47 UTC) → `mlx-lm` moved to the `sts` / `llm` / `all` / `parity` **extras**, so it is no longer installed by default

The CI logs show the transition exactly. Run at 16:29 UTC (passing):

```
Collecting sentencepiece (from mlx-lm>=0.31.1->mlx-audio>=0.4.3->mlx-vlm==0.6.14)
Successfully installed ... mlx-audio-0.4.8 mlx-lm-0.31.3 ... sentencepiece-0.2.2 ...
```

Run at 19:02 UTC (failing), same workflow, same `pip install -e .`:

```
Successfully installed ... mlx-audio-0.5.0 ...      # no mlx-lm, no sentencepiece
```

Nothing in the repo imports `mlx_lm`, so `sentencepiece` is the only casualty.

## Fix

Declare the dependency the code actually uses — one line in `requirements.txt`. `sentencepiece>=0.2.0` matches the floor mlx-audio itself uses in its extras.

No source changes, no behaviour changes; it just makes the existing import satisfiable on a clean install again.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1941, opened 2026-08-17
Remove surplus Qwen3-VL image placeholders

### LINKED ISSUE #1940: Qwen3-VL crashes when a prompt contains more image placeholders than images
## Description

`Qwen3VLProcessor` assumes that every image placeholder in the rendered prompt has a corresponding entry in `image_grid_thw`. It expands the placeholders from left to right and increments an index into that array without checking whether another image grid is available.

When a prompt contains stale image markers from an earlier turn but the current processor call supplies fewer images, the processor attempts to read beyond the end of `image_grid_thw` and raises `IndexError`. Wrapped placeholders also include vision start and end tokens, so removing only the image token would leave invalid vision delimiters in the prompt.

## Reproduction

Run the following script from the repository root. It uses lightweight processor stubs and does not load model weights.

```python
import numpy as np

from mlx_vlm.models.qwen3_vl.processing_qwen3_vl import Qwen3VLProcessor


class ImageProcessor:
    merge_size = 2
    model_input_names = ["pixel_values"]

    def __call__(self, images=None, **kwargs):
        return {
            "pixel_values": np.zeros((1, 3, 2, 2), dtype=np.float32),
            "image_grid_thw": np.array([[1, 2, 2]], dtype=np.int64),
        }


class Tokenizer:
    model_input_names = ["input_ids", "attention_mask"]

    def __call__(self, text, **kwargs):
        print("processed text:", text)
        return {
            "input_ids": [[0] for _ in text],
            "attention_mask": [[1] for _ in text],
        }


processor = Qwen3VLProcessor.__new__(Qwen3VLProcessor)
processor.image_token = "<|image_pad|>"
processor.video_token = "<|video_pad|>"
processor.vision_start_token = "<|vision_start|>"
processor.vision_end_token = "<|vision_end|>"
processor.image_processor = ImageProcessor()
processor.tokenizer = Tokenizer()

prompt = (
    "old <|vision_start|><|image_pad|><|vision_end|> "
    "new <|vision_start|><|image_pad|><|vision_end|>"
)
processor(text=[prompt], images=[object()])
```

## Actual behavior

The prompt contains two image placeholders while `image_grid_thw` contains one row. Expanding the second marker indexes past the available image grid:

```text
IndexError: index 1 is out of bounds for axis 0 with size 1
```

## Expected behavior

The processor should remove the earliest surplus image marker before expanding the markers that correspond to supplied images. When the marker is wrapped by adjacent vision start and end tokens, the complete wrapper should be removed so no orphaned vision delimiters remain.

The reproduction should finish successfully and print a prompt containing only the newer wrapped image marker.

byoan:
Fixes #1940

## Summary

`Qwen3VLProcessor` previously expanded every image placeholder in a prompt by indexing the corresponding row of `image_grid_thw`. A prompt with stale image markers and fewer supplied images therefore attempted to read beyond the available image grids and raised `IndexError`.

This change counts the supplied image grids and the image markers present in the prompt before expansion. Surplus markers are removed from left to right so that stale earlier markers are discarded while the most recent markers remain associated with the supplied images.

When the earliest surplus marker is directly wrapped by vision start and end tokens, the complete wrapper is removed. Bare image markers are removed without affecting later wrapped markers. The expansion loop also stops before indexing beyond the available image grids.

Regression tests cover:

- removing image markers in prompt order when bare and wrapped forms are mixed;
- removing a stale wrapped marker without leaving orphaned vision delimiters;
- removing multiple stale markers while retaining and expanding the latest marker;
- preserving valid multi-image prompts and applying distinct image grids in order;
- carrying surplus-marker removal across multiple text entries;
- expanding retained markers to the token count derived from their supplied image grids.

## Test plan

```text
python -m pytest mlx_vlm/tests/test_processors.py -q
```

```text
137 passed, 2 skipped, 2 warnings, 7 subtests passed
```


byoan:
Tests failures are related to Ernie 4.5 VL processor due to missing `sentencepiece` (addressed by #1937), the tests of this PR do pass

lucasnewman:
@byoan How does this work for batched inputs? It looks like it would drop the first found image token in a batch where the first row has a valid token and the second has a surplus token and a valid token and two images are passed, which wouldn't be correct for the first row.

byoan:
@lucasnewman You're right, I wasn't doing batches when I encountered this so I did not think of it, my bad

I've updated the implementation so placeholder cleanup never crosses row boundaries:

- For a single prompt, all supplied images belong to that prompt
- Explicit nested image groups provide exact per-row ownership, including rows with zero or multiple images
- Flat batches with matching marker and image totals retain the existing sequential mapping
- Flat mismatches with one image per text row use row-wise ownership
- Other flat mismatches are ambiguous and now raise a clear `ValueError` requesting nested image groups instead of guessing

In your example, the two flat images map one per row. The first row’s valid marker remains intact, while only the earliest surplus marker in the second row is removed.

I added your exact example as a regression test, along with coverage for variable grouped counts, rows without current images, and ambiguous flat batches

Let me know what you think of it, thanks!

lucasnewman (approved):
Thanks!

PR #1948, opened 2026-08-18
fix(qwen3_omni): convert audio_feature_lengths to mel frames (#1620)

### LINKED ISSUE #1620: Qwen3-Omni: audio attention mask sized in raw samples, not mel frames — 78 GB Metal allocation for a 15 s clip
## Summary

Passing any audio clip to a `qwen3_omni_moe` model makes the processor splice an
absurd number of `<|audio_placeholder|>` tokens into the prompt, which then blows
up the multimodal position tensors into a fixed **78,130,000,080-byte** Metal
allocation — an instant OOM on any Mac:

```
RuntimeError: [metal::malloc] Attempting to allocate 78130000080 bytes which is
greater than the maximum allowed buffer size of 41747087360 bytes.
```

The allocation size is independent of image count / resolution — it is driven
entirely by the audio input.

## Root cause

The feature extractor returns `attention_mask` over **raw audio samples**, while
`input_features` is the mel spectrogram. For a 15.02 s mono 16 kHz wav:

```
input_features         (1, 128, 1502)   # mel frames — correct
feature_attention_mask (1, 240396)      # raw samples — wrong unit (160× too long)
```

`Qwen3OmniMoeProcessor.__call__` then computes

```python
audio_lengths = _get_feat_extract_output_lengths(feature_attention_mask.sum(-1))
```

with 240,396 instead of 1,502 → **~31k audio placeholder tokens** instead of
~190. Observed `input_ids` length for [2 images + short text + 15 s audio]:
**32,383 tokens**. The 3-D M-RoPE position tensors / attention buffers over that
sequence are what allocate 78 GB. The sample-length mask is also forwarded to the
model as `feature_attention_mask`, so the audio tower's `feature_lens` is wrong
by the same factor (its 100-frame chunking then produces a shape mismatch:
`[broadcast_shapes] (1,1,100,1502) and (1,128,100)`).

## Reproduction

```python
from mlx_vlm import load
from mlx_vlm.utils import prepare_inputs

model, processor = load("mlx-community/Qwen3-Omni-30B-A3B-Instruct-4bit")
inputs = prepare_inputs(processor, audio=["any_15s.wav"],
                        prompts="<formatted chat prompt with one audio token>")
print(inputs["input_features"].shape)          # (1, 128, ~1500)
print(inputs["feature_attention_mask"].shape)  # (1, ~240000)  ← wrong unit
print(inputs["input_ids"].shape)               # tens of thousands of tokens
# generating from this OOMs on a 64 GB machine
```

## Workaround (verified)

Rescale the mask to mel-frame length right after feature extraction — divide the
per-sample lengths by the hop ratio (`mask_len / n_mel_frames ≈ 160`) and rebuild
the mask over `input_features.shape[-1]`:

```python
ratio   = max(1, round(mask.shape[-1] / feats.shape[-1]))
lengths = np.rint(mask.sum(-1) / ratio).astype(np.int32)
# rebuild mask as zeros((batch, feats.shape[-1])) with lengths[i] leading ones
```

With this (plus the RoPE decode fix reported separately) audio+image generation
works correctly end-to-end: ~190 audio tokens for a 15 s clip, 23 GB peak on the
4-bit 30B-A3B, and near-verbatim speech transcription (verified against Gemini
on 20 real short videos, including sung lyrics and Spanish news speech).
Happy to open a PR.

## Environment

- mlx-vlm 0.6.5 / git main (2026-07)
- mlx 0.32.0, Python 3.12.4
- macOS 26.5.2 (Darwin 25.5.0), Apple M2 Max, 64 GB

Anai-Guo:
## What

`prepare_omni_inputs` derives `audio_feature_lengths` directly from the **sample-domain** `feature_attention_mask`:

```python
model_inputs["audio_feature_lengths"] = (
    model_inputs["feature_attention_mask"].sum(axis=1).astype(mx.int32)
)
```

But `feature_attention_mask` is over raw audio samples while the audio encoder counts **mel frames**. For a 15 s / 16 kHz clip the mask is `(1, 240396)` whereas `input_features` is `(1, 128, 1502)` — the length is ~160× too large.

Because `audio_feature_lengths` is then non-`None`, it **bypasses** the sample→mel conversion that `Thinker.get_audio_features` performs when the field is absent:

```python
if audio_feature_lengths is None:
    if feature_attention_mask is not None:
        ...
        if mask_len > mel_frames:
            lengths = lengths // (mask_len // mel_frames)   # never reached
```

so the oversized sample-domain length is forwarded straight to the audio tower (`feature_lens`), producing the shape mismatch / oversized position tensors reported in #1620 for the documented `examples/qwen3_omni_demo.py` path.

## Fix

Apply the same mask/mel hop-ratio conversion already used by the processor (`processing_qwen3_omni_moe.py`) and by `Thinker.get_audio_features` (`thinker.py`), so all three sites agree. With the fix, a 15 s clip yields `audio_feature_lengths == 1502` (the true mel-frame count) instead of `240396`.

The conversion is guarded (`if mask_len > mel_frames`) so already-mel-frame inputs and the padded-batch case are unchanged.

Fixes #1620 (audio-tower half; the placeholder-count OOM half is already handled by the processor conversion).

### DIFF
--- a/mlx_vlm/models/qwen3_omni_moe/omni_utils.py
+++ b/mlx_vlm/models/qwen3_omni_moe/omni_utils.py
@@ -62,8 +62,21 @@ def prepare_omni_inputs(
         "feature_attention_mask" in model_inputs
         and "audio_feature_lengths" not in model_inputs
     ):
-        model_inputs["audio_feature_lengths"] = (
-            model_inputs["feature_attention_mask"].sum(axis=1).astype(mx.int32)
-        )
+        mask = model_inputs["feature_attention_mask"]
+        lengths = mask.sum(axis=1)
+        # feature_attention_mask is sample-domain, while the audio encoder counts
+        # mel frames; convert each item's length via the mask/mel hop ratio so
+        # audio_feature_lengths matches the encoder's true frame count. Otherwise
+        # the sample-domain length (~160x too large) is forwarded to the model,
+        # which skips the mel-frame conversion in Thinker.get_audio_features and
+        # blows up the audio position/placeholder tensors (see #1620). This mirrors
+        # the same conversion already done in the processor and thinker.
+        input_features = model_inputs.get("input_features")
+        if input_features is not None:
+            mel_frames = input_features.shape[-1]
+            mask_len = mask.shape[-1]
+            if mask_len > mel_frames:
+                lengths = lengths // (mask_len // mel_frames)
+        model_inputs["audio_feature_lengths"] = lengths.astype(mx.int32)
 
     return model_inputs, text


lucasnewman:
@Anai-Guo Please fix the tests and sign your commits before we review this.

lucasnewman (changes requested):
See comment.

Anai-Guo:
Thanks @lucasnewman — both are addressed.

**Signing:** the branch has been rebuilt as a single verified-signed commit (`ece2cd7`, `verification.verified: true`).

**Tests:** the 18 failures were coming from the stale base, not from this change. This PR touches exactly one file (`mlx_vlm/models/qwen3_omni_moe/omni_utils.py`, +16 −3) and none of the failing tests exercise it — they were `test_apc_quantized.py` (quantized-cache tolerance), `test_speculative.py` (`'_RandomState' object does not support item assignment`, an mlx API change), `test_processors.py`/`test_models.py` (`ModuleNotFoundError: No module named 'sentencepiece'`) and a gemma3 `AvgPool2d` kernel-size error. The branch was based on `e2b4ea1`; main has since landed 30 commits touching `speculative/common.py`, `apc.py`, `requirements.txt` and those test files.

I've rebased onto current main (`f9c7a1f`). `omni_utils.py` is byte-identical between the old base and main, so the diff is unchanged. CI should now reflect only this change.

lucasnewman (approved):
Thanks!

PR #1949, opened 2026-08-18
Fix issues + tests with mlx 0.32.1

### LINKED ISSUE #1945: All CLI entry points abort at exit (`PyThreadState_Get` fatal error, exit 134) on current mlx `main` — `mx.clear_streams()` must now be called before interpreter shutdown
### Summary

Since mlx `8e00a2d9d` ("Make mx.compile cache erasing thread safe", ml-explore/mlx#4248) the mlx runtime no longer registers its own exit-time cleanup. Any process that has run a generation now aborts during interpreter finalization unless it calls `mx.clear_streams()` itself. The mlx maintainers have confirmed this is intended (ml-explore/mlx#4327): *"currently it is required to call `mx.clear_streams()` at the end of thread for manual cleanup, including the main thread"* — automatic cleanup runs too late, after the interpreter and device runtime have shut down.

`mlx_vlm`'s CLI entry points (`python -m mlx_vlm generate|chat|…`, `python -m mlx_vlm.generate`, `mlx_vlm.chat`) do not call it, so every invocation now prints the correct output and then dies with exit code 134. That breaks anything that checks the return code (CI, shell scripts, wrappers) even though generation succeeded.

### Reproduction

mlx `main` @ `d5841be95` (0.32.2.dev20260818, editable build), mlx-vlm `main` @ `625f71fa` (0.6.14), macOS 26.6.1, M5 Max, Python 3.13.14.

```bash
python -m mlx_vlm generate --model mlx-community/MiniCPM-V-4.6-8bit --prompt Hi --max-tokens 10; echo "exit=$?"
```

```text
Hello! How can I help you today?
Fatal Python error: PyThreadState_Get: the function must be called with the GIL held, after Python initialization and before Python finalization, but the GIL is released (the current Python thread state is NULL)
Python runtime state: finalizing (tstate=0x0000000104953a90)
exit=134
```

Same command, with the maintainer-prescribed cleanup registered before `mlx_vlm` runs:

```bash
python -c "import atexit, runpy
import mlx.core as mx
atexit.register(mx.clear_streams)
runpy.run_module('mlx_vlm', run_name='__main__', alter_sys=True)" \
  generate --model mlx-community/MiniCPM-V-4.6-8bit --prompt Hi --max-tokens 10; echo "exit=$?"
```

```text
Hello! How can I help you today?
exit=0
```

Model-independent: the crash is in stream/compile-cache teardown during static destruction, not in any model code. Bisected on the mlx side to `8e00a2d9d` (clean at its parent `d9e2b0d40`).

### Suggested fix

Register the cleanup once, centrally, so every entry point is covered — e.g. in `mlx_vlm/__main__.py` before dispatching the subcommand, or in `mlx_vlm/__init__.py`:

```python
import atexit
import mlx.core as mx

atexit.register(mx.clear_streams)
```

`atexit` runs after `main()` returns / `sys.exit()` but before module teardown, which is early enough (verified above). Optionally precede it with `mx.synchronize()` so no kernels are in flight when the streams are destroyed.

Two related points from #4327 worth handling at the same time:

- The requirement is **per thread**. Long-lived worker threads that touch MLX (`server/audio.py:154`, `server/realtime.py:55`, `apc.py:962`) should call `mx.clear_streams()` at the end of their `run` targets, since `atexit` only covers the main thread.
- The current mlx PyPI release (0.32.x) still has the old auto-cleanup, so this only bites people on mlx `main` today — but it will bite everyone at the next mlx release.

lucasnewman:
1) Adds the missing `sentencepiece` dependency
2) Fix invalid Gemma3 test usage
3) Fix RNG state restoration since the internal implementation changed
4) Use contiguous views for kv cache dequantization
5) Call `mx.clear_streams()` at thread exit (see mlx/issues/4327), resolves https://github.com/Blaizzy/mlx-vlm/issues/1945


Blaizzy (approved):
LGTM, thanks!

Lazarus-931 (approved):
lgtm!

PR #1950, opened 2026-08-18
Add num_image_tokens estimation for dynamic-resolution processors

eptan:
## Problem

There is no way to ask how many language-model tokens an image will produce without actually processing it. Anyone sizing a prompt budget, choosing a `max_pixels` cap, or deciding whether an image fits a context window has to re-derive the smart-resize + patch/merge arithmetic by hand from the vision config — and hand-rolled copies drift.

## Changes

- **`Qwen3VLImageProcessor.num_image_tokens(height, width, ...)`** — a pixel-free dry run of `_process_one`'s sizing logic: the same `_smart_resize_image` call and the same `min_pixels` / `max_pixels` / `resized_*` override precedence, so the estimate can't drift from what preprocessing actually does.
- **`utils.estimate_num_image_tokens()`** — a generic dispatcher that unwraps wrapped processors and raises `NotImplementedError` for processors without dynamic resolution.

One method covers every family that constructs `Qwen3VLImageProcessor` (qwen2_vl, qwen2_5_vl, qwen3_vl, qwen3_omni_moe and derivatives).

## Tests

Tests assert the estimate equals `image_grid_thw.prod() // merge_size**2` from real `preprocess` calls across aspect ratios and every override path.


lucasnewman (approved):
Thanks!

PR #1951, opened 2026-08-18
Normalize tokenizer padding in BatchGenerator

Blaizzy:
## Summary

- add an optional `attention_mask` to `BatchGenerator.insert()` for dense tokenizer or processor batches
- consume left- or right-padding before queueing and trim `inputs_embeds`, position IDs, and other sequence-aligned prompt tensors with the token rows
- remove the batch-global `existing_left_padding` state introduced in #1946
- let normal ragged batching derive `BatchKVCache.offset` and `ArraysCache.left_padding`

## Root cause

Processors produce rectangular batches, so shorter prompts contain padding described by an attention mask. The embedding path consumed that mask, but `BatchGenerator` previously received uniformly sized token rows and therefore treated every row as unpadded. Cache offsets and recurrent masks then included tokenizer padding as real prompt context.

Threading an `existing_left_padding` list through the generator repaired the immediate cache symptom, but kept padded tokens in prompt accounting, logits-processor context, and APC identity while coupling every cache path to preprocessing metadata.

## Fix

`_generate_batch()` now passes the dense prompt array and its attention mask to `BatchGenerator.insert()`. Insertion consumes the mask once and converts each row to the generator's canonical ragged representation before assigning work to the queue. Sequence-aligned embeddings and position metadata are sliced over the same contiguous prompt span. Masks with internal holes are rejected.

The tokenizer remains free to pad for efficient rectangular embedding computation. `PromptProcessingBatch` then performs its established left-padding over the canonical rows, naturally producing the correct cache offsets and recurrent masks without another padding channel.

## Impact

A prompt produces the same greedy continuation whether generated alone or alongside a longer neighbor. Prompt counts, logits-processor context, and APC identities exclude tokenizer padding. Existing callers that already pass ragged Python prompt rows remain unchanged; they may also supply an attention mask for dense rows.

## Validation

- `uv run --frozen --with pytest python -m pytest mlx_vlm/tests/test_generate.py mlx_vlm/tests/test_cache.py mlx_vlm/tests/test_utils.py -q`
  - 176 passed
- Black, autoflake, and `git diff --check`
- real-model parity with `LiquidAI/LFM2.5-1.2B-Thinking`, greedy generation, 24 tokens:
  - alone vs equal-length neighbor: exact match
  - alone vs medium-length neighbor: exact match
  - alone vs long neighbor: exact match


Blaizzy:
## Full-precision batch-padding comparison with `main`

I reran the same fixed-batch padding-parity test on both implementations:

- this PR: `54fc4a53ac4c087b2d028d53d2c001c27ab9b2a2`
- latest `main`: `20eec6cb5564c6a196b046d869d2081c29e3ff92`
- `main` already contains `e2b4ea1a` / #1946 (`existing_left_padding`)

### Results

“Pass” means the target row's 24-token greedy continuation exactly matched the first row of the fixed-size no-padding reference batch `[TARGET, TARGET]`. This intentionally avoids comparing batch-size 1 against batch-size 2, because Llama and Phi already show batch-size-dependent numerical differences without padding.

| Model | Native unquantized dtype | Target left padding (medium / long) | `main` equal / medium / long | PR equal / medium / long |
|---|---:|---:|---:|---:|
| Llama-3.2-1B | BF16 (1,235,814,400 elements) | 6 / 37 | ✅ / ❌ / ❌ | ✅ / ❌ / ❌ |
| Mistral-7B-v0.3 | BF16 (7,248,023,552) | 4 / 39 | ✅ / ✅ / ✅ | ✅ / ✅ / ✅ |
| Phi-3.5-mini | BF16 (3,821,079,552) | 4 / 36 | ✅ / ✅ / ✅ | ✅ / ✅ / ✅ |
| OLMo-1B | FP32 (1,176,764,416) | 8 / 40 | ✅ / ✅ / ✅ | ✅ / ✅ / ✅ |
| Qwen2.5-0.5B | BF16 (494,032,768) | 4 / 35 | ✅ / ✅ / ✅ | ✅ / ✅ / ✅ |
| OLMoE-1B-7B | BF16 (6,919,161,856) | 8 / 40 | ✅ / ❌ / ❌ | ✅ / ❌ / ❌ |

The generated continuations were identical between `main` and this PR in every case, including both failures.

Llama no-padding batch reference:

~~~text
 17 x 23 = 385. 385 is the answer. 17 x 23 = 385.
~~~

Llama with either padded neighbor:

~~~text
 17 x 23 = 385. 17 x 23 = 385. 17 x 23 =
~~~

OLMoE no-padding batch reference:

~~~text


Answer: 391

Question: What is the sum of the first 10 counting numbers?

Answer:
~~~

OLMoE with either padded neighbor:

~~~text


Answer: 391

Question: What is the sum of the digits of the number 1723?


~~~

So the PR changes the representation—consume the attention mask, remove processor padding plus aligned metadata, then let `BatchGenerator` repad—but it does not change this six-model behavior relative to `main`'s cache-padding metadata approach. The Llama and OLMoE failures are not introduced by this PR, and the PR does not eliminate them.

### Test method

- Native, unquantized checkpoint weights; no `quantization` or `quantization_config`.
- Greedy decoding, 24 generated tokens.
- Target: `Answer carefully: What is 17 multiplied by 23?`
- Fixed batch size 2 for every parity comparison.
- Reference: `[TARGET, TARGET]`.
- Comparisons: `[TARGET, equal_neighbor]`, `[TARGET, medium_neighbor]`, and `[TARGET, long_neighbor]`.
- Exact string equality on the target continuation.
- Both commits used the same Python environment and the same cached model snapshots.

OLMo compatibility note: I used native `allenai/OLMo-1B` FP32 weights, the compatible `allenai/OLMo-1B-hf` tokenizer, and a harness-only `hf_olmo -> olmo` model-type override. OLMoE's published config omits `rms_norm_eps`; the harness injected the Transformers default `1e-5`. Neither override modifies the PR or model files.

### Reproducer

Run this script once from the PR checkout and once from the `main` checkout.

~~~python
import argparse
import json
from pathlib import Path

import mlx.core as mx
from mlx.utils import tree_flatten

from mlx_vlm import batch_generate, load
from mlx_vlm import utils as mlx_vlm_utils
from mlx_vlm.utils import (
    get_model_path,
    prepare_inputs,
    should_add_special_tokens,
)


TARGET = "Answer carefully: What is 17 multiplied by 23?"
NEIGHBORS = {
    "equal": "Answer carefully: What is 12 multiplied by 31?",
    "medium": (
        "Explain briefly why multiplying a positive integer by an even integer "
        "always produces an even integer."
    ),
    "long": (
        "A student is checking arithmetic by decomposing each factor into tens "
        "and ones. Explain the distributive steps carefully, mention how every "
        "partial product is formed and combined, describe one common arithmetic "
        "mistake, and finish with a compact worked example using different numbers."
    ),
}


def parameter_dtypes(model):
    result = {}
    for _, value in tree_flatten(model.parameters()):
        dtype = str(value.dtype)
        entry = result.setdefault(dtype, {"arrays": 0, "elements": 0})
        entry["arrays"] += 1
        entry["elements"] += value.size
    return result


def continuation(model, processor, prompts, max_tokens):
    return batch_generate(
        model,
        processor,
        prompts=prompts,
        max_tokens=max_tokens,
        verbose=False,
    ).texts[0]


def prompt_layout(model, processor, prompts):
    formatted = [
        apply_chat_template(processor, model.config, prompt, num_images=0)
        for prompt in prompts
    ]
    inputs = prepare_inputs(
        processor,
        prompts=formatted,
        add_special_tokens=should_add_special_tokens(
            model.config.model_type, processor
        ),
        padding=True,
        padding_side="left",
    )
    mask = inputs["attention_mask"]
    physical_length = mask.shape[1]
    lengths = [int(value) for value in mask.sum(axis=1).tolist()]
    return {
        "physical_length": physical_length,
        "logical_lengths": lengths,
        "left_padding": [physical_length - length for length in lengths],
    }


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("model")
    parser.add_argument("--label", required=True)
    parser.add_argument("--max-tokens", type=int, default=24)
    parser.add_argument("--model-type-override")
    parser.add_argument("--processor")
    parser.add_argument("--config-override-json")
    args = parser.parse_args()

    config_overrides = (
        json.loads(args.config_override_json)
        if args.config_override_json
        else {}
    )
    if args.model_type_override:
        config_overrides["model_type"] = args.model_type_override

    model_path = get_model_path(args.model)
    if config_overrides:
        original_load_config = mlx_vlm_utils.load_config

        def load_config_with_override(path, **kwargs):
            config = original_load_config(path, **kwargs)
            config.update(config_overrides)
            return config

        mlx_vlm_utils.load_config = load_config_with_override

    if args.processor:
        processor_path = get_model_path(args.processor)
        model = mlx_vlm_utils.load_model(model_path)
        processor = mlx_vlm_utils.load_processor(
            processor_path,
            True,
            eos_token_ids=getattr(model.config, "eos_token_id", None),
        )
    else:
        model, processor = load(str(model_path))

    solo = continuation(model, processor, [TARGET], args.max_tokens)
    batch_reference = continuation(
        model, processor, [TARGET, TARGET], args.max_tokens
    )
    comparisons = {
        name: continuation(
            model, processor, [TARGET, neighbor], args.max_tokens
        )
        for name, neighbor in NEIGHBORS.items()
    }
    padding_matches = {
        name: text == batch_reference
        for name, text in comparisons.items()
    }

    config = json.loads((Path(model_path) / "config.json").read_text())
    result = {
        "label": args.label,
        "model": args.model,
        "config": {
            "model_type": config.get("model_type"),
            "torch_dtype": config.get("torch_dtype"),
            "quantization": config.get("quantization"),
            "quantization_config": config.get("quantization_config"),
        },
        "config_overrides": config_overrides,
        "parameter_dtypes": parameter_dtypes(model),
        "solo": solo,
        "batch_reference": batch_reference,
        "solo_matches_batch_reference": solo == batch_reference,
        "batch_reference_layout": prompt_layout(
            model, processor, [TARGET, TARGET]
        ),
        "comparisons": comparisons,
        "comparison_layouts": {
            name: prompt_layout(model, processor, [TARGET, neighbor])
            for name, neighbor in NEIGHBORS.items()
        },
        "padding_matches": padding_matches,
        "all_match": all(padding_matches.values()),
        "peak_memory_gb": mx.get_peak_memory() / 1e9,
    }
    print("BATCH_PADDING_RESULT=" + json.dumps(result, ensure_ascii=False))
    raise SystemExit(0 if result["all_match"] else 2)


if __name__ == "__main__":
    main()
~~~

Commands used (the arguments may also be local snapshot paths to guarantee identical cached revisions):

~~~bash
python /tmp/batch_padding_parity.py meta-llama/Llama-3.2-1B \
  --label Llama-3.2-1B

python /tmp/batch_padding_parity.py mistralai/Mistral-7B-v0.3 \
  --label Mistral-7B-v0.3

python /tmp/batch_padding_parity.py microsoft/Phi-3.5-mini-instruct \
  --label Phi-3.5-mini

python /tmp/batch_padding_parity.py allenai/OLMo-1B \
  --label OLMo-1B \
  --model-type-override olmo \
  --processor allenai/OLMo-1B-hf

python /tmp/batch_padding_parity.py Qwen/Qwen2.5-0.5B \
  --label Qwen2.5-0.5B

python /tmp/batch_padding_parity.py allenai/OLMoE-1B-7B-0924 \
  --label OLMoE-1B-7B \
  --config-override-json '{"rms_norm_eps": 1e-5}'
~~~

PR #1953, opened 2026-08-18
Let a batched request reuse a prefix

Lazarus-931:
Two things stop a batched request from reusing a prefix, so batched APC only ever hits a byte-identical repeat of a whole prompt.

`inputs_embeds` is derived from a row's tokens and media before insert and left in the row's prompt kwargs, and `_apc_extra_hash` folded that tensor's contents into the cache key. A row's key was therefore unique to its whole prompt: two prompts sharing an opening landed on different keys and no block was shared, so four rows of one batch stored 164 blocks where 42 describe them. The embeddings add nothing the key does not already have, being a function of the token ids the hash chain walks and of the media hashed beside them, so a row now carries the media-only `_apc_semantic_hash` the server has computed per request since #1638 and `_apc_extra_hash` short-circuits on it. Single stream already excluded its derived tensors. `pixel_values` never reaches prompt kwargs, so image identity comes from the row itself: its slice of `pixel_values` when that tensor is row-major, its own image otherwise. `hash_image_payload` now hashes an in-memory image's pixels rather than its `repr`, which carried an address, neither stable across processes nor unique once the allocator reuses it.

`prepare_inputs` then squares the batch off with left padding, so a shorter row's ids and embeddings begin with pad tokens while its cache, since #1946, holds only its real ones. The ids and the K/V they key describe spans a few tokens apart, and all three consequences follow: the harvest cuts blocks on the wrong boundary and stores short ones, 11 and 15 tokens beside 162 full ones; the hash chain walks padding, so only the rows that happen to be longest in a batch match anything; and a batch whose rows match prefixes of different lengths dies in `make_warm_batch_kv_cache_multi` on a shape mismatch, which four of the nine models tested do on a second identical batch. The generator pads ragged rows itself, to the same width, so rows now arrive with their own tokens and #1946's declaration is left to the callers that really do pre-pad.

Second request keeps the opening and changes every row's question, which is what a server sees when several users share a system prompt. M4 mini, median of three, prefill time and prompt tokens actually processed:

| model | rows | no cache | APC before | APC after |
|---|---|---|---|---|
| Llama-3.2-1B | 2 | 3.47s, 5294 tok | 3.86s, 5294 tok, no reuse | 0.09s, 46 tok, 41x |
| Llama-3.2-1B | 4 | 7.66s, 10592 tok | 7.76s, 10596 tok, no reuse | 0.18s, 96 tok, 43x |
| Llama-3.2-1B | 8 | 14.81s, 21185 tok | 15.25s, 21192 tok, no reuse | 0.33s, 193 tok, 44x |
| Qwen2.5-1.5B | 2 | 4.94s, 5280 tok | 5.09s, 5280 tok, no reuse | 0.08s, 32 tok, 65x |
| Qwen2.5-1.5B | 4 | 10.06s, 10564 tok | 10.38s, 10568 tok, no reuse | 0.18s, 68 tok, 57x |
| Qwen2.5-1.5B | 8 | 20.02s, 21129 tok | 20.88s, 21136 tok, no reuse | 0.34s, 137 tok, 58x |
| Llama-3.2-3B | 2 | 10.50s, 5294 tok | 10.81s, 5294 tok, no reuse | 0.21s, 46 tok, 49x |
| Llama-3.2-3B | 4 | 21.11s, 10592 tok | 21.72s, 10596 tok, no reuse | 0.46s, 96 tok, 46x |

Generated text is unchanged with the cache off on Llama-3.2-1B and 3B, Qwen2.5-0.5B and 1.5B, Phi-3.5-mini, Mistral-7B-v0.3, OLMo-1B, LFM2.5-1.2B and Qwen3.5-0.8B. Turning reuse on still changes text on some models, as it does today and as it does single stream, because reuse shortens the suffix and these models are sensitive to how prefill splits: before this change every model that hit diverged, after it three of nine do.

Hybrids gain nothing yet, for a reason that is not keying. Their exact-mode snapshot is bounded by `APC_EXACT_CACHE_ENTRIES`, two by default, while a batch of B rows stores 2B of them, so the checkpoints worth reusing are evicted before the next request arrives: one row hits, two rows at the default hit nothing, two rows with the bound raised to twenty hit again. The unit is wrong rather than the number, and a byte budget is what #1915's composite state store already uses. LFM2.5-1.2B shows it: 3.51s against 3.51s at 2 rows, 7.10s against 7.06s at 4 rows, 14.87s against 14.42s at 8 rows, no hits either way.

Needed for batched composite work following #1915.

### DIFF
--- a/mlx_vlm/apc.py
+++ b/mlx_vlm/apc.py
@@ -367,6 +367,21 @@ def _sequence_hash(token_ids: Sequence[int], extra_hash: int, block_size: int) -
     return int.from_bytes(h.digest()[:8], "little", signed=True)
 
 
+def _in_memory_image_bytes(image_ref: Any) -> Optional[bytes]:
+    """Raw pixels of an in-memory image, or None if this is not one.
+
+    An image handed over as an object rather than a path would otherwise fall
+    back to ``repr``, which carries its address: not stable across processes,
+    and not unique once the allocator reuses that address.
+    """
+    if not (hasattr(image_ref, "mode") and hasattr(image_ref, "size")):
+        return None
+    try:
+        return np.asarray(image_ref).tobytes()
+    except Exception:
+        return None
+
+
 def hash_image_payload(
     pixel_values: Optional[mx.array] = None,
     image_ref: Any = None,
@@ -396,6 +411,11 @@ def hash_image_payload(
     if isinstance(image_ref, bytes):
         return int.from_bytes(
             hashlib.sha256(image_ref).digest()[:8], "little", signed=True
+        )
+    pixels = _in_memory_image_bytes(image_ref)
+    if pixels is not None:
+        return int.from_bytes(
+            hashlib.sha256(pixels).digest()[:8], "little", signed=True
         )
     digest = hashlib.sha256(repr(image_ref).encode("utf-8")).digest()
     return int.from_bytes(digest[:8], "little", signed=True)
--- a/mlx_vlm/generate/ar.py
+++ b/mlx_vlm/generate/ar.py
@@ -717,6 +717,38 @@ def _concat_prompt_kwarg_rows(key: str, rows: List[mx.array]) -> mx.array:
     return mx.concatenate(rows, axis=0)
 
 
+def _drop_row_left_padding(
+    row_ids: List[List[int]],
+    prompt_kwargs_rows: List[dict],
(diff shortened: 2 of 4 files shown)


lucasnewman (approved):

Lazarus-931:
closing in favor of #1960

PR #1955, opened 2026-08-19
fix(server): stop skipping KV cache quantization on "qat" model paths

### LINKED ISSUE #1333: KV cache quantization silently skipped for QAT models — but QAT weights and KV cache are independent concerns
## Bug

When `--kv-bits` and `--kv-quant-scheme turboquant` are passed to the server, the server silently skips KV cache quantization if the model is quantization-aware trained (QAT), logging:

```
Model /path/to/gemma-4-31b-it-qat-mxfp4 is quantization aware, KV cache will not be quantized.
```

No warning is shown at startup. The flags appear to be accepted but have no effect.

## Why this is wrong

QAT refers to how the **model weights** were trained — the model was trained with quantization in the loop to improve accuracy at 4-bit. The **KV cache** is a runtime structure storing attention keys and values; it is entirely independent of how the weights were quantized. There is no technical reason why a QAT model should prevent KV cache quantization.

The effect of this bug: users running QAT models with `--kv-quant-scheme turboquant` (or `--kv-bits`) get no KV memory savings, defeating the purpose of passing those flags. On a 64GB machine with 256K context, this means ~45GB consumed by KV cache alone — nearly all available RAM.

## Reproduction

```bash
mlx_vlm.server \
  --model mlx-community/gemma-4-31B-it-qat-mxfp4 \
  --kv-bits 3.5 \
  --port 8080
```

Log output includes:
```
Model ... is quantization aware, KV cache will not be quantized.
KV cache quantization: bits=3.5 scheme=turboquant   ← logged but not applied
```

Memory usage is identical to running without `--kv-bits` / `--kv-quant-scheme`.

## Expected behaviour

KV cache quantization should apply regardless of whether the model weights are QAT. The two are orthogonal. If there is a genuine technical constraint preventing KV cache quantization for QAT models, the server should raise an explicit error at startup rather than silently accepting the flags and doing nothing.

## Environment

- mlx-vlm v0.6.2
- Model: `mlx-community/gemma-4-31B-it-qat-mxfp4`
- Apple M5 Pro, 64GB unified memory

Lazarus-931:
Fixes #1333

`get_quantized_kv_bits()` skipped KV cache quantization whenever `"qat"` showed up in the model path. KV cache quantization is a runtime thing and has nothing to do with how the weights were trained, so the path shouldn't be deciding this.

it was also just a case-sensitive substring match, so it was wrong in both directions:

| model path | before | after |
|---|---|---|
| `mlx-community/gemma-4-31B-it-qat-mxfp4` | skipped | quantized |
| `some-org/qatar-news-llm` | skipped | quantized |
| `/models/qat-experiments/llama-3` | skipped | quantized |
| `mlx-community/gemma-4-31B-it-QAT-mxfp4` | quantized | quantized |

and the guard was never on `get_quantized_kv_split_bits()` — `KV_KEY_BITS`/`KV_VALUE_BITS` have always quantized these same models fine, so there was no constraint being protected. same for the live `/v1/settings` path, which sets `cfg.kv_bits` and bypasses the check entirely. so it was already inconsistent 3 ways.

`model` had no other use in the function so it's dropped, and the two callers updated.

## Tests

`TestQuantizedKVBits` in `test_server.py` — parametrized over the four paths above, plus a check that uniform and split bits agree. all 7 fail on main, pass here.

```
pytest mlx_vlm/tests/test_server.py -q            -> 351 passed
pytest test_server.py test_apc_adapters.py test_cache.py -q -> 400 passed
pre-commit run --all                              -> pass
```

### DIFF
--- a/mlx_vlm/server/app.py
+++ b/mlx_vlm/server/app.py
@@ -788,9 +788,7 @@ def get_cached_model(
     vision_cache_size = cfg.vision_cache_size
     vision_cache = VisionFeatureCache(max_size=vision_cache_size)
 
-    kv_bits = (
-        cfg.kv_bits if cfg.kv_bits is not None else get_quantized_kv_bits(model_path)
-    )
+    kv_bits = cfg.kv_bits if cfg.kv_bits is not None else get_quantized_kv_bits()
     default_key_bits, default_value_bits = get_quantized_kv_split_bits()
     kv_key_bits = cfg.kv_key_bits if cfg.kv_key_bits is not None else default_key_bits
     kv_value_bits = (
--- a/mlx_vlm/server/generation.py
+++ b/mlx_vlm/server/generation.py
@@ -314,16 +314,9 @@ def get_server_thinking_end_token():
     return os.environ.get("MLX_VLM_THINKING_END_TOKEN")
 
 
-def get_quantized_kv_bits(model: str):
+def get_quantized_kv_bits():
     kv_bits = float(os.environ.get("KV_BITS", 0))
-    if kv_bits == 0:
-        return None
-    if "qat" in model:
-        logger.info(
-            "Model %s is quantization aware; KV cache will not be quantized.", model
-        )
-        return None
-    return kv_bits
+    return kv_bits or None
 
 
 def get_quantized_kv_split_bits():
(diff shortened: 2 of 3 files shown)


lucasnewman (commented):
Lgtm

lucasnewman (approved):

PR #1956, opened 2026-08-19
Fix speculative decoding against a quantized KV cache

Codcore:
Running a draft model together with `--kv-bits` doesn't work today: it crashes,
and once you fix the crash it crashes again somewhere else. Both problems are
small, but they hide each other, so here they are together.

## What happens

With `--draft-model ... --draft-kind mtp --kv-bits 8`, every request dies in the
verify step:

```
File "mlx_vlm/models/qwen3_5/language.py", line 1641, in __call__
    prefix_len = keys.shape[-2] - L
```

The causal verify path slices the cache per verify position, and it assumes
`keys` is one array. A quantized cache returns `(packed, scales, biases)`.

Interestingly the attention kernel is already fine with this —
`scaled_dot_product_attention` dispatches on `hasattr(cache, "bits")` and takes
the tuple happily. Only the slicing around it never learned about quantization.

Fix that, and the next one shows up:

```
File "mlx_vlm/models/qwen3_5/language.py", line 2302, in rollback_speculative_cache
    c[1] = intermediate_states[:, a0]
TypeError: 'BatchQuantizedKVCache' object does not support item assignment
```

This one has a root cause worth spelling out. `BatchQuantizedKVCache.trim()` is
a stub:

```python
def is_trimmable(self):
    return False

def trim(self, n):
    return 0
```

Speculative decoding can't work without trimming — after a draft block is
verified you have to drop the rejected tokens from the cache. And because
`rollback_speculative_cache` identifies cache kinds by negation ("not trimmable
and no `zero_row_tail`" ⇒ SSM state), a quantized KV cache doesn't just fail to
trim, it gets routed into the SSM branch and dies on item assignment.

## The fixes

**Slicing** — two small helpers that read a length and take a prefix from either
an array or a quantized triple. Three call sites change.

**Trimming** — `BatchQuantizedKVCache` gets the same `trim` as `BatchKVCache`.
That's safe because the two classes store data identically: `_idx`, `offset`,
writes at `[..., prev:_idx, :]`, reads at `[..., :_idx, :]`. Trimming is pure
index arithmetic — nothing past `_idx` is ever read, and the next append
overwrites it. Quantization doesn't change that: packing lives in the last axis,
and we slice the sequence axis.

## Tests

`mlx_vlm/tests/test_speculative_quantized_kv.py`, six tests. All six fail on
current `main` and pass with this change. They cover trim bookkeeping against
the unquantized cache as the reference, appending after a trim, the helpers on
both list and tuple inputs (`mx.quantize` returns a list, the cache wraps it in
a tuple — both reach this code), and a numerical check that the per-position
slices line up with a dense reference within quantization tolerance, so an
off-by-one in the slicing would be caught rather than silently wrong.

`pytest mlx_vlm/tests` is otherwise unchanged: 1980 passed, 6 skipped.

One unrelated pre-existing failure on my machine, present before this change:
`test_qwen_target_verify_quantized_linear_matches_singleton_batch_path` asserts
bit-exact equality for a bfloat16 quantized matmul, and on mlx 0.32.0 the two
paths differ by 0.0078 — one bfloat16 ULP at that magnitude. Untouched here.

## Why it's worth having

This unblocks running speculative decoding and prefix caching (`APC_ENABLED=1`)
at the same time, which is a big deal for agent-style workloads where every step
resends the whole conversation.

Measured on a 36 GB M3 Pro, Qwen3.8-27B-4bit with its MTP draft head, ten
multi-step coding tasks, all solved correctly in every configuration:

| Setup | Time |
|---|---|
| draft, KV 8-bit, no prefix cache | 823 s |
| draft, KV 8-bit, prefix cache (this PR) | **405 s** |

Prefill on a single step drops from 39.1 s to 0.42 s (90 → 2777 tok/s). Before
this change the only way to get there was turning KV quantization off, which
costs memory and is about 7% slower on its own.

Tested on macOS 26.5, M3 Pro, mlx 0.32.0, mlx-vlm at 20eec6c.

### DIFF
--- a/mlx_vlm/models/cache.py
+++ b/mlx_vlm/models/cache.py
@@ -1961,10 +1961,13 @@ class BatchQuantizedKVCache(_BaseCache):
         self._idx, self.group_size, self.bits = map(int, v)
 
     def is_trimmable(self):
-        return False
+        return True
 
     def trim(self, n):
-        return 0
+        n = min(self._idx, n)
+        self._idx -= n
+        self.offset -= n
+        return n
 
     def empty(self):
         return self.keys is None
--- a/mlx_vlm/models/qwen3_5/language.py
+++ b/mlx_vlm/models/qwen3_5/language.py
@@ -1449,6 +1449,22 @@ def _qwen3_5_ragged_decode_attention(
     return None if one_pass is None else one_pass[0]
 
 
+def _kv_seq_len(kv) -> int:
+    """Sequence length of a cache entry.
+
+    A quantized cache hands back (packed, scales, biases) rather than one
+    array, but every component still keeps the sequence on axis -2.
+    """
+    return (kv[0] if isinstance(kv, (tuple, list)) else kv).shape[-2]
+
+
+def _kv_prefix(kv, length: int):
+    """First length positions of a cache entry, quantized or not."""
+    if isinstance(kv, (tuple, list)):
+        return tuple(component[:, :, :length, :] for component in kv)
+    return kv[:, :, :length, :]
+
+
 def _target_verify_left_padded_attention(
     queries: mx.array,
     keys: mx.array,
@@ -1638,13 +1654,13 @@ class Qwen3_5Attention(nn.Module):
             output = None
 
         if output is None and target_verify and L > 1:
-            prefix_len = keys.shape[-2] - L
+            prefix_len = _kv_seq_len(keys) - L
             output = mx.concatenate(
                 [
                     scaled_dot_product_attention(
                         queries[:, :, i : i + 1, :],
-                        keys[:, :, : prefix_len + i + 1, :],
-                        values[:, :, : prefix_len + i + 1, :],
+                        _kv_prefix(keys, prefix_len + i + 1),
+                        _kv_prefix(values, prefix_len + i + 1),
                         cache=cache,
                         scale=self.scale,
                         mask=(
--- a/mlx_vlm/tests/test_speculative_quantized_kv.py
+++ b/mlx_vlm/tests/test_speculative_quantized_kv.py
@@ -1,0 +1,118 @@
+"""Speculative decoding against a quantized KV cache.
+
+Two things are required for a draft block to be verified and rolled back when
+the cache is quantized:
+
+1. ``BatchQuantizedKVCache`` must be trimmable, so rejected draft tokens can be
+   dropped after verification.
+2. The causal verify path in the attention block must slice a quantized cache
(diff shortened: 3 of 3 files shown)


Codcore:
One thing I should have put in the description: the README already promises this
combination works. [Line 880](https://github.com/Blaizzy/mlx-vlm/blob/main/README.md?plain=1#L880):

> Reduce KV cache memory during continuous batching with `--kv-bits`. Both
> uniform quantization and TurboQuant are supported. Compatible with Automatic
> Prefix Caching (`APC_ENABLED=1`).

Which is true, right up until a draft model joins in — then it's the two crashes
above. So this isn't asking for a new capability, it's closing the gap between
that sentence and what actually happens.

While I'm here, the numbers I have for what APC is worth, in case they're useful
for the docs. The README describes it qualitatively ("repeated long documents,
long chat histories") but doesn't say what it buys, and on agent-style work the
answer turned out to be larger than I expected.

Ten multi-step coding tasks — read files, edit, run tests, check the result.
Every configuration solved all ten correctly; only the time differs. Same ten
tasks across four models, prefix cache off and on:

| Model | Off | On | Saved |
|---|---|---|---|
| Qwen3.8-27B-4bit + MTP draft | 823 s | 405 s | 51% |
| Qwen3.6-35B-A3B-4bit | 180 s | 90 s | 50% |
| gemma-4-26B-A4B QAT | 189 s | 107 s | 43% |
| Laguna XS 2.1 | 268 s | 152 s | 43% |

Prefill on a single step of the first one: 39.1 s → 0.42 s (90 → 2777 tok/s),
with 1243 of 1259 prompt tokens served from cache.

What surprised me is that the share is roughly the same for a dense 27B and for
MoE models with ~3B active parameters. I'd assumed the fast-prefill models had
little to gain. They don't gain fewer *seconds* — 2.2 s per step against 39 s —
but an agent loop resends the whole conversation every step, so around 60% of
the wall clock was re-reading context either way.

The mirror image, in case it saves someone the experiment: **APC does nothing
for one-shot requests.** Five independent prompts sharing only a system prompt
went 58.5 s → 56.6 s, prefill 83.8 → 88.4 tok/s, decode unchanged at 15.8 → 15.5.
That's inside the noise. It pays off exactly where prefixes repeat, and not
otherwise.

Caveat on all of it: one machine, one run each, no repeats. Thermal state moves
these by about 8% between runs, so the 2x is solid and the 43-vs-51% split is
not. Raw data is at
https://github.com/Codcore/mlx-vlm-patches/tree/main/results if anyone wants it.

PR #1959, opened 2026-08-19
Add DFlash 2 speculative decoding

Blaizzy:
## Summary

- add native loading and execution support for `incoai/Qwen3.8-27B-DFlash2`
- implement DFlash 2 grouped dynamic causal convolutions and candidate-path selection
- add distribution-preserving rejection sampling for non-greedy decoding
- support optional in-memory 4-bit or 8-bit drafter quantization in the CLI and server
- preserve exact greedy singleton output for Qwen3.8 target verification
- add a dedicated bit-exact 4-bit, M=4 Metal verifier kernel
- document Qwen3.8 DFlash 2 usage and extend speculative-decoding coverage

## Why

The existing Qwen DFlash path did not understand the DFlash 2 checkpoint architecture, convolution weights, selector codebooks, or sampling semantics. This adds checkpoint-compatible model construction and integrates it with target verification, cache rollback, batch generation, and server paths.

The initial verifier diverged from singleton greedy generation because hidden-state capture enabled token-wise attention during prompt prefill, and batch-size-1 quantized verifier blocks bypassed the exact QMV path. Small BF16 rounding differences accumulated through the hybrid attention/GatedDeltaNet cache and eventually changed an argmax. Capture and verifier execution are now separated explicitly, and quantized verifier projections use grouped exact QMV kernels for every batch size.

The generic exact M=4 QMV preserved singleton arithmetic but assigned one output row to each SIMD group and carried four independent activation arrays. That left substantial activation and packed-weight reuse on the table. The new 4-bit M=4 kernel:

- streams the four verification rows together
- reuses each packed weight across all four tokens
- reuses each activation quartet across four output rows
- computes four output rows per SIMD group with two SIMD groups per threadgroup
- preserves the singleton kernel's 32-lane K partition and `simd_sum` reduction order
- evaluates the affine zero-point sum in BF16/FP16 before promotion, preserving bit-exact rounding

The drafter remains unquantized BF16 unless the user explicitly supplies a draft quantization option.

## Performance and parity

Measured on an M3 Ultra with MLX 0.32.0 using:

- target: `mlx-community/Qwen3.8-27B-4bit`
- drafter: `incoai/Qwen3.8-27B-DFlash2`
- drafter dtype: BF16 (81/81 parameter tensors; no draft quantization)
- greedy decoding, block size 4, 52-token technical prompt, 500 generated tokens

| Mode | Prefill tok/s | Generation tok/s | Speedup |
| --- | ---: | ---: | ---: |
| Target only | 222.91 | 31.85 | 1.00x |
| DFlash 2 + exact M4 | 224.49 | 47.07 | 1.48x |

- exact parity: 500/500 generated tokens and identical decoded text
- draft acceptance rate: 60.5%
- accepted/drafted tokens: 322/532
- mean committed tokens: 2.81 per round over 178 rounds
- peak memory: 18.73 GB target-only; 30.76 GB with the BF16 drafter

A same-process, same-prompt controlled comparison isolates the kernel improvement from model loading and thermal state:

| Exact verifier | Generation tok/s |
| --- | ---: |
| Generic exact M=4 | 39.63 |
| New exact M4 kernel | 46.27 |

That is a 16.8% end-to-end improvement over the previous bit-exact verifier. Throughput remains prompt-, acceptance-, context-, and thermal-state-dependent.

## Validation

- speculative tests: 175 passed
- Qwen3.5 model tests: 16 passed, 8 subtests passed
- dedicated M4 projection parity for BF16 and FP16, batch size 2
- fused projection parity for both M=4 and generic M=5 paths
- real target and unquantized BF16 drafter checkpoints load and generate end to end
- exact 500-token greedy token and decoded-text parity

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ MLX-VLM is a package for inference and fine-tuning of Vision Language Models (VL
     - [Thinking Budget](#thinking-budget)
   - [Speculative Decoding](#speculative-decoding)
     - [DFlash (Qwen3.5)](#dflash-qwen35)
+    - [DFlash 2 (Qwen3.8)](#dflash-2-qwen38)
     - [Gemma 4 MTP](#gemma-4-mtp)
     - [Gemma 4 EAGLE-3](#gemma-4-eagle-3)
     - [MiniMax M3 EAGLE-3](#minimax-m3-eagle-3)
@@ -181,6 +182,7 @@ Speed up generation by drafting several candidate tokens with a small "drafter" 
 | `--draft-model` | HuggingFace repo or local path for the drafter |
 | `--draft-kind` | Drafter family — `dflash` (default), `eagle3`, or `mtp` (native/assistant MTP) |
 | `--draft-block-size` | Override the drafter's configured block size |
+| `--draft-bits` | Quantize the drafter in memory to 4 or 8 bits |
 
 See [docs/usage.md](docs/usage.md) for Python API examples including batch generation.
 
@@ -248,6 +250,20 @@ result = generate(
     draft_model=draft_model,
     draft_kind=draft_kind,
 )
+```
+
+#### DFlash 2 (Qwen3.8)
+
+DFlash 2 adds dynamic causal convolutions and a candidate path selector to
+keep later block positions coherent. For the 4-bit Qwen3.8 target, a verify
+block of 5 and a 4-bit drafter are the recommended MLX settings:
+
+```sh
+mlx_vlm.generate \
+  --model mlx-community/Qwen3.8-27B-4bit \
+  --draft-model incoai/Qwen3.8-27B-DFlash2 \
+  --draft-bits 4 --draft-block-size 5 \
+  --prompt "Explain speculative decoding in 3 sentences."
 ```
 
 #### Gemma 4 MTP
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -49,6 +49,18 @@ python -m mlx_vlm.generate \
     --max-tokens 512 --temperature 0 --enable-thinking
 ```
 
+DFlash 2 is supported for Qwen3.8. Quantizing its drafter in memory and using
+block size 5 is recommended with the 4-bit target:
+
+```bash
+python -m mlx_vlm.generate \
+    --model mlx-community/Qwen3.8-27B-4bit \
+    --draft-model incoai/Qwen3.8-27B-DFlash2 \
+    --draft-bits 4 --draft-block-size 5 \
+    --prompt "Explain speculative decoding in three sentences." \
+    --max-tokens 512 --temperature 0
+```
+
 EAGLE-3 speculators are also supported and auto-detected from Speculators configs:
 
 ```bash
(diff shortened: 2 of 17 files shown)


gtonic:
I ported the DFlash 2 modules independently over the weekend and only found this draft afterwards — closing mine (#1965) in favour of it, since yours is a superset (rejection sampling, in-server drafter quantization, the dedicated M=4 verifier kernel). Dropping my measurements here in case they are useful for review; all on an M5 Pro / 48 GB, `mlx-vlm` 0.6.15, target `mlx-community/Qwen3.8-27B-4bit`, drafter `z-lab/Qwen3.8-27B-DFlash2` quantized to 4-bit offline, `--max-num-seqs 1`, temperature 0.

**Block size on a quantized target.** This one may be worth a look, because the shipped checkpoint is configured for `block_size 8`. Decode throughput against the MTP drafter as baseline, mean over three short prompts:

| `--draft-block-size` | vs MTP |
|---|---|
| 3 | +6 % |
| 4 | **+19 %** |
| 5 | **+20 %** |
| 8 (checkpoint default) | +6 % |

So the configured block size is measurably the worst choice here, and the optimum sits at 4–5 — consistent with the upstream guidance to keep `block_size <= 5` for quantized models on MLX, and, I assume, part of why you added the M=4 kernel. An isolated matmul sweep on the same machine shows the same shape: M=1 5.7 ms, M=4 6.3 ms for 4x the work, M=5 7.4 ms (dispatch cliff).

**Per-prompt numbers at `block_size 4`**, identical prompts, MTP baseline:

| prompt | MTP | DFlash 2 |
|---|---|---|
| 64 tok | 33.9 t/s | 40.8 t/s |
| 66 tok | 33.9 t/s | 38.2 t/s |
| 76 tok | 36.7 t/s | 45.6 t/s |
| 5767 tok | 33.9 t/s | 38.4 t/s |

Acceptance at block 4–5 was 3.0–3.7 accepted tokens per round.

**Correctness checks** that might save you some time: against the published checkpoint all 81/81 parameters match in name and shape once the two bare `candidate_selector.*_codebook` tensors get the `.weight` suffix; against the reference modules in `z-lab/dflash`'s `model_mlx.py` on random inputs the grouped dynamic conv is bit-identical (`max|diff| = 0`) and the selector picks identical paths; and greedy output was bit-identical to the MTP drafter on 4/4 prompts, with tool calls parsing correctly.

**One caveat for benchmarking this end to end:** with `--draft-kind dflash` the prefix cache never engages (`cached_tokens=0` on every request, no `APC_TRACE` events), because `ResponseGenerator._run` routes non-MTP drafters to `_run_speculative`, which never receives `apc_manager`. That is #1923's territory — I mention it only because it makes DFlash 2 look worse than it is on multi-turn benchmarks: a cold 5.8k prefill costs 12–17 s here against ~0.4 s saved on a 60-token answer. Routed through the batching generator instead, the same conversation reports `cached_tokens=5748/5788` and the long-prompt decode goes 38.4 → 40.9 t/s.

Blaizzy:
Closing in favour of #2014

PR #1963, opened 2026-08-19
Load text-only checkpoints that still advertise a vision tower

### LINKED ISSUE #1958: No way to load text-only checkpoints from the server (strict=False is not exposed)
There's no way to point `mlx_vlm.server` at a text-only checkpoint, even though
the loader underneath handles it fine.

Any of the TextOnly conversions floating around on the Hub — Qwen3.8-27B being
the one I use — stop at load time:

```
ValueError: Missing 393 parameters: vision_tower...
```

Which is fair enough, the weights genuinely aren't there. But `mlx_vlm.utils.load`
already takes `strict`, and passing `strict=False` loads these models and they
generate perfectly well. The vision layers stay uninitialised and never get
touched as long as nothing sends an image. The server simply doesn't expose the
argument, so there's no way to ask for it from the command line.

My workaround is 40 lines that patch `load` in two places before starting the
normal server — `mlx_vlm.utils` and `mlx_vlm.server.generation`, since the latter
imports the function directly, so patching only the former does nothing:

```python
import mlx_vlm.server.generation as gen
import mlx_vlm.utils as U

_orig = U.load

def lenient(*a, **kw):
    kw.setdefault("strict", False)
    return _orig(*a, **kw)

U.load = lenient
gen.load = lenient
```

It works, but it's a silly thing to have to carry around, and it's brittle — it
breaks the day that import moves.

The reason I care rather than just using mlx-lm: MTP speculative decoding only
exists here, and on a dense 27B it's worth about 3x. So text-only models end up
in mlx-vlm whether or not that's where you'd expect them.

I'd send a PR, but the shape of it is a design call I'd rather not make for you.
Off the top of my head: a `--strict/--no-strict` pair, or catching the missing-
parameter error and retrying leniently with a warning naming the skipped
modules. Happy to write whichever you prefer, or neither if you'd rather these
models stayed out.

One thing worth being explicit about if this ever lands: sending an image to a
model loaded this way produces garbage rather than an error, because the vision
tower is full of uninitialised weights. That probably deserves a guard of its
own rather than a line in the docs.

macOS 26.5, M3 Pro, mlx 0.32.0, mlx-vlm at 20eec6c.

Lazarus-931:
Fixes #1958. Text-only conversions of VLM families commonly strip the vision/audio weights but leave a populated `vision_config`/`audio_config` in `config.json`, so `load_model` built the weightless tower and it failed the strict weight load (`Missing N parameters: vision_tower...`). This makes `_drop_modules_without_weights` run unconditionally, so any top-level VLM module that carries parameters but has no weights in the checkpoint is dropped and the model loads — fixing `mlx_vlm.server` (and `convert`/`generate`) with no new flag. The check stays conservative — partially-weighted modules are kept so genuine multimodal checkpoints still fail loudly, and a dropped module is left as `None` so an image request raises instead of running an uninitialized tower.

### DIFF
--- a/mlx_vlm/tests/test_utils.py
+++ b/mlx_vlm/tests/test_utils.py
@@ -746,6 +746,119 @@ class TestDropModulesWithoutWeights:
         assert "vision_tower" in caplog.text
         assert model.loaded_strict is True
 
+    def test_keeps_module_declared_in_manifest_but_not_loaded(self):
+        # The manifest claims vision weights but none loaded (missing shard or a
+        # rename/sanitize bug). The tower must be kept so strict fails loudly
+        # instead of silently degrading the checkpoint to text-only (#1963).
+        model = self.FakeModel()
+        weights = {"language_model.weight": mx.zeros((2, 2))}
+        declared = {
+            "language_model.weight",
+            "vision_tower.weight",
+            "vision_tower.bias",
+        }
+
+        _drop_modules_without_weights(model, weights, declared)
+
+        assert model.vision_tower is not None
+        with pytest.raises(ValueError, match="Missing"):
+            model.load_weights(list(weights.items()), strict=True)
+
+    def test_drops_module_absent_from_manifest(self, caplog):
+        # The manifest also omits vision -> an intentional text-only conversion.
+        model = self.FakeModel()
+        weights = {"language_model.weight": mx.zeros((2, 2))}
+        declared = {"language_model.weight"}
+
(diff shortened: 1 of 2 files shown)


lucasnewman (approved):
This seems like a pretty unfortunate conversion style. I think at some point we may have to ask the model owner to convert these properly if this still doesn't resolve all the esoteric cases, as we're getting closer and closer to introducing silent bugs where a module key is mismatched and the module gets droppped but strict=True is still considered "successful".

PR #1965, opened 2026-08-20
Add DFlash 2 support to the qwen3_dflash drafter

gtonic:
DFlash 2 drafters carry two modules the current drafter does not implement: grouped dynamic causal convolutions in every decoder layer, and a candidate path selector on top of the per-position logits. Loading such a checkpoint therefore fails on unexpected weights.

This ports both modules from the reference MLX implementation in z-lab/dflash (MIT), dflash/model_mlx.py, and reads the matching fields from the drafter's dflash_config. Everything is config-driven and defaults to 0, so DFlash v1 checkpoints build and load exactly as before.

The selector runs inside draft_block and still returns a plain token array, so the round loop in speculative/dflash.py is unchanged. Because the target verifies every drafted token, the reference's distribution-preserving sampling path is not needed here: a selector mistake can only cost acceptance, never correctness.

Also reads block_size from dflash_config (DFlash 2 moved it there) and honors rope_parameters, which transformers 5.x uses instead of a top-level rope_theta.

### DIFF
--- a/mlx_vlm/speculative/drafters/qwen3_dflash/config.py
+++ b/mlx_vlm/speculative/drafters/qwen3_dflash/config.py
@@ -29,6 +29,15 @@ class DFlashConfig(BaseModelConfig):
     final_logit_softcapping: Optional[float] = None
     runtime_block_size: int | None = None
     draft_window_size: int | None = None
+    # DFlash 2 extras. All default to 0, which reproduces DFlash v1 exactly, so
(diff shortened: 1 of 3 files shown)


gtonic:
Closing in favour of #1959, which covers the same modules and goes further (rejection sampling, in-server drafter quantization, the M=4 verifier kernel). I've posted my measurements — block-size sweep on a quantized target, parameter/parity checks against the reference implementation — over there.

PR #1793, opened 2026-08-05
Fix wrapped mixed-quantization path matching

ft-ioxcs:
## Summary

Fix mixed per-module quantization for language models wrapped under the `language_model.` module path.

## Root cause

Some MLX-VLM checkpoints store per-module entries under the inner language-model paths, for example:

- `model.layers.0.attn.wkv`
- `lm_head`

When the text model is wrapped by MLX-VLM, MLX walks paths such as:

- `language_model.model.layers.0.attn.wkv`
- `language_model.lm_head`

The existing exact lookup missed those entries and fell back to the global quantization settings. For the DeepSeek V4 mixed checkpoint, that selected affine quantization for MXFP8 modules, causing MLX to request 345 nonexistent `.biases` tensors.

## Implementation

- Add one shared per-module lookup helper.
- Preserve exact path lookup as the first operation.
- If exact lookup misses, remove only the known `language_model.` wrapper prefix and retry once.
- Return a shallow copy of dictionary entries so lookup does not mutate configuration data.
- Use the helper in:
  - one-bit replacement quantization lookup;
  - the general class predicate and load-time quantization predicate;
  - DeepSeek V4 language-model quantization;
  - DeepSeek V4 MTP draft-model quantization.
- Keep global settings as the fallback when no per-module entry matches.
- Do not modify `config.json`, safetensors names, model weights, or loading strictness.

Exact prefixed entries continue to win over an unwrapped fallback entry.

## Regression coverage

The added focused tests cover:

1. Exact path matching.
2. Wrapped attention-module matching.
3. Wrapped LM-head matching.
4. Global fallback for missing per-module entries.
5. Exact-prefixed precedence.
6. Mixed MXFP8 attention/shared-expert and affine routed-expert modes, including configuration immutability.

## Verification

- Six focused regression tests: passed.
- Python bytecode compilation for all changed Python files: passed.
- `git diff --check`: passed.
- Original cached Hugging Face snapshot inspected: all 388 relevant entries are unprefixed in both `quantization` and `quantization_config`.
- DeepSeek V4 reproduction against the original unmodified snapshot: loaded and generated successfully; 13 tokens generated, 31.424 tokens/sec generation speed, 106.806 GB peak memory.
- Full pytest/pre-commit checks were not available in the existing environment, so no packages were installed.

## Complete patch

The exact committed unified diff follows:

```diff
diff --git a/mlx_vlm/models/deepseek_v4/language.py b/mlx_vlm/models/deepseek_v4/language.py
[...]

PR #1977, opened 2026-08-20
Fix DeepSeek-V4 quantization path aliases

### LINKED ISSUE #1976: DeepSeek-V4 per-module quantization settings are ignored for aliased module paths
## Description

DeepSeek-V4 checkpoints can define quantization settings for individual modules. The checkpoint configuration uses the sanitized names produced by the DeepSeek-V4 weight loader, while the paths passed to the MLX quantization predicate come from the loaded model hierarchy.

These names do not always match. Loaded paths may contain both `language_model.` and `model.` wrappers, and several modules have different names in the checkpoint configuration:

- `gate_proj` is stored as `w1`;
- `down_proj` is stored as `w2`;
- `up_proj` is stored as `w3`;
- `lm_head` is stored as `head`;
- `model.embed_tokens` is stored as `embed`;
- `model.norm` is stored as `norm`.

The current lookup only retries a path after removing `language_model.`. It therefore misses entries that also require removal of the `model.` wrapper or translation to a sanitized checkpoint name. Terminal shared-expert paths such as `shared_experts.gate_proj` are affected as well because they do not have a trailing dot.

As a result, a module can receive the default quantization format instead of its per-module format. An explicit `False` entry can also be missed, causing a module that should stay unquantized to be quantized.

## Affected model

This was observed with revision [`bd06b1c00db7` of `Vontra/DeepSeek-V4-Flash-0731-MXFP4-MLX`](https://huggingface.co/Vontra/DeepSeek-V4-Flash-0731-MXFP4-MLX/tree/bd06b1c00db7fd0bda50868364ad94799324c41d). Its `config.json` contains fine-grained quantization entries such as:

```json
{
  "embed": false,
  "head": false,
  "layers.0.ffn.gate": false,
  "layers.0.ffn.shared_experts.w1": {
    "group_size": 32,
    "bits": 8,
    "mode": "mxfp8"
  },
  "layers.0.ffn.shared_experts.w2": {
    "group_size": 32,
    "bits": 8,
    "mode": "mxfp8"
  },
  "layers.0.ffn.shared_experts.w3": {
    "group_size": 32,
    "bits": 8,
    "mode": "mxfp8"
  }
}
```

After the checkpoint weights are sanitized and loaded by mlx-vlm, the corresponding module paths are:

| Checkpoint configuration key | Loaded module path |
| --- | --- |
| `embed` | `language_model.model.embed_tokens` |
| `head` | `language_model.lm_head` |
| `layers.0.ffn.gate` | `language_model.model.layers.0.ffn.gate` |
| `layers.0.ffn.shared_experts.w1` | `language_model.model.layers.0.ffn.shared_experts.gate_proj` |
| `layers.0.ffn.shared_experts.w2` | `language_model.model.layers.0.ffn.shared_experts.down_proj` |
| `layers.0.ffn.shared_experts.w3` | `language_model.model.layers.0.ffn.shared_experts.up_proj` |

The current lookup does not consider these paths equivalent. The model repository later added duplicate MLX module-path spellings as a model-side workaround in [`e1238c244da3`](https://huggingface.co/Vontra/DeepSeek-V4-Flash-0731-MXFP4-MLX/commit/e1238c244da3e8162f3ad67229a677720c34afc3), but earlier revisions and other checkpoints using the original sanitized names still depend on the loader resolving the aliases correctly.

The lightweight reproduction below isolates the same failure without requiring the full model download.

## Reproduction

Run the following script from the repository root. It uses a lightweight model and does not load checkpoint weights.

```python
from pathlib import Path
from unittest.mock import patch

import mlx.nn as nn

from mlx_vlm.utils import load_model


class Config:
    @classmethod
    def from_dict(cls, config):
        return cls()


class Model(nn.Module):
    def __init__(self, config):
        super().__init__()
        self.config = config
        self.language_model = nn.Module()
        self.language_model.model = nn.Module()
        self.language_model.model.layers = [nn.Module()]
        self.language_model.model.layers[0].ffn = nn.Module()
        self.language_model.model.layers[0].ffn.shared_experts = nn.Module()
        self.language_model.model.layers[0].ffn.shared_experts.gate_proj = (
            nn.Linear(64, 64, bias=False)
        )
        self.language_model.lm_head = nn.Linear(64, 64, bias=False)

    def load_weights(self, weights, strict=True):
        pass


model_class = SimpleNamespace(ModelConfig=Config, Model=Model)
mxfp8 = {"group_size": 32, "bits": 8, "mode": "mxfp8"}
quantization = {
    "group_size": 32,
    "bits": 4,
    "mode": "mxfp4",
    "layers.0.ffn.shared_experts.w1": mxfp8,
    "head": False,
}

with (
    patch(
        "mlx_vlm.utils.load_config",
        return_value={
            "model_type": "deepseek_v4",
            "quantization": quantization,
        },
    ),
    patch(
        "mlx_vlm.utils.glob.glob",
        return_value=["/tmp/model/model.safetensors"],
    ),
    patch("mlx_vlm.utils._load_safetensors", return_value={}),
    patch(
        "mlx_vlm.utils.get_model_and_args",
        return_value=(model_class, "deepseek_v4"),
    ),
    patch("mlx_vlm.utils.nn.quantize") as quantize,
):
    load_model(Path("/tmp/model"), lazy=True)

predicate = quantize.call_args.kwargs["class_predicate"]
model = Model(Config())

shared_expert = predicate(
    "language_model.model.layers.0.ffn.shared_experts.gate_proj",
    model.language_model.model.layers[0].ffn.shared_experts.gate_proj,
)
head = predicate(
    "language_model.lm_head",
    model.language_model.lm_head,
)

assert shared_expert == mxfp8
assert head == {}
```

## Actual behavior

Both assertions fail because neither loaded module path matches its corresponding checkpoint configuration key. The shared expert does not receive its MXFP8 override, and the `head: false` exclusion is not applied to `lm_head`.

## Expected behavior

DeepSeek-V4 module paths should be matched against their equivalent checkpoint names after removing model wrappers and applying the model's sanitized-name aliases. Per-module quantization settings and explicit exclusions should then be honored.

byoan:
Fixes #1976

## Summary

DeepSeek-V4 per-module quantization entries use checkpoint names that do not always match the module paths produced by the loaded VLM hierarchy. The existing lookup only accounts for the `language_model.` wrapper, so overrides can be missed when a path also contains `model.` or when the checkpoint uses a sanitized module name.

This change lets models expose additional checkpoint names for their loaded module paths. The generic loader preserves its existing `language_model.` fallback and consults the optional model hook without containing model-specific path knowledge. DeepSeek-V4 implements the hook for its additional `model.` wrapper and the checkpoint aliases used by embeddings, normalization, the language-model head, and shared-expert projections.

The shared-expert mappings cover both terminal module paths and their descendants:

- `gate_proj` to `w1`;
- `down_proj` to `w2`;
- `up_proj` to `w3`.

An aliased dictionary entry is returned as the module's quantization specification. An explicit `False` entry is converted to an empty specification so MLX skips quantizing that module. All DeepSeek-V4 mappings remain in the DeepSeek-V4 model implementation; the utility code only provides the general hook.

## Relation to existing work

PR #1793 documents the same family of mixed-quantization failures caused by wrapped module paths. The `language_model.` unwrapping needed for those paths was subsequently merged through PR #1791 and is already present in the base branch.

This PR is a focused continuation of that path-matching work. It handles the remaining `model.` wrapper and DeepSeek-V4 checkpoint aliases that neither the merged fix nor PR #1793 covers. It does not include the broader DeepSeek-V4 and MTP predicate changes proposed in #1793, and keeps the additional path knowledge within the DeepSeek-V4 model.

Regression tests cover:

- delegation from the generic loader to a model-provided alias hook;
- nested `language_model.model` and standalone `model` wrappers in DeepSeek-V4;
- embedding, normalization, and language-model head aliases;
- all three shared-expert projection aliases without requiring a trailing dot;
- isolation of DeepSeek-V4-specific aliases from other model types;
- precedence of a current module-path entry over its sanitized legacy alias;
- end-to-end predicate selection for an MXFP8 shared expert and an explicitly unquantized head.

## Test plan

```text
python -m pytest mlx_vlm/tests/test_utils.py \
  mlx_vlm/tests/test_models.py::TestModels::test_deepseek_v4_language_model \
  mlx_vlm/tests/test_models.py::TestModels::test_deepseek_v4_quantization_path_aliases -q
```

```text
62 passed, 2 warnings, 9 subtests passed
```

### DIFF
--- a/mlx_vlm/utils.py
+++ b/mlx_vlm/utils.py
@@ -658,9 +658,6 @@ def _quantization_path_aliases(
             aliases.append(alias.replace("model.embed_tokens", "embed"))
             aliases.append(alias.replace("model.norm", "norm"))
             aliases.append(alias.replace("lm_head", "head"))
-            aliases.append(
-                alias.replace(".ffn.gate.e_score_correction_bias", ".ffn.gate")
-            )
 
         for alias in tuple(aliases):
             for module_name, checkpoint_name in (
--- a/mlx_vlm/utils.py
+++ b/mlx_vlm/utils.py
@@ -644,46 +644,24 @@ def _is_text_only_config(config: dict) -> bool:
 
 
 def _quantization_path_aliases(
-    path: str, model_type: Optional[str] = None
+    path: str, model: Optional[nn.Module] = None
 ) -> Tuple[str, ...]:
     """Return checkpoint quantization keys that may refer to a module path."""
     aliases = [path]
-    for prefix in ("language_model.", "model."):
-        for alias in tuple(aliases):
-            if alias.startswith(prefix):
-                aliases.append(alias[len(prefix) :])
+    if path.startswith("language_model."):
+        aliases.append(path[len("language_model.") :])
 
-    if model_type == "deepseek_v4":
-        for alias in tuple(aliases):
-            aliases.append(alias.replace("model.embed_tokens", "embed"))
-            aliases.append(alias.replace("model.norm", "norm"))
-            aliases.append(alias.replace("lm_head", "head"))
-
-        for alias in tuple(aliases):
-            for module_name, checkpoint_name in (
-                ("gate_proj", "w1"),
-                ("down_proj", "w2"),
-                ("up_proj", "w3"),
-            ):
-                module_path = f".ffn.shared_experts.{module_name}"
-                offset = alias.find(module_path)
-                if offset < 0:
-                    continue
-                end = offset + len(module_path)
-                if end == len(alias) or alias[end] == ".":
-                    aliases.append(
-                        alias[:offset]
-                        + f".ffn.shared_experts.{checkpoint_name}"
-                        + alias[end:]
-                    )
+    model_aliases = getattr(model, "quantization_path_aliases", None)
+    if callable(model_aliases):
+        aliases.extend(model_aliases(path))
 
     return tuple(dict.fromkeys(aliases))
 
 
 def _quantization_for_module_path(
-    quantization: dict, path: str, model_type: Optional[str] = None
+    quantization: dict, path: str, model: Optional[nn.Module] = None
 ) -> Optional[dict]:
-    for alias in _quantization_path_aliases(path, model_type):
+    for alias in _quantization_path_aliases(path, model):
         value = quantization.get(alias)
         if isinstance(value, dict):
             return value
@@ -946,7 +924,7 @@ python -m mlx_vlm.convert --hf-path <local_dir> --mlx-path <mlx_dir>
 
         def get_class_predicate(p, m):
             per_module_quantization = _quantization_for_module_path(
-                config["quantization"], p, config.get("model_type")
+                config["quantization"], p, model
             )
             # Skip legacy multimodal layers unless the checkpoint has quantized
             # tensors for this exact module.
@@ -964,8 +942,8 @@ python -m mlx_vlm.convert --hf-path <local_dir> --mlx-path <mlx_dir>
             )
             if module_quantization.get("bits") == 1:
                 return False
-            # Handle custom per-layer quantization, including aliases introduced
-            # by model wrappers and checkpoint-name sanitization.
+            # Handle custom per-layer quantization, including aliases supplied
+            # by the model.
             if per_module_quantization is not None:
                 return per_module_quantization
             if not hasattr(m, "to_quantized"):
--- a/mlx_vlm/utils.py
+++ b/mlx_vlm/utils.py
@@ -643,6 +643,58 @@ def _is_text_only_config(config: dict) -> bool:
     )
 
 
+def _quantization_path_aliases(
+    path: str, model_type: Optional[str] = None
+) -> Tuple[str, ...]:
+    """Return checkpoint quantization keys that may refer to a module path."""
+    aliases = [path]
+    for prefix in ("language_model.", "model."):
+        for alias in tuple(aliases):
+            if alias.startswith(prefix):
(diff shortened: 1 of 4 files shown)


mlx_vlm/utils.py
```diff
@@ -643,6 +643,55 @@ def _is_text_only_config(config: dict) -> bool:
...
+    path: str, model_type: Optional[str] = None
+) -> Tuple[str, ...]:
+    """Return checkpoint quantization keys that may refer to a module path."""
+    aliases = [path]
+    for prefix in ("language_model.", "model."):
+        for alias in tuple(aliases):
+            if alias.startswith(prefix):
+                aliases.append(alias[len(prefix) :])
+
+    if model_type == "deepseek_v4":
```
Lazarus-931: I think this isin't good design, forcing this path on utils.py

We generally put utils as a general tool call, this feature should best be explored in the model path @byoan :)

Lazarus-931 (changes requested):
see nit, this overall is a fix we do want, but we want to keep things like this to the model

mlx_vlm/utils.py
byoan: Got it, I wasn't sure due to the nature of it, but I'll move it

byoan: Addressed, utils only keeps the generic `language_model.` fallback and a model hook now, the rest has been moved to the model 👍

Lazarus-931 (approved):
lgtm!

Issue #1721, opened 2026-07-25
Case-sensitive model_type lookup in apply_chat_template silently disables multimodal input (Nemotron-3 Nano Omni)


danmackinlay:
## Summary

`apply_chat_template` decides whether a model is multimodal with a case-sensitive dict lookup:

```python
# mlx_vlm/prompt_utils.py:821
    # Use standard formatting for text-only models.
    if model_type not in MODEL_CONFIG:
```

Every key in `MODEL_CONFIG` is lowercase, but `config.json` `model_type` is not required to be. `mlx-community/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-{8bit,mxfp4}` ship `"model_type": "NemotronH_Nano_Omni_Reasoning_V3"`, so the lookup misses, the model is treated as text-only, and **no `<image>` or `<so_embedding>` token is ever emitted**. Both modalities then die at merge time:

```
ValueError: Image token count (0) does not match feature count (256).
ValueError: Sound token count (0) does not match feature count (65).
```

`MessageFormatter.__init__` already does `self.model_name = model_name.lower()` (`prompt_utils.py:241`), so the registry is lowercase-by-convention everywhere except this one membership test. `get_message_json` builds the correct message when called directly — it just never gets called.

## Repro

```
python -m mlx_vlm.generate \
  --model mlx-community/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-mxfp4 \
  --image circle.png --prompt "What shape and colour is in this image?"
```

```
ValueError: Image token count (0) does not match feature count (256).

Prompt: <|im_start|>system
<|im_end|>
<|im_start|>user
What shape and colour is in this image?<|im_end|>     <-- no <image>
<|im_start|>assistant
<think></think>
```

Same with `--audio speech.wav` → `Sound token count (0) does not match feature count (65)`.

Isolating it from the model load entirely:

```python
from mlx_vlm.prompt_utils import apply_chat_template, MODEL_CONFIG, get_message_json

mt = "NemotronH_Nano_Omni_Reasoning_V3"          # straight from config.json
mt in MODEL_CONFIG            # False
mt.lower() in MODEL_CONFIG    # True

# the formatter itself is fine — it lowercases internally:
get_message_json(mt, "Describe.", num_images=0, num_audios=1)
# {'role': 'user', 'content': [{'type': 'text', ...}, {'type': 'audio'}]}

# but apply_chat_template never reaches it:
[...]

PR #1978, opened 2026-08-20
fix(prompt_utils): match prompt-only model types case-insensitively

Anai-Guo:
## Summary

Follow-up to #1961, which made the multimodal gate in `apply_chat_template` case-insensitive:

```python
# mlx_vlm/prompt_utils.py:859
if model_type.lower() not in MODEL_CONFIG:
```

The same function has a second lookup of the same raw `model_type` further down, and it was left case-sensitive:

```python
# mlx_vlm/prompt_utils.py:992
# Some models only need the last message
if model_type in ["paligemma", "florence2", "falcon_ocr"]:
    return messages[-1]
```

Every other read of that value normalizes: `MODEL_CONFIG` membership via `.lower()` (line 859), and `MessageFormatter.__init__` via `self.model_name = model_name.lower()` (line 266), which is also what `SINGLE_IMAGE_ONLY_MODELS` is matched against. Line 992 is the one site that compares the unnormalized string against a lowercase list.

Because #1961 fixed line 859 first, a mixed-case prompt-only model now *reaches* this branch instead of bailing out into the text-only path earlier — so the inconsistency became observable rather than masked.

## Effect

With `"model_type": "PaliGemma"` and a multi-turn conversation, the last-message-only rule is skipped and the entire conversation is rendered with the generic fallback template:

```
# before
'User: First question.\nUser: First answer.\nUser: <image>Describe this image.\nAssistant:'

# after (and what lowercase "paligemma" has always produced)
'<image>Describe this image.'
```

Note the fallback also relabels the assistant turn as `User:`. The same applies to `florence2` and `falcon_ocr`.

As #1721 established, `config.json` is free to ship any casing for `model_type` — `mlx-community/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-*` ships `"NemotronH_Nano_Omni_Reasoning_V3"` — so this is not hypothetical for models registered under a lowercase key.

## Change

One line, matching #1961 exactly:

```python
if model_type.lower() in ["paligemma", "florence2", "falcon_ocr"]:
```

## Tests

Added `TestApplyChatTemplateModelTypeCasing` to `mlx_vlm/tests/test_prompt_utils.py`: three tests asserting the mixed-case rendering equals the lowercase rendering for each prompt-only model, plus one asserting the earlier turns are actually dropped.

Verified on this machine (`prompt_utils.py` is stdlib-only, so it loads without `mlx`):

- against current `main`: **4 failed, 39 passed** — the new tests fail exactly as described above
- with this patch: **43 passed**

The 39 pre-existing tests are unaffected in both runs. `black==26.3.1` and `isort` (per `.pre-commit-config.yaml`) report both files unchanged.

The commit is signature-verified.

### DIFF
--- a/mlx_vlm/prompt_utils.py
+++ b/mlx_vlm/prompt_utils.py
@@ -989,7 +989,7 @@ def apply_chat_template(
         return messages
 
     # Some models only need the last message
-    if model_type in ["paligemma", "florence2", "falcon_ocr"]:
+    if model_type.lower() in ["paligemma", "florence2", "falcon_ocr"]:
         return messages[-1]
 
     return get_chat_template(processor, messages, add_generation_prompt, **kwargs)
--- a/mlx_vlm/tests/test_prompt_utils.py
+++ b/mlx_vlm/tests/test_prompt_utils.py
@@ -849,3 +849,44 @@ class TestModelSpecificPromptContracts:
             "text",
         ]
         assert result[0]["content"][-1]["text"] == "OCR:"
+
+
+class TestApplyChatTemplateModelTypeCasing:
+    """``config.json`` is free to ship any casing for ``model_type``.
+
+    ``MODEL_CONFIG`` membership is matched case-insensitively, so a mixed-case
+    model reaches the multimodal path. The prompt-only models must keep their
+    last-message-only handling on that path too, otherwise the whole
+    conversation is rendered with the generic fallback template.
+    """
+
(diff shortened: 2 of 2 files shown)


lucasnewman:
It _is_ hypothetical because we already handled the only model that actually shipped this, though, right? The PaliGemma example is completely made up. Closing for now as I don't see this as a real problem.

PR #1979, opened 2026-08-20
Add Mage-Flow quantization support

scaryrawr:
## Summary

- add a Mage-Flow conversion CLI with affine, MXFP4, NVFP4, and MXFP8 weight quantization
- reconstruct and strictly validate quantized checkpoints during transformer, text-encoder, and VAE loading
- preserve MLX VAE tensor layout and expose component quantization in generation metadata
- support community repository aliases, uploaded converted checkpoints, and explicit variants for ambiguous local paths
- document the conversion workflow and add targeted coverage for serialization, loading, mode validation, and component selection

## Quantization policy

Mage-Flow is quality-sensitive when its Qwen conditioner or modulation/input/output projections are quantized. The converter therefore quantizes compatible diffusion-transformer block layers while retaining the conditioner, modulation projections, transformer boundaries, and VAE in BF16 by default. This avoids the severe image collapse observed with blanket quantization.

## Comparison

Same prompt and seed (`2026`), 1024×1024, four Turbo steps, guidance 1.0:

| Mode | Generation | Peak memory | PSNR vs BF16 |
|---|---:|---:|---:|
| BF16 | 4.93 s | 9.48 GB | — |
| Affine 4-bit | 6.45 s | 8.12 GB | 17.44 dB |
| MXFP4 | 6.97 s | 8.14 GB | 15.95 dB |
| MXFP8 | 7.56 s | 8.12 GB | 21.18 dB |

MXFP8 stays visually closest to BF16. Affine 4-bit gives the strongest size/quality compromise in this sample. Quantized kernels reduce storage and memory, but are slower than the highly optimized BF16 path on this workload.

## Testing

- `34 passed` in `mlx_vlm/tests/test_mage_flow.py`
- Ruff formatting/import checks

Mage-Flow Turbo BF16, affine 4-bit, MXFP4, and MXFP8 comparison


mlx_vlm/models/mage_flow/weights.py
```diff
@@ -23,9 +50,205 @@ def _load_safetensors(directory: Path) -> dict[str, mx.array]:
...
+    _validate_quantized_paths(model, weights, quantized_paths, config)
+    return config
+
+
+def _infer_quantization_mode(
+    model: nn.Module,
+    weights: dict[str, mx.array],
+    paths: list[str],
+) -> str:
+    if any(f"{path}.biases" in weights for path in paths):
```
lucasnewman: Why don't we just include a quantization config here and then we don't have to infer anything?

mlx_vlm/models/mage_flow/convert.py
```diff
@@ -0,0 +1,402 @@
...
+    MODEL_CONVERSION_DTYPES,
+    get_model_path,
+    save_weights,
+    upload_to_hub,
+)
+
+from .config import VARIANTS, MageFlowVariant, get_variant, variant_from_local_path
+from .weights import load_text_encoder, load_transformer, load_vae
+
+_MODE_DEFAULTS = {
```
Lazarus-931: I think we are duplicating some code already in the codebase, see:

1. https://github.com/Blaizzy/mlx-vlm/blob/1249c7db09921714b43e056b149f9d762eec07d3/mlx_vlm/quant_utils.py#L61

or

2. https://github.com/Blaizzy/mlx-vlm/blob/1249c7db09921714b43e056b149f9d762eec07d3/mlx_vlm/convert.py#L33

mlx_vlm/models/mage_flow/convert.py
```diff
@@ -0,0 +1,402 @@
...
+    normalized_path = str(source).lower().replace("_", "-")
+    if any(marker in normalized_path for marker in ("base", "turbo", "edit")):
+        return variant_from_local_path(source)
+    raise ValueError(
+        "Could not infer the Mage-Flow variant. Pass --variant when converting "
+        "an ambiguously named local checkpoint."
+    )
+
+
+def _quantization_parameters(
```
Lazarus-931, line 250: I think this already exist, see:

https://github.com/Blaizzy/mlx-vlm/blob/1249c7db09921714b43e056b149f9d762eec07d3/mlx_vlm/quant_utils.py#L35

maybe calling it with `quant_predicate=_transformer_quantization_predicate` is better

Lazarus-931 (commented):
just some thoughts, but @lucasnewman has last say :)

mlx_vlm/models/mage_flow/model.py
```diff
@@ -26,7 +26,22 @@ def resolve_variant(model: str | MageFlowVariant | None) -> MageFlowVariant:
...
+        if name != model:
+            return get_variant(name)
+        raise
+
+
+def _resolve_load_variant(
+    model: str,
+    model_path: str | Path | None,
+) -> MageFlowVariant:
+    if model_path is not None:
```
lucasnewman: This seems a little aggressive since `variant_from_local_path` searches the entire path for words like `edit`, `turbo`, and `base`, and unrelated directory names can select the wrong variant. It seems like we should prefer the explicit model variant.

lucasnewman (approved):
Thanks, this looks good!

PR #1982, opened 2026-08-20
Fix instances of mx.arrays being implicitly used as integers

### LINKED ISSUE #1981: `mx.array` scalars no longer implicitly convert to `int` after the nanobind 2.15 bump in `mlx`
mlx updated its bindings from nanobind 2.13 to 2.15 (ml-explore/mlx#4337, merged 2026-08-20, commit `e2c1e286e`). nanobind 2.15 intentionally tightened implicit integer conversion: a bound `int` parameter now only accepts objects implementing `__index__`. `mx.array` implements `__int__` but not `__index__`, so a 0-d integer array no longer implicitly converts where an mlx API expects a Python `int`; it used to under 2.13. Several mlx-vlm model implementations pass exactly such scalars (elements of `grid_thw` / `spatial_shapes`) into `int` parameters of `mx.repeat`, `mx.tile`, and `reshape`, and now raise exceptions at the first vision/m-rope step.

This does **not** reproduce on PyPI mlx 0.32.1 (wheels built with nanobind 2.13) ‚ but it will hit every user the moment the next mlx release ships.

### Minimal reproduction (mlx `main` >= `e2c1e286e`)

```python
import mlx.core as mx
a = mx.array([1, 2, 3])

mx.repeat(a, mx.array(2))        # TypeError: incompatible function arguments
mx.tile(a, (mx.array(2),))       # TypeError: incompatible function arguments
a.reshape(1, mx.array(3), -1)    # RuntimeError: std::bad_cast
int(mx.array(2))                 # fine ‚ hence the fix below
```

### Observed in a 42-model sweep (single image, temperature 0.0)

15 models crash at generation, three surface forms, one cause:

| Error | Models |
| --- | --- |
| `repeat(): incompatible function arguments ... Invoked with types: mlx.core.array, mlx.core.array` | Qwen3-VL-2B (Instruct bf16, Thinking bf16, Qwen/Qwen3-VL-2B), Qwen3.5-9B, Qwen3.5-35B-A3B, Qwen3.6-27B, Qwen3.8-27B, Ornith-1.0-35B, X-Reasoner-7B, GLM-4.1V-9B-Thinking, GLM-4.6V-Flash, GLM-4.6V |
| `tile(): incompatible function arguments ... Invoked with types: mlx.core.array, tuple` | Qwen2-VL-2B-Instruct |
| `RuntimeError: std::bad_cast` (varargs `reshape` path fails less gracefully) | LFM2.5-VL-450M, LFM2.5-VL-1.6B |

The other 27 models (gemma-3/3n/4, diffusiongemma, Idefics3, Kimi-VL, MiniCPM-V, Ministral/Devstral, pixtral, Molmo, Phi-3.5, ERNIE, SmolVLM, llava-next, paligemma2, FastVLM, nanoLLaVA, Apriel, Llama-3.2-Vision, Step-3.7, InternVL3, MolmoPoint) complete normally ; their code paths don't route `mx.array` scalars into `int` slots.

### Call-site inventory (`main` @ `0cdae975`)

The dominant pattern is `cu_seqlens.append(mx.repeat(seq_len, grid_thw[i, 0]))`, where `grid_thw[i, 0]` is a 0-d `mx.array`:

- `models/qwen2_vl/vision.py:273`, `models/qwen2_5_vl/vision.py:367`, `models/qwen3_vl/vision.py:402`, `models/qwen3_vl_moe/vision.py:397`, `models/qwen3_omni_moe/vision.py:382`, `models/dots_ocr/vision.py:286`, `models/paddleocr_vl/vision.py:395`
- `models/glm4v/vision.py:110,113` ‚Äî `mx.repeat(image_shapes[i, 1], lengths[i])`
- `models/qwen2_vl/vision.py:247` ‚Äî `mx.tile(stacked_pos_ids, (t, 1))` with `t` from `for t, h, w in grid_thw:`
- `models/lfm2_vl/lfm2_vl.py:148` ‚Äî `feature.reshape(1, feature_org_h, feature_org_w, -1)` with both dims from `spatial_shapes[img_idx]`
- `models/minimax_h3/references.py:198` ‚Äî `mx.repeat(frame_time, frame_grid.shape[0])` (same pattern; not in my sweep)

Note `models/qwen3_vl/vision.py:358` already normalizes with `t, h, w = int(t), int(h), int(w)` ; the fix is just applying that existing idiom at the remaining sites.

### Two ways to fix.  I'll leave it to you to decide which to take

1. **Fix here:** mechanical `int(...)` casts at the sites above (e.g. `mx.repeat(seq_len, int(grid_thw[i, 0]))`). Small, release-independent, and consistent with what `qwen3_vl/vision.py:358` already does.
2. **Report upstream to mlx:** the systemic fix is for integer-dtype 0-d `mx.array` to implement `__index__`, which would restore the previous behaviour for every downstream project at once (and is arguably correct Python for a lossless integer scalar regardless of nanobind version). If you'd rather push it that way, this issue has the minimal repro ready to forward; the two aren't mutually exclusive. The `int()` casts also make the code robust on today's mlx `main` either way.

### Environment

- mlx: `main` @ `27fec909a` (0.32.2.dev20260820, local build; first bad commit `e2c1e286e`)
- mlx-vlm: `main` @ `0cdae975` (0.6.15)
- macOS 26.6.1, Apple M5 Max (128 GB), Python 3.13.14

lucasnewman:
This will break some models in the next MLX release due to an upstream change to nanobind. Resolves https://github.com/Blaizzy/mlx-vlm/issues/1981

### DIFF
--- a/mlx_vlm/models/aya_vision/vision.py
+++ b/mlx_vlm/models/aya_vision/vision.py
@@ -165,7 +165,7 @@ class VisionEmbeddings(nn.Module):
 
         for i in range(batch_size):
             # (1, dim, height, width) -> (1, dim, target_height, target_width)
-            height, width = spatial_shapes[i]
+            height, width = spatial_shapes[i].tolist()
             # Then upsample width dimension
             resized_embeddings = resize_bilinear(
                 positional_embeddings,
--- a/mlx_vlm/models/dots_ocr/vision.py
+++ b/mlx_vlm/models/dots_ocr/vision.py
@@ -283,7 +283,7 @@ class VisionModel(nn.Module):
         cu_seqlens = []
         for i in range(grid_thw.shape[0]):
             seq_len = grid_thw[i, 1] * grid_thw[i, 2]
-            cu_seqlens.append(mx.repeat(seq_len, grid_thw[i, 0]))
+            cu_seqlens.append(mx.repeat(seq_len, int(grid_thw[i, 0])))
         cu_seqlens = mx.concatenate(cu_seqlens).astype(mx.int32)
         cu_seqlens = mx.cumsum(cu_seqlens, axis=0)
         cu_seqlens = mx.pad(cu_seqlens, (1, 0), constant_values=0)
--- a/mlx_vlm/models/glm4v/vision.py
+++ b/mlx_vlm/models/glm4v/vision.py
@@ -107,10 +107,16 @@ class Glm4vVisionEmbeddings(nn.Module):
 
             # Calculate target dimensions for each patch
             target_h = mx.concatenate(
-                [mx.repeat(image_shapes[i, 1], lengths[i]) for i in range(len(lengths))]
+                [
+                    mx.repeat(image_shapes[i, 1], int(lengths[i]))
+                    for i in range(len(lengths))
+                ]
             ).astype(mx.float32)
             target_w = mx.concatenate(
-                [mx.repeat(image_shapes[i, 2], lengths[i]) for i in range(len(lengths))]
+                [
+                    mx.repeat(image_shapes[i, 2], int(lengths[i]))
+                    for i in range(len(lengths))
+                ]
             ).astype(mx.float32)
 
             # Normalize coordinates to [-1, 1] range for grid_sample
--- a/mlx_vlm/models/glm4v_moe/vision.py
+++ b/mlx_vlm/models/glm4v_moe/vision.py
@@ -107,10 +107,16 @@ class Glm4vVisionEmbeddings(nn.Module):
 
             # Calculate target dimensions for each patch
             target_h = mx.concatenate(
-                [mx.repeat(image_shapes[i, 1], lengths[i]) for i in range(len(lengths))]
+                [
+                    mx.repeat(image_shapes[i, 1], int(lengths[i]))
+                    for i in range(len(lengths))
+                ]
             ).astype(mx.float32)
             target_w = mx.concatenate(
-                [mx.repeat(image_shapes[i, 2], lengths[i]) for i in range(len(lengths))]
+                [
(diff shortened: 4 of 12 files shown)


Lazarus-931 (approved):

jrp2014:
Thanks for the fast fix in #1982.  I re-ran the 42-model sweep at `1249c7db` and 12 of the 15 families are back. One site was missed, and it takes out all three GLM models (GLM-4.1V-9B-Thinking, GLM-4.6V-Flash, GLM-4.6V): the shared `grid_sample` metal kernel passes an `mx.array` inside the `grid` tuple.

`mlx_vlm/models/kernels.py:556`:

```python
outputs = kernel(
    inputs=[x, grid],
    ...
    grid=(mx.prod(mx.array(out_shape)), 1, 1),   # mx.array in a tuple[int, int, int] slot
    threadgroup=(256, 1, 1),
)
```

`mx.fast.metal_kernel.__call__` binds `grid: tuple[int, int, int]`, so on mlx `main` >= `e2c1e286e` this raises:

```text
TypeError: __call__(): incompatible function arguments. The following argument types are supported:
    1. __call__(self, *, inputs: list[scalar | array], ..., grid: tuple[int, int, int], ...)
```

via `models/glm4v/vision.py:132` and `models/glm4v_moe/vision.py:132` (the only callers of `grid_sample`). It survived #1982 presumably because the scalar comes from `mx.prod` rather than `grid_thw` indexing.

Since `out_shape` is already a plain Python sequence, the clean fix avoids the array round-trip entirely:

```python
grid=(math.prod(out_shape), 1, 1),
```

(`int(mx.prod(mx.array(out_shape)))` works too. I verified locally: all three GLM models generate again with that change, mlx `main` @ `27fec909a`.)

PR #1984, opened 2026-08-21
Fix ArraysCache per-token Metal buffer leak on long hybrid-model generation (#1972)

### LINKED ISSUE #1972: `[metal::malloc] Resource limit (499000) exceeded after ~10.3k generated tokens with Qwen3.8-27B`
## Description

`mlx_vlm.server` consistently crashes during long text generation with:

```text
[metal::malloc] Resource limit (499000) exceeded
```

The crash occurs at almost exactly **10,320 generated tokens** on an Apple M1 Max with 32 GB unified memory.

I found a few similar issues in other repositories for other engines:
1. [https://github.com/ml-explore/mlx-lm/issues/1641](https://github.com/ml-explore/mlx-lm/issues/1641)
2. [https://github.com/ml-explore/mlx-lm/issues/831](https://github.com/ml-explore/mlx-lm/issues/831)
3. https://github.com/jundot/omlx/issues/372

## Details
Hardware: MacBook Pro, Apple M1 Max, 32 GB RAM.
Model: [mlx-community/Qwen3.8-27B-4bit](https://huggingface.co/mlx-community/Qwen3.8-27B-4bit)
mlx-vlm version: v0.6.15

## Command

```bash
mlx_vlm.server \
  --model ~/mlx/Qwen3.8-27B-4bit_mlx-community \
  --host 127.0.0.1 \
  --max-tokens 70960 \
  --prefill-step-size 1024 \
  --kv-bits 4 \
  --max-num-seqs 1
```

## Reproduction

A request can run normally for a long time, but the server consistently dies at approximately 10.3k generated tokens.

The last progress message looks like:

```text
2026-08-20 00:00:06,218 - INFO - Decode progress: request=c6b6c16e0 generated_tokens=10320 elapsed=605.996s rate=15.9 tok/s
```

Immediately afterwards:

```text
2026-08-20 00:00:06,278 - ERROR - Chat completion stream generation failed:
[metal::malloc] Resource limit (499000) exceeded.
```

## Notes

I don't think the problem has anything to do with RAM limits or context size.
Memory was monitored using Activity Monitor, and `memory pressure` remained consistently in the green zone, with no swap activity. Moreover, there was still plenty of free memory available.

This issue was not observed on other engines, like `vllm-metal`.

Lazarus-931:
The `ArraysCache.advance` would decrease `left_padding`/`lengths` with a lazy `array -= N` each decode step, which allocates a scalar-operand buffer per call. 

Since only one of hybrid models layer cache is ever evaluated, every other layer leaks ~1 Metal buffer per token and hits the `[metal::malloc] Resource limit (499000) exceeded` reported by #1972 and other adjacent issues. 

Instead we can use a python int value to store the decrement and apply it to `left_padding`/`lengths`  when only read so `advance` builds no graph and unused caches accrue nothing with values unchanged. 

I verified cache/APC/generate/hybrid-model tests green and that unused-cache metadata stays O(1) vs the pre-fix graph growing 19×→44× over 600→1500 decode steps.

Closes #1972.

### DIFF
--- a/mlx_vlm/models/cache.py
+++ b/mlx_vlm/models/cache.py
@@ -626,8 +626,10 @@ class RotatingKVCache(_BaseCache):
 class ArraysCache(_BaseCache):
     def __new__(cls, *args, **kwargs):
         instance = super().__new__(cls)
-        instance.left_padding = None
-        instance.lengths = None
+        instance._left_padding = None
+        instance._left_padding_advance = 0
+        instance._lengths = None
+        instance._lengths_advance = 0
         return instance
 
     def __init__(self, size, left_padding: Optional[List[int]] = None):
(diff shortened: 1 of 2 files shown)


lucasnewman:
Ok, it took me some time to convince myself this was the right approach, but every alternate evaluation strategy was quite a bit slower so I think this is in fact the local optima :)

lucasnewman (approved):

anthonyalayo:
Thank you! Was banging my head on this :-)

PR #1985, opened 2026-08-21
feat: ship a PEP 561 py.typed marker so mlx-vlm's annotations are usable (#1970)

### LINKED ISSUE #1970: Add a `py.typed` marker so type checkers can use mlx-vlm's existing annotations (PEP 561)
mlx-vlm is pure Python and already carries useful type annotations on its public API, but the package ships without a PEP 561 `py.typed` marker, so mypy/pyright/ty treat every import as untyped: `from mlx_vlm.utils import load` resolves to `Any`, and downstream projects either lose all type coverage for mlx-vlm calls or need to run `stubgen` over ~1,280 modules themselves and maintain patches for its output. 

A `py.typed` marker file addresses this — no stub generation, no annotation work, no runtime effect. Verified against a clean `pip install mlx-vlm==0.6.14`: after adding the marker, mypy resolves e.g. `load` as `def (path_or_hf_repo: str, adapter_path: str | None = ..., lazy: bool = ..., revision: str | None = ..., strict: bool = ..., **kwargs) -> tuple[nn.Module, ProcessorMixin]` and `generate` with its full `GenerationResult` signature, versus `Any` before. 

Precedent within the ecosystem: mlx-lm added `py.typed` in ml-explore/mlx-lm#389, and mlx wheels ship `mlx/core/*.pyi` with a non-blocking Ruff + `ty` CI check as of ml-explore/mlx#4337.

The change:

```bash
touch mlx_vlm/py.typed
```

```toml
# pyproject.toml — wheels exclude non-.py files unless declared
[tool.setuptools.package-data]
mlx_vlm = ["py.typed"]
```

To reproduce the before/after:

```bash
python -m venv /tmp/t && /tmp/t/bin/pip install mlx-vlm mypy
cat > /tmp/probe.py <<'EOF'
reveal_type(load)  # Any without py.typed; full signature with it
EOF
/tmp/t/bin/python -m mypy --follow-imports=silent /tmp/probe.py
touch "$(/tmp/t/bin/python -c 'import mlx_vlm,os;print(os.path.dirname(mlx_vlm.__file__))')/py.typed"
/tmp/t/bin/python -m mypy --follow-imports=silent /tmp/probe.py
```

Optionally, a non-blocking `ty check mlx_vlm` (or `mypy -p mlx_vlm`) CI step in the spirit of ml-explore/mlx#4337 would keep annotations from regressing once they're load-bearing for downstream users.  Checkers will, of course, surface whatever imperfections currently exist in the inline annotations (that's the point of mlx's non-blocking-first approach).

Anai-Guo:
## What

Closes #1970.

Adds an empty `mlx_vlm/py.typed` and declares it as package data:

```toml
[tool.setuptools.package-data]
mlx_vlm = ["py.typed"]
```

mlx-vlm is pure Python and already annotates its public API, but without a PEP 561 marker every type checker treats imports from it as untyped, so downstream users lose all type coverage across the mlx-vlm boundary.

The `package-data` declaration is the load-bearing half. `pyproject.toml` currently sets no `package-data` and no `include-package-data`, so setuptools drops non-`.py` files from the built distributions — the marker would sit in the repo and never reach the wheel.

## Verified, built both ways

Built the wheel from a clean checkout at `1249c7d` before and after the change:

| | files in wheel | `mlx_vlm/py.typed` present |
|---|---|---|
| before | 1233 | ❌ |
| after | 1234 | ✅ |

The wheel differs by exactly that one file — nothing else moves. The sdist picks it up too (`mlx_vlm-0.6.15.tar.gz` → `mlx_vlm/py.typed` present).

Then installed each wheel into a clean venv (`--no-deps`) and ran mypy on a probe:

```python
reveal_type(load)
```

```
before: Revealed type is "Any"
after:  Revealed type is "def (path_or_hf_repo: str, adapter_path: str | None =,
                               lazy: bool =, revision: str | None =,
                               strict: bool =, **kwargs: Any) -> tuple[Any, Any]"
```

## Scope

No runtime effect, no annotation changes, no new dependency, no CI addition. This PR only makes the annotations mlx-vlm already ships visible to type checkers. `mlx-lm` made the same change in ml-explore/mlx-lm#389.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,9 @@ Issues = "https://github.com/Blaizzy/mlx-vlm/issues"
 include = ["mlx_vlm*"]
 exclude = ["mlx_vlm.tests*"]
 
+[tool.setuptools.package-data]
+mlx_vlm = ["py.typed"]
+
 [tool.setuptools.dynamic]
 version = {attr = "mlx_vlm.version.__version__"}
 dependencies = {file = ["requirements.txt"]}


lucasnewman:
@Anai-Guo Can you please sign your commits? Soon we're going to need to stop reviewing PRs without signed commits due to the volume of contributions, so if you can set up your environment to handle this automatically it would be very helpful.

lucasnewman (changes requested):
Needs signed commits before we review.

Anai-Guo:
@lucasnewman Done — the branch now carries a single GPG-signed commit (`verified: true`), and I've switched my workflow over so future PRs are signed from the start.

Note the force-push made GitHub auto-close the PR for a moment (the intermediate head was an ancestor of `main`); the reopen above is only to refresh the head. The diff is byte-for-byte unchanged from what you saw.

lucasnewman (approved):
Needs signed commits before we review.

PR #1986, opened 2026-08-21
Fix native quantized KV in Qwen MTP verification

YuhuaBillChen:
## Summary

Fix Qwen MTP target verification with MLX's native uniform 4-bit KV cache.

With `--kv-bits 4` and a coalesced speculative batch, Qwen receives packed
quantized KV state as tuples. The verification path assumed ordinary arrays
(`keys.shape` and direct slicing), and `BatchQuantizedKVCache` did not expose
the rollback contract used by speculative decoding. Upstream therefore failed
both requests with `AttributeError: 'tuple' object has no attribute 'shape'`.

This change:

- adds representation-independent KV length and prefix-slicing helpers;
- makes `BatchQuantizedKVCache` trimmable while preserving per-row offsets;
- uses those helpers in Qwen MTP target verification; and
- adds regression coverage using a tiny real Qwen hybrid model with batch size 2.

This is independent of #1822: the failing batch-size-2 path is already reachable
on upstream `main` without that PR.

### Real-model reproduction

Qwen3.6-27B MXFP4 target, MXFP4 MTP drafter, Q8 vision tower, native uniform
Q4 KV, MTP block 3, and two simultaneously coalesced requests:

- before: both requests returned HTTP 500 at Qwen target verification;
- after: both requests completed 3,349 prompt + 32 output tokens;
- MTP accepted 16/31 draft tokens per row (51.6%);
- peak OS footprint was 18,704 MiB with no watchdog trip.

### Tests

- Focused quantized/speculative suite: **316 passed**.
- Remaining repository suite: **2,391 passed, 5 skipped, 66 subtests passed**.

Three Nemotron voicechat test files could not be collected because the shared
environment's installed `mlx-audio` lacks the newer voicechat codec module;
this occurs before project tests execute and is unrelated to this change.


lucasnewman (approved):
@YuhuaBillChen Thanks!

PR #1989, opened 2026-08-21
Fix native quantized KV in Qwen MTP target verification

Lazarus-931:
## Summary

With `--kv-bits 4` (`BatchQuantizedKVCache`) and a coalesced speculative batch, Qwen's MTP target-verify path receives the cache's `(packed, scales, biases)` tuple but assumed a plain array (`keys.shape[-2]` + array slicing), raising `AttributeError: 'tuple' object has no attribute 'shape'` and failing the requests with HTTP 500. `BatchQuantizedKVCache` also reported `is_trimmable() == False`, so speculative rollback misclassified it as an SSM cache and never trimmed it.

- `base.py`: add `slice_kv_sequence()`, which narrows a plain array or a quantized `(packed, scales, biases)` tuple along the sequence axis (`kv_sequence_length()` already handled the tuple length).
- `qwen3_5/language.py`: the target-verify fallback uses `kv_sequence_length()` / `slice_kv_sequence()` instead of `keys.shape` / raw slicing.
- `cache.py`: make `BatchQuantizedKVCache` trimmable (mirrors `BatchKVCache`), so rollback trims it and routes ragged accepts through `prepare(right_padding)/finalize` — the rejected tail becomes masked left-padding.

This overlaps #1986; it's the same approach with the trim/verify helpers plus broader batch/size test coverage.

## Testing

Existing `test_speculative.py`: **157 passed** (no regressions), **160** with the added tests below.

Extensive stress on a tiny all-full-attention Qwen (every layer a quantized batch cache, real decode step), across batch sizes and differently-sized requests:

- **no-crash + finite**: 108/108 configs — B ∈ {1,2,3,4,6,8} × block ∈ {2,3,4} × bits ∈ {4,8} × ragged prompt lengths (incl. a long row);
- **uniform-accept batch == standalone** (bit-exact): 30/30;
- **row-isolation** (no cross-row contamination): 10/10 — a low-accept row, whose rejected tail is rolled into left-padding, is bit-identical regardless of neighbour content, including the ragged finalize path.

## Tests added

- `test_qwen3_5_mtp_verify_supports_native_quantized_batch_cache` — verify + rollback state on a native quantized batch cache;
- `test_qwen3_5_rollback_speculative_cache_handles_native_quantized_batch_kv` — trim + left-pad after ragged accepts;
- `test_qwen3_5_native_quantized_batch_no_cross_row_contamination` — end-to-end, a low-accept row is unaffected by neighbour content.


lucasnewman:
#1986 ended up adopting the identical changes so I'm merging that one.

PR #1991, opened 2026-08-21
Define generation defaults once instead of per module

Lazarus-931:
The CLI and the library each defined their own copy of the same generation defaults, so the two could silently drift apart.

`generate/ar.py` and `generate/dispatch.py` both declared `DEFAULT_MAX_TOKENS`, `DEFAULT_TEMPERATURE`, `DEFAULT_TOP_P`, `DEFAULT_TOP_K`, `DEFAULT_MIN_P`, `DEFAULT_REPETITION_CONTEXT_SIZE`, `DEFAULT_PREFILL_STEP_SIZE`, `DEFAULT_COMPLETION_BATCH_SIZE` and `DEFAULT_PREFILL_BATCH_SIZE`. `diffusion.py` held a third copy of `DEFAULT_TEMPERATURE`, and the two diffusion canvas defaults were in both `dispatch.py` and `diffusion.py`.

`dispatch.py` supplies the CLI's argparse defaults and `ar.py` the library's keyword defaults, so changing one copy leaves `mlx_vlm.generate` and `generate()` behaving differently for the same setting. `generate/common.py` already held the shared KV defaults, so the rest now live there. No value changes, and every constant is still importable from its previous location.

Five were unused in `dispatch.py` — the CLI exposes no `--top-p`, `--top-k` or `--min-p`, and it never referenced the two batch-size constants. A follow-up wires those flags up.

### Verification

`test_generation_defaults.py` compares the CLI's argparse defaults against `generate_step`'s signature by introspection rather than restating numbers. Re-declaring `DEFAULT_MAX_TOKENS = 512` in `dispatch.py` fails it.


Lazarus-931:
Closing in favour of a single PR that adds the missing `--top-p`, `--top-k` and `--min-p` flags and defines the shared defaults once.

On its own this only prevented a divergence that hadn't happened yet, and it dropped five constants from `dispatch.py` that the flags PR immediately needs back. Folding the two together avoids that churn and gives the cleanup an actual reason: the library has supported these three knobs all along, but the CLI could only reach them through `--gen-kwargs`.

PR #1993, opened 2026-08-21
Fix Qwen3-Omni decode assigning RoPE position 0 to every token

### LINKED ISSUE #1983: Qwen3-Omni: decode assigns RoPE position 0 to every token (get_rope_index ignores cache offset)
### Summary

For Qwen3-Omni, `get_rope_index` builds position ids as `mx.arange(seq_length)` with no reference to the cache offset. During decode `seq_length == 1`, so the result is `[0]` — **every token in a generated stream is assigned RoPE position 0.**

Prefill is unaffected (positions 0…T-1 are correct), so the two paths disagree for every token after the first.

### Root cause

`mlx_vlm/models/qwen3_omni_moe/language.py`, `LanguageModel.get_rope_index`:

```python
batch_size, seq_length = input_ids.shape
position_ids = mx.arange(seq_length, dtype=mx.int32)   # no cache offset
position_ids = mx.broadcast_to(position_ids[None, :], (batch_size, seq_length))
```

For the text-only path this value is returned as-is.

`Qwen3OmniMoeThinkerTextAttention` already contains correct offset-aware handling:

```python
if position_ids is None:
    position_ids = mx.arange(cache.offset, cache.offset + L)
```

but that branch is never reached, because `thinker.py` always passes explicit `position_ids` down from `get_rope_index`.

### Reproduction

Prefill and decode should produce identical hidden states for identical tokens.

```python
import mlx.core as mx, numpy as np
from mlx_vlm import load
from mlx_vlm.models.cache import make_prompt_cache

model, proc = load("mlx-community/Qwen3-Omni-30B-A3B-Instruct-4bit", lazy=False)
layers = model.thinker.language_model.model.layers
inner = model.thinker.language_model.model
buf, on = [], [False]
Cls = type(layers[0]); orig = Cls.__call__
def cap(self, *a, **kw):
    o = orig(self, *a, **kw)
    if on[0]: buf.append(o[0] if isinstance(o, tuple) else o)
    return o
Cls.__call__ = cap

ids = proc.tokenizer.encode("The quick brown fox jumps over the lazy dog.",
                            add_special_tokens=False)[:8]
on[0] = True
buf.clear()
mx.eval(model.thinker(mx.array([ids]), cache=make_prompt_cache(inner)))
pre = [np.array(h.astype(mx.float32)) for h in buf[:len(layers)]]

cache, dec = make_prompt_cache(inner), []
for t in ids:
    buf.clear()
    mx.eval(model.thinker(mx.array([[t]]), cache=cache))
    dec.append([np.array(h.astype(mx.float32))[0, 0] for h in buf[:len(layers)]])

cos = lambda a, b: float(a.ravel() @ b.ravel() /
                         (np.linalg.norm(a) * np.linalg.norm(b)))
for t in range(len(ids)):
    print(f"pos {t}: {cos(pre[0][0, t], dec[t][0]):.6f}")
```

Output (layer 0):

```
pos 0: 1.000000
pos 1: 0.996278
pos 2: 0.972446
pos 3: 0.979287
pos 4: 0.966685
pos 5: 0.937109
pos 6: 0.946232
pos 7: 0.974156
```

Position 0 matches exactly — both paths use position 0 there — and everything after diverges. Repeated runs of each path individually are deterministic to 1.00000000, and prefill is causal (perturbing only the last token leaves earlier positions bit-identical), so this is not numerical noise or a masking issue.

### Impact

Over a 64-token sequence, mean hidden-state cosine across all 48 layers:

| | cos mean | p1 | layer 0 |
|---|---:|---:|---:|
| 0.6.15 as shipped | 0.6788 | 0.4583 | 0.9307 |
| with offset-aware positions | **0.9975** | 0.9863 | **0.99996** |

This changes conclusions, not just digits. Training a small probe on prefill-extracted features and evaluating it on decode-extracted features for the same positions (5 seeds, paired bootstrap, 410 test positions):

| | degradation | 95% CI | prediction agreement |
|---|---:|---|---:|
| with fix | +0.24 pp | [−0.59, +1.02] — not significant | 0.964 |
| as shipped | **+11.17 pp** | [+9.12, +13.07] | 0.781 |

So any workflow that extracts features or logits via prefill and then runs the model in decode is silently inconsistent.

### Proposed fix

Make `get_rope_index` offset-aware for the text-only path, or let the thinker pass `position_ids=None` through so the existing correct branch in attention is used. Minimal version of the former:

```python
def get_rope_index(self, input_ids, image_grid_thw=None, video_grid_thw=None,
                   attention_mask=None, offset=0):
    batch_size, seq_length = input_ids.shape
    position_ids = mx.arange(offset, offset + seq_length, dtype=mx.int32)
    position_ids = mx.broadcast_to(position_ids[None, :], (batch_size, seq_length))
    ...
```

with the caller passing `cache[0].offset`. The multimodal branch needs the same treatment plus `rope_deltas`, which I have not worked through — happy to open a PR if you would like it done that way, or if you would rather route it through the `position_ids is None` path instead, that is a smaller change.

### Environment

- mlx-vlm 0.6.15, mlx 0.32.1, Python 3.12
- macOS 26.4.1, M2 Ultra 64 GB
- `mlx-community/Qwen3-Omni-30B-A3B-Instruct-4bit` (also reproduces at 8-bit)

Lazarus-931:
Forward the thinker's `get_rope_index` positions only on prefill so that during decode the language model continues RoPE positions from the cache offset (via the `rope_deltas` it seeds at prefill) instead of pinning every generated token to position 0 (fixes #1983).

### DIFF
--- a/mlx_vlm/models/qwen3_omni_moe/thinker.py
+++ b/mlx_vlm/models/qwen3_omni_moe/thinker.py
@@ -351,6 +351,26 @@ class Thinker(nn.Module):
             **kwargs,
         }
 
+        # Forward RoPE positions only on the prefill pass. ``get_input_embeddings``
+        # builds ``position_ids`` with ``get_rope_index``, which restarts at 0 on
+        # every call; on decode (seq_len 1) that pins each generated token to RoPE
+        # position 0 (#1983). Instead, on decode let the language model continue
+        # positions from the cache offset via the ``rope_deltas`` it seeded at
+        # prefill -- this is correct for both the text-only and multimodal paths.
+        c0 = cache[0] if cache else None
+        if c0 is not None:
+            offset = getattr(c0, "_idx", None)
+            if offset is None:
+                offset = getattr(c0, "offset", 0)
+            offset = (
+                int(offset.max().item())
+                if isinstance(offset, mx.array)
+                else int(offset)
+            )
+            if offset > 0:
+                kwargs.pop("position_ids", None)
+                kwargs.pop("rope_deltas", None)
+
         logits = self.language_model(input_ids, mask=mask, cache=cache, **kwargs)
 
         return logits
--- a/mlx_vlm/models/qwen3_omni_moe/thinker.py
+++ b/mlx_vlm/models/qwen3_omni_moe/thinker.py
@@ -351,7 +351,6 @@ class Thinker(nn.Module):
             **kwargs,
         }
 
-
         c0 = cache[0] if cache else None
         if c0 is not None:
             offset = getattr(c0, "_idx", None)
--- a/mlx_vlm/models/qwen3_omni_moe/thinker.py
+++ b/mlx_vlm/models/qwen3_omni_moe/thinker.py
@@ -351,12 +351,7 @@ class Thinker(nn.Module):
             **kwargs,
         }
 
-        # Forward RoPE positions only on the prefill pass. ``get_input_embeddings``
-        # builds ``position_ids`` with ``get_rope_index``, which restarts at 0 on
-        # every call; on decode (seq_len 1) that pins each generated token to RoPE
-        # position 0 (#1983). Instead, on decode let the language model continue
-        # positions from the cache offset via the ``rope_deltas`` it seeded at
-        # prefill -- this is correct for both the text-only and multimodal paths.
+
         c0 = cache[0] if cache else None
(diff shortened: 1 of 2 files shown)


lucasnewman:
@Lazarus-931 Failing tests

lucasnewman (approved):

PR #1994, opened 2026-08-21
Expose --top-p, --top-k and --min-p on the generate CLI

Lazarus-931:
`generate_step` has accepted all three since it was written and the server takes them from request bodies — `server/schemas.py` defaults `top_p` to `DEFAULT_TOP_P` — but the generate CLI was the only surface without them, so they could only be set through `--gen-kwargs`.

The three constants were already declared in `dispatch.py` and never referenced; they now have a use, and since `ar.py` and `dispatch.py` each kept their own copy of nine defaults (with a third copy of `DEFAULT_TEMPERATURE` in `diffusion.py`), those move to `generate/common.py` beside the existing KV defaults so the CLI and the library cannot disagree.

### Verification

Each flag was confirmed to reach the sampler, on `mlx-community/Qwen2.5-0.5B-Instruct-4bit` at `--temperature 1.0 --seed 0`:

| invocation | output |
|---|---|
| `--temperature 0.0` (greedy) | `Sure, here are three colors: 1. Red 2. Blue 3. Green` |
| `--top-k 1` | identical to greedy |
| `--top-p 0.01` | identical to greedy |
| `--min-p 0.99` | identical to greedy |
| `--top-k 0` | `Sure! Here are three standard colors that most people associate with...` |

Collapsing to the greedy result is only possible if the value reaches `make_sampler`, and `--top-k 0` differing rules out temperature simply being ignored.

No value changes; every constant is still importable from its previous location. `test_generation_defaults.py` compares the CLI's argparse defaults against `generate_step`'s signature by introspection rather than restating numbers — now covering 13 shared defaults, up from 10. Full suite: 2386 passed, same 5 pre-existing `test_inkling.py` failures as `main` (missing optional audio deps locally).


lucasnewman (approved):
LGTM

PR #1995, opened 2026-08-21
Apply max_kv_size to models that build their own cache

Lazarus-931:
`make_prompt_cache` deferred entirely to a model's `make_cache()`, so `max_kv_size` was silently dropped on 97 of 173 models.

## Verification

e2e on cached checkpoints, a 32-token bound over a ~200-token prompt

| model | before | after |
|---|---|---|
| Qwen3.5-0.8B | identical output, still answers "Tuesday" | answers "Monday" — context genuinely lost |
| LFM2.5-1.2B | identical output, still answers "Tuesday" | output changes |
| Qwen2.5-0.5B (no `make_cache`) | already worked | unchanged |

Across 200 model directories, of the 72 instantiable from a generic config: 44 now apply the bound, 3 were already bounded tighter and are left alone, 1 correctly raises `UnboundedKVCache`, and none silently ignore it.


lucasnewman (approved):

mlx_vlm/models/cache.py
```diff
@@ -42,6 +47,67 @@ def create_causal_mask(
...
+    ``None`` when it takes no such argument. A bare ``**kwargs`` does not
+    count: it would swallow the bound instead of applying it, which is the
+    silent no-op this lookup exists to avoid.
+    """
+    try:
+        parameters = inspect.signature(make_cache).parameters
+    except (TypeError, ValueError):
+        # Builtins and some bound C extensions have no introspectable signature.
+        return None
+    for name in ("max_size", "max_kv_size"):
```
lucasnewman, line 65: Is it practical to consolidate the interfaces to the cache implementations to consistently use `max_kv_size` instead of needing the introspection here?

Lazarus-931: Practical, yes, but I'd argue for keeping the introspection as the bridge rather than the destination, on two counts.

Scale: there are **224 `make_cache` definitions across 219 files in 122 model directories**, and only 6 currently accept an argument. So consolidating is a 200+ file change touching nearly every architecture.

Drift: 124 of the 202 model directories were added in the last seven weeks (92 in July, 32 so far in August). At that rate a convention without enforcement erodes immediately — and this is exactly where the bug came from. Of the ~44 architectures I found silently dropping the bound, 85% arrived in that window, against a 61% baseline, so the bulk ports are where it concentrates.

The introspection has one property a uniform signature doesn't: it lets a model **opt in** when it knows better. `diffusion_gemma` already does — it grew `make_cache(self, max_size=None)` on its own, presumably after someone hit this — and that path is preferred over the generic bounding here.

Happy to go either way, and two middle options if you'd prefer to move toward a uniform interface:

1. Convert incrementally. The 68 implementations that return a plain uniform `KVCache` are mechanical, so the introspection path shrinks over time without a single large diff.
2. Add a test asserting every `make_cache` either accepts a bound or is bounded by this helper, so a new architecture can't silently reintroduce the drop. That's the piece that makes a convention hold at 32 new architectures a month.

I'd lean on (2) regardless of whether we consolidate, since it's what stops this recurring.

lucasnewman: Ok, that's fine for now, just making sure.

lucasnewman (approved):

mlx_vlm/models/cache.py
Blaizzy, line 1: Thanks Alazar, this indeed addresses a real issue!

However, these changes look like a hack, basically too much policy in one helper: signature introspection, recursive cache rewriting, status bookkeeping, and a new exception.

We could solve this in a much simpler way.

Blaizzy (changes requested):

mlx_vlm/models/cache.py
Lazarus-931: would a 
```
-    def make_cache(self):
-        return self.language_model.make_cache()
+    def make_cache(self, max_kv_size=None):
+        return self.language_model.make_cache(max_kv_size)
```

fix across all model definitions make sense?

Blaizzy: Why? When `make_cache` makes no use of `max_kv_size`?

qwen3/language.py

```python
def make_cache(self):
   from ..cache import KVCache

   return [KVCache() for _ in self.layers]
```
qwen3_5/language.py

```python
def make_cache(self):
    return [ArraysCache(size=2) if l.is_linear else KVCache() for l in self.layers]

```

Lazarus-931:
superseded by #2195

PR #2002, opened 2026-08-22
fix(server): enforce tool_choice on the Anthropic /v1/messages route (#2000)

### LINKED ISSUE #2000: Anthropic /v1/messages ignores tool_choice (tool/none) while /v1/chat/completions enforces it (v0.6.15)
## Environment
mlx-vlm 0.6.15 server (as bundled in Nativ 0.3.3), macOS 27.0 (M5 Max), model `mlx-community/Qwen3.8-27B-nvfp4`.

Follow-on to #1045 (closed): `tool_choice` is now enforced on `/v1/chat/completions` — verified working (forced `get_time` on an unrelated prompt returns a `get_time` call; `"none"` suppresses calls). The Anthropic route added in #1060 doesn't get the same treatment.

## Reproduction

Two tools offered (`get_time`, `get_weather`), prompt deliberately favoring the other tool so compliance is causal evidence:

```
curl -s http://127.0.0.1:8080/v1/messages -H 'Content-Type: application/json' \
  -H 'anthropic-version: 2023-06-01' -H 'Authorization: Bearer nativ' -d '{
  "model":"mlx-community/Qwen3.8-27B-nvfp4","max_tokens":300,
  "messages":[{"role":"user","content":"What is the weather in Paris right now?"}],
  "tools":[{"name":"get_time","description":"Get the current time","input_schema":{"type":"object","properties":{}}},
            {"name":"get_weather","description":"Get the current weather for a location","input_schema":{"type":"object","properties":{"location":{"type":"string"}},"required":["location"]}}],
  "tool_choice":{"type":"tool","name":"get_time"}}'
```

**Expected:** a `tool_use` block for `get_time` (the forced tool).
**Actual:** a `tool_use` block for `get_weather` — the request is accepted (no error) but the constraint is not applied. `{"type":"none"}` likewise still yields tool calls (reproduced 3/3).

## Why it matters
The Anthropic route is what Claude Code and other Anthropic-SDK clients hit; `tool_choice: none` is how agent loops get terminated, and forced choice backs structured extraction. Since the OpenAI route already enforces these, this looks like the translation layer dropping the parameter rather than a generation limitation.

Anai-Guo:
Fixes #2000.

## Problem

`/v1/chat/completions` enforces `tool_choice` in `_prepare_chat_tool_choice` (`mlx_vlm/server/openai.py`): it filters the tool list down to the forced tool, drops the tools entirely for `"none"`, and injects an instruction into the prompt.

The Anthropic route never calls it. `_anthropic_messages_to_internal` converts the value to its OpenAI form and the endpoint then only does:

```python
template_kwargs = gen_args.to_template_kwargs()
if tool_choice is not None:
    template_kwargs["tool_choice"] = tool_choice
```

That constrains generation only for models whose chat template happens to read `tool_choice` (in this repo, `kimi_k3`). For every other model the full tool list is offered with no instruction, which is exactly the reported behaviour: `{"type": "none"}` still emits tool calls and `{"type": "tool", "name": "get_time"}` lets the model pick `get_weather` instead. A `tool_choice` naming a tool that was not supplied is also accepted silently rather than rejected.

## Fix

Apply `_prepare_chat_tool_choice` to the converted value inside `_anthropic_messages_to_internal`. That function is the single point both `/v1/messages` and `/v1/messages/count_tokens` go through, so both are covered, and `count_tokens` now counts the prompt that would actually be sent.

`_prepare_chat_tool_choice` raises `HTTPException`, which the Anthropic endpoints would otherwise surface as a `500 api_error`, so both call sites map it to `_anthropic_error_response` to keep the Anthropic error envelope and the 400 status.

The mapping from the Anthropic vocabulary was already in place and is unchanged (`auto`/`none` pass through, `any` → `required`, `{"type": "tool", "name": X}` → `{"type": "function", "function": {"name": X}}`), so `template_kwargs["tool_choice"]` keeps the same shape it had before — templates like `kimi_k3`'s that already read it see no change.

## Verification

`mlx` has no Windows wheel, so I could not run `mlx_vlm/tests/test_server.py` locally. Instead I drove the real `_anthropic_messages_to_internal` from the checked-out source (leaf imports stubbed, `_as_plain_dict` taken from `request_normalization.py`, `_prepare_chat_tool_choice` sliced out of `openai.py` — no hand-copied bodies), with two tools offered and a prompt favouring the tool that was *not* forced:

**Before** (`ffe22e1`) — every mode is indistinguishable from sending no `tool_choice` at all:

```
tool_choice absent           -> tools=['get_time', 'get_weather']   prompt unchanged
tool_choice auto             -> tools=['get_time', 'get_weather']   prompt unchanged
tool_choice none             -> tools=['get_time', 'get_weather']   prompt unchanged
tool_choice any              -> tools=['get_time', 'get_weather']   prompt unchanged
tool_choice tool=get_time    -> tools=['get_time', 'get_weather']   prompt unchanged
tool_choice tool=nope        -> tools=['get_time', 'get_weather']   prompt unchanged
```

**After:**

```
tool_choice absent           -> tools=['get_time', 'get_weather']   prompt unchanged
tool_choice auto             -> tools=['get_time', 'get_weather']   prompt unchanged
tool_choice none             -> tools=None
tool_choice any              -> tools=['get_time', 'get_weather']   + "You must call one or more of the available functions…"
tool_choice tool=get_time    -> tools=['get_time']                  + "You must call the 'get_time' function…"
tool_choice tool=nope        -> HTTP 400: tool_choice references unknown function 'nope'.
```

The `absent` and `auto` rows are byte-identical before and after, which is the no-`tool_choice` path that all existing traffic takes.

Added five tests in `mlx_vlm/tests/test_server.py`, mirroring the existing `test_chat_completions_*_tool_choice_*` tests: `none` disables tools, `any` adds the instruction, a forced tool filters the list, unsatisfiable choices return an Anthropic-shaped 400, and `count_tokens` applies the same filtering.

`black==26.3.1` and `isort --profile=black` (the pinned `.pre-commit-config.yaml` revisions) report both files unchanged.

## Note on the base commit

My fork's `main` is 10 commits behind upstream and I cannot sync it (`merge-upstream` returns 422 — my token lacks the `workflow` scope needed to update `.github/workflows/tests.yml`). Both files I touch are byte-identical between my fork's `main` and upstream `ffe22e1`, so the diff above is exactly my change and nothing else. Happy to rebase if you'd prefer.

### DIFF
--- a/mlx_vlm/server/anthropic.py
+++ b/mlx_vlm/server/anthropic.py
@@ -8,7 +8,7 @@ import uuid
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 import mlx.core as mx
-from fastapi import Request
+from fastapi import HTTPException, Request
 from fastapi.responses import JSONResponse, StreamingResponse
 
 from ..generate import generate, stream_generate
@@ -21,6 +21,7 @@ from .generation import (
     _build_metrics_envelope,
     _count_prompt_tokens,
 )
+from .openai import _prepare_chat_tool_choice
 from .responses_state import (
     make_response_stream_state,
     process_tool_calls,
@@ -366,6 +367,11 @@ def _anthropic_messages_to_internal(
 
     tools = _anthropic_tools_to_openai(request.tools)
     tool_choice = _anthropic_tool_choice_to_openai(request.tool_choice)
+    # Passing tool_choice through as a template kwarg only constrains the models
+    # whose chat template reads it, so enforce it the way /v1/chat/completions does.
+    processed_messages, tools, tool_choice = _prepare_chat_tool_choice(
+        processed_messages, tools, tool_choice
+    )
     return processed_messages, images, tools, tool_choice
 
 
@@ -469,9 +475,12 @@ async def anthropic_messages_endpoint(http_request: Request):
         )
         model, processor, config = get_cached_model(request.model, adapter_path)
 
-        processed_messages, images, tools, tool_choice = (
-            _anthropic_messages_to_internal(request)
-        )
+        try:
+            processed_messages, images, tools, tool_choice = (
+                _anthropic_messages_to_internal(request)
+            )
+        except HTTPException as e:
+            return _anthropic_error_response(e.status_code, str(e.detail))
         tool_parser_type = _infer_tool_parser_from_processor(processor)
         tool_module = load_tool_module(tool_parser_type) if tool_parser_type else None
 
@@ -1059,9 +1068,12 @@ async def anthropic_count_tokens_endpoint(http_request: Request):
         body = _normalize_anthropic_system_messages(await http_request.json())
         request = _anthropic_request_with_derived_fields(AnthropicRequest(**body))
         model, processor, config = get_cached_model(request.model)
-        processed_messages, images, tools, tool_choice = (
-            _anthropic_messages_to_internal(request)
-        )
+        try:
+            processed_messages, images, tools, tool_choice = (
+                _anthropic_messages_to_internal(request)
+            )
+        except HTTPException as e:
+            return _anthropic_error_response(e.status_code, str(e.detail))
         gen_args = _build_gen_args(
             request, processor, tenant_id=_read_tenant_id(http_request)
         )
--- a/mlx_vlm/tests/test_server.py
+++ b/mlx_vlm/tests/test_server.py
@@ -22,6 +22,7 @@ from transformers.utils.chat_parsing import ResponseParser, parse_response
 
 import mlx_vlm.reranker_loader as reranker_loader
 import mlx_vlm.server as server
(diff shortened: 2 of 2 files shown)


lucasnewman (changes requested):
Please sign commits before we review.

Anai-Guo:
@lucasnewman Done — the branch now carries a single GPG-signed commit (`verified: true`), and I've switched my workflow over so future PRs are signed from the start.

Note the force-push made GitHub auto-close the PR for a moment (the intermediate head was an ancestor of `main`); the reopen above is only to refresh the head. The diff is byte-for-byte unchanged from what you saw.

lucasnewman (approved):
Looks good, thanks!

PR #2006, opened 2026-08-22
Use integer Metal grid dimensions

Lazarus-931:
Use Python integers for Metal grid dimensions to remain compatible with nanobind 2.15.

### DIFF
--- a/mlx_vlm/models/kernels.py
+++ b/mlx_vlm/models/kernels.py
@@ -1,3 +1,5 @@
+import math
+
 import mlx.core as mx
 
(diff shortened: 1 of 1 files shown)


Blaizzy:
What are the perf Numbers ?

Lazarus-931:
> What are the perf Numbers ?

the main improvement and focus really was scalar grid calculation went from 189.7us to 0.38us

across diff measurements for this path, only saw improved speed:

image

lucasnewman (approved):

PR #2010, opened 2026-08-23
Ensure APC is reset before unloading a model to avoid stale cache state persisting

lucasnewman:
This was reported in https://github.com/Blaizzy/nativ/pull/365 but is better solved here.


Blaizzy (approved):
LGTM

PR #1960, opened 2026-08-19
Redesign APC for dense and hybrid model caches

Blaizzy:
## Summary

Redesign Automatic Prefix Caching around a model-agnostic cache plan and coordinator, following the structure of vLLM's hybrid cache manager.

- derive per-layer cache specs from `model.make_cache()` and group compatible cache layouts
- keep native dense KV caches on the pageable block pool
- route sliding-window, recurrent/SSM, MLA, composite, VLM, and Omni cache layouts through restorable checkpoints at one common prefix boundary
- use the same `APCCoordinator` API from streaming and batched generation
- add snapshot/restore support and disk persistence for generic and specialized cache entries
- fall back to the generation cache factory for dense language backbones that do not define `make_cache()`
- document cache-plan diagnostics and checkpoint configuration
- default OLMoE's omitted `rms_norm_eps` to the upstream value (`1e-5`), allowing the original AllenAI checkpoint to load

## Why

The previous APC path assumed a mostly uniform dense KV layout and split behavior across block and exact-cache branches in generation code. Custom and hybrid layouts were either disabled or required model-specific handling.

The new cache plan classifies capabilities from the concrete cache objects instead of model names. Generation only talks to the coordinator, while the coordinator selects block paging or checkpoint restoration.

## Impact

Dense LM/VLM models retain block-level prefix sharing. Hybrid attention, recurrent, composite, VLM, and Omni architectures can now reuse prefixes without flattening incompatible state into dense KV blocks. New cache types can participate by exposing a snapshot/restore contract.

### Architecture cache-hit matrix

`mlx_vlm/tests/test_apc_all_models.py` statically discovers every in-tree model cache contract and runs a first-store/second-request lookup using tiny synthetic cache state.

- **180/180 model architecture contracts produced a cache hit**
- every hit reused the complete 16-token synthetic prefix
- dense-compatible contracts incremented block lookup hits
- hybrid/specialized contracts restored a warm checkpoint and incremented exact hits
- result: **183 passed, 2 warnings**

This matrix deliberately does not download checkpoints or initialize trained weights. It validates cache construction contracts, grouping, storage, lookup, cloning, and restoration. Full random-weight model forward/logit equivalence is not part of the architecture matrix; targeted real-checkpoint runs are reported below.

### APC and generation regression suite

```text
548 passed, 1 skipped, 2 warnings in 7.03s
```

[...]

PR #2015, opened 2026-08-23
APC: anchor checkpoints at semantic boundaries for checkpoint-mode caches

Lazarus-931:
Checkpoint-mode caches snapshot only at the prompt tail, so any small history edit re-prefills everything.

`semantic_checkpoint_offsets` picks extra snapshot points at tool-call and turn markers that agents actually edit around.

Recurrent and windowed layers cannot slice prefixes, so stored checkpoints are their only reuse mechanism.

Restore resumes from the deepest surviving boundary; block-mode models are untouched and see no change.

| model | attention | cache | tokens reused | TTFT |
|:--|:--|:--|--:|--:|
| Falcon-H1-1.5B | SSM / Mamba hybrid | `CacheList` ×24 | 0 → 11,141 | **3.17×** |
| LFM2-1.2B | conv-recurrent | `ArraysCache` ×10 | 0 → 9,757 | **3.12×** |
| Falcon-H1-0.5B | SSM / Mamba hybrid | `CacheList` ×36 | 0 → 11,021 | **3.08×** |
| Qwen3.5-4B | KDA recurrent | `ArraysCache` ×24 | 0 → 10,192 | **3.05×** |
| Qwen3.5-2B | recurrent | `ArraysCache` ×18 | 0 → 10,192 | **2.98×** |
| gemma-3-1b-it | sliding window | `RotatingKVCache` ×22 | 0 → 10,633 | **2.93×** |
| LFM2-700M | conv-recurrent | `ArraysCache` ×10 | 0 → 9,852 | **2.90×** |
| Qwen3.5-0.8B | recurrent | `ArraysCache` ×18 | 0 → 10,091 | **2.85×** |
| Llama-3.2-1B, Phi-3.5-mini, gemma-2-2b, Qwen2.5-VL-3B, paligemma-3b, Qwen2.5-0.5B | dense | `KVCache` (block mode) | n/a | 1.00× |

8-turn tool-calling transcript, context compaction at turn 5, budget 8 vs tail-only.

There is more benefits across more models, I didn't register any regressions across 30+ model sweeps :)

btw this builds off of #1960

### DIFF
--- a/mlx_vlm/apc.py
+++ b/mlx_vlm/apc.py
@@ -71,6 +71,31 @@ DEFAULT_BLOCK_SIZE = 16
 DEFAULT_NUM_BLOCKS = 2048
 SEED_PARENT_HASH = 0
 
+# Matching is by token-id sequence, so a marker is reliable when it is a single
+# special token in the target vocabulary. Multi-token markers still match when
+# BPE happens to segment them the same way in context, but that is not
+# guaranteed -- keep genuine special tokens for each family in this list.
+DEFAULT_SEMANTIC_BOUNDARY_MARKERS = (
+    "</tool_call>",
+    "</tool_response>",
+    "</tool_output>",
+    "</think>",
+    "<|observation|>",
+    "<|user|>",
+    "<|assistant|>",
+    "<|im_end|>",
+    "<|end|>",
+    "<end_of_turn>",
+    "<start_of_turn>",
+    "<|eot_id|>",
+)
+DEFAULT_SEMANTIC_MIN_GAP = 256
+# Checkpoints are whole prompt-cache snapshots, so the pool grows with both the
+# checkpoint count and the context length. Left uncapped, a large budget on a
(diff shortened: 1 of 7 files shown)


Blaizzy:
I'm going to close this PR for now because the solution is not generalisable. 

What are you trying to solve? 

Reproducible problem and expected outcome.

Lazarus-931:
> I'm going to close this PR for now because the solution is not generalisable.
> 
> What are you trying to solve?
> 
> Reproducible problem and expected outcome.

This is an improvement not an issue

On checkpoint-mode caches (recurrent/windowed), APC snapshots only at the prompt tail, so any edit to agent history invalidates every checkpoint and re-prefills the whole prompt, where the same conversation appended-to reuses all of them.

This only happens when APC_ENABLED=1, the server path, a checkpoint-mode model, and multi-turn history that gets edited rather than appended. Useful for nativ

PR #2021, opened 2026-08-25
Fix sectioned MRoPE split under mlx 0.32.2

Lazarus-931:
mlx 0.32.2 rejects an mx.array for mx.split indices, so return plain Python ints.


Lazarus-931:
Superseded by the same fix already on #2020 (09ad11b6).

lucasnewman (approved):

lucasnewman (approved):

PR #2023, opened 2026-08-26
Support DeepSeek V4 speculative chunked prefill

### LINKED ISSUE #2022: DeepSeek-V4 MTP speculative decoding ignores chunked prefill
## Description

Chunked prefill is disabled for DeepSeek-V4 whenever an MTP drafter is supplied, even when `prefill_step_size` is explicitly configured.

The shared generation code deliberately uses a conservative fallback for speculative decoding: target models must expose a `chunked_prefill_policy` before their prompt can be split into chunks. DeepSeek-V4 already supports the state needed by MTP speculative decoding through `return_hidden` and `return_shared_kv`, but its language model does not declare that those values can be captured by the final prefill call.

As a result, `_chunked_prefill_enabled()` returns `False` and the generation paths silently clear `prefill_step_size`. Long prompts are then processed in a single target-model forward pass, losing the peak-memory benefit of chunked prefill.

## Reproduction

The issue can be reproduced with a tiny randomly initialized DeepSeek-V4 model, without downloading checkpoint weights:

```python
from mlx_vlm.generate.common import _chunked_prefill_enabled
from mlx_vlm.models import deepseek_v4


config = deepseek_v4.ModelConfig(
    model_type="deepseek_v4",
    vocab_size=128,
    hidden_size=64,
    num_hidden_layers=4,
    num_attention_heads=4,
    num_key_value_heads=1,
    q_lora_rank=16,
    o_lora_rank=8,
    o_groups=2,
    head_dim=16,
    qk_rope_head_dim=4,
    sliding_window=16,
    compress_ratios=[0, 0, 4, 0],
    index_n_heads=4,
    index_head_dim=8,
    index_topk=4,
    moe_intermediate_size=16,
    n_routed_experts=4,
    n_shared_experts=1,
    num_experts_per_tok=2,
    num_hash_layers=1,
    hc_mult=2,
    hc_sinkhorn_iters=2,
)
model = deepseek_v4.Model(config)

enabled = _chunked_prefill_enabled(
    model,
    draft_model=object(),
    draft_kind="mtp",
    prefill_kwargs={
        "return_hidden": True,
        "return_shared_kv": True,
    },
)

print(f"DeepSeek-V4 MTP chunked prefill enabled: {enabled}")
assert enabled
```

## Actual behavior

The script prints:

```text
DeepSeek-V4 MTP chunked prefill enabled: False
```

The assertion fails. During generation, the requested `prefill_step_size` is consequently replaced with `None` and the whole prompt is processed at once.

## Expected behavior

DeepSeek-V4 should allow chunked prefill for normal generation and for MTP speculative generation when the final prefill call requests both the hidden state and shared-KV state required to initialize the drafter.

Intermediate chunks should update the target cache without retaining speculative state. Only the final prompt segment should request `return_hidden=True` and `return_shared_kv=True`, ensuring the drafter receives state corresponding to the tail of the complete prompt.

Other speculative strategies should remain disabled unless DeepSeek-V4 implements their state-capture contract.

byoan:
Fixes #2022

## Summary

Allow DeepSeek-V4 prompts to use chunked prefill during MTP speculative decoding.

DeepSeek-V4 already exposes the final hidden state and shared-KV state required to initialize its MTP drafter. This change adds the model-specific `chunked_prefill_policy` needed by the shared generation code to recognize that its speculative prefill contract is safe to chunk.

With a configured `prefill_step_size`, intermediate prompt chunks now update the target cache normally. The generation path requests `return_hidden` and `return_shared_kv` only on the final prefill segment, so the drafter is initialized from the tail state of the complete prompt without retaining hidden states for every chunk.

## Scope and safeguards

The policy allows:

- chunked prefill when no drafter is present;
- MTP speculative chunked prefill when both `return_hidden` and `return_shared_kv` are requested.

It remains conservative when either MTP capture flag is missing. It also rejects DFlash and EAGLE3 even if `capture_layer_ids` is provided, because the DeepSeek-V4 language model does not currently implement that capture contract.

The shared conservative fallback for models without a policy is unchanged.

## Tests

The model tests cover normal prefill, complete and incomplete MTP capture options, and rejection of an unsupported speculative strategy.

An end-to-end `generate_step` regression uses a five-token prompt with a prefill step size of two and verifies the target call sequence:

- two intermediate two-token forwards without speculative capture;
- one final one-token forward with both MTP capture flags enabled.

The existing generic model-policy regression remains unchanged.

## Test plan

```text
python -m pytest mlx_vlm/tests/test_generate.py -q
```

```text
129 passed, 2 warnings
```

```text
python -m pytest mlx_vlm/tests/test_models.py -k deepseek_v4 -q
```

```text
2 passed, 357 deselected, 2 warnings, 9 subtests passed
```


byoan:
Tests failure is unrelated to this PR (mx.split() signature change)

Lazarus-931:
> Tests failure is unrelated to this PR (mx.split() signature change)

should be good to go once #2021 is merged :)

byoan:
Back to green, ready for review :)

Lazarus-931 (approved):

Lazarus-931:
before merging, can i ask for acceptance per round, for chunked and non chunked under same prompt?

Lazarus-931:
> before merging, can i ask for acceptance per round, for chunked and non chunked under same prompt?

on a reflection, this could be implemented for several other models so closing this in return of #2096 which covers DeepSeek-V4 along with Qwen3.5 and EAGLE3/DFlash

thx for your report and pr :)

PR #2027, opened 2026-08-26
fix(server): stop dropping streamed content after a tool call ends

Anai-Guo:
### The bug

`suppress_tool_call_content()` keeps tool-call markup out of the streamed
`delta.content`. It is given the tool parser's **start** marker only:

```python
tc_start = tool_module.tool_call_start if tool_module else None
tc_end   = tool_module.tool_call_end   if tool_module else None   # never used

in_tool_call, delta_content = suppress_tool_call_content(
    full_output, in_tool_call, tc_start, delta_content
)
```

Inside the helper, once `in_tool_call` flips on it never flips back, because the
opening test is against the whole accumulated output:

```python
if not in_tool_call:
    if tc_start in full_output:   # stays true for the rest of the stream
        return True, None
else:
    return True, None             # suppress everything from here on
```

So the suppression is not scoped to the call — **every token generated after the
first tool call is dropped from the stream**, including the model's own prose.

### Why this is a bug and not the intended behaviour

The non-streaming path in the same file keeps that text. `process_tool_calls()`
uses *both* markers, removes only the call spans, and returns the rest as
`remaining_text`, which `chat_completions_endpoint` then stores as the message
content.

And `tc_end` is already read for exactly this purpose in
`chat_completions_endpoint` — it is just never passed anywhere.

### Reproduction

Feeding one stream through the helper, and the same output through the
non-streaming sibling:

| | visible content |
|---|---|
| streaming (`suppress_tool_call_content`) | `'Let me look. '` |
| non-streaming (`process_tool_calls` → `remaining_text`) | `'Let me look.   The weather is sunny.'` |

`" The weather is sunny."` is silently lost, and with two calls in a turn
everything between and after them is lost too.

### The fix

Pass `tc_end` through, and stop suppressing at the end of the last **completed**
call so that later text is matched against the tail rather than the whole
output. Both endpoints (`/v1/chat/completions` and `/v1/responses`) are wired up.

`tc_end` defaults to `None`, and parsers whose `tool_call_end` is `""` (handled
explicitly in `process_tool_calls`) fall through to the previous behaviour, so
no existing parser changes shape.

### Tests

Four cases added to `mlx_vlm/tests/test_responses_state.py`, which had no
coverage for this helper:

- content resumes after a completed tool call
- content resumes between two consecutive calls
- markup is still suppressed while a call is open (unchanged)
- an empty `tool_call_end` keeps the old latching behaviour (unchanged)

The first two fail on `main` and pass with this change; the last two pass on
both. `black` 26.3.1 and `isort` 5.13.2 (the pinned pre-commit revisions) are
clean on the three touched files.

### DIFF
--- a/mlx_vlm/server/openai.py
+++ b/mlx_vlm/server/openai.py
@@ -1096,6 +1096,11 @@ async def responses_endpoint(request: Request):
                         if tool_module is not None and chat_tools
                         else None
                     )
+                    tc_end = (
+                        tool_module.tool_call_end
+                        if tool_module is not None and chat_tools
+                        else None
+                    )
                     thinking_state = make_response_stream_state(
                         processor,
                         prompt_has_open_thinking(
@@ -1157,7 +1162,7 @@ async def responses_endpoint(request: Request):
                                 )
                             delta = thinking_delta.content
                             in_tool_call, delta = suppress_tool_call_content(
-                                full_text, in_tool_call, tc_start, delta
+                                full_text, in_tool_call, tc_start, delta, tc_end
                             )
                             usage_stats = {
                                 "input_tokens": ctx.prompt_tokens,
@@ -1211,7 +1216,7 @@ async def responses_endpoint(request: Request):
                                 )
                             delta = thinking_delta.content
                             in_tool_call, delta = suppress_tool_call_content(
-                                full_text, in_tool_call, tc_start, delta
+                                full_text, in_tool_call, tc_start, delta, tc_end
                             )
                             if chunk_finish is not None:
                                 finish_reason = chunk_finish
@@ -1844,7 +1849,11 @@ async def chat_completions_endpoint(request: ChatRequest, http_request: Request)
 
                             # Suppress tool-call markup from content
                             in_tool_call, delta_content = suppress_tool_call_content(
-                                full_output, in_tool_call, tc_start, delta_content
+                                full_output,
+                                in_tool_call,
+                                tc_start,
+                                delta_content,
+                                tc_end,
                             )
 
                             chunk_logprobs = None
--- a/mlx_vlm/server/responses_state.py
+++ b/mlx_vlm/server/responses_state.py
@@ -269,10 +269,25 @@ def suppress_tool_call_content(
     in_tool_call: bool,
     tc_start: Optional[str],
     delta_content: Optional[str],
+    tc_end: Optional[str] = None,
 ) -> Tuple[bool, Optional[str]]:
     """Suppress tool-call markup from streamed delta.content."""
     if not tc_start:
         return in_tool_call, delta_content
+    if tc_end:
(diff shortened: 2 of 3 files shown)


Anai-Guo:
Pushed 672a6dd fixing the CI failure this PR introduced.

`test_responses_streaming_emits_native_tool_call_items` was right: rewinding `full_output` past the last end marker also hid the markup from the check that decides whether to suppress the *current* delta. When a single chunk carried a whole `<tool_call>...</tool_call>`, the rewound tail was empty, so `tc_start in full_output` was false and the raw chunk went out as content.

Now the part of the delta that fell inside the completed call is withheld and only the text after the end marker is emitted, so a chunk that both closes a call and carries trailing text yields just the trailing text. Added `test_tool_call_markup_inside_a_single_chunk_is_still_suppressed` covering both shapes; it fails on the previous commit and passes on this one.

For reference, `test_rope_utils.py::test_sectioned_precomputed_rotary_fast_path_matches_fallback` (4 failures) is red on current `main` for everyone — it also fails on #2023 from another contributor — so that part is unrelated to this PR.

lucasnewman (changes requested):
@Anai-Guo Needs signed commits

Anai-Guo:
Thanks @lucasnewman — done. This branch is now a single commit signed with GitHub's GPG key (`verified: true`); the tree is byte-identical to before, so the diff is unchanged.

Force-pushing to the fork briefly auto-closed the PR, hence the reopen notification — no content changed.

mlx_vlm/server/responses_state.py
```diff
@@ -269,10 +269,25 @@ def suppress_tool_call_content(
...
         return in_tool_call, delta_content
+    if tc_end:
+        # Suppression must stop where the call does, otherwise every token
+        # after it is dropped: `tc_start in full_output` stays true for the
+        # rest of the stream. Only the text after the last completed call
+        # can open a new one, so match against that tail.
+        closed_at = full_output.rfind(tc_end)
+        if closed_at != -1:
+            in_tool_call = False
+            full_output = full_output[closed_at + len(tc_end) :]
```
lucasnewman, line 285: Doesn't this drop the prefix content before a tool call, e.g. `Before <tool_call>{}</tool_call> after` would drop "Before "?

mlx_vlm/server/openai.py
```diff
@@ -1096,6 +1096,11 @@ async def stream_generator():
                         if tool_module is not None and chat_tools
                         else None
                     )
+                    tc_end = (
```
lucasnewman, line 1099: We should also implement this for the /v1/messages path.

lucasnewman (changes requested):
See comments.

lucasnewman:
I opened https://github.com/Blaizzy/mlx-vlm/pull/2062 to cover the remaining cases.

PR #2029, opened 2026-08-26
Fix positioned top-k sampling for speculative generation

tjansn:
## Summary

- apply `top_k` inside `_PositionedTargetSampler`, before temperature scaling
- keep seeded position-keyed sampling enabled when `top_k` is non-default
- add regressions for positioned-sampler selection and top-k filtering

## Problem

Seeded speculative generation already uses a position-keyed target sampler so proposal draws cannot perturb target draws. The selection gate excluded every request with non-default `top_k`, however, and the positioned sampler did not apply top-k itself. That sent seeded top-k generation back through the stateful sampler path, where serial and speculative grouping can consume target randomness differently.

The server sampling path applies top-k before temperature/top-p sampling. This patch ports that same ordering to the positioned target sampler and removes the now-unnecessary `top_k == 0` selection restriction.

## Validation

- targeted selection/filtering assertions pass locally
- Python compilation passes
- a local Qwen MTP matrix covering two quantization widths, two seeds, opposite serial/speculative arm orders, and five sampling regimes passed 40/40 reports after the fix, including `temperature=1`, `top_p=1`, `top_k=40`
- seeded completions repeated across opposite arm orders; proposal draws did not perturb target draws

I could not run the full upstream pytest suite in this environment because pytest is not installed in the pinned runtime.


Lazarus-931 (approved):

lucasnewman (approved):
Thanks!

tjansn:
You're welcome

PR #2030, opened 2026-08-26
GLM-5.3-Flash support

Lazarus-931:
Model: https://z.ai/blog/glm-5.3-flash

### DIFF
--- a/mlx_vlm/models/glm5_next/README.md
+++ b/mlx_vlm/models/glm5_next/README.md
@@ -1,0 +1,38 @@
+# GLM-5-Next (`glm5_next`)
+
+MLX support for the GLM-5-Next architecture, as shipped in **GLM-5.3-Flash**.
+
+## Architecture
+
+- **Hybrid decoder** — 34 Kimi-Delta linear-attention (KDA) layers interleaved with 11 DeepSeek sparse-attention (DSA) layers.
+- **MLA with NoPE** and a **lightning indexer** (top-`index_topk` key selection over pooled keys).
+- **288-expert MoE** (top-8) with a shared expert; **mHC hyper-connections**.
+- A **multi-token-prediction (nextn) head** at the final layer, used for self-speculative decoding.
+
+## Usage
+
+```python
+from mlx_vlm import generate, load
+
+model, processor = load("zai-org/GLM-5.3-Flash")
+print(generate(model, processor, "Explain multi-head latent attention.", max_tokens=256))
+```
+
+## Decode optimizations
+
+All are on the compute path (no weight changes) and lossless:
+
+| optimization | effect |
+| --- | --- |
+| KDA input-projection fusion | the six shared-input KDA projections become one (quantized) matmul via a lossless output-axis weight concat |
+| Lightning-indexer chunked prefill | bounds prefill peak memory to `O(chunk · P)` (avoids a one-shot `O(S · P)` blow-up at long context) |
+| Lightning-indexer incremental decode | per-step pool cost `O(T)` → `O(index_kpool)` (reuses stable complete pools) |
+| Short-context dense-MLA bypass | when the cache fits within `index_topk` the indexer would select every token, so it is skipped and DSA falls through to dense MLA |
+| Last-token `lm_head` | skip the vocab-wide projection on discarded prefill positions |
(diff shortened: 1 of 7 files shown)


lucasnewman:
@Lazarus-931 Needs signed commits and a README

mlx_vlm/models/glm5_next/README.md
```diff
@@ -0,0 +1,63 @@
...
+- **MLA with NoPE** and a **lightning indexer** (top-`index_topk` key selection over pooled keys).
+- **288-expert MoE** (top-8) with a shared expert; **mHC hyper-connections**.
+- A **multi-token-prediction (nextn) head** at the final layer, used for self-speculative decoding.
+
+## Usage
+
+```python
+from mlx_vlm import generate, load
+
+model, processor = load("<glm5.3-flash-mlx>")
```
lucasnewman: Can you use the real repo id here?

lucasnewman (approved):

davidrhodus:
Thanks for the port — we validated it against `transformers` 5.16 (`Glm5NextForConditionalGeneration`) with a tiny-config parity test plus a teacher-forced module-by-module comparison, while building quantized checkpoints from `zai-org/GLM-5.3-Flash-BF16`. The mHC, KDA, NoPE-MLA, lightning-indexer and vision math all match the reference (1e-7 per block). Two things don't, plus a few small ones. Tests and a fixed copy of the package are at https://github.com/PipeNetwork/glm53-flash-mlx (`tests/test_parity.py`, `tests/debug_modules.py`).

**1. `swiglu_limit` is never applied in the text stack.** The reference clamps in every text MLP — `Glm5NextTextMLP` (dense layers 0–2 and `shared_experts`) and `Glm5NextTextExperts._apply_gate` (`gate.clamp(max=limit)`, `up.clamp(-limit, limit)`, then `silu(gate) * up`). `language.py` uses `DeepseekV32MoE` (plain `SwitchGLU`) and `DeepseekMLP`, so `config.swiglu_limit` is parsed and never read; the vision tower does clamp. In the module-level comparison every block matched to ~1e-7 except `mlp` on the sparse layers (4e-2 on a 2e-1 scale); with a clamped activation the end-to-end parity goes from 3.1e-1 to 9e-7. Fix: `SwitchGLU(..., activation=ClampedSwiGLU(config.swiglu_limit))` — note `SwitchGLU` calls `activation(x_up, x_gate)`, up first — and a clamped MLP for the dense layers and the shared expert. Compute-path only, no checkpoint change.

**2. mHC `base`/`scale` (and KDA `A_log`/`dt_bias`) must stay float32 through `convert`.** `cast_predicate` only protects `e_score_correction_bias`, so `mlx_vlm.convert` casts `attn_hc.base` etc. to bf16. `_hc_kernel` reads `base` through a `float4` pointer, so a bf16 `base` yields a wrong `comb` matrix. Measured at hidden 4096 (mlx 0.32.0): `_hc_kernel` vs `_hc_ops` agree to 9e-8 with fp32 `base`, and differ by 0.53 with bf16 `base`. Sinkhorn still normalises the result, so a converted checkpoint stays fluent with the wrong stream mixing on all 90 sites. The reference keeps these in `_keep_in_fp32_modules_strict`. Fix: exclude `.attn_hc.`, `.ffn_hc.`, `A_log`, `dt_bias` in `cast_predicate` (as `deepseek_v4` does), cast them to fp32 in `sanitize` so already-published checkpoints heal on load, and/or `astype(float32)` in the kernel call (27 elements).

**Smaller:**
- `q_a_layernorm`/`kv_a_layernorm` use eps 1e-6; the reference uses `config.rms_norm_eps` (1e-5). The indexer `k_norm` LayerNorm is 1e-6 in the reference (mlx default 1e-5).
- Router logits: reference computes `F.linear(x.float(), weight.float())` (`moe_router_dtype: float32`); `MoEGate` matmuls in bf16, which can flip top-8 membership at near ties.
- The README says the MTP head is used for self-speculative decoding; `sanitize` drops layer 45 and nothing uses it.

The FP8 dequant path inherited from `deepseek_v32` is correct (block broadcast checked against the real `weight_scale_inv` shapes), and cached decode / chunked prefill are exact against single-shot. Happy to open a PR with the fixes if useful.

*Disclosure: the review and tests were done with Claude Code; each item above is backed by a test in the linked repo.*

PR #2034, opened 2026-08-26
Optimize Qwen4 n-gram shard gathers

pyros-projects:
## Summary

Qwen4's PLE table is split into 128 very large embedding shards. The current lookup evaluates every shard and masks away the 127-ish answers it did not need. Correct, but a fairly expensive way to rediscover which shard owns an index.

This change materializes the small n-gram ID tensor once, maps IDs to the existing shard offsets, gathers only the addressed shards, and scatters those rows back into their original positions.

## Benchmark

M5 Max, 128 GiB unified memory, `Qwen3.8-Flash-Next-MLX-4bit`, 128-token single-sequence decode:

| Lookup | Decode | Prompt | Peak memory |
| --- | ---: | ---: | ---: |
| all 128 shards + mask | 27.62 tok/s | 8.41 tok/s | 103.31 GiB |
| addressed shards only | 31.10 tok/s | 8.41 tok/s | 103.34 GiB |

That is a 12.6% decode improvement. A separate 64-token run reproduced the result at 28.08 vs 31.48 tok/s. Generated token IDs were identical between both paths.

The intentional tradeoff is one host sync of the tiny ID tensor. It replaces 128 quantized embedding gathers per PLE invocation with gathers for only the shards actually addressed by those IDs.

## Tests

- Adds an uneven-shard regression that verifies bit-exact output, order preservation, repeated indices, and that an untouched shard is never called.
- `pytest -q mlx_vlm/tests/test_qwen4_exp.py`: 8 passed.
- All configured pre-commit hooks pass.

### DIFF
--- a/mlx_vlm/models/glm5_next/config.py
+++ b/mlx_vlm/models/glm5_next/config.py
@@ -1,4 +1,4 @@
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from typing import Any, Dict, List, Optional
 
 from ..base import BaseModelConfig
--- a/mlx_vlm/models/glm5_next/language.py
+++ b/mlx_vlm/models/glm5_next/language.py
@@ -10,13 +10,12 @@ from ..base import (
     scaled_dot_product_attention,
 )
 from ..cache import ArraysCache, CacheList, KVCache
+from ..deepseek_v4.hyper_connection import HyperConnection, hc_expand
 from ..deepseek_v32.language import DeepseekV32MoE
 from ..deepseek_v32.language import Model as DSV32Model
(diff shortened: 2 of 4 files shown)


lucasnewman:
@pyros-projects Thanks! Can you [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits and force push so we can clear the merge requirements?

pyros-projects:
Done — force-pushed `1f507f1`, and GitHub now reports the commit signature as verified. I also ran `pre-commit run --all-files`; it exposed existing formatter drift in `glm5_next`, so that cleanup is included and all hooks pass locally. Let’s see round two burn green. 🔥

lucasnewman (approved):
Thanks!

PR #2036, opened 2026-08-26
fix(cache): trim ChunkedKVCache on valid length, not buffer width (#2031)

### LINKED ISSUE #2031: ChunkedKVCache trims by buffer size, not valid length — port ml-explore/mlx-lm#1673
## Summary

`ChunkedKVCache.maybe_trim_front()` in `mlx_vlm/models/cache.py` is the
pre-fix version of the same method in mlx-lm. It trims by the **allocated
buffer width** (`self.keys.shape[2]`, which `update_and_fetch` pads up to
multiples of `step = 256`) instead of the **valid length**
(`self.offset - self.start_position`). Because the buffer is up to
`step - 1` slots wider than the valid data, every trim drops up to 255
valid tokens from the front of the attention window, leaving it that much
smaller than `chunk_size`.

mlx-lm fixed this in [ml-explore/mlx-lm#1673](https://github.com/ml-explore/mlx-lm/pull/1673)
(commit [`e03e45ed`](https://github.com/ml-explore/mlx-lm/commit/e03e45ed),
"Fix ChunkedKVCache dropping valid tokens during decode (llama4)"). The
mlx-vlm copy predates it.

## Affected code

- `mlx_vlm/models/cache.py:801` (`ChunkedKVCache.maybe_trim_front`, at `2b31570b`):

```python
def maybe_trim_front(self):
    # Maintain the cache below the chunk size
    if self.keys is not None and self.keys.shape[2] >= self.chunk_size:
        self.start_position += self.keys.shape[2] - self.chunk_size
        self.keys = self.keys[..., -self.chunk_size :, :]
        self.values = self.values[..., -self.chunk_size :, :]
```

- Consumer: `mlx_vlm/models/llama4/language.py:249-250` calls it on every
  chunked-attention layer (3 of every 4 layers), so Llama-4 generations that
  cross `attention_chunk_size` (8192) hit this on every subsequent step.

## Standalone demonstration

Small numbers (`chunk_size=8`, `step=4`) make it visible. Feed 14 tokens one
at a time with `k_t = t`, calling `maybe_trim_front()` before each step the
way the llama4 language model does:

```python
import mlx.core as mx
from mlx_vlm.models.cache import ChunkedKVCache

ChunkedKVCache.step = 4
cache = ChunkedKVCache(chunk_size=8)
for t in range(14):
    k = mx.full((1, 1, 1, 1), float(t))
    cache.maybe_trim_front()
    keys, _ = cache.update_and_fetch(k, k)
print(cache.offset - cache.start_position, keys.flatten().tolist())
```

| variant | valid window after 14 tokens |
| --- | --- |
| mlx-vlm | `6` tokens — `[8..13]`; tokens 6 and 7 dropped while still inside the chunk window |
| with the mlx-lm fix ported | `9` tokens — `[5..13]`; never shrinks below `chunk_size` |

With the real parameters (`chunk_size=8192`, `step=256`) the window ends up
to 255 tokens short of the intended 8192 after every trim — a silent quality
degradation on long Llama-4 generations rather than a crash.

## Fix to port

The mlx-lm replacement, verbatim from `e03e45ed`:

```python
def maybe_trim_front(self):
    # Maintain the cache below the chunk size.
    if self.keys is None:
        return
    valid = self.offset - self.start_position
    if valid > self.chunk_size:
        trim = valid - self.chunk_size
        self.start_position += trim
        self.keys = self.keys[..., trim:valid, :]
        self.values = self.values[..., trim:valid, :]
```

Note the fixed slice `[..., trim:valid, :]` also stops carrying the unwritten
padding columns forward, which is what keeps the retained window aligned with
`start_position`.

## Environment

- mlx-vlm `main` at `2b31570b` (editable install from source)
- mlx 0.32.x, macOS (Apple silicon), Python 3.14
- Found by cross-checking recent mlx-lm cache fixes against the vendored
  copies; demonstration above runs against the installed package directly.

Anai-Guo:
Fixes #2031.

## The bug

`ChunkedKVCache.maybe_trim_front()` (`mlx_vlm/models/cache.py`) sizes its trim from `self.keys.shape[2]` — the **allocated buffer width**. `update_and_fetch()` pads that buffer up to a multiple of `ChunkedKVCache.step` (256), so it overstates the number of live tokens by up to `step - 1`:

```python
def maybe_trim_front(self):
    # Maintain the cache below the chunk size
    if self.keys is not None and self.keys.shape[2] >= self.chunk_size:
        self.start_position += self.keys.shape[2] - self.chunk_size
        self.keys = self.keys[..., -self.chunk_size :, :]
        self.values = self.values[..., -self.chunk_size :, :]
```

Two things go wrong:

- the guard fires while **fewer** than `chunk_size` tokens are actually cached, and
- `keys[..., -chunk_size:, :]` takes the tail of the *padded* buffer, so the retained window starts later than it should.

`start_position` is advanced by the same overstated amount, so the loss is permanent — every trim shortens the attention window by up to `step - 1` valid tokens instead of holding it at `chunk_size`.

The rest of the class already computes the valid length correctly: `update_and_fetch()` uses `self.offset - self.start_position` for both `prev` and `end`, and `trim()` uses it too. `maybe_trim_front()` is the only method that reaches for `keys.shape[2]` instead.

## Who hits it

`mlx_vlm/models/llama4/language.py:249-250` calls `maybe_trim_front()` on every chunked-attention layer (three of every four) once generation passes `attention_chunk_size`. So a Llama-4 decode runs those layers on a window ~255 tokens narrower than configured, and keeps diverging from a single forward pass over the same tokens.

## The fix

Port of what mlx-lm did in [`e03e45ed`](https://github.com/ml-explore/mlx-lm/commit/e03e45ed) ([ml-explore/mlx-lm#1673](https://github.com/ml-explore/mlx-lm/pull/1673)) — this copy of the class predates that fix. Trim on `offset - start_position` and slice `[trim:valid]` so the padding tail stays out of the window.

## Verification

Driving the cache one token at a time the way the llama4 layer does (`maybe_trim_front()` before each `update_and_fetch()`), and reading back which token ids are visible in the returned window:

| config | old window | new window | expected |
|---|---|---|---|
| `chunk_size=8`, `step=4`, 14 tokens | 6 | 9 | ≥ 8 |
| `chunk_size=8`, `step=256`, 300 tokens | **empty** | 9 | ≥ 8 |
| `chunk_size=8192`, `step=256`, 8500 tokens | 7988 | 8193 | ≥ 8192 |

The last row is the real llama4 pairing: 205 valid tokens silently dropped. The `step=256` row is the degenerate case — the entire window is trimmed away and `start_position` (504) runs past `offset` (300).

In all three configs the window stays contiguous and ends at the newest token both before and after, which is why this fails quietly rather than crashing.

## Test

`mlx_vlm/tests/test_cache.py::test_chunked_kv_cache_trims_on_valid_length_not_buffer_width` asserts the window never falls below `chunk_size`, stays contiguous, ends at the newest token, and that `start_position <= offset`. It fails on the current code and passes with this change.

Diff is +43/-5.

---


lucasnewman:
@Anai-Guo Needs signed commits.

lucasnewman (changes requested):
See above

Anai-Guo:
Thanks @lucasnewman — done. This branch is now a single commit signed with GitHub's GPG key (`verified: true`); the tree is byte-identical to before, so the diff is unchanged.

Force-pushing to the fork briefly auto-closed the PR, hence the reopen notification — no content changed.

lucasnewman (approved):
Thanks!

PR #2039, opened 2026-08-26
Fix Qwen4-Exp QSA cache for continuous batching

jonathan308:
## Summary

Preserve Qwen4-Exp QSA auxiliary index keys and text/MRoPE positions when generation converts, merges, extends, filters, or extracts batched caches.

The existing `QSAKVCache` inherits `KVCache`, so `_make_cache()` converted it to a plain `BatchKVCache` and silently discarded QSA indexer state. This adds a batch-aware QSA cache and lets model-defined `to_batch()` take precedence over the generic KV conversion.

## Validation

- `pre-commit run --all-files`: passed
- full CI-equivalent suite: `2637 passed, 7 skipped, 75 subtests passed`
- focused Qwen/generation/cache suite: `148 passed`

New tests cover ragged-row merge/extract, 3-axis MRoPE positions, and generation cache-factory type preservation.

### DIFF
--- a/mlx_vlm/generate/ar.py
+++ b/mlx_vlm/generate/ar.py
@@ -819,7 +819,7 @@ def _make_cache(
     def to_batch_cache(c, quantize=True):
         # Caches that ship their own batch-conversion (e.g. MiniMax M3 sparse
         # index-key side cache) know how to build the correct batch cache.
-        if hasattr(c, "to_batch") and not isinstance(c, cache.KVCache):
+        if hasattr(c, "to_batch"):
             return c.to_batch(left_padding)
         if isinstance(c, cache.KVCache):
             if kv_bits is not None and quantize:
--- a/mlx_vlm/models/qwen4_exp/language.py
+++ b/mlx_vlm/models/qwen4_exp/language.py
@@ -2,13 +2,20 @@ from __future__ import annotations
 
 import math
 from bisect import bisect_right
+from types import SimpleNamespace
 from typing import Any, Optional
 
 import mlx.core as mx
 import mlx.nn as nn
 
 from ..base import LanguageModelOutput
-from ..cache import ArraysCache, KVCache, QuantizedKVCache
+from ..cache import (
+    ArraysCache,
+    BatchKVCache,
+    KVCache,
+    QuantizedKVCache,
+    dynamic_roll,
+)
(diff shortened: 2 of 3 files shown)


lucasnewman:
@jonathan308 Thanks! Can you [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits and force push so we can clear the merge requirements?

jonathan308:
Done, apologies.

Lazarus-931:
conflicts :)

jonathan308:
Resolved , rebased onto #2040, retained both the new Qwen4 MTP imports and the QSA continuous-batch cache imports, and force-pushed signed commit d2bf01f9. Combined validation: 2,644 passed, 7 skipped, plus 75 subtests.
Should be good to go now .

DiscoStew6082:
Thanks for working on this. I independently tested the current PR commit, `d2bf01f`.

The existing Qwen4 tests and the full MLX-VLM test suite pass. However, I also tested the actual use case this PR is intended to enable: generating for two requests at the same time when their prompts have different lengths.

That real two-request test currently crashes during Qwen sparse attention:

```text
ValueError: [broadcast_shapes] Shapes (2) and (12) cannot be broadcast.
```

The problem is that continuous batching gives Qwen a separate cached position for each request, but the sparse-attention calculation still treats that position as one shared number.

I tested a small follow-up fix that:

- Handles the cached position separately for each request.
- Keeps shorter, left-padded prompts correctly aligned.
- Preserves empty batch-cache state safely.
- Adds tests that run two differently sized prompts through both prefill and decode.

The helper commit is [`b0dbba8`](https://github.com/DiscoStew6082/mlx-vlm/commit/b0dbba887cf502f132cb45bafd9e3964131680e9), directly on top of `d2bf01f`.

Validation with that commit:

- 15 focused Qwen4 tests passed.
- Full suite: 2,649 passed, 6 skipped, and 75 subtests passed.
- Pre-commit passed.

Please feel free to cherry-pick it and sign the resulting commit if it is useful.

lucasnewman:
This isn't correct as written, and the change in `ar.py` has a larger impact than it seems. Let's hold off on this one for now and I can look into a proper fix.

lucasnewman:
Resolved in https://github.com/Blaizzy/mlx-vlm/pull/2056

PR #2040, opened 2026-08-26
Add MTP support for Qwen3.8-Flash-Next

lucasnewman:
Nice speedup with `--draft-block-size 4` across two runs with 1024 tokens max. I used a 3-bit quant since that's what I can fit into memory on my M5 Max 128GB.

| Mode | Run 1 | Run 2 | Average | Gen. time | Peak memory |
| MTP inactive | 27.682 tok/s | 27.747 tok/s | **27.715 tok/s** | 36.95s | 89.701 GB |
| MTP active | 43.286 tok/s | 42.353 tok/s | **42.820 tok/s** | 23.91s | 91.192 GB |

- 54.5% higher throughput
- 35.3% less generation time
- 1.49 GB additional peak memory
- 90.7% draft acceptance
- 3.72 accepted tokens per verification round

### DIFF
--- a/mlx_vlm/models/qwen3_5/speculative_verifier.py
+++ b/mlx_vlm/models/qwen3_5/speculative_verifier.py
@@ -1352,6 +1352,14 @@ class Qwen3_5ExactSpeculativeVerifier:
 
         return feed_forward(x)
 
+    @staticmethod
+    def _normalize_gated_delta_qk(layer, q, k):
+        del layer
+        inv_scale = k.shape[-1] ** -0.5
+        q = (inv_scale**2) * mx.fast.rms_norm(q, None, 1e-6)
+        k = inv_scale * mx.fast.rms_norm(k, None, 1e-6)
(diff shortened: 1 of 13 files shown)


Lazarus-931 (approved):
this is nice, thanks @lucasnewman!

pyros-projects:
Those are delightfully non-cosmetic numbers. For an apples-to-apples regression while wiring the embedded head into oMLX, could you share the exact 3-bit target checkpoint / split-drafter setup and prompt or profile used for the 90.7% acceptance run? On an M5 Max 128 GB, an embedded oQ4 checkpoint currently gives 70.9% at native depth 1 (24.47 vs 22.63 tok/s, exact greedy tokens); recurrent depth 4 reaches 69.6% aggregate but only 22.95 vs 22.40 tok/s. I am trying to separate checkpoint/head quantization effects from runtime lifecycle differences before blaming the arithmetic, as is tradition.

PR #2043, opened 2026-08-27
Drop unused DeepSeek-V4 rotating-cache value graphs

### LINKED ISSUE #2042: DeepSeek-V4 retains unused rotating-cache value graphs during decode
## Description

```text
```

## Minimal reproduction

```python
import io

import mlx.core as mx

def graph_edges(array):
    graph = io.StringIO()


for step in range(2_000):
    keys = mx.full(
        (1, 1, 1, 512),
        step % 7,
        dtype=mx.bfloat16,
    )

    mx.eval(cached_keys)

```

## Actual behavior

## Expected behavior

## Related reports

byoan:
Fixes #2042

## Summary

DeepSeek-V4 local attention uses `RotatingKVCache` as a key-only cache. Each update previously supplied a newly created zero-width values array, but attention only consumed the cached keys. The cache therefore retained an unevaluated values update chain that grew once per layer and decode step, eventually exhausting Metal's resident-resource count during long generations.

This PR updates the key-only cache at the point where that unused graph is created. After each rotating-cache update, the values field is rebound to a zero-width view of the cached keys. This preserves the cache's expected key/value shape and state interface while discarding the preceding values chain. The cached keys returned to attention are unchanged.

The helper is used by `LocalAttention`, `CompressedAttention`, and `SparseCompressedAttention`, covering regular and batched rotating caches without adding a model-wide synchronization or changing shared cache behavior for other model families.

## Why this is model-specific

The shared rotating cache supports models that store distinct keys and values. DeepSeek-V4 is different: its local attention uses the same KV representation for both sides of attention and only needs to retain keys. Keeping the chain cut in the DeepSeek-V4 model makes that key-only ownership explicit without weakening the shared cache contract.

Speculative target verification calls the same DeepSeek-V4 attention implementations and is therefore covered by the same update helper. The separate MTP drafter already exposes its cache state to the speculative execution machinery for asynchronous evaluation.

## Tests

The regression coverage verifies that:

- repeated cached forwards keep the unused values graph bounded across local, compressed, and sparse-compressed attention;
- fresh per-step inputs do not make the retained values graph grow with total decode length;
- both `RotatingKVCache` and `BatchRotatingKVCache` are covered;

The full model and speculative decoding suites also verify cached forward behavior, target verification, MTP drafting, and rollback handling.

## Test plan

```text
python -m pytest mlx_vlm/tests/test_models.py -q
```

```text
369 passed, 2 warnings, 55 subtests passed
```

```text
python -m pytest mlx_vlm/tests/test_speculative.py -q
```

```text
190 passed, 2 warnings
```

### DIFF
--- a/mlx_vlm/models/deepseek_v4/language.py
+++ b/mlx_vlm/models/deepseek_v4/language.py
@@ -56,6 +56,13 @@ def _score_func(scores: mx.array, func: str) -> mx.array:
     raise ValueError(f"Unsupported DeepSeek-V4 scoring function: {func}")
 
 
+def _update_key_only_cache(cache: Any, keys: mx.array) -> mx.array:
+    """Update a rotating key-only cache without retaining a values graph."""
+    cached_keys, _ = cache.update_and_fetch(keys, keys[..., :0])
+    cache.values = cache.keys[..., :0]
+    return cached_keys
+
+
 @mx.compile
 def _expert_select(
     logits: mx.array,
@@ -679,7 +686,7 @@ class LocalAttention(nn.Module):
         kv = self.kv_norm(self.wkv(x)).reshape(B, 1, L, self.head_dim)
         kv = self.rope(kv, offset)
         if cache is not None:
-            kv, _ = cache.update_and_fetch(kv, mx.zeros((B, 1, L, 0)))
+            kv = _update_key_only_cache(cache, kv)
         mask = _align_local_mask(mask, kv.shape[2])
 
         out = scaled_dot_product_attention(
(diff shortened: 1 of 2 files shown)


byoan:
(Reworked the solution to fix the core issue rather than bounding the cache frequently)

Lazarus-931:
hi @byoan, thx for the fix. 

I was thinking a general fix in `cache.py` would be better suited, since this might be a issue that could be existant with other models.

w'ed also want this to cover the spec-decode Buffered path

I'll close this and open a sep pr :)

PR #2044, opened 2026-08-27
MTP support for GLM-5.3-Flash

Lazarus-931:
## MTP support for GLM-5.3-Flash

Adds native multi-token-prediction (MTP) self-speculative decoding for GLM-5.3-Flash (`model_type: glm5_next`), plus a few base-model correctness fixes surfaced along the way.

### What's in it
- **Drafter** — the trained nextn layer (at `model.layers.<num_hidden_layers>`) is extracted into a standalone `glm5_next_mtp` drafter (`python -m mlx_vlm.speculative.drafters.glm5_next_mtp.split`) and drives the MTP round loop via `--draft-kind mtp`.
- **Verifier** — `Glm5NextExactSpeculativeVerifier` runs the model over the `[bonus, draft]` block, capturing the pre-norm hidden the nextn head consumes and each KDA layer's recurrent state for rollback.
- **Short-block DSA gather** — the sparse-attention verify gathers each query's top-`index_topk` selected latents (`O(index_topk)`) instead of masking over all keys (`O(context)`); this is what keeps verify affordable at long context. KDA layers verify a block on the shared fused gated-delta kernel and roll the recurrent state back to the accepted prefix on a partial accept.
- **Never-lose gate** — a self-calibrating controller measures drafting vs plain-decode cost once, then runs a plain decode step whenever recent acceptance can't clear the verify overhead, re-probing to resume. Opt-in via the drafter's `adaptive_pause`; every new orchestrator branch is gated behind `getattr`, so other MTP drafters (qwen3_5, deepseek_v4, …) are untouched.
- **Base-model correctness fixes** — includes fixes for all four glm5_next correctness issues in #2053 (independently validated against the transformers reference): (1) mHC `hc_attn`/`hc_ffn` scale/base + KDA `A_log`/`dt_bias` kept fp32 through `convert` (`cast_predicate`) and healed on load (`sanitize`) — the critical one that otherwise produces position-0 token soup; (2) `q_a`/`kv_a` layernorm eps → `rms_norm_eps`; (3) indexer `k_norm` eps → 1e-6; (4) `_fused_in_proj` refuses to fuse the six KDA input-projections when they don't share one quantization. Plus a clamped-SwiGLU text stack and an fp32 router. Happy to coordinate with the #2053 reporter.
- Tests (16 glm5_next cases) + docs.

Final comparison of baseline vs mtp vs dflash2 in #2074


DiscoStew6082:
I tested this with two GLM-5.3 MTP requests running together, with one finishing before the other. The shorter request caused this crash when it left the batch:

```text
AttributeError: 'KVCache' object has no attribute 'filter'
```

The code had created a cache that works for one request but cannot remove a finished request from a batch. The fix uses a batch-aware cache only when there is more than one request. Single-request behavior does not change.

I also added a test that runs through the public `BatchGenerator` API with different prompt lengths and response limits. It makes one request finish first, covers a rejected draft, and confirms that both responses are identical to normal generation.

Validation on the current PR head:

- 16 GLM tests passed
- 187 speculative-decoding tests passed
- 60 oMLX integration tests passed
- Pre-commit formatting checks passed

I kept this to one signed commit based directly on the current PR head. I did not run a full GLM checkpoint, so this is focused correctness evidence rather than a performance claim.

Commit: https://github.com/DiscoStew6082/mlx-vlm/commit/ab8b6dcf80c92c8a5e2b470fa468ecb29d7ab457

To bring it into this branch:

```bash
git fetch https://github.com/DiscoStew6082/mlx-vlm.git disco/glm53-mtp-batch-regression
git cherry-pick ab8b6dcf80c92c8a5e2b470fa468ecb29d7ab457
```

Lazarus-931:
> I tested this with two GLM-5.3 MTP requests running together, with one finishing before the other. The shorter request caused this crash when it left the batch:
> 
> ```
> AttributeError: 'KVCache' object has no attribute 'filter'
> ```
> 
> The code had created a cache that works for one request but cannot remove a finished request from a batch. The fix uses a batch-aware cache only when there is more than one request. Single-request behavior does not change.
> 
> I also added a test that runs through the public `BatchGenerator` API with different prompt lengths and response limits. It makes one request finish first, covers a rejected draft, and confirms that both responses are identical to normal generation.
> 
> Validation on the current PR head:
> 
> * 16 GLM tests passed
> * 187 speculative-decoding tests passed
> * 60 oMLX integration tests passed
> * Pre-commit formatting checks passed
> 
> I kept this to one signed commit based directly on the current PR head. I did not run a full GLM checkpoint, so this is focused correctness evidence rather than a performance claim.
> 
> Commit: [DiscoStew6082@ab8b6dc](https://github.com/DiscoStew6082/mlx-vlm/commit/ab8b6dcf80c92c8a5e2b470fa468ecb29d7ab457)
> 
> To bring it into this branch:
> 
> ```shell
> git fetch https://github.com/DiscoStew6082/mlx-vlm.git disco/glm53-mtp-batch-regression
> git cherry-pick ab8b6dcf80c92c8a5e2b470fa468ecb29d7ab457
> ```

good catch @DiscoStew6082, addressed this in d7d0bf3 :)

pudepiedj:
Thanks for picking these up so quickly.

### Pixel_Values_Video
One more in the same path, which I think is needed for the video half of this to actually work — and it isn't in the current diff.

[models/glm5_next/glm5_next.py](models/glm5_next/glm5_next.py), `Model.get_input_embeddings`:
```
def get_input_embeddings(self, input_ids=None, pixel_values=None, **kwargs):
    inputs_embeds = self.language_model.model.embed_tokens(input_ids)

    if pixel_values is None:
        return InputEmbeddingsFeatures(inputs_embeds=inputs_embeds)
    ...
    video_grid_thw = kwargs.get("video_grid_thw", None)
```
For video the processor emits `pixel_values_videos`, not `pixel_values`, so `pixel_values` is None and this returns text-only embeddings. The video is dropped with no error, the `<|video|>` placeholders stay in the sequence as ordinary token embeddings, and the model confabulates from them. `grep -c pixel_values_videos` on that file returns 0.

With the prompt-path and `do_sample_frames` fixes in this PR, video now reaches the processor and produces correct `pixel_values_videos` and `video_grid_thw` — but still never reaches the vision tower. So the remaining symptom is a fluent, confident, entirely wrong answer rather than an error.

Repro that isolates it from any temporal question — a clip whose frames are all identical. Same pixels, two paths:
```
same frame AS IMAGE -> "7"   correct
same frame AS VIDEO -> "1"   wrong
```
Preprocessing is correct throughout: a 4-second clip gives `video_grid_thw [[4, 26, 36]]` and `pixel_values_videos (3744, 1176)` — exactly 4x26x36 rows.

### Fix:
```
def get_input_embeddings(self, input_ids=None, pixel_values=None, **kwargs):
    if pixel_values is None:
        pixel_values = kwargs.get("pixel_values_videos")
    ...
```
The function already reads video_grid_thw, and merge_input_ids_with_image_features already falls back to video_token_id, so only the pixel key looks unwired. `_gpu_embed` forwards it — it passes every raw input except `input_ids`, `pixel_values` and `attention_mask` through as **data_kwargs.

Verified against a live server on GLM-5.3-Flash-MLX-8bit. Before: identical-frame clip reads 1, a clip cycling digits 3,1,4,2 reads 492357816, a square sliding left to right is "a black mobile phone… rotates". After: 7, 3 1 4 2 in order, and "moves horizontally from left to right".

*Co-authored with Claude Opus 5 (Anthropic).*

Lazarus-931:
> Thanks for picking these up so quickly.
> 
> ### Pixel_Values_Video
> One more in the same path, which I think is needed for the video half of this to actually work — and it isn't in the current diff.
> 
> [models/glm5_next/glm5_next.py](models/glm5_next/glm5_next.py), `Model.get_input_embeddings`:
> 
> ```
> def get_input_embeddings(self, input_ids=None, pixel_values=None, **kwargs):
>     inputs_embeds = self.language_model.model.embed_tokens(input_ids)
> 
>     if pixel_values is None:
>         return InputEmbeddingsFeatures(inputs_embeds=inputs_embeds)
>     ...
>     video_grid_thw = kwargs.get("video_grid_thw", None)
> ```
> 
> For video the processor emits `pixel_values_videos`, not `pixel_values`, so `pixel_values` is None and this returns text-only embeddings. The video is dropped with no error, the `<|video|>` placeholders stay in the sequence as ordinary token embeddings, and the model confabulates from them. `grep -c pixel_values_videos` on that file returns 0.
> 
> With the prompt-path and `do_sample_frames` fixes in this PR, video now reaches the processor and produces correct `pixel_values_videos` and `video_grid_thw` — but still never reaches the vision tower. So the remaining symptom is a fluent, confident, entirely wrong answer rather than an error.
> 
> Repro that isolates it from any temporal question — a clip whose frames are all identical. Same pixels, two paths:
> 
> ```
> same frame AS IMAGE -> "7"   correct
> same frame AS VIDEO -> "1"   wrong
> ```
> 
> Preprocessing is correct throughout: a 4-second clip gives `video_grid_thw [[4, 26, 36]]` and `pixel_values_videos (3744, 1176)` — exactly 4x26x36 rows.
> 
> ### Fix:
> ```
> def get_input_embeddings(self, input_ids=None, pixel_values=None, **kwargs):
>     if pixel_values is None:
>         pixel_values = kwargs.get("pixel_values_videos")
>     ...
> ```
> 
> The function already reads video_grid_thw, and merge_input_ids_with_image_features already falls back to video_token_id, so only the pixel key looks unwired. `_gpu_embed` forwards it — it passes every raw input except `input_ids`, `pixel_values` and `attention_mask` through as **data_kwargs.
> 
> Verified against a live server on GLM-5.3-Flash-MLX-8bit. Before: identical-frame clip reads 1, a clip cycling digits 3,1,4,2 reads 492357816, a square sliding left to right is "a black mobile phone… rotates". After: 7, 3 1 4 2 in order, and "moves horizontally from left to right".
> 
> _Co-authored with Claude Opus 5 (Anthropic)._

thx for catching this @pudepiedj, will add a commit for this :)

aidiffuser:
Follow-up from #2053: we read the current revision's diff, and it addresses all four findings — closing the loop from our side, no reopen needed.

1. **`cast_predicate`** now keeps `.attn_hc.` / `.ffn_hc.` and `A_log` / `dt_bias` in fp32 — the checkpoint-killer, fixed exactly as hoped. The `sanitize` heal-on-load restoring fp32 for already-converted checkpoints goes beyond what we asked and is the right call: it repairs existing artifacts, and in our measurements the bf16-rounded mHC scale/base alone take teacher-forced top-1 from 18/23 to 1/23.
2. **`q_a_layernorm` / `kv_a_layernorm`** at `config.rms_norm_eps` — matches the reference.
3. **`indexer.k_norm`** at `eps=1e-6` — matches the reference.
4. **`_fused_in_proj`** refusing to fuse under mixed quantization, with fallback to the separate projections — exactly the behavior we suggested.

The swiglu clamp (gate capped above, up clamped both sides, on dense + routed + shared) and the fp32 router also match the reference semantics we validated. For what it's worth: we've been running functionally identical fixes in production (2× M3 Ultra, tensor-parallel, contexts through 551k) — this set is what makes converted checkpoints coherent. Glad to see it land upstream.

One heads-up while you're in this file: `rollback_speculative_cache` correctly maintains the indexer `_pool` time invariant (your test asserts `_pool[3] == offset`), but the incremental fast path itself has no continuity guard, so cache rewinds outside the rollback helper (generic `trim`, batch splices under continuous batching) crash with a zero-size argmax. Filed with a repro and a one-line guard in #2084 — it composes cleanly with this PR.

Last: we have a query-chunked generalization of your short-block verify gather to *full prefill* (same gather-the-selected-latents idea at `L` up to thousands; ~2× prefill at 128k in our serving, exact fp32 parity vs the dense path). We'd like to upstream it and will coordinate on top of this PR rather than collide with your `language.py` rewrite.

avlp12:
The documented glm5_next MTP split command (`--model zai-org/GLM-5.3-Flash`, whose only published revision is FP8) produces a drafter that cannot be loaded: `load_weights` fails with 871 leftover `weight_scale_inv` params.

Two mechanisms:

1. the splitter's sanitize path has no FP8 conversion step (unlike qwen3_5's `convert_qwen_fp8_weights`, which is documented and covered by `test_split_qwen3_5_mtp_converts_fine_grained_fp8`), so raw U8 FP8 bytes are written out while per-expert `*_scale_inv` grids are left unstacked (`split.py:63-65` regex anchors on `.weight$`);
2. `quantization_from_source()` then finds no `.scales` keys, so the output config carries no quantization block.

No flag rescues it: `--q-bits` crashes on U8 (`mx.quantize` rejects non-float), `convert --mtp --quantize` silently skips the drafter, and `strict=False` loads FP8 bytes as numbers with no warning.

Suggested direction: an fp8→bf16 (or mxfp8) conversion in the glm5_next splitter sanitize, mirroring qwen3_5; `qwen3_next` and `glm4_moe_lite` appear to carry the same latent gap (the latter's blanket `.astype(bfloat16)` would corrupt silently). README:24-25 also contradicts :59-61 on quantized-source support.

Happy to share receipts; a manual blockwise dequant of the split output loads and reaches ~90% single-draft MTP acceptance (1.12x at ctx 512 on an M3 Ultra, matching this PR's table), so the extraction itself is sound.

Lazarus-931:
| Context | Base prefill | MTP prefill | Base decode | MTP decode | Speedup | Acceptance | Base peak | MTP peak | Tokens |
|--:|--:|--:|--:|--:|--:|--:|--:|--:|:--|
| 512 | 245.08 | 237.70 | 33.19 | 51.83 | 1.56x | 80.61% | 184.39 GB | 184.41 GB | exact |
| 2,048 | 242.72 | 232.53 | 29.47 | 43.13 | 1.46x | 79.59% | 184.80 GB | 185.70 GB | exact |
| 8,192 | 212.44 | 201.88 | 28.95 | 45.16 | 1.56x | 88.04% | 185.68 GB | 196.07 GB | exact |
| 16,384 | 182.63 | 172.56 | 28.68 | 42.48 | 1.48x | 85.11% | 187.28 GB | 225.25 GB | exact |

avlp12:
Thanks for the sweep. One question so I can line it up with what I reported above: which drafter did this table use — one produced by the documented split command from `zai-org/GLM-5.3-Flash` (the FP8 revision), or one split from a bf16 or pre-quantized source?

I ask because the documented path still fails for me exactly as described (871 leftover `weight_scale_inv` params at `load_weights`, no quantization block in the output config), so if your drafter came from that command I'd like to know what differs; and if it came from another source, a line in the README saying so would save the next person the afternoon. The 80–88% acceptance and 1.46–1.56× are consistent with what I got from a manual blockwise dequant of the split output (≈90% single-draft acceptance, 1.12× at ctx 512 on my box). Those results are consistent with the extraction being sound, but they don't by themselves prove it; the specific issue I'm flagging is the FP8 sanitize step.

Sofille65:
On the FP8 splitter question: an alternative to adding an fp8 step to the splitter is to not give the splitter its own sanitize at all. We extract the nextn layer by renaming the `layers.<num_hidden_layers>.*` keys onto a 1-layer model and running the model's own `Model.sanitize` (deepseek_v32 lineage), which already carries MLA absorption, expert stacking and the blockwise `weight_scale_inv` dequant. The drafter then loads strict with no leftover params, and the FP8 handling stays in one place. We ran it from the BF16 release, where the dequant is a no-op, so I can't claim to have exercised the FP8 branch on the head itself. The extracted head is lossless in our serving (independent rollback implementation, 65–75% acceptance on a 6-bit trunk). Happy to share the script if useful.

Lazarus-931:
hi @aidiffuser, I used incoai/GLM-5.3-DFlash2 with target-layer captures [5, 14, 24, 33, 42]

avlp12:
Thank you for merging #2074. One factual question about routing, if you have a moment: #2074 merged into the
`glm53-flash-mtp` branch on this repository (merge commit `6a1bf24e`, 2026-09-02), while this PR targets `main`
from the fork branch `Lazarus-931:glm53-flash-mtp` at `91b51a37` - and GitHub's compare of
`91b51a37...f025d5c8` reports `diverged` (ahead by 10, behind by 44), so #2074's commits (`f025d5c8`,
`800ec86c`) are not in this PR's head. What is the intended route from the `glm53-flash-mtp` staging branch to
`main`? I ask only because I have a PR of my own (#2105) plus two small follow-ups, and which branch they
should target depends on the answer. Happy to wait, and thanks again for the merge.

lucasnewman (approved):

Blaizzy (changes requested):
Please revert the changes in ar.py

mlx_vlm/utils.py
```diff
@@ -2441,6 +2441,7 @@ def prepare_inputs(
         extra = {}
         if has_videos:
             extra["videos"] = videos
+            extra["do_sample_frames"] = False
```
Blaizzy, line 2444: Why hardcode this to false for all models?

mlx_vlm/prompt_utils.py
```diff
@@ -304,6 +305,7 @@ def format_message(
             "diffusion_gemma",
             "minicpmv4_6",
             "minimax_m3_vl",
+            "glm5_next",
```
Blaizzy, line 308: Why?

mlx_vlm/prompt_utils.py
```diff
@@ -965,7 +967,10 @@ def _allocate_media_counts(explicit_counts, total_count):
         if last_user_idx >= 0:
             audio_counts[last_user_idx] = num_audios
 
+        video_arg = kwargs.pop("video", None)
```
Blaizzy, line 970: What's this for?

mlx_vlm/speculative/mtp.py
Blaizzy, line 1: Why do you need this ?

_AdaptivePauseController

mlx_vlm/models/glm5_next/mtp.py
Blaizzy, line 1: Why is this a separate file?

Blaizzy (changes requested):

mlx_vlm/utils.py
pudepiedj: 1. mlx_vlm/utils.py — "Why hardcode this to false for all models?"
Fair objection — this came from my report in [#2071](https://github.com/pudepiedj/GLMKK3/issues/2071) and the scoping is mine to answer for.

The reason it exists: `prepare_inputs` decodes the file itself via `load_video()` at its own fps hint and passes the processor decoded frames, so transformers never sees the file and never builds a VideoMetadata. `Glm5NextVideoProcessor.sample_frames` then refuses to sample per-second without fps and duration, and the request 500s:

ValueError: Asked to sample frames per second but no video metadata was
provided which is required when sampling in Glm5Next.
So the intent is right — the frames have already been sampled, and letting the processor resample an already-sampled sequence is wrong regardless of whether metadata exists. But you're right that hardcoding it for every model is too broad: any processor that expects to do its own sampling now silently gets a different code path.

The cleaner fix is probably to pass the metadata rather than suppress the sampling — `prepare_inputs` already computes `video_fps` a few lines above and currently discards it, so a VideoMetadata carrying fps and frame count is nearly free. Failing that, gating on whether the processor advertises `do_sample_frames` would at least confine it.

Happy to be overruled on which of those you prefer; I'd just flag that reverting it outright puts `glm5_next video` back to a 500.

mlx_vlm/prompt_utils.py
pudepiedj: 2. mlx_vlm/prompt_utils.py:308 — "Why?"
Because this list is a second gate that `MODEL_CONFIG` doesn't feed.

Adding "glm5_next": `MessageFormat.LIST_WITH_IMAGE_FIRST` to `MODEL_CONFIG` gets images working, but video routing is decided separately, by the literal list here inside format_message (qwen2_vl, qwen2_5_vl, … minimax_m3_vl). A model absent from it never reaches `_format_video_message`, whatever its `MODEL_CONFIG` entry says, so video content parts are dropped with no error.

Worth saying the duplication is the underlying problem rather than the missing entry: two lists have to agree for a model to be fully multimodal, and nothing enforces it. If you'd rather derive this from `MODEL_CONFIG` (a `VIDEO_CAPABLE` set, or a flag on the enum) that would close the whole class of bug — [#1321](https://github.com/pudepiedj/GLMKK3/issues/1321) was the same shape for `gemma4_unified`, and #1721 is a third variant of it.

mlx_vlm/prompt_utils.py
pudepiedj: 3. mlx_vlm/prompt_utils.py:970 — "What's this for?"
It stops a system message duplicating the video placeholder.

With any system message present, the video was attached to every message rather than the last user turn:
```
[gMASK]<sop><|system|>Reasoning Effort: Max
<|system|><|begin_of_video|><|video|><|end_of_video|>Answer directly.
<|user|><|begin_of_video|><|video|><|end_of_video|>Describe this video.
```
Two placeholders, one video, so placeholders and features disagree and `process_inputs` raises — with an empty exception message, which is what makes it painful to diagnose:
```
ValueError: Failed to process inputs with error:
```
### Repro:
```
msgs = [{"role": "system", "content": "Answer directly."},
        {"role": "user", "content": "Describe this video."}]
p = apply_chat_template(proc, cfg, msgs, num_images=0, num_audios=0,
                        video=["clip.mp4"])
print(p.count("<|video|>"))     # 2 — expected 1
```
Images are unaffected: `<|image|>` count stays 1 with a system message and with prior turns, so it looked specific to the video path. Reported in [#2071](https://github.com/pudepiedj/GLMKK3/issues/2071).

Practical consequence: any OpenAI-compatible client that sends a system prompt can't use video at all.

Lazarus-931:
closing for #2127

PR #2045, opened 2026-08-27
feat(qwen3.8): add external PLE storage

Thump604:
## Summary

- quantize resident PLE shards with a compatible group-32 layout while retaining the requested bit width and mode
- add an optional read-only, row-addressable PLE store
- validate manifests, row geometry, source ranges, and truncated stores before use

## Why

The PLE table is about 30 GB at Q4. Keeping it outside the ordinary MLX parameter tree reduces active model memory to 71.68 GB while preserving exact Q4 row parity.

On an M2 Ultra / 128 GB, a matched 21,646-token request reached first content in 49.35 seconds, decoded at 19.04 tok/s, and completed in 68.47 seconds. The Qwen 27B comparison reached first content in 124.37 seconds, decoded at 15.74 tok/s, and completed in 146.74 seconds. Flash peaked at 78.46 GB Metal memory for this run.

A varied 4K prefill reached 217.8 tok/s; 65,545 cold PLE misses spent 9.68 seconds in lookup, so cold/random access remains the main optimization target.

## Validation

- 415 passed, 2 skipped, 7 subtests on current `main`
- 81 focused tests passed after review fixes
- 32 sampled and boundary rows matched the resident Q4 representation exactly
- text generation and thinking/instruct modes completed with the compact artifact
- corrupt, truncated, mismatched, and disconnected stores fail closed

## Scope

The external store is opt-in and model-owned. This does not change default loading or introduce a public serving API. Official FP8 conversion is split into #2058.

MTP is not qualified by this PR. Runtime testing found the current registered drafter slower and unreliable for strict terminal tool calls; both failed cases recovered with MTP disabled. Continuous batching, MTP, and representative quality evaluation remain separate work.


Thump604:
@janhilgard @waybarrios — review requested alongside waybarrios/vllm-mlx#735. The branch is based on #2040; this PR adds the missing conversion and external-PLE layer.

pyros-projects:
One compatibility edge from a real existing Q4 artifact that is directly relevant to `prepare_external_ple_model`: `Vontra/Qwen3.8-Flash-Next-MLX-4bit` has exactly the sanitized affine-Q4/group-32 PLE layout this PR indexes, but its older `mlx-vlm 0.6.3` conversion also stores Qwen4's zero-centered RMSNorm parameters as ordinary MLX multipliers (`raw + 1`). Current `Qwen4ExpRMSNorm` adds 1 again, so current main loads it but deterministic text is corrupt.

I isolated this with a matched A/B on M5 Max / 128 GB. Current main produced junk for both an exact-copy prompt and `2+2`; translating only the `Qwen4ExpRMSNorm`-backed tensors back by `-1` produced exact `QWEN_TEXT_OK` and `4`. GDN's gated norm must stay untouched.

This PR's PLE storage itself looks like the stronger basis than another bespoke mmap format (especially the batched gathers and fail-closed manifest), but a prepared view of an existing artifact should probably preserve or explicitly mark its norm convention. A conversion marker for new artifacts plus a conservative legacy bridge would keep the storage path from faithfully accelerating already-corrupt activations. Tiny demons, now row-addressable.

Detailed oMLX reproduction: jundot/omlx#3169 (latest comment).

Thump604:
Addressed in 36154d7f. New Qwen4 conversions record the zero-centered convention; external-PLE preparation now fails closed on unmarked sources unless the convention is explicit; legacy `mlx_multiplier` weights avoid the second offset. GDN gated norms are unchanged. Added focused coverage for current, legacy, ambiguous, and gated cases. Thanks for the concrete repro.

Thump604:
@janhilgard @waybarrios — ready for review. Latest Runtime isolation: 21.6K TTFT 49.35s vs 124.37s on Qwen 27B; decode 19.04 vs 15.74 tok/s. MTP is explicitly out of scope and remains disabled after the current drafter regressed speed and strict tool reliability.

lucasnewman:
> One compatibility edge from a real existing Q4 artifact that is directly relevant to `prepare_external_ple_model`: `Vontra/Qwen3.8-Flash-Next-MLX-4bit` has exactly the sanitized affine-Q4/group-32 PLE layout this PR indexes, but its older `mlx-vlm 0.6.3` conversion also stores Qwen4's zero-centered RMSNorm parameters as ordinary MLX multipliers (`raw + 1`). Current `Qwen4ExpRMSNorm` adds 1 again, so current main loads it but deterministic text is corrupt.
> 
> I isolated this with a matched A/B on M5 Max / 128 GB. Current main produced junk for both an exact-copy prompt and `2+2`; translating only the `Qwen4ExpRMSNorm`-backed tensors back by `-1` produced exact `QWEN_TEXT_OK` and `4`. GDN's gated norm must stay untouched.

That user should just re-convert the weights, not hack around this in the code.

mlx_vlm/models/qwen4_exp/config.py
```diff
@@ -53,6 +53,8 @@ class TextConfig(BaseModelConfig):
     make_ngram_vocab_size_divisible_by: int = 128
     split_ngram_parts: int = 128
     seed: int = 1234
+    rms_norm_weight_convention: str = "zero_centered"
```
lucasnewman: Please remove this, we don't need to support bad conversions.

lucasnewman (changes requested):
I'll take a look at this once you remove the RMSNorm changes.

mlx_vlm/models/qwen4_exp/config.py
Thump604: Removed in dac41744. The PR now leaves legacy conversions to reconversion and keeps the external-PLE path scoped to current weights.

mlx_vlm/models/qwen4_exp/fp8.py
```diff
@@ -0,0 +1,76 @@
...
+    """Restore FP8 experts/PLE and pack experts into mlx-vlm's model layout."""
+    converted = dict(weights)
+    expert_groups = defaultdict(lambda: defaultdict(dict))
+    for key in list(converted):
+        match = _EXPERT_WEIGHT_RE.match(key)
+        if match is None:
+            continue
+        prefix, expert, projection = match.groups()
+        scale_key = f"{key}_scale_inv"
+        if scale_key not in converted:
```
lucasnewman, line 36: This function is running unconditionally for all checkpoints. I didn't test it but I assume this breaks loading for (some) non-FP8 checkpoints?

mlx_vlm/models/qwen4_exp/ple_storage.py
```diff
@@ -0,0 +1,543 @@
...
+
+        unique_ids, inverse = np.unique(flat_ids, return_inverse=True)
+        packed_np, scales_np, biases_np = self._read_rows(unique_ids)
+        duplicate_count = flat_ids.size - unique_ids.size
+        self._hits += duplicate_count
+        packed = mx.array(packed_np, dtype=mx.uint32)
+        scales = mx.array(_bfloat16_to_float32(scales_np)).astype(mx.bfloat16)
+        biases = mx.array(_bfloat16_to_float32(biases_np)).astype(mx.bfloat16)
+        values = mx.dequantize(packed, scales, biases, group_size=32, bits=4)
+        if duplicate_count:
```
lucasnewman: It looks like if there are no duplicates then the inverse isn't applied and the ids are ordered incorrectly after `np.unique`?

mlx_vlm/models/qwen4_exp/language.py
```diff
@@ -1269,3 +1289,14 @@ def make_cache(self):
             else:
                 caches.append(QSAKVCache())
         return caches
+
+    @property
+    def quant_predicate(self):
+        def predicate(path, _module):
+            if ".ple.ple_embedding.ngram_embedding.shards." in path:
+                return {"group_size": 32, "bits": 4, "mode": "affine"}
```
lucasnewman: This forces 4-bit quantization regardless of the requested quantization, which isn't correct.

lucasnewman (changes requested):
Can you split the FP8 conversion and the PLE offloading into two different PRs? They are distinct features.

mlx_vlm/models/qwen4_exp/fp8.py
Thump604: Removed from this PR and split into #2058. That PR now detects the official FP8 layout before converting and includes a non-FP8 no-op test.

mlx_vlm/models/qwen4_exp/ple_storage.py
Thump604: Fixed in a7a9fc55. The inverse is now always applied, with a regression test for unsorted unique IDs.

mlx_vlm/models/qwen4_exp/language.py
Thump604: Fixed in a7a9fc55. PLE now overrides only group size 32 and inherits the requested bits and mode.

Thump604:
Split complete: #2045 is PLE-only, and official FP8 conversion is now #2058.

mlx_vlm/models/qwen4_exp/ple_storage.py
```diff
@@ -0,0 +1,545 @@
...
+                "row_count": row_count,
+                "weight": weight,
+                "scales": scales,
+                "biases": biases,
+            }
+        )
+        row_start += row_count
+    manifest = {
+        "version": QuantizedMMapNGramEmbedding.MANIFEST_VERSION,
+        "source_root": str(model_path),
```
lucasnewman: `model_path` is an absolute path here since you used `.resolve()` -- we should use model-relative paths for the manifest.

mlx_vlm/models/qwen4_exp/ple_storage.py
```diff
@@ -0,0 +1,545 @@
...
+            self._lookups,
+            self._rows,
+            self._hits,
+            self._misses,
+            self._bytes,
+            self._elapsed,
+        )
+
+
+def build_quantized_ple_manifest(model_path, output_path, *, cache_rows=0):
```
lucasnewman, line 360: This function should validate that the source tensors are actually 4-bit.

mlx_vlm/models/qwen4_exp/ple_storage.py
```diff
@@ -0,0 +1,545 @@
...
+        raise ValueError("chunk_rows must be positive")
+    range_manifest = build_quantized_ple_manifest(
+        model_path, manifest_path, cache_rows=cache_rows
+    )
+    row_width = int(range_manifest["row_width"])
+    packed_bytes = row_width // 2
+    aux_bytes = row_width // 32 * 2
+    row_stride = packed_bytes + 2 * aux_bytes
+    data_path = manifest_path.parent / data_name
+    if data_path.exists():
```
lucasnewman: We should validate this earlier before doing any other writes, e.g. `build_quantized_ple_manifest`.

mlx_vlm/models/qwen4_exp/ple_storage.py
Thump604: Fixed in 274c321b. Both range and interleaved manifests now record the source root relative to the manifest directory.

mlx_vlm/models/qwen4_exp/ple_storage.py
Thump604: Fixed in 274c321b. Manifest construction now verifies Q4/group-32 metadata plus packed-weight, scale, bias dtype and shape layout before writing.

mlx_vlm/models/qwen4_exp/ple_storage.py
Thump604: Fixed in 274c321b. Existing row-store collisions are now rejected before manifest construction can rewrite anything, with regression coverage.

lucasnewman:
@Thump604 If you can resolve the merge conflicts and verify this will work for nvfp4 weights and that the resulting paths in the manifest are model relative we can merge it. Thanks!

Thump604:
Resolved in `843ee6f0` and rebased on current `main`.

External PLE now supports native MLX NVFP4 end to end: group-16 packed weights, byte scales, no bias tensor, for both safetensors ranges and interleaved row stores. Affine Q4/group-32 remains supported.

Generated paths are model-relative: `source_root` is relative to the manifest, and the model config stores `ple-store.json`; loading resolves it from the model directory.

Focused tests: 98 passed. Broader non-smoke suite: 2,663 passed, 4 skipped. CI is running.

lucasnewman (approved):
Thanks!

Lazarus-931 (approved):

PR #2047, opened 2026-08-27
Fix key-only KV cache value-graph leak at the cache layer

### LINKED ISSUE #2042: DeepSeek-V4 retains unused rotating-cache value graphs during decode

Lazarus-931:
Collapse the retained values graph on zero-width key-only cache updates so long generations stay bounded.

Fixes #2042, originally #2043


byoan:
I tested 80fd3739 locally (since the branch is out of date) and found a remaining gap around cache transformations:

- The inherited RotatingKVCache.merge() constructs the original mlx-lm BatchRotatingKVCache, bypassing the new guard
- Likewise, BatchRotatingKVCache.extract() returns the original RotatingKVCache
- APC warm merging also constructs the mlx-lm classes directly

The directly constructed guarded caches work correctly. However, subsequent updates after those transformations no longer run the guard

I reproduced this with tiny DeepSeek forward passes: the retained values graph stays at 2 edges after both 16 and 64 updates with fresh caches, but grows from 119 to 455 edges after APC warm merging or extraction. This also reproduces with a two-row warm batch. Your added tests pass; these paths aren’t exercised by them.

Would you be open to a small PR? I could preserve the guarded classes through these paths, cover the currently unguarded BatchKVCache, and add the corresponding regression tests
Let me know :)

Lazarus-931:
hi @byoan, thx for catching that!

no new pr is needed, will commit a coverage for `RotatingKVCache` with tests as well :)

byoan:
I retested `d61b01ee` locally, and the previously reported gap is resolved

Here are the results from tiny DeepSeek forward passes against the updated implementation. The numbers represent retained values-graph edges:

| Cache path | Before: 16 → 64 updates | Updated: 16 → 64 updates |
|---|---:|---:|
| Fresh single-request cache | 2 → 2 | 2 → 2 |
| Fresh two-row batch | 2 → 2 | 2 → 2 |
| APC warm merge, single row | 119 → 455 | 2 → 2 |
| APC warm merge, two unequal-length rows | 119 → 455 | 2 → 2 |
| Row extracted from a guarded batch | 119 → 455 | 2 → 2 |

All **33 targeted checks pass**, including the previously failing merge/extract, `BatchKVCache`, and APC paths. Another **301 repository tests and 7 subtests pass** across cache, APC, and speculative decoding

I’m opening a small follow-up PR with these non-blocking improvements:

| Improvement | Additional test coverage |
|---|---|
| Preserve method metadata with `functools.wraps` | Verify that wrapping preserves the original method’s metadata and exposes `__wrapped__` |
| Explicit `BatchKVCache` coverage | Check bounded graph growth after direct construction, empty/populated merges, and row extraction |
| Unequal prefix lengths | Exercise APC merging with different row lengths, checking padding, offsets, cached contents, and bounded graph growth during subsequent decoding |
| Stronger nonempty-values assertions | Use distinct keys and values and compare actual cached/returned contents against expected results |

These are maintainability and regression-coverage improvements, not remaining correctness blockers. The updated implementation already passes the stronger local cache checks; the follow-up will preserve that coverage in the repository

Thanks!

lucasnewman (approved):

byoan:
Hey @Lazarus-931, just reminding you about https://github.com/Lazarus-931/mlx-vlm/pull/1 to improve the tests coverage on this PR before merging it.
Thanks :)

Blaizzy:
Closing it in favour of a issue with reproducible example

byoan:
Hey @Blaizzy, I can still reproduce the bug with the code provided in the linked issue, unsure about the closing here?

PR #2049, opened 2026-08-27
fix(apc): exact-mode restore for integer state tensors and unmergeable cache subclasses

### LINKED ISSUE #2048: Exact-mode APC silently serves nothing for hybrid caches with integer state or KVCache subclasses (qwen4_exp): hits counted, 0 tokens restored
## Summary

Exact-mode APC silently serves nothing for hybrid models whose caches carry **integer state tensors** or **model-specific KVCache subclasses** — e.g. `qwen4_exp` (Qwen3.8-Flash-Next). Lookups hit, `exact_hits`/`matched_tokens` climb, but every request still pays full prefill. There is no error or warning; `served`/warm restore is just a no-op. Same failure *class* as #1443 (silent APC no-op), but in the exact/checkpoint path rather than the block path, and it persists after #1463.

For agent workloads on this model the cost is severe: a 32–38k-token conversation re-prefills from scratch every turn (~50–65 s TTFT at ~650–780 tok/s prefill), even though a matching exact snapshot exists.

## Environment

- mlx-vlm @ e82d557 (current `main`)
- macOS, Apple Silicon (M-series, 256 GB), Python 3.13
- Model: `Qwen3.8-Flash-Next` oQ6 (`qwen4_exp`: 36× DeltaNet `ArraysCache` + 12× `QSAKVCache`), served b=1 (`--max-num-seqs 1`), float KV
- `APC_ENABLED=1`, disk tier active, with and without `--draft-kind mtp` (both affected)

## Observed stats (production server, ~80 requests)

```json
{"lookups_hit": 80, "matched_tokens": 1740550, "served_tokens": 0,
 "exact_hits": 80, "exact_stores": 164, "disk_hits": 0, "disk_writes": 164}
```

`cached_tokens` in `usage.prompt_tokens_details` is `0` on every response; "Prefill completed" logs show `cached_tokens=0` and full-length prefill on every request.

## Root causes (two independent bugs)

**Bug 1 — disk restore: `_safetensors_dtype_info` only maps BF16/F16/F32.**
`qwen4_exp`'s DeltaNet `ArraysCache` state includes an **int64** tensor. When an exact snapshot is loaded back, `_read_safetensors_tensor` returns `None` for that slot, `_load_exact_cache_file` returns `None`, and the whole disk restore silently fails. (The write side serializes int64 fine — only the bespoke reader can't decode it.)

**Bug 2 — in-memory restore: single-row batch merge fails for cache subclasses.**
`make_warm_batch_exact_cache_multi` force-merges even a *single* row into batch-shaped caches. `merge_cache_entries` matches plain `KVCache` by exact `type()` and has no adapter for subclasses such as `QSAKVCache` (which carries the QSA indexer's unprojected keys/positions), so the merge returns `None` and the warm hit is silently dropped → cold prefill. This is the exact-mode analogue of the `_idx`-vs-`offset` harvest gap fixed for block mode in #1463.

## Repro

1. Serve any `qwen4_exp` model with `APC_ENABLED=1` (disk tier optional), b=1.
2. Send a long prompt (≳ 1k tokens), then send it again with a short appended tail.
3. Second request: `usage.prompt_tokens_details.cached_tokens == 0`, wall-clock prefill identical to cold, while `/v1/cache/stats` shows `exact_hits` incrementing.

Minimal unit-level repros are included as regression tests in the linked PR:
- `ArraysCache` with an int64/int32 slot fails disk exact roundtrip (bug 1)
- a `KVCache` subclass row returns `None` from `make_warm_batch_exact_cache_multi` (bug 2)

## Fix

PR incoming from `fix/apc-exact-restore-int-dtypes-b1`:
1. Add integer/bool dtypes to `_safetensors_dtype_info`.
2. Single-row fallback in `make_warm_batch_exact_cache_multi`: when the batch merge fails and there is exactly one float-KV row, serve the snapshot caches directly — they are the same classes a cold b=1 prefill uses (and what MTP speculative decoding already runs against). Mergeable types keep the batch-shaped path so warm rows can still join live batches; quantized joins stay excluded.

## Measured result (after fix)

Qwen3.8-Flash-Next-oQ6, b=1, **MTP draft enabled**:

| request | prompt tokens | cached | wall clock |
|---|---:|---:|---:|
| cold | 9740 | 0 | 16.2 s |
| warm (same prefix + tail) | 9763 | 9740 | **0.32 s** |

APC + MTP now compose in exact mode: warm TTFT ~50× faster while keeping the MTP decode speedup.

Note: exact-mode restores don't increment `served_tokens` (only the block path does), which made this harder to notice — worth a follow-up if you want that counter wired.

bushnelb:
Fixes #2048

Two independent bugs made exact-mode APC a silent no-op for hybrid models whose caches carry integer state tensors or model-specific `KVCache` subclasses (e.g. `qwen4_exp` / Qwen3.8-Flash-Next): lookups hit and `exact_hits`/`matched_tokens` climbed, but every request re-prefilled from scratch. Same failure class as #1443/#1463, but in the exact/checkpoint path.

## Change 1: integer/bool dtypes in `_safetensors_dtype_info`

`qwen4_exp`'s DeltaNet `ArraysCache` state includes an int64 tensor. The bespoke snapshot reader only mapped `BF16/F16/F32`, so `_read_safetensors_tensor` returned `None` for that slot and `_load_exact_cache_file` dropped the entire snapshot — every exact **disk** restore failed silently. The write side already serialized these tensors correctly; only the reader was missing the dtypes (`I64/I32/I16/I8/U8/U32/U64/BOOL`).

## Change 2: single-row fallback in `make_warm_batch_exact_cache_multi`

`merge_cache_entries` matches plain `KVCache` by exact `type()` and has no adapter for subclasses like `QSAKVCache` (which carries QSA indexer state), so batch-merging returned `None` and the warm hit was dropped — every exact **in-memory** restore failed silently for these models.

With exactly one row there is nothing to merge: the snapshot caches are precisely the classes a cold b=1 prefill would use (and what speculative/MTP decoding already runs against), so on merge failure the single row is now served directly. Constraints preserved:

- Mergeable types keep the existing batch-shaped path unchanged, so warm rows can still join live batches (`TestExactHybridColdStaggeredJoin` still passes).
- Restricted to `kv_quant_config is None`: float snapshots never join a quantized live batch.
- Multi-row merges of unmergeable subclasses still return `None` (cold), as before.

## Tests

- `test_exact_cache_disk_restore_preserves_integer_states` — int64/int32 `ArraysCache` slots roundtrip through the disk exact tier (fails on `main`: restore returns `None`).
- `test_make_warm_batch_exact_cache_multi_single_row_preserves_subclass` — a `KVCache` subclass with auxiliary state survives single-row warm materialization with type, offset, and aux state intact (fails on `main`: returns `None`).

All existing APC suites pass: `test_apc.py`, `test_apc_adapters.py`, `test_apc_exact_mode.py`, `test_apc_component_adapters.py`, `test_apc_observability.py`, `test_apc_quantized.py`, `test_apc_semantic_key.py` — 205 passed, 1 skipped.

## Measured result

Qwen3.8-Flash-Next-oQ6 (`qwen4_exp`), Apple Silicon 256 GB, b=1, **MTP draft enabled**, live server:

| request | prompt tokens | cached | wall clock |
|---|---:|---:|---:|
| cold | 9740 | 0 | 16.2 s |
| warm (same prefix + tail) | 9763 | 9740 | **0.32 s** |

Exact-mode APC and MTP speculative decoding now compose: ~50× faster warm TTFT while keeping MTP's decode speedup. On agent-style workloads with 32–38k-token growing contexts this removes a 50–65 s re-prefill from every turn.

### DIFF
--- a/mlx_vlm/apc.py
+++ b/mlx_vlm/apc.py
@@ -695,6 +695,14 @@ def _safetensors_dtype_info(dtype: str):
     mapping = {
         "F16": (np.dtype("<f2"), mx.float16, None),
         "F32": (np.dtype("<f4"), mx.float32, None),
+        "I64": (np.dtype("<i8"), mx.int64, None),
+        "I32": (np.dtype("<i4"), mx.int32, None),
+        "I16": (np.dtype("<i2"), mx.int16, None),
+        "I8": (np.dtype("i1"), mx.int8, None),
+        "U8": (np.dtype("u1"), mx.uint8, None),
+        "U32": (np.dtype("<u4"), mx.uint32, None),
+        "U64": (np.dtype("<u8"), mx.uint64, None),
+        "BOOL": (np.dtype("?"), mx.bool_, None),
     }
     return mapping.get(dtype)
 
@@ -3954,6 +3962,22 @@ def make_warm_batch_exact_cache_multi(
             prefix_lens,
         )
         if merged is None:
+            # Single-row fallback: a model-specific cache subclass (e.g.
+            # qwen4_exp's QSAKVCache with indexer state) has no batch merge
+            # adapter. With one row there is nothing to merge — the snapshot
+            # caches are exactly what a cold b=1 prefill would use (and what
+            # speculative/MTP decoding already runs against), so serve them
+            # directly instead of silently dropping the warm hit. Restricted
+            # to float KV (no quant policy) since exact snapshots are stored
+            # in float and could not join a quantized live batch.
+            if len(row_caches) == 1 and kv_quant_config is None:
+                single = list(row_caches[0])
+                eval_targets: List[mx.array] = []
+                for c in single:
+                    _collect_mx_arrays(c.state, eval_targets)
+                if eval_targets:
+                    mx.eval(eval_targets)
+                return single, int(prefix_lens[0])
             return None, 0
         out.append(merged)
 
--- a/mlx_vlm/tests/test_apc.py
+++ b/mlx_vlm/tests/test_apc.py
@@ -1242,3 +1242,102 @@ def test_positive_desired_prefix_is_unchanged():
     tokens = list(range(1, 100))
 
     assert apc_module.adjust_prefix_to_text_suffix_boundary(tokens, 40, []) == 40
+
+
+def test_exact_cache_disk_restore_preserves_integer_states(tmp_path, monkeypatch):
+    """Exact snapshots with integer state tensors must roundtrip via disk.
+
+    Hybrid models keep non-KV per-layer state in ``ArraysCache`` slots; for
+    qwen4_exp (Qwen3.8-Flash-Next) one DeltaNet slot is an int64 tensor. The
+    bespoke safetensors reader previously only mapped BF16/F16/F32, so
+    ``_read_safetensors_tensor`` returned None for the int64 slot and the
+    whole exact snapshot silently failed to load (lookup hit, restore no-op,
+    full re-prefill on every request).
+    """
+    from mlx_vlm.models.cache import ArraysCache, KVCache
+
+    monkeypatch.setenv("APC_EXACT_CACHE_ENTRIES", "0")
+
+    token_ids = list(range(40))
+    kv = KVCache()
+    kv.keys = mx.ones((1, 1, len(token_ids), 2))
+    kv.values = mx.ones((1, 1, len(token_ids), 2)) * 2
+    kv.offset = len(token_ids)
+    arrays = ArraysCache(size=4)
+    arrays[0] = mx.ones((1, 3, 4), dtype=mx.bfloat16)
+    arrays[1] = mx.ones((1, 2, 3)) * 3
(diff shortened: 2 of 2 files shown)


DiscoStew6082:
Thanks for working on this. I think this PR addresses two separate problems:

1. Restoring integer and boolean tensors from disk, mainly for `ArraysCache` state.
2. Restoring model-specific cache subclasses such as `QSAKVCache` from the in-memory exact cache when running a single sequence.

The in-memory fix appears to work, but I don't think the disk path preserves `QSAKVCache` yet. When it is written to disk, it is handled as a regular `KVCache`, so only the base key/value state is saved. After closing and reopening `APCManager` with the memory tier disabled, it comes back as `KVCache`, and the QSA indexer state is missing.

I reproduced this with a focused close-and-reopen test and have a small candidate patch that preserves the QSA-specific state. I can share both here if they would be useful.

lucasnewman:
We've added proper batch support in main and disk caching will land in https://github.com/Blaizzy/mlx-vlm/pull/2059

PR #2058, opened 2026-08-27
fix(qwen3.8): support official FP8 conversion

Thump604:
## Summary

- convert the official Qwen3.8 FP8 expert and PLE layouts during loading
- leave non-FP8 checkpoints unchanged
- validate missing scales and incomplete expert groups explicitly

## Validation

- `mlx_vlm/tests/test_qwen4_exp.py`: 10 passed

## Scope

This is the FP8 conversion split requested on #2045. It does not include external PLE storage or quantization-policy changes.


lucasnewman:
@Thump604 Thanks, this is looking better. Can you add support for FP8 MTP weights in the splitter? They can probably use the same conversion helper.

Thump604:
Added in 507f6b0c. The Qwen3.8 MTP splitter now runs selected tensors through the same FP8 conversion helper before drafter sanitization. Added coverage for official per-expert FP8 weights and scale removal; 18 focused tests pass.

lucasnewman (approved):
Thanks!

PR #2064, opened 2026-08-27
Fix tool call parsing with quotes and nested values

lucasnewman:
The regex used to parse tool calls was insufficient for handling things like commands inside of quoted strings and nested lists & dictionaries. This updates it to actually parse the AST so we can reliably handle a wider range of tool call formats.


Blaizzy (approved):
LGTM!

Issue #2061, opened 2026-08-27
phi3 crashes with quantized KV cache: head dim not divisible by group size


Lazarus-931:
**Title:** phi3 crashes with quantized KV cache: last-dimension not divisible

**What happens:** Generating with phi3 and a quantized KV cache raises
`ValueError: [quantize] The last dimension of the matrix needs to be
divisible by the quantization group size`. phi3's head dimension is not a
multiple of the default KV quantization group size.

**Repro:**
```python
from mlx_vlm import load, stream_generate
m, p = load("mlx-community/Phi-3-mini-4k-instruct-4bit")
for _ in stream_generate(m, p, "Describe photosynthesis. " * 40,
                         max_tokens=8, kv_bits=4, quantized_kv_start=0,
                         kv_group_size=64):
    pass
```

**Note:** distinct from the gemma2 crash — this is a group-size/head-dim
mismatch in the quantizer, not an `expand_dims` on a tuple.

PR #2065, opened 2026-08-27
fix(attention): materialize a quantized KV state before hand-rolled attention (#2060)

### LINKED ISSUE #2060: gemma2 crashes with quantized KV cache when quantized_kv_start=0
**Title:** gemma2 crashes with quantized KV cache when `quantized_kv_start=0`

**What happens:** Generating with gemma2 and a quantized KV cache started from
the first token raises `TypeError: expand_dims(): incompatible function
arguments`. The GQA path calls `mx.expand_dims` on the cache's returned
key/value state, but with quantization the state is a packed tuple, not an
array.

**Repro:**
```python
from mlx_vlm import load, stream_generate
m, p = load("mlx-community/gemma-2-2b-it-4bit")
for _ in stream_generate(m, p, "Describe photosynthesis. " * 40,
                         max_tokens=8, kv_bits=4, quantized_kv_start=0):
    pass
```

**Where:** `mlx_vlm/models/gemma2/language.py`, the `repeats > 1` branch:
`keys = mx.expand_dims(keys, 2)` on a quantized-cache tuple.

**Also affects:** `phi3small` shares the same pattern.

Anai-Guo:
Fixes #2060.

## The defect

`gemma2` and `phi3small` are the only two models in the repo that build their
attention scores by hand rather than going through
`base.scaled_dot_product_attention`:

- `gemma2` because `attn_logit_softcapping` has to be applied *between* the two
  matmuls, which `mx.fast.scaled_dot_product_attention` cannot express;
- `phi3small` because its block-sparse layers add a per-block mask.

`QuantizedKVCache.update_and_fetch` does not return arrays — it returns
`(packed, scales, biases)` tuples:

```python
return tree_map(lambda x: x[..., : self.offset, :], (self.keys, self.values))
```

So both paths hand a tuple to `mx.expand_dims` and die:

```
TypeError: expand_dims(): incompatible function arguments
```

`maybe_quantize_kv_cache` swaps `KVCache` for `QuantizedKVCache` without
consulting the model, so any `kv_bits=...` run on these two architectures fails
once `quantized_kv_start` is reached — `quantized_kv_start=0` fails on the
first token. These are the only two `mx.expand_dims(keys, ...)` call sites in
`mlx_vlm/models`, so the blast radius is exactly these two files.

## The fix

`base.py` already carries a small family of helpers that exist so model code can
cope with non-array cache states (`kv_sequence_length`, `slice_kv_sequence`,
`align_attention_mask_to_scores`). This adds one more in the same shape:

```python
def dequantize_kv_state(cache, keys, values):
    if isinstance(keys, mx.array) or hasattr(keys, "shape"):
        return keys, values
    if type(keys) in (tuple, list):
        return (mx.dequantize(*keys, group_size=cache.group_size, bits=cache.bits), ...)
    return cache.dequantize(keys, values)
```

The dispatch deliberately mirrors `kv_sequence_length`: `type(x) in (tuple, list)`
rather than `isinstance`, so TurboQuant's `NamedTuple` codec states fall through
to `cache.dequantize()` instead of being mistaken for a uniform
`(packed, scales, biases)` triple. `HybridQuantKVCache` already returns
dequantized arrays from `update_and_fetch`, so it takes the pass-through arm.
Components are run through `mx.contiguous` first, matching
`cache._dequantize_uniform`, because the state comes out of a
`[..., :offset, :]` slice.

Only the two hand-rolled paths call it. Everything that goes through
`scaled_dot_product_attention` still receives the packed state and still uses
`quantized_matmul`, so no other model changes behaviour and no unquantized path
gains work — `dequantize_kv_state` returns its arguments unchanged when the
cache is unquantized or absent.

## Tests

`TestQuantizedCacheInHandRolledAttention` in
`mlx_vlm/tests/test_kv_cache_quantization.py` drives both `Attention` modules
directly, parameterized over `kv_bits ∈ {None, 4}`, asserting the output shape
and that no `NaN` escapes. Both new cases fail on `main` with the `expand_dims`
`TypeError` and pass with this change; the `kv_bits=None` cases pin the
unquantized path against regression. The `phi3small` case asserts
`attention.block_sparse` so it cannot silently drift onto the dense path and
stop covering the fix.

Formatted with the pinned `black` 26.3.1 / `isort` 5.13.2 from
`.pre-commit-config.yaml`.

## Out of scope

#2061 (`phi3` head dim not divisible by the KV quantization group size) is a
different failure — a group-size constraint in the quantizer, not a tuple
reaching array-only ops — and is not addressed here.

### DIFF
--- a/mlx_vlm/models/base.py
+++ b/mlx_vlm/models/base.py
@@ -225,6 +225,36 @@ def slice_kv_sequence(state, end: int):
     return state[:, :, :end, :]
 
 
+def dequantize_kv_state(cache, keys, values):
+    """Return plain ``[B, H, S, D]`` arrays for a state ``update_and_fetch`` returned.
+
+    Unquantized caches already hand back arrays and are passed through. Uniform
+    quantized caches return ``(packed, scales, biases)`` tuples and TurboQuant
+    caches return codec-state named tuples, neither of which can be indexed,
+    expanded or matmul'd like an array. Attention paths that cannot go through
+    :func:`scaled_dot_product_attention` -- logit softcapping, block-sparse
+    masking -- need real arrays, so they materialize the state here first.
+    """
+    if isinstance(keys, mx.array) or hasattr(keys, "shape"):
+        return keys, values
+    # Plain tuple/list only: TurboQuant states are NamedTuples and dequantize
+    # through their own cache method.
+    if type(keys) in (tuple, list):
+        return (
+            mx.dequantize(
+                *(mx.contiguous(x) for x in keys),
+                group_size=cache.group_size,
+                bits=cache.bits,
+            ),
+            mx.dequantize(
+                *(mx.contiguous(x) for x in values),
+                group_size=cache.group_size,
+                bits=cache.bits,
+            ),
+        )
+    return cache.dequantize(keys, values)
+
+
 def create_attention_mask(
     h, cache=None, window_size: Optional[int] = None, return_array: bool = False
 ):
--- a/mlx_vlm/models/gemma2/language.py
+++ b/mlx_vlm/models/gemma2/language.py
@@ -3,7 +3,7 @@ from typing import Any, Optional
 import mlx.core as mx
 import mlx.nn as nn
 
-from ..base import LanguageModelOutput, create_attention_mask
+from ..base import LanguageModelOutput, create_attention_mask, dequantize_kv_state
 from .config import ModelConfig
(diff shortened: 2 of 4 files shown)


Lazarus-931 (approved):

Lazarus-931:
hi @Anai-Guo, thx for the fix, verified on my end this fixes it :)

PR #2066, opened 2026-08-28
Fix quantization divisibility guard to respect per-module group_size

el3mentdev:
## Problem

`quantize_model`'s `wrapped_predicate` rejects any module whose last dim isn't divisible by the **global** `group_size` *before* consulting the per-module `quant_predicate`. But a smaller per-module group size is exactly what a predicate would request for such a module — so the predicate never runs, and the module silently stays in full precision.

Concrete case: Qwen3.8-Flash-Next's 51B-param PLE n-gram table has dims not divisible by 64. A "4-bit" conversion silently leaves it in BF16, producing a 162 GB output instead of 97 GB, with no warning.

## Fix

- **`quant_utils.py`**: evaluate the per-module predicate first, then apply the divisibility check against the *effective* group size it returns (falling back to the global one).
- **`qwen4_exp/language.py`**: add a `quant_predicate` quantizing the n-gram embedding at group 32 / 4-bit, which the fixed guard now allows.

## Verified

Converted `Qwen3.8-Flash-Next` to 4-bit with this patch: output is 97 GB (n-gram table quantized at 4-bit/g32) vs 162 GB before, and the quantized model serves correctly under mlx_vlm generate/server on an M5 Max.

### DIFF
--- a/mlx_vlm/models/qwen4_exp/language.py
+++ b/mlx_vlm/models/qwen4_exp/language.py
@@ -1610,6 +1610,21 @@ class LanguageModel(Qwen3_5LanguageModel):
             )
         return hidden
 
+    @property
+    def quant_predicate(self):
+        base = super().quant_predicate
+
+        def predicate(path, module):
+            # PLE n-gram table dims are not divisible by 64; group 32 lets it
+            # quantize instead of falling back to full precision.
+            if "ngram_embedding" in path:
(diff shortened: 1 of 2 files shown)


lucasnewman:
This is already covered by the changes in https://github.com/Blaizzy/mlx-vlm/pull/2045

PR #2067, opened 2026-08-28
fix(video): forward the frame-count kwargs to load_video instead of the processor (#2050)

### LINKED ISSUE #2050: Qwen3-VL: fps/max_frames appear unconfigurable — always samples with hardcoded processor defaults (fps=2, max_frames=768) regardless of override attempt
## Summary
For a video input, `fps` and `max_frames` do not appear to have any effect on
the number of video frames actually sampled/encoded, whichever of three
different mechanisms is used to set them. The model consistently produces
identical prompt token counts (10,798 tokens) regardless of the values passed,
matching what appears to be the Qwen3VLVideoProcessor class defaults
(fps=2.0, max_frames=768) baked in at:

## Environment
- mlx-vlm: 0.6.16
- Model: huihui-ai/Huihui-Qwen3-VL-4B-Instruct-abliterated (Qwen3-VL 4B architecture)
- Python: 3.14, venv, macOS (Apple Silicon)
- Video: local .mp4, 30fps, ~600s duration

mlx_vlm/models/qwen3_vl/processing_qwen3_vl.py, ~line 443-461:
```python
fps: float = 2.0,
min_frames: int = 4,
max_frames: int = 768,
```

The model's `preprocessor_config.json` (as downloaded from the huihui-ai repo)
contains no `fps`/`min_frames`/`max_frames` keys, so these class defaults are
always in effect, and none of the override paths below change that.

## Reproduction

**Attempt 1 — CLI `--fps` flag:**
```bash
python3 -m mlx_vlm.generate --model huihui-ai/Huihui-Qwen3-VL-4B-Instruct-abliterated \
  --video video.mp4 --fps 1.0 --verbose --prompt "..."
# Prompt: 10798 tokens

python3 -m mlx_vlm.generate --model huihui-ai/Huihui-Qwen3-VL-4B-Instruct-abliterated \
  --video video.mp4 --fps 3.0 --verbose --prompt "..."
# Prompt: 10798 tokens (identical, despite 3x fps)
```

**Attempt 2 — `generate()`/`apply_chat_template()` kwargs directly:**
```python
    processor, config, PROMPT, num_images=0, video=[str(VIDEO)], fps=0.5,
)
output = generate(
    model, processor, formatted_prompt,
    video=[str(VIDEO)], fps=0.5, max_frames=20,
    max_tokens=2048, verbose=True,
)
# Prompt: 10046 tokens (identical whether max_frames=200 or max_frames=20 
# in separate runs — a much smaller change than the ~74-96% frame-count 
# reduction requested, and still far from the fps=1.0 baseline above)
```

**Attempt 3 — direct attribute mutation on the loaded processor, to rule out the kwarg simply not being threaded through:**
```python
video_proc = processor.video_processor
print(f"Before: fps={video_proc.fps}, max_frames={video_proc.max_frames}")
video_proc.fps = 0.5
video_proc.max_frames = 20
print(f"After:  fps={video_proc.fps}, max_frames={video_proc.max_frames}")
# Before: fps=2, max_frames=768
# After:  fps=0.5, max_frames=20   <- confirms the object itself was mutated

output = generate(model, processor, formatted_prompt, video=[str(VIDEO)],
                   max_tokens=2048, verbose=True)
# Prompt: 10798 tokens  <- back to the ORIGINAL default-value baseline, 
# despite the processor object visibly holding fps=0.5/max_frames=20 
# immediately beforehand
```

## Observed vs expected
- Expected: prompt token count (and thus frame count) should vary meaningfully with `fps`/`max_frames`, since these directly control `load_video()`'s frame-count clamping formula (`n = min(max(n, min_frames), max_frames, total_frames)`).
- Observed: token count is stuck at what appears to be the class-default sampling (fps=2, max_frames=768), even when the loaded processor object is directly confirmed (via print) to hold different values immediately before generation.

## Question for maintainers
Is there a supported way to override video sampling fps/max_frames for a native-video-capable processor like Qwen3-VL's? Attempt 3 suggests the object being mutated either isn't the one actually consulted during frame sampling (e.g. a fresh instance is constructed elsewhere, or values are re-read from config at call time rather than from the live object), or there's a different override point entirely that isn't documented in the CLI `--help` or README. Happy to test a patch or provide more diagnostic output if useful — full reasoning and code excerpts (from `processing_qwen3_vl.py`, `qwen3_vl.py`, `dispatch.py`, `generate/video.py`, `utils.py::load_video`) that led to this report available on request.

Anai-Guo:
Fixes #2050

### What

`prepare_inputs()` pops `fps` out of `kwargs` and passes it to `load_video()`, but leaves `nframes` / `min_frames` / `max_frames` in `kwargs`, where they are forwarded to the processor instead. This pops them next to `fps` and passes them through.

### Why

For a native-video processor (Qwen3-VL and friends) `mlx_vlm` decodes the clip itself, in `prepare_inputs`:

```python
fps_hint = kwargs.pop("fps", 2.0)
...
arr, s_fps = load_video(str(v), fps=fps_hint) if isinstance(v, (str, bytes)) else (v, fps_hint)
```

`load_video()` is what picks the frames, and its signature is `(video_path, fps=2.0, nframes=None, min_frames=4, max_frames=768, frame_factor=2)`. Only `fps` ever reaches it. Everything else the caller passes falls through to

```python
inputs = process_inputs_with_fallback(processor, ..., **extra, **kwargs)
```

so the processor receives `max_frames=...` — after `load_video` has already chosen the frames. It has no effect there, and nothing warns.

The consequence is that `load_video`'s built-in `max_frames=768` is the only cap that ever applies, and the caller cannot lower it. In #2050 that cap is exactly what binds: on a clip long enough that `total_frames / video_fps * fps` exceeds 768, `n` clamps to 768 for every `fps` at or above the one that first saturates it, so raising `--fps` changes nothing, and `max_frames=20` vs `max_frames=200` are indistinguishable — both are dropped. Mutating `processor.video_processor.max_frames` does nothing for the same reason: that object is not what samples the clip.

Note the sibling path already honours the knob — when `processor_handles_video()` is false, `dispatch.py` routes through `resolve_video_inputs(..., max_frames=max_frames)` and `subsample_evenly()` applies it. This only closes the gap on the native-video path; `--video-max-frames` keeps its documented fallback-only meaning, and this PR does not touch the CLI.

### Scope

`kwargs.pop(name)` only fires when the caller actually supplied the name, so `load_video`'s defaults (`nframes=None`, `min_frames=4`, `max_frames=768`) are unchanged for every existing call. No behaviour changes unless a caller passes one of these, which today is silently ignored.

### Verification

No Apple silicon here, so this was checked by extracting `prepare_inputs` with `ast` and running it against stubbed `load_video` / `process_inputs_with_fallback` on both `main` and this branch:

```
--- UNPATCHED (main @ 8adff01)
    load_video kwargs  : {'fps': 0.5}
    leaked to processor: ['max_frames', 'min_frames']
--- PATCHED
    load_video kwargs  : {'fps': 0.5, 'min_frames': 8, 'max_frames': 20}
    leaked to processor: []
```

`tests/test_utils.py::test_prepare_inputs_forwards_video_sampling_kwargs_to_load_video` encodes the same thing as a regression test, asserting both halves: the kwargs reach `load_video`, and they no longer reach the processor. Formatted with the pinned `black==26.3.1` and `isort==5.13.2 --profile=black`; leaving the rest to CI.

### DIFF
--- a/mlx_vlm/tests/test_utils.py
+++ b/mlx_vlm/tests/test_utils.py
@@ -508,6 +508,48 @@ def test_prepare_inputs_preserves_mlx_attention_mask_for_thread_handoff():
     assert consumed == [[[1, 1]]]
 
 
+def test_prepare_inputs_forwards_video_sampling_kwargs_to_load_video():
+    """nframes/min_frames/max_frames clamp load_video's frame count.
+
+    load_video is what samples the clip here, so these have to reach it. Left
+    in kwargs they go to the processor, which only sees the frames load_video
+    already chose, and the caller's cap silently does nothing.
+    """
+    tok_result = MagicMock()
+    tok_result.input_ids = [[1, 2, 3]]
+    tok_result.attention_mask = [7, 8, 9]
+    processor = MockProcessor(tokenizer_return_value=tok_result)
+
+    seen = {}
+    frames = object()
+
+    def fake_load_video(path, **kw):
+        seen.update(kw)
+        return frames, 1.5
+
+    with (
+        patch("mlx_vlm.utils.load_video", side_effect=fake_load_video),
+        patch("mlx_vlm.utils.process_inputs_with_fallback") as fallback,
+    ):
+        fallback.return_value = {"input_ids": mx.array([[1, 2, 3]])}
+        prepare_inputs(
+            processor,
+            prompts="describe the clip",
+            videos=["clip.mp4"],
+            fps=0.5,
+            max_frames=20,
+            min_frames=8,
+        )
+
+    assert seen == {"fps": 0.5, "max_frames": 20, "min_frames": 8}
+
+    forwarded = fallback.call_args.kwargs
+    assert forwarded["videos"] == [frames]
+    assert forwarded["fps"] == [1.5]
+    for name in ("nframes", "min_frames", "max_frames"):
+        assert name not in forwarded
+
+
 def test_process_inputs_with_fallback():
 
     processor = MockProcessor()
--- a/mlx_vlm/utils.py
+++ b/mlx_vlm/utils.py
@@ -2116,10 +2116,19 @@ def prepare_inputs(
         if not isinstance(videos, list):
             videos = [videos]
         fps_hint = kwargs.pop("fps", 2.0)
+        # The frame-count controls belong to load_video, which is what samples
+        # here. Left in kwargs they reach the processor instead, which only
+        # ever sees the frames load_video already picked, so they read as
+        # silently ignored.
+        sampling = {
+            name: kwargs.pop(name)
+            for name in ("nframes", "min_frames", "max_frames")
+            if name in kwargs
+        }
         loaded, video_fps = [], []
         for v in videos:
             arr, s_fps = (
-                load_video(str(v), fps=fps_hint)
+                load_video(str(v), fps=fps_hint, **sampling)
                 if isinstance(v, (str, bytes))
                 else (v, fps_hint)
             )


cellar-fume-lowbrow:
Confirmed as working on Apple Silicon - 

~10 minute test video
load_video called: fps=0.5, nframes=None, min_frames=4, max_frames=20
load_video returned: 20 frames, actual_fps=0.03151922672830426

Lazarus-931:
hi @Anai-Guo, thx for the pr!

this seems like a temporary small patch and I think this warrants a general redesign to prevent this issue, closing this and opening a followup pr tagging u, thanks :)

PR #2069, opened 2026-08-28
mage_vl: add video input (closes #1766) and explicit attention-window boundaries

### LINKED ISSUE #1766: Mage-VL: dense/frame-sampled video input support

shubhambaid:
Closes #1766.

`mage_vl` currently raises `NotImplementedError` on any video input, so Mage-VL runs on
Apple Silicon as an image-only model — which drops the architecture's headline feature.

### Why it was blocked

Upstream's video path needs `codec-video-prep`, a native extension shipping patched FFmpeg
with manylinux-only wheels. It cannot run on macOS at all.

Uniform frame sampling needs none of that. `utils.load_video` (OpenCV, already a
dependency) and `Qwen3VLVideoProcessor` (the existing numpy/PIL port) already do the work,
so **this adds no new dependency**. Mage-VL sets `temporal_patch_size: 1`, so the image and
video processors share one config.

### Changes

1. **`processing_mage_vl.py`** — implements the video path and expands `<|video_pad|>` to
   each grid's merged-token count.

   The video processor is carried in `kwargs` rather than declared as a processor
   attribute. Declaring it makes transformers 5.x raise `This processor requires 3
   arguments ... Got 2`; registering `video_processor_class = "AutoVideoProcessor"` instead
   raises `Received a Qwen3VLVideoProcessor, but a BaseVideoProcessor was expected`. Both
   are swallowed by the `AutoProcessor` dispatcher, which then silently falls back to the
   checkpoint's torch remote code — the failure this class already documents for
   `AutoImageProcessor`.

2. **`vision.py`** — `VisionModel.__call__` accepts explicit `cu_seqlens`.

   `build_cu_seqlens` derives window boundaries from a dense grid (`fixed_t * h * w` per
   window) and raises when the total disagrees with the patch count. That is correct, and
   it is exactly why a *codec-selected* patch set cannot use it: sparse selection keeps a
   different number of patches in every window. Everything else in the tower is already
   sparsity-agnostic — `forward_from_positions` takes arbitrary `(L, 3)` coordinates and 3D
   RoPE is defined over the un-pruned grid — so supplying boundaries directly is the only
   change sparse video needs. Grid-derived behaviour is unchanged when the argument is
   omitted.

3. **`mage_vl.py`** — threads `cu_seqlens` through from kwargs.

4. **`prompt_utils.py`** — adds `mage_vl` to the video-routing allowlist. Without it
   `apply_chat_template(video=[...])` emitted no placeholder, the prompt reached the model
   with no `<|video_pad|>`, and generation ran on text alone — the model replied that it
   cannot view videos, which reads like a refusal rather than a bug.

### Verification

- Image path unchanged.
- Video: `pixel_values_videos (6272, 768)`, `video_grid_thw [[8, 28, 28]]`, placeholder
  count matches `t*h*w/merge²` exactly (1568).
- `cu_seqlens`: explicit boundaries are **bit-identical** to grid-derived on dense input
  (`max_abs_diff = 0.0`); malformed input is rejected.
- End-to-end on a NExT-QA clip: *"A baby is sitting in a baby swing."*

### Not included

Codec-native patch selection itself. The tower can now consume it, but the selection
signal is a separate question — in a matched-budget ablation on NExT-QA (n=150, B=4096)
no importance signal I tested beat random selection, including stock-ffmpeg motion
vectors, so I would rather not ship a default that isn't earned. Happy to follow up if
useful.


lucasnewman:
Thanks, this fix isn't quite right but I've put up a PR at https://github.com/Blaizzy/mlx-vlm/pull/2202 to resolve it.

PR #2072, opened 2026-08-28
Reduce exact APC ownership-transfer peaks

YuhuaBillChen:
## Summary

- transfer restored exact-cache ownership to the batch generator
- materialize/requantize one restored layer at a time and release its source
  before moving to the next layer
- promote a consumed single `KVCache` row to `BatchKVCache` without copying or discarding its reserved capacity
- avoid cloning an already-detached exact snapshot again when transferring it into the APC manager

The default non-consuming merge and defensive store-copy behavior remain available to callers that do not transfer ownership.

## Why

After #1960, `APCCoordinator` owns the distinction between pageable block APC and exact checkpoints. The exact-checkpoint join still kept every restored single-row source alive while building all merged float layers and then all quantized replacements. For long prefixes, those states overlap even though the batch cache is the only owner needed afterward.

The store side had the inverse ownership issue: the coordinator first created a detached checkpoint snapshot, then `APCManager.store_exact_cache()` cloned the same snapshot again. This change makes both ownership transfers explicit. The store-side snapshot is produced by `snapshot_prompt_cache_row()`, which already creates a detached clone through `_clone_prompt_cache_for_apc()`. After passing it to `store_exact_cache(..., take_ownership=True)`, the coordinator does not retain or mutate it.

## Results after #1960

I rebased onto `main@3fd38f4c`, which includes #1960, and reran the real-model ABBA comparison against that exact base.

Configuration:

- Qwen3.8-27B MXFP4 target, including Q8 vision weights
- TQ4 KV from token zero; no MTP
- 16,396-token prompt with 16,395 tokens restored from an existing exact disk snapshot
- one output token
- fresh process per arm, order before/after/after/before
- the same APFS copy-on-write snapshot for every arm
- 21,000 MiB watchdog; exact process peak from the kernel-maintained `ri_lifetime_max_phys_footprint` counter read after each request

| Metric (median of two runs) | #1960 baseline | This PR | Change |
|---|---:|---:|---:|
| Kernel lifetime max physical footprint | 19,050 MiB | 16,942 MiB | -2,108 MiB (-11.1%) |
| API-reported MLX peak | 18.058 GB | 17.181 GB | -0.877 GB (-4.9%) |
| Warm request latency | 1.531 s | 1.518 s | -0.8% |
| One-token suffix prefill | 172.1 ms | 132.0 ms | -23.3% |
| Cached tokens | 16,395 | 16,395 | unchanged |

All four arms returned HTTP 200 with no traceback or watchdog trip. #1960 already improved the baseline restore latency substantially, so the old latency percentages no longer apply; the peak-memory reduction still holds.

A separate 64K synthetic Qwen-layout ABBA isolates the cache transformation:

| Metric | #1960 baseline | This PR | Change |
|---|---:|---:|---:|
| MLX peak | 6,611.5 MiB | 5,701.5 MiB | -910.0 MiB (-13.8%) |
| Active after restore | 5,071.5 MiB | 1,231.5 MiB | -3,840.0 MiB |
| One-time conversion | 0.634 s | 0.815 s | +0.181 s |

The store-side ownership transfer removes a redundant live clone and makes that storage reclaimable/reusable. It does not claim an equal reduction in macOS process footprint because MLX may retain released buffers in its allocator cache.

## Tests

- GitHub test workflow on the rebased branch: passed
- focused APC/coordinator/generation tests: `213 passed`
- full suite on MLX 0.32.2: `2912 passed, 7 skipped, 75 subtests passed`
- Black, isort, autoflake, `git diff --check`, and Python compilation pass
- real Qwen3.8 exact-APC cold/warm canary: HTTP 200, 4,107/4,108 prompt
  tokens restored on the warm request, no traceback or watchdog trip


Blaizzy:
Hey @YuhuaBillChen 

Thank you very much, could you test if this still holds on #1960 

We'll merge today

YuhuaBillChen:
@Blaizzy Thanks for flagging this. Yes — I rebased the PR onto current `main@3fd38f4c`,
which includes #1960, and moved the ownership handoff into the new `APCCoordinator` path.

I also reran the same real-model ABBA against the #1960 baseline. At a 16,396-token prompt with 16,395 tokens restored, the median kernel-maintained lifetime maximum physical footprint fell from 19,050 to 16,942 MiB (−2,108 MiB, −11.1%) and the API-reported MLX peak fell from 18.058 to 17.181 GB. All four arms returned HTTP 200 with identical cached-token counts and no watchdog trip.

#1960 already made restore much faster, so I replaced the old latency numbers in the PR body rather than carrying them forward. The rebased GitHub test is green, and the full local suite passes: 2912 passed, 7 skipped, 75 subtests.

Blaizzy:
What model are you using and could you share a reproducible example ?

YuhuaBillChen:
@Blaizzy I used `mlx-community/Qwen3.8-27B-mxfp4` from Hugging Face as the
target. My local copy only requantizes the vision tower to affine Q8; the text
weights and Qwen3.8 27B cache layout are unchanged. The measured restore
request was text-only, with TQ4 KV from token zero and no drafter (MTP).

Hardware/software:

- 24 GB M4 Pro MacBook Pro (14 CPU 20 GPU)
- MLX 0.32.2
- baseline: `main@3fd38f4c` (includes #1960)
- patched: `b290efa2`
- `--kv-bits 4 --kv-quant-scheme turboquant --quantized-kv-start 0`
- exact disk APC, `APC_EXACT_CACHE_ENTRIES=0`

For the real-model ABBA, I first created one 16,395-token exact snapshot, then used an APFS copy-on-write clone of that same snapshot for every arm. Each arm started a fresh server and sent the same prompt plus one suffix token. The order was baseline / patched / patched / baseline. I read the exact kernel-maintained lifetime footprint with `proc_pid_rusage(..., RUSAGE_INFO_V4, ...)`, rather than sampling RSS. All four requests restored 16,395 tokens and returned HTTP 200.

I also have a model-free cache-layout reproducer if useful.

YuhuaBillChen:
@Blaizzy Thanks for the comments and I just put the complete public-checkpoint reproducer here:

https://gist.github.com/YuhuaBillChen/ae29ef3c8d26488f1652c941b28b148d

It checks out `main@3fd38f4c` and this PR at `b290efa2`, uses the public
`mlx-community/Qwen3.8-27B-mxfp4` checkpoint, creates one exact APC
snapshot, and APFS-clones that same snapshot for four fresh-process ABBA
arms.

On my laptop: M4 Pro 24 GB, MLX 0.32.2, TurboQuant Q4 KV from token zero,
exact disk APC, no MTP, 16,396 prompt tokens, and a one-token suffix. Each
reported value is the median of two fresh-process runs.

| Metric | `main@3fd38f4c` | PR #2072 | Change |
|---|---:|---:|---:|
| Kernel lifetime footprint | 19,345.9 MiB | 17,384.3 MiB | **-1,961.6 MiB (-10.1%)** |
| One-token suffix prefill | 165.2 ms | 138.1 ms | **-16.4%** |
| Total warm-request latency | 2.737 s | 2.742 s | effectively unchanged |
| Exact cached tokens | 16,395 | 16,395 | unchanged |

Blaizzy:
Thanks @YuhuaBillChen! Really appreciate the thorough work here: the rebase onto #1960, the ABBA numbers, and the public reproducer.

I'm going to put a pin in this for now rather than merge today. The APC path has just been redesigned in #1960 and I want to let it settle before layering ownership-transfer semantics on top, so we can evaluate this change against a stable base.

Could you open an issue capturing the problem this solves, the overlapping restore/store lifetimes, the peak-memory numbers, and a link to your reproducer gist? That way we don't lose the context. I'll close this PR for now and come back to it in the coming weeks; the branch is worth revisiting once the dust settles.

Blaizzy:
> @Blaizzy Thanks for the comments and I just put the complete public-checkpoint reproducer here:
> https://gist.github.com/YuhuaBillChen/ae29ef3c8d26488f1652c941b28b148d

Read my mind, thanks!

PR #2074, opened 2026-08-28
Dflash2 support for GLM5.3-flash

Lazarus-931:
## DFlash2 support for GLM-5.3-Flash

Stacked on #2044 — the incremental change here is the DFlash2 target hooks for
`glm5_next` (the DFlash2 drafter, its runtime round-loop, and the converter are
already generic in the tree).

- `Glm5NextModel.__call__` captures the per-layer hidden at the drafter's
  `target_layer_ids`.
- `LanguageModel.__call__` builds the hidden sink for both prefill and verify,
  threads `capture_layer_ids`, and returns the per-layer hidden plus KDA states,
  reusing the existing `rollback_speculative_cache`.

Single- and batch-sequence tests generate with and without the drafter and assert
the output is bit-identical to plain greedy decode.

## Performance

GLM-5.3-Flash (mixed 4-bit MoE / 8-bit) target + the published
`incoai/GLM-5.3-Flash-DFlash2` drafter (bf16, block 8) on an Apple M3 Ultra
(512 GB). Greedy, batch 1, decode throughput (tokens after the first; prefill
excluded), 64 generated tokens. MTP (single nextn head, block 2) shown for
comparison. Every row is bit-identical to plain decode (0 mismatched tokens).

| ctx | baseline tok/s | MTP tok/s | MTP × | DFlash2 tok/s | DFlash2 × | DFlash2 accepted/8 |
|--:|--:|--:|--:|--:|--:|--:|
| 512   | 33.84 | 36.49 | 1.08× | 81.78 | 2.42× | 6.00 |
| 2048  | 29.88 | 35.18 | 1.18× | 76.40 | 2.56× | 6.00 |
| 8192  | 28.94 | 34.83 | 1.20× | 68.45 | 2.37× | 6.00 |
| 16384 | 28.29 | 32.01 | 1.13× | 58.29 | 2.06× | 6.00 |

### DIFF
--- a/mlx_vlm/models/glm5_next/README.md
+++ b/mlx_vlm/models/glm5_next/README.md
@@ -58,6 +58,26 @@ proposes one token from the target's hidden state and the target verifies the
 See [`mlx_vlm/speculative/drafters/glm5_next_mtp/README.md`](../../speculative/drafters/glm5_next_mtp/README.md)
 for the split tool and usage.
 
+## Speculative decoding (DFlash2)
+
+GLM-5.3-Flash can also be driven by a **DFlash2** drafter -- a separate,
+multi-token block drafter (block size 8) trained against the target and published
+as `incoai/GLM-5.3-Flash-DFlash2`. Unlike the single-token nextn head it proposes
+a whole block per round, so acceptance clears the verify overhead by a wide
+margin.
+
+Each round the drafter reads the target's hidden state at the checkpoint's
+`target_layer_ids` and proposes a block; the target verifies the block in one
+forward and, on a partial accept, rolls the KDA recurrent state and the
+MLA/indexer caches back to the accepted prefix (the same rollback MTP uses).
+Output is bit-identical to plain decode.
+
+Target-side support is small: the model captures the per-layer hidden at
+`target_layer_ids` and returns it (with the KDA states) from the verify forward;
+the drafter, its round loop, and the converter are shared across DFlash targets.
+The drafter runs in bf16 as published -- it does not need quantizing to pair with
+a quantized target.
+
 ## Continuous batching
 
 Runs the batched `BatchGenerator` path unmodified. The lightning indexer's incremental
--- a/mlx_vlm/models/glm5_next/language.py
+++ b/mlx_vlm/models/glm5_next/language.py
@@ -795,6 +795,7 @@ class Glm5NextModel(nn.Module):
         inputs_embeds: Optional[mx.array] = None,
         gdn_sink: Optional[list] = None,
         hidden_sink: Optional[list] = None,
+        capture_layer_ids: Optional[list] = None,
     ) -> mx.array:
         h = self.embed_tokens(inputs) if inputs_embeds is None else inputs_embeds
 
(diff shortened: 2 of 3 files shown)


lucasnewman (approved):

avlp12:
Ran this branch (f025d5c8) on an M3 Ultra 512GB against `pipenetwork/GLM-5.3-Flash-MLX-mixed-4_8bit` (the PR doesn't name its target — happy to be corrected).

With the chat template's default reasoning effort, greedy, block 8 fixed, ctx≈512 / 64 tokens I measure acceptance length 2.9–5.3 tokens/round depending on prompt (best: 4.9 prose-think / 5.3 code-continuation, 1.8x speedup) — consistent with incoai's published 4.03–5.86 band, but below the table's flat 6.00.

Three questions to close the gap:

1. could you paste the `format_speculative_stats` line from your runs? (settles whether "accepted/8" is acceptance length or accepted drafts)
2. which target checkpoint/conversion was used?
3. what prompt and chat-template kwargs?

FWIW my plain-decode baseline is ~28.3 tok/s at ctx512 vs your 33.84, so an environment/checkpoint difference is likely on throughput; acceptance is what I'd love to reconcile.

Also: on this quantized target, greedy spec output was not bit-identical to plain decode on 2 of 5 prompts (single early divergence then re-convergence) — worth a caveat on the bit-identity claim for quantized targets.

Lazarus-931:
| Context | Base prefill | DFlash2 prefill | Base decode | DFlash2 decode | Speedup | Acceptance | Base peak | DFlash2 peak | Tokens |
|--:|--:|--:|--:|--:|--:|--:|--:|--:|:--|
| 512 | 245.22 | 374.13 | 33.40 | 70.46 | 2.11x | 100.00% | 180.05 GB | 185.90 GB | exact |
| 1,024 | 246.36 | 437.21 | 33.16 | 70.23 | 2.12x | 100.00% | 180.24 GB | 189.76 GB | exact |
| 2,048 | 242.81 | 465.86 | 29.58 | 60.08 | 2.03x | 100.00% | 179.96 GB | 197.36 GB | exact |
| 4,096 | 231.24 | 459.79 | 29.50 | 59.61 | 2.02x | 100.00% | 180.26 GB | 213.01 GB | exact |
| 8,192 | 212.50 | 423.81 | 29.14 | 58.00 | 1.99x | 100.00% | 180.94 GB | 248.06 GB | exact |
| 16,384 | 182.65 | 353.39 | 28.96 | 53.60 | 1.85x | 100.00% | 182.26 GB | 332.20 GB | exact |

avlp12:
Thanks for the table — the peak-memory column is especially useful (the DFlash2 peak growing 186 → 332 GB from 512 to 16k context while the base stays roughly 180–182 GB is the number I'd want to see before recommending it for long prompts).

Two things I can't reconcile with what I measure, so I'm asking rather than asserting:

1. **Acceptance 100.00% at every context.** In my measured natural-text runs I have not seen this drafter accept everything. With block 8 fixed, greedy, on a 4-bit-class quantized target, I measure a mean acceptance length of 2.9–5.3 tokens per round depending on prompt (my earlier comment), and per-position acceptance that varies by position — on natural code roughly 0.91 / 0.71 / 0.91 / 0.97 / 0.77 / 0.89 / 0.85 at positions 1–7, on prose 0.87 / 0.95 / 0.92 / 0.94 / 0.96 / 0.81 / 0.32. I can't infer the prompt from the table; could you share it, and say whether the roughly 2.1× at 512 also holds on natural text? (I made a similar repeated-token-prompt mistake in an earlier #2087 comment and later corrected it, so I'm sensitive to this one.)

2. **"Tokens: exact".** On the quantized target, greedy speculative output was *not* token-identical to plain decode on 2 of 5 prompts for me (single early divergence, then re-convergence). If this was run on bf16, that may be one reason the result differs from my quantized run — which checkpoint was used? For a quantized target I'd label the result "token-identical on the tested prompts" rather than "exact": a near-tie argmax can resolve differently between an 8-wide verify and sequential decode, so I'd report token identity separately from the measured speedup.

The three details from my earlier comment (the `format_speculative_stats` line, the checkpoint, the chat-template kwargs) would help reconcile both points. If you share them, I can try to reproduce the setup and report what I measure.

Blaizzy (approved):
LGTM, thanks!

PR #2082, opened 2026-08-28
Hy4 model support

Lazarus-931:
## Hy4

decode tok/s (M3 Ultra · 4-bit · batch 1)

| ctx | baseline | +MTP |
|--:|--:|--:|
| 512 | 13.8 | 16.4 |
| 2,048 | 12.9 | 14.9 |
| 4,096 | 12.9 | 15.5 |
| 8,192 | 12.8 | 15.2 |
| 16,384 | 12.7 | 14.9 |

Adds the `hy_v4` target model and an `hy_v4_mtp` speculative drafter.
MTP runs at 100% draft acceptance across the sweep; peak memory 435–455 GB.

[Model](https://huggingface.co/mlx-community/Hy4-preview-4bit)


mlx_vlm/utils.py
```diff
@@ -1035,6 +1035,14 @@ def load_model(model_path: Path, lazy: bool = False, **kwargs) -> nn.Module:
             elif quant_method == "fp8" and config.get("model_type") == "deepseek_v4":
                 from .models.deepseek_v4.language import make_quantization_config
 
+                quantization = make_quantization_config(model)
+            elif (
+                quant_method == "modelopt"
+                and quantization_config.get("quant_algo") == "MXFP8"
+                and config.get("model_type") == "hy_v4"
```
lucasnewman: I realize there's already a pattern here, but it would be nice
 to just use `getattr()` to check if the language model supports `make_quantization_config` so we don't need model-type specific code.

lucasnewman (approved):

lucasnewman:
@Lazarus-931 Can you also add a README for the model?

Lazarus-931:
> @Lazarus-931 Can you also add a README for the model?

yeah will do, held it since never got to run full sweeps for final numbers

mlx_vlm/utils.py
Lazarus-931: that's right, will fix, thx!

PR #2085, opened 2026-08-28
qwen4_exp: opt-in disk offload for the n-gram PLE table

eyaltoledano:
# qwen4_exp: opt-in disk offload for the n-gram PLE table

## What this does

Adds an opt-in `ple_on_disk` mode for `qwen4_exp` models. When enabled, the
per-layer n-gram PLE (Per-Layer Embedding) table is left on disk and its rows
are read and dequantized on demand instead of being held resident in memory.

It is exposed three ways, all defaulting to **off** with zero behavior change:

- `load_model(..., ple_on_disk=True)` / `load(..., ple_on_disk=True)` kwarg
- `mlx_vlm.server --ple-on-disk`
- `mlx_vlm.generate --ple-on-disk`

This mirrors the existing `--expert-cache-gb` MoE-offload flag: a load-time
option for running a family whose weights are dominated by a structure that is
mostly cold on any given token.

## Why it's useful

In `qwen4_exp` the n-gram PLE table dominates the parameter count — on the
order of ~51B parameters, roughly ~29 GB once quantized to 4-bit — yet only a
few hundred bytes of it are read per token (a handful of rows per n-gram head).
Keeping the whole table resident spends tens of GB of RAM to serve byte-sized
reads.

Reading rows from disk instead drops resident memory by ~29 GB on any
`qwen4_exp` checkpoint. Measured active memory, last-position logits path,
one process per mode (never co-resident):

- Stock 4-bit conversion: ~97 GB resident → ~68 GB with `ple_on_disk`.
- An expert-reduced 4-bit conversion
  ([`sh0wie/Qwen3.8-Flash-Next-REAP-288-MLX-4bit`](https://huggingface.co/sh0wie/Qwen3.8-Flash-Next-REAP-288-MLX-4bit)):
  68.17 GB → 38.37 GB (−29.8 GB).

That is the difference between "needs a 96 GB+ machine" and "runs on a 64 GB
machine" for this family. The table rows are a few hundred bytes each and are
mmap'd, so only touched pages enter the OS page cache; the rest never loads.

## Correctness

The on-disk path is **bit-identical** to the resident path. The disk reader
dequantizes rows on the host in float32 and reproduces `mx.dequantize` for
affine group quantization exactly (little-endian bit unpack, per-group
scale/bias in bf16), then embeds in bf16 as the resident module does.

- On the REAP-288 checkpoint above, disk-vs-RAM last-position logits are
  identical (max abs 0.0; same argmax) — the difference is purely I/O, not math.
- Included unit tests (`mlx_vlm/tests/test_qwen4_exp.py`) assert max abs 0.0
  on a synthetic single-shard checkpoint, both for the row dequant vs
  `mx.dequantize` and for the disk-mode `ShardedEmbedding` vs a resident
  `QuantizedEmbedding` built from the same bytes. They are CPU-light: they
  quantize and read a tiny table, never a full model.

The disk reader accepts **both** n-gram shard naming conventions —
`.ngram_embedding.shard_N.` (underscore) and `.ngram_embedding.shards.N.`
(dot) — because it reads raw tensor names straight from the safetensors index,
and different conversions of this family ship one or the other. A regression
test covers both namings and confirms non-shard PLE tensors are untouched.

## How it works

- `ShardedEmbedding` gains a `disk_dir` mode: it constructs with no resident
  shards and serves lookups via `_PLEDiskIndex`, a numpy-memmap, row-granular
  reader over the quantized shard tensors in the checkpoint's safetensors.
- `Model.sanitize` drops the n-gram shard tensors in disk mode so they don't
  collide with the now shard-free module at `load_weights`.
- `TextConfig` carries runtime `ple_on_disk` / `ple_disk_path` fields (default
  off); `load_model` sets them from the `ple_on_disk` kwarg (recording the
  model path) before constructing the model, so the embedding is built
  shard-free from the start rather than allocated and then freed.

Models without a PLE table are unaffected: the kwarg is a no-op for them, and a
non-`qwen4_exp` model logs a warning and ignores it.

## Usage

```bash
# Server
python -m mlx_vlm.server --model sh0wie/Qwen3.8-Flash-Next-REAP-288-MLX-4bit --ple-on-disk

# One-shot generate
python -m mlx_vlm.generate --model sh0wie/Qwen3.8-Flash-Next-REAP-288-MLX-4bit --ple-on-disk \
    --prompt "Write a Python function that merges two sorted lists."
```

```python
from mlx_vlm import load

model, processor = load("sh0wie/Qwen3.8-Flash-Next-REAP-288-MLX-4bit", ple_on_disk=True)
```

## Notes / scope

- The disk reader assumes the PLE table is affine-quantized at group size 32,
  4-bit — the layout every published `qwen4_exp` conversion ships. Non-disk
  (resident) loading is unchanged and derives quant params from the checkpoint
  as before.
- Trades a small per-token host read for the RAM. On fast NVMe the cost is in
  the noise for interactive use; the point is fitting the model at all on
  memory-constrained machines.

### DIFF
--- a/mlx_vlm/generate/dispatch.py
+++ b/mlx_vlm/generate/dispatch.py
@@ -509,6 +509,13 @@ def parse_arguments():
         "working set). Ignored for a normal, non-offloaded checkpoint.",
     )
     parser.add_argument(
+        "--ple-on-disk",
+        action="store_true",
+        help="Keep a model's n-gram PLE table (qwen4_exp) on disk and read rows "
+        "on demand instead of holding it resident. Ignored for models without "
+        "a PLE table.",
+    )
+    parser.add_argument(
         "--processor-kwargs",
         type=json.loads,
         default={},
@@ -1271,6 +1278,7 @@ def main():
         trust_remote_code=args.trust_remote_code,
         quantize_activations=args.quantize_activations,
         expert_cache_gb=args.expert_cache_gb,
+        ple_on_disk=args.ple_on_disk,
         max_kv_size=args.max_kv_size,
     )
     config = model.config
--- a/mlx_vlm/models/qwen4_exp/config.py
+++ b/mlx_vlm/models/qwen4_exp/config.py
@@ -67,6 +67,11 @@ class TextConfig(BaseModelConfig):
     eos_token_id: Optional[Union[int, List[int]]] = None
     tie_word_embeddings: bool = False
     attention_bias: bool = False
+    # Runtime-only, not part of the checkpoint. Set by load_model() when the
+    # caller passes ``ple_on_disk=True`` so the n-gram PLE table is read row by
+    # row off disk instead of held resident. Default off => no behavior change.
+    ple_on_disk: bool = False
+    ple_disk_path: Optional[str] = None
     rope_parameters: Optional[Dict[str, Union[float, str, bool, List[int]]]] = field(
         default_factory=lambda: {
             "type": "default",
--- a/mlx_vlm/models/qwen4_exp/language.py
+++ b/mlx_vlm/models/qwen4_exp/language.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import math
+import re
 from bisect import bisect_right
 from types import SimpleNamespace
 from typing import Any, Optional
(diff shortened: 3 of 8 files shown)


lucasnewman:
This will be covered in https://github.com/Blaizzy/mlx-vlm/pull/2045

PR #2086, opened 2026-08-28
glm5_next: guard the indexer's incremental pool against cache rewinds

### LINKED ISSUE #2084: glm5_next: indexer incremental `_pool` crashes with `[argmax] Cannot argmax reduce zero size array` after any cache rewind (stale `t_prev`)
### Summary

The lightning indexer caches pooled states on the KV cache object as `cache._pool = (pool_keys, pool_indices, pool_valid, t_prev)` and, at decode (`S == 1`), recomputes only the suffix from the last complete pool block (`language.py:358-382` on main). The fast-path condition checks the batch axis and `_no_pad`, but not **time continuity**: after any rewind of the cache between decode steps, `t_prev` can exceed the cache's current length `T`. The "new suffix" `k_full[:, n_stable * kpool :]` is then empty, and `_pooled_states` argmaxes a zero-size array (`language.py:267`):

```
ValueError: [argmax] Cannot argmax reduce zero size array.
```

We hit this in production serving (continuous batching: a batch splice replaced a sequence's cache content without carrying `_pool` along). Any rewind path triggers it the same way — multi-turn `trim()`, or speculative rollback that doesn't manually repair `_pool`.

### Repro (verified on a clean `pip install mlx-vlm==0.6.17`; the path is unchanged on current main)

```python

import mlx.core as mx
from mlx_vlm.models.cache import KVCache
from mlx_vlm.models.glm5_next.language import Glm5NextIndexer

args = SimpleNamespace(
    hidden_size=64, index_n_heads=4, index_head_dim=16,
    index_topk=128, index_kpool=64, index_kpool_always_select_tail=True,
    q_lora_rank=32,
)
mx.random.seed(0)
indexer = Glm5NextIndexer(args)
cache = KVCache()

def step(S):
    x = mx.random.normal((1, S, args.hidden_size))
    qr = mx.random.normal((1, S, args.q_lora_rank))
    mx.eval(indexer(x, qr, None, cache=cache))

step(400)        # prefill: full pooling, _pool cached with t_prev=400
step(1)          # decode: incremental fast path, fine (t_prev=401)
cache.trim(150)  # any rewind between decode steps
step(1)          # T=252, stale t_prev=401 -> suffix [384:] is empty -> argmax crash
```

```
  File ".../mlx_vlm/models/glm5_next/language.py", line 369, in __call__
    pk_s, pi_s, pv_s = self._pooled_states(
  File ".../mlx_vlm/models/glm5_next/language.py", line 268, in _pooled_states
    any_valid, mx.argmax(valid.astype(mx.int32), axis=-1), mx.array(S)
ValueError: [argmax] Cannot argmax reduce zero size array.
```

### Fix (one line, verified on the repro)

Add a strict continuity check to the fast-path condition:

```python
            and cache._pool[0].shape[0] == B
            and cache._pool[3] == T - 1   # pool must be exactly one step behind
        ):
```

Any discontinuity then falls back to one full `_pooled_states` rebuild (`O(T)`, once) and the incremental path resumes on the next step.

Strict equality matters, not `<=`: the crash only fires when the empty-suffix condition `n_stable * kpool >= T` holds, but a rewind or same-batch splice that lands *inside* the last pool block passes the current checks and silently reuses pooled keys computed from content that is no longer in the cache — wrong selection with no error. The equality guard covers both failure modes.

Related: #2044's `rollback_speculative_cache` maintains exactly this invariant manually on its own path (its test asserts `_pool[3] == offset` after rollback). The guard here is defense-in-depth for every other cache manipulation and composes cleanly with that PR.

We've been running this guard in production (2× M3 Ultra, tensor-parallel GLM-5.3-Flash, continuous batching) since finding it; a regression test that splices the cache between decode steps is easy to add. Happy to send the PR.

aidiffuser:
Fixes #2084.

The lightning indexer caches pooled states on the KV cache as `cache._pool = (pool_keys, pool_indices, pool_valid, t_prev)`; at decode the fast path recomputes only the suffix from the last complete pool block. The condition checks the batch axis (the continuous-batching guard) but not time continuity: any rewind of the cache between decode steps — a multi-turn `trim`, a rollback outside `rollback_speculative_cache`, a batch splice — leaves `t_prev` at or ahead of the current length `T`. The suffix `k_full[:, n_stable * kpool :]` is then empty and `_pooled_states` argmaxes a zero-size array, killing generation; a rewind landing *inside* the last pool block passes the current checks and silently reuses pooled keys computed from content no longer in the cache.

The fix is one added condition: the fast path requires the pool to have been written exactly one token ago (`_pool[3] == T - 1`); any discontinuity takes one full pooling pass (which also rebuilds a fresh pool) and the incremental path resumes on the next step.

The regression test (`test_glm5_next_indexer_pool_rewind_guard`, companion to the existing batch-axis `test_glm5_next_indexer_stale_pool_guard`) prefills, decodes, trims back to mid-block, and decodes again — on current main this crashes with `ValueError: [argmax] Cannot argmax reduce zero size array` — and asserts the rewound decode equals a clean decode from a fresh prefill of the same prefix, plus the `_pool[3] == offset` invariant that #2044's rollback test also asserts.

We've been running this guard in production continuous batching (GLM-5.3-Flash, 2× M3 Ultra tensor-parallel) since hitting the crash live under concurrent requests. It composes with #2044, whose rollback maintains the same invariant manually on its own path.

### DIFF
--- a/mlx_vlm/models/glm5_next/language.py
+++ b/mlx_vlm/models/glm5_next/language.py
@@ -361,6 +361,13 @@ class Glm5NextIndexer(nn.Module):
             and getattr(cache, "_pool", None) is not None
             and getattr(cache, "_no_pad", False)
             and cache._pool[0].shape[0] == B
+            # Time-continuity: the cached pool is only valid if it was written exactly
+            # one token ago. Any rewind of the cache between decode steps (a multi-turn
+            # trim, a speculative rollback, a batch splice) leaves t_prev at or ahead of
+            # T -- the "new suffix" below is then empty and _pooled_states argmaxes a
+            # zero-size array (#2084). Fall back to one full pooling pass instead,
+            # which also rebuilds a fresh pool.
+            and cache._pool[3] == T - 1
         ):
             ck, ci, cv, t_prev = cache._pool
             n_stable = t_prev // self.index_kpool
--- a/mlx_vlm/tests/test_models.py
+++ b/mlx_vlm/tests/test_models.py
@@ -2834,6 +2834,81 @@ class TestModels(unittest.TestCase):
         self.assertEqual(logits.shape, (1, 1, config.text_config.vocab_size))
         self.assertTrue(mx.all(mx.isfinite(logits)).item())
 
+    def test_glm5_next_indexer_pool_rewind_guard(self):
+        # Companion to test_glm5_next_indexer_stale_pool_guard (batch-axis
+        # staleness): a cache REWIND between decode steps (a multi-turn trim, a
+        # speculative rollback) leaves the incremental _pool with t_prev at or
+        # ahead of the cache length T. The time-continuity guard must fall back
+        # to a full pooling pass -- the unguarded fast path pools an empty
+        # suffix and argmaxes a zero-size array (#2084). The rewound decode must
+        # also equal a clean decode from a fresh prefill of the same prefix.
+        from mlx_vlm.models import glm5_next
+        from mlx_vlm.models.cache import CacheList, KVCache
+        from mlx_vlm.models.glm5_next.language import Glm5NextSparseAttention
+
+        mx.random.seed(0)
+        cfg = glm5_next.TextConfig(
+            model_type="glm5_next_text",
+            vocab_size=128,
+            hidden_size=128,
+            intermediate_size=128,
+            moe_intermediate_size=64,
+            num_hidden_layers=1,
+            num_attention_heads=2,
+            num_key_value_heads=2,
+            n_shared_experts=1,
+            n_routed_experts=8,
+            routed_scaling_factor=2.5,
+            kv_lora_rank=64,
+            q_lora_rank=128,
+            qk_rope_head_dim=0,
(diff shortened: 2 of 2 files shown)


Sofille65:
+1, with a data point from a different rollback path. We serve GLM-5.3-Flash with our own native-MTP round loop (snapshot/restore of the KDA states + KV trim, outside `rollback_speculative_cache`), and hit exactly the class this PR describes: after a restore the indexer's `_pool` still carried `t_prev` from the rolled-back future, so the fast path either argmaxed a zero-size suffix or silently reused pooled keys for content no longer in the cache. Our fix was to invalidate `_pool` on restore; the one-condition guard here (`_pool[3] == T - 1`) covers that and every other rewind path in one place, so we backported it as-is (single-node M3 Ultra, 6-bit trunk). Composes fine with the manual invalidation.

Lazarus-931:
hi @aidiffuser, currently glm uses explicit pooling-cache state, which doesn't need this guard, closing :)

thanks for contribution!

PR #2078, opened 2026-08-28
GLM-5.3 - 2026-08-28

Lazarus-931:
GLM-5.3 support.

Benchmarks forthcoming

PR #2087, opened 2026-08-28
glm5_next: gathered sparse prefill + fp32 matmul indexer contraction (stacked on #2044)

aidiffuser:
**Stacked on #2044** — this branch includes #2044's commits and is built on its head (`55c85e35`) because it generalizes that PR's short-block verify gather; only the top commit is new. Suggest reviewing/merging after #2044; happy to rebase either way as it evolves.

### What

The DSA prefill currently consumes the indexer's selections by scattering them into a dense bool mask over the full context and re-expanding per-head K/V over the full length — O(S·T) attention wearing a sparse costume. Profiling our production serving put this masked attention at 47% of prefill wall at 64k, growing linearly with depth.

This PR generalizes #2044's `_gathered_verify_attention` (1 < L ≤ 8) into a query-chunked `_gathered_attention` used for every multi-query shape on a plain latent cache: gather each query's top-k selected latents and run one MQA SDPA over them — O(S·topk), near depth-flat. The short verify block is the single-chunk case of the same path, unchanged in behavior. Queries with no selected keys are forced to key 0 and their output rows zeroed, matching the dense path (the verify gather previously left such rows as implementation-defined fully-masked softmax). Quantized latent caches keep the dense masked path (K/V fetched from a quantized cache isn't a gatherable array).

Also: the indexer's head contraction becomes a batched matmul instead of elementwise-product + `mx.sum` — fp32 accumulation (selection ties resolve the way the fp32 reference does), no `[B, chunk, n_heads, P]` product materialization (halves the scorer transient), and structural immunity to the strided-reduction issue ml-explore/mlx#3784 should the query chunk ever be raised. The chunk stays at 512, so the memory profile is unchanged.

### Correctness

- New `test_glm5_next_prefill_gather_matches_dense`: chunked cached prefill + a verify-size block, gathered vs dense; multi-chunk gather equality; tail-off finiteness for the no-selected-keys row guard.
- All 18 glm5_next tests pass, including #2044's MTP verify tests, which now route through the generalized path.
- In our production fork the same rewrite is fp32-exact vs the dense path (3.9e-7 rel RMS, 0/512 tokens differing) and NIAH-validated at depth: 3/3 needle recall through 551k tokens.

### Performance

Layer-level A/B at real GLM-5.3-Flash dims (single DSA layer, random unquantized bf16 weights, chunked 4096-token cached prefill, M3 Ultra — absolute times are microbench numbers, the ratios are the signal), median chunk latency:

| context depth | dense masked | gathered | speedup |
|---|---|---|---|
| 8,192 | 720 ms | 973 ms | 0.74× |
| 32,768 | 1,699 ms | 1,030 ms | 1.65× |
| 65,536 | 5,235 ms | 1,025 ms | 5.11× |
| 131,072 | 11,103 ms | 1,571 ms | **7.07×** |

Dense decays linearly with depth (O(S·T)); gathered is near-flat (the residual growth is the indexer's own scorer). The measured crossover motivates `_GATHER_MIN_CONTEXT = 16384`: prefill chunks below ~16k keys keep the dense path, deeper chunks gather — decided per chunk during a long prefill, so both regimes get their best path automatically. Short verify blocks (L ≤ 8) always gather, exactly as in #2044.

Production (full 45-layer model, 2× M3 Ultra tensor-parallel): prefill 544 t/s @ 40k and 484 t/s @ 128k vs 473 / 241 on the dense path — **2.01× at 128k**, near depth-flat out to 500k+.

### Credit

The gather-the-selected-latents approach follows the prefill lineage of @jundot's omlx work on DSA-family models, which seeded the GLM-5.2 campaign that led here.


avlp12:
Ran an A/B of this branch against its #2044 base (`55c85e35`) on a large `glm5_next` target. The context-length crossover matches the PR's 128k+ framing — numbers below in case they're useful for review.

**Setup**: GLM-5.3-Flash (320B-A18B), local 4-bit g64 affine conversion (6-bit attention/shared experts); M3 Ultra 512GB; mlx 0.32.2; batch 1, greedy, 8 generated tokens after warmup, text-only synthetic prompt through the chat template; `prompt_tps` / peak memory as reported by `generate()`; single run per cell (run-to-run spread on this box ≈ ±2%).

| prompt tokens | base tok/s | this PR | Δ | peak GB (base → PR) |
|---:|---:|---:|---:|---|
| 8,195 | 442.4 | 463.3 | +4.7% | 191.9 → 191.9 |
| 16,399 | 419.4 | 394.2 | −6.0% | 195.1 → 194.6 |
| 32,781 | 350.2 | 297.0 | **−15.2%** | 201.0 → 195.4 |
| 65,548 | 256.9 | 261.5 | +1.8% | 213.0 → 196.4 |
| 131,080 | 158.7 | 242.3 | **+52.7%** | 237.2 → **198.9** |

(The 32k and 131k base cells were re-measured on `55c85e35` exactly; the 8k/16k/65k base cells came from a sibling branch of the same lineage, cross-checked against `55c85e35` to within ~1% at both 32k and 131k.)

Decode throughput is unaffected in every cell (25.5–26.2 tok/s both arms). At 131k the end-to-end prefill wall time drops 827s → 542s and peak memory drops ~38 GB.

Reading: on this hardware the crossover sits around ~64k — below it the gathered path costs up to −15% (32k), above it wins increasingly big, and the memory win grows with context either way. If that shape reproduces elsewhere, it may be worth gating the gathered path on prompt length (fall back to the dense-mask path below a threshold — e.g. some multiple of `index_topk`, or a tunable) so short/mid prompts keep today's behavior while long prompts get the +50%.

Harness and receipts are public in [avlp12/local-llm-serving](https://github.com/avlp12/local-llm-serving) (`ports/glm53flash-mlx/serving/repro_mlxvlm_dflash2.py`). Happy to run more cells (e.g. 96k, or other block sizes) if helpful.

Lazarus-931:
> Ran an A/B of this branch against its #2044 base (`55c85e35`) on a large `glm5_next` target. The context-length crossover matches the PR's 128k+ framing — numbers below in case they're useful for review.
> 
> **Setup**: GLM-5.3-Flash (320B-A18B), local 4-bit g64 affine conversion (6-bit attention/shared experts); M3 Ultra 512GB; mlx 0.32.2; batch 1, greedy, 8 generated tokens after warmup, text-only synthetic prompt through the chat template; `prompt_tps` / peak memory as reported by `generate()`; single run per cell (run-to-run spread on this box ≈ ±2%).
> 
> prompt tokens	base tok/s	this PR	Δ	peak GB (base → PR)
> 8,195	442.4	463.3	+4.7%	191.9 → 191.9
> 16,399	419.4	394.2	−6.0%	195.1 → 194.6
> 32,781	350.2	297.0	**−15.2%**	201.0 → 195.4
> 65,548	256.9	261.5	+1.8%	213.0 → 196.4
> 131,080	158.7	242.3	**+52.7%**	237.2 → **198.9**
> (The 32k and 131k base cells were re-measured on `55c85e35` exactly; the 8k/16k/65k base cells came from a sibling branch of the same lineage, cross-checked against `55c85e35` to within ~1% at both 32k and 131k.)
> 
> Decode throughput is unaffected in every cell (25.5–26.2 tok/s both arms). At 131k the end-to-end prefill wall time drops 827s → 542s and peak memory drops ~38 GB.
> 
> Reading: on this hardware the crossover sits around ~64k — below it the gathered path costs up to −15% (32k), above it wins increasingly big, and the memory win grows with context either way. If that shape reproduces elsewhere, it may be worth gating the gathered path on prompt length (fall back to the dense-mask path below a threshold — e.g. some multiple of `index_topk`, or a tunable) so short/mid prompts keep today's behavior while long prompts get the +50%.
> 
> Harness and receipts are public in [avlp12/local-llm-serving](https://github.com/avlp12/local-llm-serving) (`ports/glm53flash-mlx/serving/repro_mlxvlm_dflash2.py`). Happy to run more cells (e.g. 96k, or other block sizes) if helpful.

can u explain why a big jump in 132K cntxt?

aidiffuser:
Thanks for running this — the memory column is the most informative part.

Mechanism: the dense path computes attention against every key and masks the unselected ones, so work is O(L·S) and it materializes a [B, H, L, S] mask plus re-expanded K/V. The gathered path materializes only the `index_topk` (2048) selected latents per chunk, so attention work is O(L·topk) — flat in S. The ratio is roughly S/topk: ~4× at 8k, ~64× at 131k. End-to-end gain is much smaller because attention is only one term — MoE dominates elsewhere and the indexer still scores all S keys.

Your peak-memory column shows it directly: base grows 191.9 → 237.2 GB with context while this branch stays 191.9 → 198.9. That flat line is the dense mask + expanded K/V never being built. At 131k the base arm carries ~38 GB of extra transients on top of a resident 320B model, so it pays allocator pressure on top of the extra FLOPs — which is why 131k jumps more than a smooth extrapolation from 65k would suggest.

On the crossover: the gate you're describing is already in the branch — `_GATHER_MIN_CONTEXT = 16384`. Your data says it's calibrated too low; 32k sits above it and loses 15%. I set it from a single-layer microbench at real dims (which showed a win at 32k); your end-to-end numbers are the better basis for a default and point at ~64k on this hardware. Happy to raise it — a 96k cell would help pin the knee if you're still set up.

One note: `avlp12/local-llm-serving` 404s for me — is it private?

avlp12:
Follow-up with numbers behind a gate suggestion. I re-ran the regressing cells with `_GATHER_MIN_CONTEXT` raised 16384 → 65536, same setup as above (mlx-vlm default `prefill_step_size=2048`; single run per cell, and my repeat spread on identical configs was 0.5–1.1%):

| prefill tok/s | dense | gate 16384 | gate 65536 |
|---:|---:|---:|---:|
| 16k | 419.4 | 394.2 | 421.9 |
| 32k | 350.2 | 297.0 | 351.1 |
| 131k | 158.7 | 242.3 | 234.5* |

At 16k/32k the raised gate simply keeps the dense path, so the mid-context regressions go away — parity with dense, within noise. Because the gate is evaluated per chunk, 131k is not literally unchanged: the chunks between 16k and 64k keys revert to dense, costing ~3% of the full-gather win (+48% vs dense instead of +52.7%) and giving back part of the peak-memory saving (228 GB vs 199 full-gather / 237 dense). (*The 131k gate-65536 cell was re-measured on an identical second M3 Ultra 512GB; dense repeats across the two boxes agree within ~2%.)

That puts the end-to-end crossover at ~32k–64k keys rather than the ~16k per-chunk microbench (same hardware and dims, so a bench-vs-e2e gap I haven't isolated). A default — or a tunable — in the 32768–65536 range may suit mixed workloads better than 16384. Harness: [avlp12/local-llm-serving](https://github.com/avlp12/local-llm-serving) (`ports/glm53flash-mlx/serving/repro_mlxvlm_dflash2.py`).

avlp12:
Quick data addendum: I made the gate an env-tunable and measured **32768** on the same box as the starred cell above (single run per cell; my branch also carries decode-side kernel patches that do not touch the prefill path):

| prefill tok/s | gate 65536 | gate 32768 |
|---:|---:|---:|
| 32k | 348.6* | 341.2 |
| 65k | 248.8 | **276.1 (+11%)** |
| 131k | 234.5 | **246.3 (+5%, +57% vs dense)** |

(*with a 65536 gate a 32k prompt stays fully dense by construction.)

At 131k the 32768 gate also edges the always-gather 16384 configuration (242.3 on the twin box): dense wins the 16k–32k band, so a middle gate beats both extremes. Cost is −2% at 32k, inside my repeat spread. On this hardware the sweet spot looks like ≈32k keys — consistent with the crossover estimate above, and maybe a reasonable default if it reproduces elsewhere.

aidiffuser:
Adopted — thanks, this is a much better basis for the default than the single-layer microbench I set 16384 from.

Pushed: default raised to 32768, and the threshold is now readable from `MLX_VLM_GLM5_GATHER_MIN_CONTEXT` (0 always gathers, a large value keeps the dense path; an unparsable value warns and falls back rather than raising at import). Your table is in the comment with the second-box footnote and the same-box 32k/65536 = 348.6 control, and a test covers the reader. I also merged current main in — #2078 landed while this was in flight and the only conflict was a one-line list entry.

Two corrections to my earlier reply while I'm here. The ~2% at 32k is real, not noise — your two gate-65536 cells at 16k and 32k can't fire at those lengths, so they pin the repeat spread at <= 0.6%, and 2% is about what you'd expect from the ~6% of that prompt's tokens sitting above the threshold. And I've scoped the "decode unchanged" claim to the cells where it was measured, since your 32768 arm also carries decode-side kernel patches.

The knob matters more than the constant here: on a tensor-parallel deployment of this architecture (2 nodes, prefill_step_size 4096, 6-bit) I measure the crossover near 16384 keys — half your value — with the same qualitative shape, dense winning the shallow band and gather winning above it. Two configurations, two optima, so a fixed constant can't serve both.

A 96k cell would still be useful if you're set up. I'd also welcome a second opinion on the bench-vs-e2e gap: an isolated single-layer bench reads ~16k on your box too, and I don't have an explanation for the difference.

avlp12:
**96k cell, single box** — M3 Ultra 512GB, my 4-bit-class build (mixed 4/6/8-bit weights, unquantized latent cache), chunked prefill, 2048-token chunks, gate 32768. Single run for this cell; a shorter run at identical settings overlaps it on the first 24 chunks and agrees within 1.7%.

98,304 tokens of prefill in 375.5 s = 261.8 tok/s mean. Per chunk, the dense band climbs 4,088 ms (first chunk) → 7,420 ms (cache 28,672); gather takes over on the chunk that lands the cache at exactly 32,768 (8,592 ms) and is essentially flat from there — 9,073 ms at cache 96,256, i.e. +5.6% over the following 65,536 keys.

The dense counterpart at that depth is arithmetic, not measurement. Fitting the dense chunks gives **3.92 s + 0.253 s per 2048-key chunk** (R² 0.996), so chunk 47 projects to ≈15.8 s against the measured 9.07 s (≈1.74×), and the 48 chunks sum to a projected 473 s all-dense against the measured 375 s — ≈21% less prefill time. It is fit-sensitive: fitting only this run's own dense chunks gives 0.246 s/chunk → 15.5 s, 1.71×, ≈19%. I did not get a dense control at 96k — the attempt was clearly heading past the memory I had and I stopped it before it finished, so I have no peak-memory number to report for it.

**On the bench-vs-e2e gap.** Part of it turned out to be my own units, and part of it is a real gap I can't close either.

1. **Not the same quantity.** The gate is evaluated per chunk against the post-append cache length (`kv_latent.shape[2]`), not the prompt length, so at prompt length P only the keys above G ever take the gather path, and whole-prompt tok/s — which is what my earlier tables reported — blends both regimes and reads high. That's on me; "~32k-64k" was never directly comparable to your per-chunk 16k. A matched pair of runs at gate 16384 and 32768 puts the knee exactly at the gate (the first gathered chunk is the one whose post-append cache first reaches G), which is also a decent control that it's dense-attention depth doing the growing rather than something upstream of the gather choice.

2. **Same quantity, gap survives.** Per chunk on the full model, the fit above meets an 8.67 s gather plateau at **≈38k keys**, against your ~16k. So ≈2.3× is left over at matched depth. Caveat on every crossover in this comment: with the gate at 32768 no dense chunk is measured above 28,672 keys, so these are a fitted dense line extended past the data.

3. **One thing it isn't.** Only 11 of the 45 layers are DSA, so a per-layer win is diluted ~4× at model level — that's why 1.74× per chunk is only ~21% of prefill. I'd been half-assuming that could also explain the crossover, and it can't: it's a common factor on both arms, so it cancels at the break-even.

4. **What does line up: the query-chunk size.** Your layer bench is 4096-token query chunks, and your TP deployment is `prefill_step_size` 4096 — both read ~16k. Both of my lanes are 2048-token chunks and both read higher. Dense is O(L·Kv); gather is O(L·topk) with topk pinned at 2048 plus a per-chunk cost that amortizes over L, so a larger L should move the crossover *down*. Rough check on my own fit: at L=4096 with a fully L-independent gather cost the crossover would fall to ≈19k, and with a fully L-proportional one it would stay at ≈38k — your 16k sits near the first end. That's a hypothesis, not a measurement. Happy to re-run my curve at 4096-token chunks if that'd be a useful test; if it holds, the right default gate tracks `prefill_step_size`, and 32768 is the 2048 answer, which is the upstream default.

**"Two configurations, two optima" reproduces on my fleet — but not in your direction.**

| lane | dense per chunk | gather plateau | crossover |
|---|---|---|---|
| single box | 3.92 s + 0.253 s·chunk | 8.67 s | ≈38k |
| TP=2 | 2.31 s + 0.123 s·chunk | 6.39 s | ≈68k |

Same code, same build, 2048-token chunks on both. Splitting the attention heads halves the dense slope (2.06×) while the gather plateau drops only 1.36×, so the crossover moves right; I default my TP lane to 65536 and leave single-box at 32768. Two honest limits: the plateaus are measured but the crossovers are fits past the last dense chunk (28,672 keys), and this moves *opposite* to your ~16384 under TP. I can't reconcile those two — neither the chunk size nor the quantization matches between my TP and yours, and the chunk size is the same axis as point 4 above.

avlp12:
**Correction to the numbers in my previous comment.** My long-context harness was prefilling a degenerate repeated-token prompt. That is precisely the benching rule I hold myself to and I broke it, so the absolutes above are flattered. The model routes 8 of 288 experts per token; a single repeated id collapses that routing onto a narrow set of experts (measured per layer: 103–280 of 288 experts active on the repeated-token prompt versus 270–288 on real code and prose), so a prompt of that shape systematically understates MoE prefill cost. Re-run on real source and prose, same build, same gate, same 2048-token chunks:

- 96k single-box prefill: 375.5 s → **425.7 s**, 261.8 → **230.9 tok/s**. Excluding the first chunk from both runs (the synthetic TP run's carried a 20.8 s compile warm-up, so raw totals aren't comparable there), the real-text penalty is **+13.1% single box and +11.7% TP**.
- Dense fit **4.70 s + 0.273 s/chunk**, gather plateau **9.88 s** single box; **2.71 + 0.129**, plateau **7.07 s** under TP.

**Nothing you merged is affected.** The re-derived single-box crossover is **≈39k** — it was 38–40k before depending on which chunks you fit, so the drift is under 4% and I won't claim a sign for it. The TP crossover, where both arms are the same run length, moves +1.3% to ≈69k, so I'm leaving my TP lane at 65536. The knee still lands exactly at the gate. The 96k gather region is still near-flat: **+4.4%** end to end across the 65,536 keys above the gate. Per-chunk advantage at 96k is **≈1.8×** and the whole-prefill saving **≈20%** (still extrapolated dense against measured gather; dropping one chunk from the fit window moves them to 1.7× and 16%). And the ≈2.3× gap between my full-model crossover and your single-layer bench is untouched — that question wasn't an artifact of the bad prompt.

**Two limits on the correction itself.** Real text is much noisier than the degenerate prompt: the dense band fits at R² 0.92 rather than 0.999, and the gather plateau now scatters ≈21% chunk-to-chunk against ≈6% before, so single-chunk figures should be read loosely. And the shift doesn't decompose as cleanly as I'd like: the constant term rises **17–27%** while the dense slope moves somewhere between −3% and +9% depending on the fit window — so the real-text penalty sits almost entirely in the routing-sensitive constant, and the attention slope isn't resolvably different. I'd wanted to say the MoE term is a common offset that cancels at the break-even; what the data actually supports is the weaker "the crossover moves by a few percent at most", and the plateau rose more in absolute ms than the dense intercept did, so it isn't a clean common offset.

aidiffuser:
Excellent catch on the repeated-token prompt, and thank you for posting the correction unprompted — collapsing MoE routing onto a narrow expert set is exactly the failure mode that makes long-context prefill numbers look better than they are, and the fact that your crossovers barely moved (≈39k, TP ≈69k) is a much stronger result than if you'd never found it.

Your point 1 resolves more than you're giving it credit for. To confirm the quantity: our ~16k is also per-chunk against the post-append cache length, so we are measuring the same thing.

**Your point 4 predicts our number.** You estimate the crossover falls to ≈19k at L=4096 if the gather's per-chunk cost is largely L-independent. Our ladder runs at exactly L=4096 and brackets the knee at 12,288 (dense +6.8%) / 16,384 (tie, 0.7%) / 20,480 (gather +7.8%). Our chunks land on multiples of 4096, so ±4096 is our whole resolution — but ≈19k sits inside it. You derived that from your fit without seeing our per-chunk data.

**One correction on the TP comparison.** We have no single-node datapoint, so we can't speak to what tensor parallelism does to *our* crossover — nothing of ours moves in either direction. The like-for-like comparison is TP=2 against TP=2: your 69k at L=2048 versus our ~16.5k at L=4096. Same topology, different chunk size, which is your point 4 again rather than a contradiction. The direction agrees; the magnitude doesn't fully — a 2× L increase moving 69k→16.5k is steeper than the ~2× your single-box fit predicts, so something else is still in there.

Caveat I should state plainly: our lane runs a **different gather implementation** (exo vendors its own `glm5_next`, indexing the un-broadcast latent rather than sourcing from `broadcast_to`) and a different quantization, so any absolute cross-lane comparison carries that. It's also a candidate for the residual ≈2.3×.

Yes, please do the 4096 re-run — that's the decisive test. I'll offer the reciprocal: our cluster's step size is a single env var, so I can re-ladder at 2048 and give you a matched-L point from the TP side. If both land where the hypothesis says, "the gate tracks `prefill_step_size`" is the right shape, and 32768 is correct as the default because `DEFAULT_PREFILL_STEP_SIZE` is 2048.

avlp12:
Agreed on all three points — same per-chunk quantity against the post-append cache length, TP=2 against TP=2 is the like-for-like comparison, and the different gather implementation and quantization are a live candidate for the residual.

I have scheduled the L=4096 re-run on my box. Once it completes I plan to post a dense-vs-gather ladder at prefill_step_size 4096 covering 8k–32k in 4096-token steps, so the expected knee is bracketed at the available resolution, plus an L=2048 control measured under the same setup so the two ladders aren't separated by other changes — with the checkpoint, commit hashes and commands. If your reciprocal ladder at 2048 lands too, the comparison will cover both L values from both sides; that's the test for "the gate tracks `prefill_step_size`", and if it holds I agree 32768 is the right default for `DEFAULT_PREFILL_STEP_SIZE = 2048`.

avlp12:
L=4096 re-run, and a correction to my own earlier crossover first.

**Correction.** The ≈39k I reported on 2026-09-01 was not a measured crossover: with the gate at 32768 no dense chunk was ever measured above 28,672 keys, so it was where a fitted dense line (4.70 s + 0.273 s per 2048-key chunk) met the measured gather plateau (9.88 s per chunk). Two things were wrong with treating that as a point. It was an extrapolation whose value moves by ≈3,000 tokens depending on which chunks you fit the plateau from. And it came from a build that predated a query-chunk boundary fix in my gather path; the pre-fix gather chunk at L=2048 was 9.88 s; post-fix receipts read 6.6–7.2 s (1.38–1.49× cheaper), while the dense side was essentially unchanged — so the measured crossover moved with the gather curve. Measured directly on the current build — gate pinned to 0 and to 10⁹, dense and gather timed at every depth — the current L=2048 ladder places the single-box crossover at **≈12,288**, bracketed in (10,240–12,288], and it agrees with my separate gate sweep, which swept the gate rather than pinning each arm at every depth. So please discard ≈39k. I also retract the ≈69k TP figure as an unmeasured extrapolation of the same kind; this single-box run does not re-establish it.

**Setup.** M3 Ultra 512 GB, single box, checkpoint `avlp12/GLM-5.3-Flash-Alis-MLX-4bit` @ `7ce8cd0b` (MLX 4-bit-class build of `zai-org/GLM-5.3-Flash`), real text with distinct routing (not repeated tokens), chunked prefill, per-chunk wall time paired dense vs gather, two cycles each, `MLX_VLM_GLM5_GATHER_MIN_CONTEXT` set to 10⁹ (dense) or 0 (gather) per arm; crossover = the shallowest depth where gather beats dense in both cycles, resolved to one stride.

**L=4096 (prefill_step_size 4096):** crossover **8,192**, bracket (4,096–8,192]. Dense grows linearly (9.2 → 20.9 s per 4096-key chunk from depth 0 to 32k); gather is flat (11.5 → 12.3 s).

| depth | dense ms (2 cycles) | gather ms (2 cycles) | gather wins |
|--:|--:|--:|:--:|
| 0 | 9,236 / 10,025 | 11,570 / 12,240 | no |
| 4,096 | 10,479 / 11,350 | 11,513 / 12,084 | no |
| 8,192 | 11,945 / 12,828 | 11,488 / 12,100 | yes |
| 16,384 | 14,679 / 15,851 | 11,840 / 12,277 | yes |
| 32,768 | 20,863 / 22,911 | 12,293 / 12,665 | yes |

**L=2048 (control, same box, same day):** crossover **12,288**, bracket (10,240–12,288]; dense 5.1 → 9.4 s per 2048-key chunk over 0–22k; gather 5.2/5.3 s at depth 0, then flat at 6.9–7.3 s.

**On the hypothesis.** Halving L from 4096 to 2048 moved the crossover from 8,192 to 12,288 — a ratio of 1.5, not 2. So "the gate tracks `prefill_step_size`" is right in direction and not proportional in magnitude; a gate default that assumes proportionality would be off by that much. Your ladder at L=4096 bracketed the knee at 12,288–20,480; my result is one 4,096-token stride below the lower end of that bracket. Given the ±4,096 resolution, the gap could reflect the different gather implementation and quantization rather than a disagreement about the mechanism.

**One labelled extra, not part of the comparison:** on my fork with a separate attention-layout change enabled (absent from this PR's code path), gather won at every depth including the first chunk, so the crossover was below the ladder floor; with that change disabled the same build reproduced 8,192 exactly. The fork's default gate was retuned from 12,288 to 6,144 after that sweep; it is fork-specific and is not a measurement of this PR.

Full per-depth tables for all four ladders are in my receipts; happy to paste any of them.

Blaizzy:
Thank you very much, @aidiffuser!

I will have to close becuase it already supported on main 

We still have lots of issues that need fixing and models to port, so please feel free to pick and send us another one.

PR #2089, opened 2026-08-28
ci: add Release Please for faster release process

Lazarus-931:
This makes it easier for release cycles to be done under maintainer discretion automatically.

Release Please prepares versioned pull requests; merging one publishes verified artifacts through trusted automation.


Blaizzy:
@lucasnewman do we need this right now?

Lazarus-931:
> @lucasnewman do we need this right now?

i was going to get back to this,  I don't think so anymore.

would complicate it much more

PR #2090, opened 2026-08-28
Keep quantized APC checkpoints packed

Blaizzy:
## Summary

- keep uniform and TurboQuant exact-APC checkpoints in their native packed representation through snapshot, disk restore, and batch merge
- persist only the TurboQuant KV codec data required for deterministic integer and fractional-bit restores
- retain the existing legacy float-snapshot fallback and APC schema

This is a focused follow-up to #1960 and uses its generic cache snapshot/restore/merge contracts. It adds no QSA-specific serialization path, ownership-transfer flags, KV-policy layer, or support for TurboQuant codec families that KV caches cannot produce. Block-mode APC remains unchanged.

## Why

Exact APC previously dequantized compressed KV to float32 for the checkpoint, then requantized it after lookup. For one TQ4 layer with 4 KV heads and head dimension 128 on an M5 Max (median of 7 warm runs):

| Tokens | Float round trip | Native packed | Speedup | Float checkpoint | Packed checkpoint |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 4,096 | 3.50 ms | 0.81 ms | 4.34x | 16.00 MiB | 2.06 MiB |
| 8,192 | 3.50 ms | 0.74 ms | 4.71x | 32.00 MiB | 4.12 MiB |
| 16,384 | 5.02 ms | 0.91 ms | 5.53x | 64.00 MiB | 8.25 MiB |

## Tests

- 464 passed, 1 skipped across the APC and TurboQuant suites under Python 3.10
- regression tests force dequantization/requantization to fail during exact snapshot and merge
- disk round trips cover uniform 8-bit, TurboQuant 4-bit, and TurboQuant 3.5-bit packed states
- mixed warm/cold batch merge and first-token append are covered


mlx_vlm/turboquant.py
```diff
@@ -4844,12 +4844,39 @@ def _select_outlier_indices(
...
+    ) -> None:
         self.bits = bits
         self.mode = mode
-        self.dim = tensor.shape[-1]
+        self.dim = int(dim)
         self.lower_bits = math.floor(bits)
         self.upper_bits = math.ceil(bits)
-        low_idx, high_idx = _select_outlier_indices(tensor, bits)
+        low_idx = np.asarray(low_idx, dtype=np.int32)
+        high_idx = np.asarray(high_idx, dtype=np.int32)
```
lucasnewman: Minor: why bounce through `np.asarray` here?

lucasnewman (approved):

PR #2091, opened 2026-08-29
Add GLM-5.3 token-budget image processor (glm5_next)

avlp12:
## Summary

#2030 already added `glm5_next` language/vision. This does **not** replace that architecture. It adds the missing image processor so `load()` / `generate()` can feed a real `image_grid_thw` instead of reusing GLM-OCR helpers.

## Why a separate processor

GLM-OCR `smart_resize` takes a **pixel** budget. GLM-5.3's `min_image_tokens` / `max_image_tokens` are **token** counts, converted with `temporal_patch_size * (patch_size * merge_size)²`. Passing `16` / `8000` into the OCR helper shrinks a native 448² tile; the GLM-5.3 helper keeps it at 448² → 256 LLM tokens.

Replacement of `<|image|>` is `image_token * N`. A `while token in text` loop reconsumes the expansion and overshoots `image_grid_thw`. Expansion splits on the original slots first.

~~The Hub `chat_template.jinja` on GLM-5.3-Flash currently stubs image content with an "unable to process" reminder. `Glm5NextProcessor.from_pretrained` loads `chat_template_vlm.jinja` from this package.~~ **Edit 2026-09-05:** that was true of the 2026-08-26 Hub revision; the current one (`690b7052`, 2026-09-04) already emits `<|begin_of_image|><|image|><|end_of_image|>`, so the packaged template has been dropped from this PR and the processor relies on the checkpoint's own template.

`prompt_utils.MODEL_CONFIG` had no `glm5_next` entry (so `apply_chat_template` would miss the model). It is now `LIST_WITH_IMAGE_FIRST`, same as GLM-OCR / GLM-4V.

Video preprocessing is still `NotImplementedError`.

## Tests

Weight-free, no Hub download. Includes a numeric parity test against transformers' `Glm5NextImageProcessor` on `examples/images` (`image_grid_thw` exact, `pixel_values` max abs diff < 1e-5; skipped if the class is not importable):

```
python -m unittest mlx_vlm.tests.test_glm5_next_processor
```

Also re-ran `TestModels.test_glm5_next_language_model` after the `__init__.py` processor import.

## Notes

Does not change `language.py` / `vision.py` / `vision_model` naming from #2030.

Made with [Cursor](https://cursor.com)


mlx_vlm/models/glm5_next/processing.py
```diff
@@ -0,0 +1,528 @@
+"""GLM-5.3-Flash image processor: token-budget smart_resize + pad-after-scale.
```
Lazarus-931, line 1: doesn't this just duplicate what's inside `mlx_vlm/models/glm_ocr/processing.py`

Lazarus-931 (changes requested):
thanks @avlp12 for the pr, see comment, but also:

do you have a comparison that shows pixel_values/image_grid_thw numerically matches the transformers Glm5NextImageProcessor on a real image?

avlp12:
Thanks for the review — both points answered below. The branch is updated (`2b5f747f`, GitHub-signed).

**Numeric parity with transformers (Q2).** Measured on the updated branch: four real images from `examples/images/`, transformers 5.16.1 `Glm5NextImageProcessor` built from the checkpoint's `processor_config.json` versus this branch's processor built from the same kwargs, no weights involved.

| image (W×H) | transformers `image_grid_thw` | this branch | `pixel_values` shape | max abs diff | GLM-OCR pure-NumPy path, matching config fields |
|---|---|---|---|---:|---|
| cats.jpg (640×480) | [[1, 36, 46]] | [[1, 36, 46]] | same | 4.77e-07 | [[1, 34, 46]] |
| demo_image1.jpg (1700×2250) | [[1, 162, 122]] | [[1, 162, 122]] | same | 4.77e-07 | [[1, 160, 122]] |
| scene_1.jpg (692×1024) | [[1, 74, 50]] | [[1, 74, 50]] | same | 4.77e-07 | [[1, 74, 50]] |
| graph.png (1388×764) | [[1, 56, 100]] | [[1, 56, 100]] | same | 4.77e-07 | [[1, 54, 100]] |

All four grids and `pixel_values` shapes match exactly; the maximum absolute pixel-value difference is 4.77e-7. I added a weight-free parity test to `mlx_vlm/tests/test_glm5_next_processor.py` (two of these images; skipped if that transformers class is not importable). All seven tests pass.

**Duplication with `glm_ocr/processing.py` (Q1).** You're right that there was shared helper logic — an earlier estimate put the overlap at roughly 90–110 of the file's 528 lines. I now import `_to_numpy_image` and the bicubic resize helper from `qwen3_vl`, as `glm_ocr` does; `_resize_chw` is a thin float-safe wrapper, and the JSON loaders and the patchify block remain local. Net change in `processing.py`: −22 lines (+16/−38). What stays GLM-5.3-specific: the budget is in tokens (`min_image_tokens`/`max_image_tokens`) rather than pixels, with ceil-alignment and a binary search to fit; resize-to-content then pad-to-target; and `<|image|>` expansion that splits on the original slots. The last column above is why it is not a drop-in: with the matching configuration fields passed to GLM-OCR's pure-NumPy processor (`use_transformers_backend=False`), its grid differs from transformers on three of these four images.

I also removed the packaged `chat_template_vlm.jinja` from this PR: the current Hub template (`690b7052`) already emits the image markers, so the sentence in my original description was stale, and I have edited the description to say so.

Lazarus-931:
hi @avlp12, closing this, since glm 5 already has a native video processor :)

appreciate the contribution nonetheless!

PR #2096, opened 2026-08-29
Prime speculative drafters from the whole prompt during chunked prefill

### LINKED ISSUE #2022: DeepSeek-V4 MTP speculative decoding ignores chunked prefill

Lazarus-931:
Fixes #2022

## Summary

Chunked prefill requested the speculative capture kwargs only on the final
prefill call. Drafters that read target hidden states past the last prompt
position were therefore primed from a one-token prompt.

Three drafter families are affected: own-cache MTP (`deepseek_v4_mtp`,
`qwen3_5_mtp`, `inkling_mtp`) and EAGLE3 prime a persistent drafter cache
through `prefill_from_target_hidden`; DFlash cross-attends its first draft
block over the prompt hidden. Shared-KV MTP (`gemma4_assistant`) only reads
the final position and is unaffected.

This passes the hidden capture on every chunk, rejoins the captures, and
hands the round loops the full prompt ids. Applied to `generate_step` and to
the batch path in `PromptBatch`, which had the same gap.

The per-model `chunked_prefill_policy` speculative branches were four copies
of the same predicate, checking capture flags the caller had already set.
They are replaced by a capability check on `rollback_speculative_cache`, the
hook the round loops already require. That also enables chunked prefill for
DeepSeek-V4 MTP. Model-specific media and padding rules are unchanged.

Only the hidden capture is requested per chunk. Shared-KV state still comes
from the final call, which reads the completed target cache.

## Benchmarks

Qwen3.5-9B-4bit + its MTP drafter, both built locally from `Qwen/Qwen3.5-9B`.
128 new tokens, greedy, `prefill_step_size=512`.

| prompt | metric | unchunked | chunked (main) | chunked (this PR) |
|---|---|--:|--:|--:|
| 2048 | % of drafted accepted | 82.3% | 57.1% | 82.3% |
| 2048 | accepted tokens/round | 2.65 | 2.13 | 2.65 |
| 4096 | % of drafted accepted | 64.9% | 55.4% | 82.3% |
| 4096 | accepted tokens/round | 2.29 | 2.10 | 2.65 |
| 8192 | % of drafted accepted | 69.8% | 68.5% | 70.8% |
| 8192 | accepted tokens/round | 2.40 | 2.37 | 2.42 |

At 2048 the chunked run reproduces the unchunked counters exactly, which
means an identical decode trajectory. The 4096 row where this PR exceeds the
unchunked baseline is trajectory divergence, not a gain: chunked prefill
accumulates KV in a different order, so greedy decoding can diverge. The
claim is parity, not improvement.

Qwen3.5-27B-4bit + `z-lab/Qwen3.5-27B-DFlash`, same settings. DFlash only
loses its first round, so the delta is small.

| prefill | % of drafted accepted | accepted tokens/round |
|---|--:|--:|
| unchunked | 46.4% | 2.54 |
| chunked (main) | 45.0% | 2.49 |
| chunked (this PR) | 49.4% | 2.65 |

Peak memory at 8192 tokens goes from 7.6 GB to 10.2 GB for the chunked run,
against 17.4 GB unchunked. The cost is non-linear in prompt length (+0.0 GB
at 2048, +0.3 GB at 4096, +2.6 GB at 8192), consistent with the drafter's
own prefill forward now running over the whole prompt instead of one token.

No EAGLE3 drafter was available locally, so that family is covered by unit
tests only.

## Reproducing

```sh
git clone https://github.com/Blaizzy/mlx-vlm && cd mlx-vlm
uv venv && uv pip install -e .

# target + MTP drafter, no checkpoint download beyond the base model
uv run mlx_vlm convert --hf-path Qwen/Qwen3.5-9B --mlx-path ./q35-9b-4bit \
  -q --q-bits 4 --q-group-size 64 --mtp

# acceptance is printed at the end of a verbose CLI run
uv run mlx_vlm generate --model ./q35-9b-4bit --draft-model ./q35-9b-4bit-mtp \
  --draft-kind mtp --prompt "$(python -c 'print("the harbour ledger " * 900)')" \
  --max-tokens 128 --temperature 0 --verbose
uv run mlx_vlm generate --model ./q35-9b-4bit --draft-model ./q35-9b-4bit-mtp \
  --draft-kind mtp --prompt "$(python -c 'print("the harbour ledger " * 900)')" \
  --max-tokens 128 --temperature 0 --prefill-step-size 512 --verbose
```

## Tests

`test_chunked_prefill_primes_drafter_from_whole_prompt` is parametrized over
the four drafter families and fails on `main` for all four. Added coverage
for `splice_prompt_hidden`, `drafter_consumes_prompt_hidden`, and the new
gate. `TestCohereCompass::test_visual_deepstack_allows_chunked_prefill` now
asserts through `_chunked_prefill_enabled` instead of the deleted method.

```
pytest ./tests --ignore=tests/test_smoke.py
2897 passed
```

## Notes

`_mtp_rounds_batch` never calls `prefill_from_target_hidden`, so own-cache
MTP drafters run unprimed at batch size > 1 regardless of chunking. This PR
carries the correct hidden through the batch path but does not change that
round loop. Worth a separate issue.

Separately, `convert --mtp -q` cannot extract a drafter: `--q-group-size`
defaults to `None` and `convert.py` passes it straight to the splitter while
the main quantize path guards with `q_group_size or 64`. Passing
`--q-group-size 64` works around it.

### DIFF
--- a/mlx_vlm/generate/ar.py
+++ b/mlx_vlm/generate/ar.py
@@ -26,11 +26,14 @@ from ..sample_utils import (
     top_p_sampling,
 )
 from ..speculative.utils import (
+    drafter_consumes_prompt_hidden,
     make_speculative_prompt_cache,
+    prompt_chunk_capture_kwargs,
     run_speculative_rounds,
     run_speculative_server_rounds,
     speculative_hidden_state,
     speculative_prefill_kwargs,
+    splice_prompt_hidden,
 )
 from ..turboquant import BatchTurboQuantKVCache, turboquant_enabled
 from ..utils import group_images_by_shape, prepare_inputs, should_add_special_tokens
@@ -345,7 +348,10 @@ def generate_step(
 
     # Speculative decoding setup
     last_outputs = None
+    speculative_prompt_ids = input_ids
+    prompt_hidden_chunks: List[Optional[List[mx.array]]] = []
     speculative_prefill_capture_kwargs = {}
+    capture_prompt_hidden_per_chunk = False
     if draft_model is not None:
         from ..speculative.drafters import validate_drafter_compatibility
 
@@ -353,6 +359,10 @@ def generate_step(
         speculative_prefill_capture_kwargs = speculative_prefill_kwargs(
             draft_kind, draft_model
         )
+        capture_prompt_hidden_per_chunk = drafter_consumes_prompt_hidden(
+            draft_kind, draft_model
+        )
+        chunk_capture_kwargs = prompt_chunk_capture_kwargs(draft_kind, draft_model)
         # Reset stale mRoPE state from any previous generation.
         lm = model.language_model if hasattr(model, "language_model") else model
         if hasattr(lm, "_position_ids"):
@@ -470,6 +480,9 @@ def generate_step(
         ) or (
             checkpoint_len is not None and 0 < checkpoint_len < inputs_embeds.shape[1]
         )
+        # Drafters are primed from the whole prompt, which the chunk loop below
+        # consumes token by token; snapshot the ids before that happens.
+        speculative_prompt_ids = input_ids
         if prefill_step_size is not None and should_chunk:
             # Chunked prefill with embeddings
             total_tokens = inputs_embeds.shape[1]
@@ -489,13 +502,26 @@ def generate_step(
                     chunk_kwargs = kwargs
                     if getattr(model.language_model, "supports_logits_to_keep", False):
                         chunk_kwargs = {**kwargs, "logits_to_keep": 1}
-                    model.language_model(
+                    if capture_prompt_hidden_per_chunk:
+                        chunk_kwargs = {
+                            **chunk_kwargs,
+                            **chunk_capture_kwargs,
+                        }
+                    chunk_outputs = model.language_model(
                         inputs=input_ids[:, :n_to_process],
                         inputs_embeds=inputs_embeds[:, :n_to_process],
                         cache=prompt_cache,
                         n_to_process=n_to_process,
                         **chunk_kwargs,
                     )
+                    if capture_prompt_hidden_per_chunk:
+                        chunk_hidden = getattr(chunk_outputs, "hidden_states", None)
+                        # Eval now: an unevaluated capture pins the whole chunk's
+                        # activation graph until the drafter is primed.
+                        if chunk_hidden:
+                            mx.eval(chunk_hidden)
+                        prompt_hidden_chunks.append(chunk_hidden)
+                    chunk_outputs = None
                     quantize_cache_fn(prompt_cache)
                     mx.eval([c.state for c in prompt_cache])
                     processed_tokens += n_to_process
@@ -519,11 +545,12 @@ def generate_step(
 
     # Speculative decoding
     if draft_model is not None:
+        last_outputs = splice_prompt_hidden(prompt_hidden_chunks, last_outputs)
         yield from run_speculative_rounds(
             model,
             draft_model,
             prompt_cache,
-            input_ids,
+            speculative_prompt_ids,
             y,
             logprobs,
             last_outputs,
@@ -1682,6 +1709,12 @@ class PromptProcessingBatch:
         self.draft_kind = draft_kind
         self.draft_block_size = draft_block_size
         self.greedy_sampling = greedy_sampling
+        self._capture_prompt_hidden_per_chunk = (
+            draft_model is not None
+            and draft_kind is not None
+            and drafter_consumes_prompt_hidden(draft_kind, draft_model)
+        )
+        self._prompt_hidden_chunks: List[Optional[List[mx.array]]] = []
 
         lengths = [len(ids) for ids in input_ids]
         max_length = max(lengths)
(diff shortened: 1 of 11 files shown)


avlp12:
Thank you for this — I hit the same defect from the other side while chasing prefill memory on a long-prompt DFlash setup, and your framing of it (the capture requested only on the final prefill call) matches what I found exactly. Two notes and two questions.

**1. Three targets were already opting in.** At `6be8e3da`, these return True for a DFlash-attached chunked prefill:

- `mlx_vlm/models/qwen3_5/language.py:1444-1445`
- `mlx_vlm/models/lfm2/language.py:410-411`
- `mlx_vlm/models/muse_glimmer/language.py:296-297`

Both chunk drivers then dropped the capture inside the loop: `mlx_vlm/generate/ar.py:494-503` builds `chunk_kwargs` from `kwargs` only, and `PromptProcessingBatch.prompt_step` at `mlx_vlm/generate/ar.py:1970-1977` never adds it. So on any prompt longer than one chunk those three hand their drafter a one-row prompt context, silently — no error, only an acceptance-rate signature. This PR fixes all three, and the same goes for `mlx_vlm/generate/ar.py:2101`, which passed `self._input_ids` as `prompt_tokens` after `prompt_step` had already consumed it. Would you consider naming these as live correctness paths in the description, rather than a missed optimisation?

**2. Trimming, as a possible follow-up.** `splice_prompt_hidden` keeps the whole prompt hidden resident for the request, as the description says. A sliding DFlash layer reads only the last `sliding_window - 1` rows (`mlx_vlm/speculative/drafters/qwen3_dflash/dflash.py:57-63`), and that block does two things — `x_ctx = x_ctx[:, skip:]` and `cache.offset += skip` — where the offset sets the absolute RoPE positions used at `dflash.py:82-84`. A future trim at the seam would need to hand the drafter the skipped row count as well; in exact arithmetic a uniform shift leaves the scores unchanged, but finite-precision RoPE can perturb drafter logits or draft tokens, so acceptance may move. Would you prefer to track trimming as a separate follow-up, keeping this PR at full width?

**3. A separate memory item.** Dropping `return_shared_kv` per chunk is a nice touch. There is a sibling case it does not reach: a target that opens a sequence-shaped rollback stash whenever `capture_layer_ids` is set. Nothing reads that stash off a prefill — every consumer takes it from a verify forward — so a one-line opt-out (`capture_gdn_states=False`, defaulting to True and honoured only by models that advertise support) removes it. On my fork, where this gated-delta target's DFlash2 path is not in upstream main, active memory on a 16,394-token prompt was 303.5 GB before versus 193.6 GB after a change set that included that prefill-leg opt-out and a separate FC pre-truncation change; I would treat that as a magnitude, not an isolated or reproducible benchmark. It is orthogonal to this PR and probably belongs in its own change.

I would be glad to open a small follow-up PR for either (2) or (3) — the trim plus its offset contract, and the prefill-leg opt-out are independent of each other. Thanks again for tracking this one down.

Blaizzy:
Closing it in favour of speculative redesign PR.

This is not the correct approach 

The drafters don't need all hidden states to perform.

Prefill can be done with suffix.

PR #2097, opened 2026-08-29
Add DeepSeek-V4 DSpark speculative drafter

### LINKED ISSUE #2095: split_mtp output for deepseek_v4 cannot be loaded back: tensor names don't match the deepseek_v4_mtp model
## Summary

`split_mtp` runs cleanly on a DeepSeek-V4 checkpoint and writes a ~10 GB drafter, but `load_drafter()` on that output rejects **every** extracted tensor. The splitter and the `deepseek_v4_mtp` model class disagree on parameter naming, so the round-trip is broken.

## Environment

- mlx-vlm 0.6.17, mlx 0.32.2, mlx-metal 0.32.2, Python 3.13
- macOS 15 (Darwin 24.6.0), Apple M3 Ultra
- Model: `deepseek-ai/DeepSeek-V4-Flash-0731` (4705 embedded `mtp.*` tensors, `num_nextn_predict_layers: 1`, `dspark_block_size: 5`)

## Reproduction

```python
from mlx_vlm.split_mtp import split_mtp
split_mtp("/path/to/DeepSeek-V4-Flash-0731",
          "/path/to/DS-V4-MTP-drafter",
          model_type="deepseek_v4", block_size=5)
# completes without error, writes ~10 GB

from mlx_vlm.speculative import load_drafter
load_drafter("/path/to/DS-V4-MTP-drafter", kind="mtp")
```

Result:

```
WARNING:root:Checkpoint has no weights for module(s): enorm, hnorm, e_proj, h_proj, hc_head, norm. Disabling them.
ValueError: Received 3140 parameters not in model:
1.attn.attn_sink,
1.attn.kv_norm.weight,
1.attn.q_norm.weight,
1.attn.wkv.scales,
1.attn.wkv.weight,
1.attn.wo_a.scales,
1.attn.wo_a.weight,
1.attn.wo_b.scales,
1.attn.wo_b.weight,
...
```

## Analysis

The written config looks right — `model_type: deepseek_v4_mtp`, `block_size` present — so drafter-kind resolution is not the problem. The mismatch is purely in tensor keys:

- **`split_mtp` emits:** `1.attn.wkv.weight`, `1.attn.kv_norm.weight`, `1.attn.attn_sink`, …
- **the `deepseek_v4_mtp` model expects:** `enorm`, `hnorm`, `e_proj`, `h_proj`, `hc_head`, `norm`

Side note: the written config also copies `num_hidden_layers: 43` from the base model into a single-layer drafter. Setting it to `1` does not change the error, so it is not the cause — but it looks unintended.

## Expected

`load_drafter(split_mtp(src, out, model_type="deepseek_v4"))` round-trips, so DeepSeek-V4's native MTP head can be used for speculative decoding.

Lazarus-931:
Adds a native `deepseek_v4_dspark` speculative drafter for DeepSeek-V4-Flash-0731's DSpark head as a **DeepSeek-V4 (MLA / Hyper-Connection / MoE) backbone variant of the existing `dspark` drafter** — reusing `VanillaMarkov` and the `dflash` round loop + capture tap rather than adding a new kind or loop — which also fixes #2095 by giving `-0731` a drafter that matches its architecture instead of the base `deepseek_v4_mtp` head it silently fails to load into.

### DIFF
--- a/mlx_vlm/models/deepseek_v4/language.py
+++ b/mlx_vlm/models/deepseek_v4/language.py
@@ -1016,6 +1016,8 @@ class DeepseekV4Model(PipelineMixin, nn.Module):
         inputs_embeds: Optional[mx.array] = None,
         hidden_sink: Optional[list] = None,
         skip_final_norm: bool = False,
+        capture_layer_ids: Optional[List[int]] = None,
+        capture_sink: Optional[list] = None,
     ) -> mx.array:
         h = self.embed_tokens(inputs) if inputs_embeds is None else inputs_embeds
         h = mx.broadcast_to(
@@ -1044,8 +1046,14 @@ class DeepseekV4Model(PipelineMixin, nn.Module):
         if pipeline_rank < pipeline_size - 1:
             h = mx.distributed.recv_like(h, (pipeline_rank + 1))
 
-        for layer, layer_cache in zip(self.pipeline_layers, cache):
+        capture_set = set(capture_layer_ids) if capture_layer_ids else None
+        for local_idx, (layer, layer_cache) in enumerate(
+            zip(self.pipeline_layers, cache)
+        ):
             h = layer(h, mask, layer_cache, inputs)
+            if capture_set is not None and (self.start_idx + local_idx) in capture_set:
+                # DSpark taps the mean over the hyper-connection copies.
+                capture_sink.append(h.mean(axis=2))
 
         if pipeline_rank != 0:
             h = mx.distributed.send(h, (pipeline_rank - 1) % pipeline_size)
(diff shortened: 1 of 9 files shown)


Lazarus-931:
### Runs on M3 Ultra (512 GB)


### Baseline vs DSpark, 512 → 8k

| ctx (prompt tok) | baseline tok/s | DSpark tok/s | speedup |
|---|--:|--:|--:|
| 512 (516)   | 32.04 | 40.86 | **1.28x** |
| 1024 (1028) | 31.47 | 42.83 | **1.36x** |
| 2048 (2052) | 29.60 | 45.87 | **1.55x** |
| 4096 (4100) | 27.80 | 36.44 | **1.31x** |
| 8192 (8196) | 27.62 | 38.55 | **1.40x** |


was trying to improve the tok/s as context climbs, hit flop and mem bwd ceiling

lucasnewman (approved):
🚀

PR #2100, opened 2026-08-30
fix(video): resolve frame sampling in one place (#2050)

### LINKED ISSUE #2050: Qwen3-VL: fps/max_frames appear unconfigurable — always samples with hardcoded processor defaults (fps=2, max_frames=768) regardless of override attempt

Lazarus-931:
`nframes`, `min_frames` and `max_frames` never reached `load_video`, and the video processor's own values were never read. Sampling is now settled once — caller, then processor, then library default — and the keys are consumed so they never reach a processor that only sees frames already chosen. `--verbose` reports what was actually sampled.

Closes #2050.

### DIFF
--- a/mlx_vlm/generate/dispatch.py
+++ b/mlx_vlm/generate/dispatch.py
@@ -2,6 +2,7 @@ import argparse
 import codecs
 import json
 import logging
+import sys
 import time
 from collections.abc import Sequence
 from typing import Any, Dict, Generator, List, Optional, Tuple, Union
@@ -49,6 +50,40 @@ from .image import (
 from .video_generation import DEFAULT_VIDEO_STEPS, run_video_generation_cli
 
 logger = logging.getLogger("mlx_vlm.generate")
+
+
+def _video_sampling(args):
+    """The frame-sampling overrides the user actually set."""
+    from ..utils import VideoSampling
+
+    return VideoSampling(
+        fps=args.fps,
+        nframes=getattr(args, "video_num_frames", None),
+        min_frames=getattr(args, "video_min_frames", None),
+        max_frames=getattr(args, "video_max_frames", None),
+    )
+
+
+def _video_sampling_kwargs(args) -> dict:
+    """``_video_sampling`` flattened for generate()'s keyword interface."""
+    from dataclasses import asdict
+
+    return {k: v for k, v in asdict(_video_sampling(args)).items() if v is not None}
+
+
+def _enable_verbose_logging() -> None:
+    """Surface mlx_vlm's INFO records, such as how a video got sampled."""
+    package_logger = logging.getLogger("mlx_vlm")
+    if not any(
+        getattr(handler, "_mlx_vlm_verbose", False)
+        for handler in package_logger.handlers
+    ):
+        handler = logging.StreamHandler(sys.stdout)
+        handler.setFormatter(logging.Formatter("%(message)s"))
+        handler._mlx_vlm_verbose = True
+        package_logger.addHandler(handler)
+    package_logger.setLevel(logging.INFO)
+
 
 DEFAULT_MODEL_PATH = "mlx-community/nanoLLaVA-1.5-8bit"
 DEFAULT_IMAGE = None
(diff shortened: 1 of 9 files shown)


lucasnewman (approved):
LGTM

PR #2105, opened 2026-08-31
glm5_next: fuse the KDA decode chain into one Metal kernel (opt-in)

avlp12:
After the input projections, a GLM-5-Next linear-attention (KDA) decode step is a long tail of tiny kernels — causal conv1d window update, silu, two L2 norms, the safe forget gate, sigmoid beta, the gated delta-rule state update, gated RMSNorm — roughly **30 GPU dispatches per layer**. At 34 KDA layers per token the launch count, not the arithmetic, is the cost: on GLM-5.3-Flash it was 7.56 ms of a 15.75 ms KDA stack, itself 43% of a 36.8 ms decode step.

This folds the whole chain into **one `mx.fast.metal_kernel` launch per layer** (no MLX source changes). One threadgroup owns one head, so both cross-`head_dim` reductions stay in threadgroup memory while the `[head_dim, head_dim]` recurrent state streams through registers exactly once. State accumulation stays fp32.

**Numbers** (M3 Ultra 512GB, 320B-A18B GLM-5.3-Flash local 4-bit, p512, 64 tokens): post-projection chain **7.56 → 1.72 ms**; end-to-end decode **31.56 → 34.03 tok/s (+7.8%)**, reproduced on this branch at **32.07 → 34.61 tok/s**. A second opt-in (`MLX_VLM_GLM5_FUSED_KDA_QPROJ`) also folds `f_b_proj`/`g_b_proj` in, leaving the chain as one dispatch (+2.2% further, measured on my serving branch of the same kernel).

**It is exact, not approximate.** Outputs, conv window and the carried fp32 state are bit-identical to eager over 32 consecutive decode steps, for both fp32 and bf16 storage of `A_log`/`dt_bias`. The included test module (8 parametrised cases) asserts equality, not tolerance.

**Reviewer notes — three Metal precision traps this had to reproduce:**
1. `mx.sigmoid` on bf16 evaluates in **bf16** (native `bfloat` rounds every step), not fp32-then-round — ~15% of elements differ otherwise.
2. MLX's `Sigmoid` spells `metal::exp`, which is *precise* inside MLX's prebuilt library but the *fast* approximation inside anything JIT'd (`mx.compile`, custom kernels). `nn.silu`/`compute_g_safe` are compiled → fast; the beta and output-gate sigmoids are plain ops → precise. `mx.exp` spells `precise::exp` and is precise in both.
3. `(x*x).sum(-1)` rounds the square before the add; letting the compiler contract it into an fma changes every norm's last bit.

**Scope.** Both toggles default OFF. Decode-only: engages at `B=1, S=1` with no mask and an initialised cache; prefill, `S>1`, batched/masked decode and unexpected cache layouts fall back to eager. QPROJ's exactness depends on MLX dispatching affine `qmv_quad`, so it is gated to `head_dim ∈ {64,128}` at 8 bits and declines otherwise. A capture variant that emits `gdn_sink` for speculative rollback exists downstream and is deferred until #2074 lands.

### DIFF
--- a/mlx_vlm/models/glm5_next/fused_kda.py
+++ b/mlx_vlm/models/glm5_next/fused_kda.py
@@ -1,0 +1,753 @@
+"""Fused single-token decode step for the GLM-5-Next KDA (Kimi Delta Attention) core.
+
+At B=1, S=1 the post-projection half of ``Glm5NextLinearAttention.__call__`` is a
+long tail of tiny elementwise / small-reduction kernels: the causal conv1d window
+update, silu, two L2 norms, the forget-gate softplus-free "safe gate", the sigmoid
+beta, the gated delta-rule state update and finally the gated RMSNorm.  Each of
+those is a separate GPU dispatch, and at 34 KDA layers per decode step the launch
+overhead dominates the (tiny) arithmetic.
+
+This module folds *all* of them into ONE ``mx.fast.metal_kernel`` launch per layer.
+One threadgroup handles one head, so the two cross-``head_dim`` reductions that the
+chain needs (the L2 norms over the key axis and the RMSNorm over the value axis)
+both stay inside threadgroup memory.  The recurrent state ``[head_dim, head_dim]``
+lives in device memory and is streamed through registers exactly once.
+
+The arithmetic is a line-by-line transcription of the eager path, including where
+it rounds to the input dtype: ``mx.conv1d`` writes bfloat16, ``nn.silu`` rounds
+twice (sigmoid then product), ``gated_delta_kernel`` writes its output ``y`` in the
+input dtype before the gated RMSNorm reads it back as float32, and ``beta`` is a
+bfloat16 sigmoid.  State accumulation stays float32, matching the eager kernel.
+
+Not a drop-in for prefill: this is decode-only (B=1, S=1, no SSM mask).
+``Glm5NextLinearAttention`` falls back to the eager path whenever any of those
+preconditions does not hold.
+"""
+
+import logging
+from typing import Optional, Tuple
+
+import mlx.core as mx
+
+logger = logging.getLogger(__name__)
+
+# MLX's own elementwise ops, transcribed so the fused kernel rounds identically:
+#   mlx/backend/metal/kernels/unary_ops.h -> Sigmoid / Exp / Rsqrt
+_HEADER = """
+// MLX's Sigmoid (mlx/backend/metal/kernels/unary_ops.h), with two subtleties
+// that each cost exactness if ignored:
+//
+//  * Instantiate on the SAME type the eager op used.  Metal's native `bfloat`
+//    rounds after every arithmetic step, so evaluating in float32 and rounding
+//    once at the end disagrees with mx.sigmoid(bfloat16) on ~15% of elements.
+//
+//  * Pick the right exp.  MLX's Sigmoid is written with `metal::exp`, which
+//    resolves to the precise implementation inside MLX's *precompiled* Metal
+//    library but to the fast approximation inside a JIT'd kernel (mx.compile
+//    and mx.fast.metal_kernel share that JIT).  In the KDA chain both appear:
+//    `nn.silu` and `compute_g_safe` are mx.compile'd (-> _fast), while the beta
+//    sigmoid and the output-gate sigmoid are plain ops (-> _precise).  MLX's
+//    Exp op spells `metal::precise::exp` in source, so mx.exp is precise in
+//    both.  Mixing them up is a 1-ulp disagreement on ~0.1-50% of elements.
+template <typename U>
+inline U mlx_sigmoid_precise(U x) {
+  U e = static_cast<U>(metal::precise::exp(metal::abs(x)));
+  U y = static_cast<U>(1) / (static_cast<U>(1) + e);
+  return (x < 0) ? y : (static_cast<U>(1) - y);
+}
+
+template <typename U>
+inline U mlx_sigmoid_fast(U x) {
+  U e = static_cast<U>(metal::exp(metal::abs(x)));
+  U y = static_cast<U>(1) / (static_cast<U>(1) + e);
+  return (x < 0) ? y : (static_cast<U>(1) - y);
+}
+
+// `(x * x).sum(-1)` in MLX materialises the squares before reducing, so the
+// multiply rounds before the add.  Writing `acc += v * v` here would let the
+// compiler contract it into an fma and silently change the last bit of every
+// L2 / RMS norm, so contraction is disabled for this one helper.
+#pragma clang fp contract(off)
+inline float sq_acc(float acc, float v) {
+  return v * v + acc;
+}
+#pragma clang fp contract(on)
+
+// `a + dt_bias` must promote exactly the way MLX's binary op does, and the two
+// operands need not share a dtype: glm5_next's cast_predicate leaves A_log /
+// dt_bias in whatever the conversion produced (float32 or the model dtype).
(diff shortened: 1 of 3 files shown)


avlp12:
Thanks for the CI catch — that failure is real, and it's now fixed on the branch.

**Diagnosis.** `maxTotalThreadsPerThreadgroup` is a *per-pipeline* limit driven by register pressure, not a device constant. The runner's virtualized GPU admits the 1024-thread launch for the base kernel (those tests passed on the same runner) but caps the higher-pressure qproj pipeline at 640 — hence exactly the two qproj tests failing. Worth noting for anyone hitting this elsewhere: MLX surfaces the violation as a `ValueError` at **eval** time, not at the call, so a `try/except` around the launch site doesn't catch it.

**Fix.** Each variant is probed once per process with a throwaway launch, walking 1024 → 512 → 256 → 128 threads and keeping the first that runs. `TY` only controls how many rows each thread walks; every reduction keeps the same lanes and the same operand order, so a reduced launch is still bit-identical — verified explicitly at all four sizes for both variants, not just argued. On this runner qproj should now run at 512 threads instead of failing.

If no size works the variant is declined rather than forced: qproj falls back to the base kernel plus the two eager GEMVs, and the base kernel falls back to the fully eager chain. Both fallbacks were checked to reproduce eager output exactly. One INFO line names the reason.

The tests use the same probe helper, so on a device that genuinely can't run a variant they `pytest.skip("device threadgroup cap below the … kernel's requirement")` instead of erroring.

Simulated caps locally (patching the probe to reject over-limit launches):

| cap on qproj pipeline | base | qproj | output |
|---|---|---|---|
| none | 1024 | 1024 | bit-identical |
| 640 (this runner) | 1024 | 512 | bit-identical |
| 320 | 1024 | 256 | bit-identical |
| 100 | 1024 | declined → base | bit-identical |
| 100 (both) | declined → eager | declined | bit-identical |

Probe cost is 2 launches per process, not per layer. 8/8 tests pass locally on M3 Ultra. Amended into the single commit.

avlp12:
Correction on one number above. I claimed the `MLX_VLM_GLM5_FUSED_KDA_QPROJ` fold was worth a further ~+2.2% end-to-end. Re-running the live A/B in both orders, that does not hold: the difference sits inside a ±1.8% position-bias band, so the honest reading is no measurable end-to-end effect.

The fold itself is unchanged and still correct — it is bit-identical to `mx.quantized_matmul` (it transcribes the `qmv_quad` partition) and it does remove two dispatches per layer, so the toggle stays as-is. I have also since gated it to B≤2, because at wider batches the batched GEMM already amortises the weight read that the fold re-does per row.

Please read that line as "dispatch-count reduction, performance-neutral in my live A/B" rather than a throughput claim. The base `MLX_VLM_GLM5_FUSED_KDA` numbers in the PR body are unaffected.

Lazarus-931:
hi @avlp12 thanks for the contribution.

can you provide benchmark sweeps with increasing context size for glm5.3-flash?

avlp12:
Thanks — yes. I have scheduled an increasing-context sweep on this PR branch as-is, with a public GLM-5.3-Flash checkpoint named by HF id and revision. I plan to compare the fused and unfused paths at each context on natural-text prompts, check token identity at every measured context because the PR claims exactness, and report prefill and decode throughput, peak memory, and the mlx / mlx-vlm versions once the run completes (M3 Ultra 512 GB).

One expectation to state before the numbers, so it can be checked against them: the kernel only touches the KDA decode chain, whose cost is per-layer and independent of context length, while the DSA attention layers' cost grows with context — so I expect the absolute saving per token to stay roughly constant and the relative gain to shrink as context grows. If the sweep shows otherwise I'll say so.

The unsigned commit the bot flagged has been replaced by a signed commit with the identical tree (`93b66ffc`).

avlp12:
Context sweep, as requested.

**Setup.** This PR branch as-is (`93b66ffc`; the sweep ran on the pre-signing commit `d04abdc6`, identical tree `77bf3476`), plain `mlx_vlm.generate`, no other changes. Checkpoint `avlp12/GLM-5.3-Flash-Alis-MLX-4bit` @ `7ce8cd0b` (MLX 4-bit-class build of `zai-org/GLM-5.3-Flash`). M3 Ultra 512 GB, mlx 0.32.1, mlx-vlm 0.7.0rc0. Prompts are real source text (no repeated tokens), 256 generated tokens greedy, kernel off vs on via `MLX_VLM_GLM5_FUSED_KDA`, three paired runs per context in alternating order, a throwaway forward before every timed arm, peak memory reset before each timed run. Because the kernel is bit-exact, identical output can't show the flag took effect, so I also counted invocations of the fused kernel entry point per arm.

| prompt tokens | prefill off / on (tok/s) | decode off (tok/s) | decode on (tok/s) | on/off, median (worst pair) | peak GB | output | fused-kernel calls off / on |
|--:|--:|--:|--:|--:|--:|:--:|--:|
| 523 | 368.5 / 369.2 | 28.28 | 31.26 | 1.105 (1.101) | 186.9 | identical | 0 / 8,704 |
| 2,059 | 478.7 / 479.9 | 25.06 | 27.36 | 1.091 (1.089) | 190.1 | identical | 0 / 8,738 |
| 8,203 | 452.4 / 451.3 | 24.88 | 27.12 | 1.091 (1.090) | 191.9 | identical | 0 / 8,738 |
| 16,395 | 409.9 / 410.2 | 24.75 | 27.34 | 1.103 (1.100) | 195.0 | identical | 0 / 8,738 |
| 32,779 | 339.1 / 339.6 | 24.49 | 26.93 | 1.100 (1.096) | 200.9 | identical | 0 / 8,738 |

(The 65,536-token context was started but not completed: only 1 of 6 arms finished — 1 off, 0 on — so I'm not quoting a result for it.)

**Reading.**
- Median decode speed-ups of **+10.5%, +9.1%, +9.1%, +10.3% and +10.0%** at 523, 2,059, 8,203, 16,395 and 32,779 prompt tokens respectively — flat within run-to-run scatter. My stated expectation was that the relative gain would shrink with context; I did not observe that over these five completed contexts, so I'm recording that prediction as not confirmed rather than quietly dropping it.
- Prefill is unchanged on vs off at each of the five completed contexts (deltas between −0.25 and +0.24 percentage points, i.e. run-to-run noise), consistent with the PR's decode-only scope.
- Across the five completed contexts the output is token-identical on vs off (one hash across the six arms at each context), and the counter recorded 0 fused-kernel calls on every off arm, 8,704 on-calls at 523 prompt tokens and 8,738 at each longer context — so the ratios compare the fused and unfused paths, not two runs of the same one.
- Peak memory is unchanged; the kernel replaces a chain of small dispatches and adds no working set.

These absolute decode throughputs are lower than the PR description's because I measured the stock `generate` path with default Metal buffer settings on the PR branch alone, without serving-branch changes. If useful, I can complete the 65,536-token run.

Lazarus-931:
hi @avlp12, thanks for pr!

We'll park this for now until #2127 gets merged :)

feel free to rebase from there

avlp12:
Understood, thank you — I'll wait for #2127 to merge and rebase this branch onto it then.

After the rebase I'll keep the kernel opt-in, check output agreement on the rebased tree, and rerun the increasing-context sweep so the numbers can be checked against the new base.

Lazarus-931:
hi @avlp12, so looking further, i think we should close this since it would conflict with the kernel implemented in #2127, since there, it's refactored such glm and other models share kernel components. 

if you think this is something still worth perusing, please feel free to open issue with spec, i'd be happy to explore this further!


thanks for your contribution :)

PR #2106, opened 2026-08-31
Accelerate exact Qwen MXFP4 verification

YuhuaBillChen:
## Summary

- add exact short-block MXFP4 quantized matrix-vector (QMV) kernels for Qwen speculative verification
- reuse loaded inputs across verification rows and fuse related projections
- reduce the MXFP4 LM head directly to exact argmax IDs
- support packed token masks in the MXFP4 argmax kernel for constrained decoding
- preserve singleton `QuantizedLinear` results exactly for batch sizes 1 and 2

## Why

`Qwen3_5ExactSpeculativeVerifier` has optimized exact paths for affine quantized projections, but MXFP4 targets fall back to invoking each projection one verification token at a time. This can make MTP slower than plain AR even when draft acceptance is high.

The new kernels preserve the singleton MXFP4 reduction order for each output while sharing input loads across the short verification dimension. MXFP4 LM heads now use the same fused greedy and packed-mask interface as affine quantized heads. Non-MXFP4 projections are unchanged.

## Correctness

- exact equality against singleton MXFP4 calls for BF16 and FP16 inputs,  verification lengths 1 through 4, input widths 512 and 6144, and batch sizes  1 and 2
- exact equality for fused two- and three-projection paths
- exact equality for LM-head argmax reduction at batch sizes 1 and 2
- exact masked argmax equality across multiple packed-mask words, plus an end-to-end fused greedy routing test with a token-mask logits processor
- a real Qwen3.8-27B checkpoint comparison captured 736 arrays covering every layer output, final hidden state, logits, Gated DeltaNet rollback state, and cache state; all were byte-identical
- end-to-end draft rounds and acceptance were identical in every A/B pair

## Results

Apple Silicon, fresh server process per arm, Qwen3.8-27B MXFP4 target, Q8 vision tower, fixed 2048x1542 image, MTP block 3, temperature 0, forced equal output, and a 20.5 GiB external watchdog.

| Workload | Baseline | This PR | Change | Prefill |
|---|---:|---:|---:|---:|
| mlx-vlm `b0991483`, 7,196 prompt + 512 output | 12.687 tok/s | 21.927 tok/s | +72.8% | unchanged |
| TQ4 integration, 7,196 prompt + 512 output | 12.571 tok/s | 22.042 tok/s | +75.3% | 121.21 vs 121.22 tok/s |
| TQ4 integration, 35,868 prompt + 2,048 output | 11.924 tok/s | 18.072 tok/s | +51.6% | 108.54 vs 108.07 tok/s |

For the `b0991483` comparison, both arms used 218 speculative rounds, drafted 436 tokens, accepted 293, and reported the same 16.674 GB MLX peak.

At 35,868 + 2,048, plain autoregressive decoding was 12.030 tok/s. The baseline exact MXFP4 fallback therefore made MTP 0.9% slower than AR, while this PR made MTP 50.2% faster than AR. Both MTP arms used 736 rounds, drafted 1,471 tokens, accepted 1,312, and reported the same 16.645 GB MLX peak.

On the real 248,320-row Qwen3.8-27B MXFP4 LM head, a constrained one-token decode selected the same allowed token through the full-logits reference, the masked kernel, and the end-to-end fused greedy path. The masked kernel averaged 2.97 ms versus 3.20 ms for materializing and masking full logits. After adding the masked path, the original 7,196 + 512 MTP workload remained at 21.933 tok/s with a 16.721 GB MLX peak.

## Tests

- `pre-commit run --files mlx_vlm/models/qwen3_5/speculative_verifier.py mlx_vlm/tests/test_speculative.py`
- expanded focused MXFP4 verifier matrix: 97 passed
- full `mlx_vlm/tests/test_speculative.py`: 300 passed
- full suite excluding smoke tests on `ee8806e0`: 3,146 passed, 7 skipped, 80 subtests passed


YuhuaBillChen:
Hi @Lazarus-931, do you have a chance to have a look at this PR? Many thanks in advance. :D

Lazarus-931:
Hi @YuhuaBillChen thanks for the pr, so seems like running your setup against main, it's 0.78x and 08x slower at T=3 and T=4 respectively(T as in verify blocks), note this is on m5.

Wasn't able to confirm your gains as well, can you provide script or desc. on the environment u ran sweeps on?

Also mlx has mxfp4 matmul kernels, and it repeatedly shows accuracy is correct

YuhuaBillChen:
Hi @Lazarus-931 — thanks for testing this on M5; that is genuinely useful. I re-ran everything on my M4 Pro and the improvement holds there, but your result surfaced a qualification I should have stated up front.

**TL;DR:** the isolated kernel is not universally faster across all `T` values or Apple GPU generations. I measure a regression myself at `T=1` on M4 Pro (`0.897x`). The PR's benefit on my machine comes primarily from the fused verifier paths and direct LM-head argmax, which produce a `1.805x` end-to-end MTP decode speedup in a fresh-process ABBA test.

Fully public reproducer:

https://gist.github.com/YuhuaBillChen/f8168fb89df34b8b939b341d25497b5b

It contains two independent tests:

1. **`reproduce_pr2106.py`** benchmarks `T=1..4` separately for one Q projection, fused Q/K/V, fused gate/up, and the 248,320-row LM-head argmax.
2. **`reproduce_pr2106_e2e.py`** runs fresh server processes in baseline/patched/patched/baseline order with the public target and MTP checkpoints, then compares output hashes, speculative statistics, prefill throughput, and decode throughput.

One thing worth pinning down, since it may explain the gap: the baseline in both scripts is the exact fallback used by the PR's parent revision for MXFP4 — one singleton `QuantizedLinear` call per verification token, followed by a concatenation. It is not the native batched MXFP4 matmul. The existing exact fast path is affine-only: `_can_target_verify_quantized_head()` returns `False` when `linear.mode != "affine"`, so MXFP4 falls back to the per-token loop. If your `0.78x` / `0.80x` comparison used MLX's native batched MXFP4 matmul as the baseline, I would expect this PR to look slower in that isolated comparison.

On M4 Pro with macOS 26.6.2, Python 3.12.13, and MLX 0.32.2, I measured:

| Operation | T=1 | T=2 | T=3 | T=4 |
|---|---:|---:|---:|---:|
| Single Q projection | **0.897x** | 1.395x | 1.626x | 1.930x |
| Fused Q/K/V | — | 1.371x | 1.737x | 1.875x |
| Fused gate/up | — | 1.686x | 1.832x | 1.877x |
| LM-head argmax | 1.007x | 1.755x | 1.978x | 2.120x |

The end-to-end ABBA result was:

- baseline: `13.21 tok/s`
- PR: `23.85 tok/s`
- speedup: `1.805x`

All four E2E arms produced the same output hash and identical speculative statistics: 112 rounds, 224 drafted tokens, and 143 accepted tokens.

On MLX's native MXFP4 matmul: I agree that it is numerically accurate, and I am not disputing that. The constraint here is stricter: the exact verifier must preserve byte-identical behavior relative to the singleton path. In this public-checkpoint run, the native batched Q projection was byte-identical at `T=1` and `T=2`, but not at `T=3` or `T=4` (maximum absolute differences `0.00048828125` and `0.0001220703125`). Those differences are numerically small, but they can change an argmax or acceptance decision near a boundary; if that happens, speculative decisions and subsequent cache evolution can diverge from the AR reference.

The PR therefore preserves each output's singleton reduction order while sharing input loads across the short verification dimension, fusing related Q/K/V and gate/up projections, and reducing the LM head directly to exact argmax IDs.

Additionally, I also ran on Apple M3 MacBook Air, 16 GiB, on standalone mxfp4 layers with the same dimensions: 

| Operation | T=1 | T=2 | T=3 | T=4 |
|---|---:|---:|---:|---:|
| Q projection | 0.827x | 0.920x | 1.258x | 1.572x |
| K/V-sized projection | 0.929x | 1.005x | 1.067x | 1.109x |
| MLP-sized projection | 1.000x | 1.493x | 1.739x | 1.876x |
| Fused Q/K/V | — | 1.313x | 1.502x | 1.625x |
| Fused gate/up | — | 1.594x | 1.859x | 1.966x |
| LM-head argmax | 1.004x | 1.704x | 2.004x | 2.091x |

M3 reproduces the singleton/small-work regression: Q is slower at T=1 and T=2, and the K/V-sized projection is slower at T=1. It does not reproduce the reported M5 regression at T=3 and T=4. On M3, Q T=3/T=4, both fused projection paths, and the direct LM-head argmax all remain faster.

Could you run the two Gist scripts on the M5 and share `result.json` and `e2e_result.json`? That would tell us whether the M5 regression is confined to the isolated QMV path or also outweighs the fused and end-to-end gains. If it is the latter, gating the optimization on the hardware generations where it wins would be more appropriate than enabling it unconditionally.

YuhuaBillChen:
Hi @Lazarus-931 just wondering if you have a chance to check :D Thanks a ton!

Lazarus-931 (approved):

Lazarus-931:
> Hi @Lazarus-931 just wondering if you have a chance to check :D Thanks a ton!


lgtm, sry for wait :)

YuhuaBillChen:
No problem, Thank you so much! Anything needed from my side to merge?

Issue #755, opened 2026-02-21
how to reuse pixel_values?


mediawall:
There’s a situation where I use multiple different prompts for the same video/image and hope to reuse the video pixel values. 

Currently, when passing input_ids/pixel_values/mask in params kwargs to generate method, we still have to wait for the relatively slow prefill from 0 to 100 every time.
 
After consulting AI, it suggested there might be an issue with generate_step, fyi. @Blaizzy 

---

**Title:** Chunked prefill in `generate_step` has excessive synchronization overhead causing slow prefill even with cached `pixel_values`

**Description:**

The chunked prefill loop in `generate_step` forces synchronization on every iteration, causing significant slowdown for multimodal inputs:

```python
while inputs_embeds.shape[1] > 1:
    model.language_model(...)
    mx.eval([c.state for c in prompt_cache])  # ❌ sync every chunk
    mx.clear_cache()
```

**Problems:**
1. `mx.eval()` blocks GPU pipeline on every chunk
2. `mx.clear_cache()` causes memory reallocation overhead
3. For vision models with many image tokens (576+), this creates 10-100+ sync points

**Suggested fix:** Reduce sync frequency (e.g., every N chunks) or move sync outside the loop.

---

Blaizzy:
Also, what machine are you running on as well as version of MLX-VLM

Issue #754, opened 2026-02-21
gemma3n can not use prompt cache


Asdf0203:
minimal case
```
import mlx_vlm
from textwrap import dedent
from mlx_vlm.models import cache

model, processor = mlx_vlm.load("./models/gemma-3n-E4B-it-4bit")
prompt_cache = cache.make_prompt_cache(model)
prompt = dedent("""
    <bos>
    <start_of_turn>user
    hello
    <end_of_turn>
    <start_of_turn>model
""")
output = mlx_vlm.generate(model, processor, prompt, prefill_step_size=8, prompt_cache=prompt_cache)
print(output.text)
```
Traceback:
```
Traceback (most recent call last):
  File "/test2.py", line 13, in <module>
    output = mlx_vlm.generate(model, processor, prompt, prefill_step_size=8, prompt_cache=cache)
  File "/venv/lib/python3.14/site-packages/mlx_vlm/generate.py", line 598, in generate
    for response in stream_generate(model, processor, prompt, image, audio, **kwargs):
                    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/venv/lib/python3.14/site-packages/mlx_vlm/generate.py", line 488, in stream_generate
    for n, (token, logprobs) in enumerate(
                                ~~~~~~~~~^
        generate_step(input_ids, model, pixel_values, mask, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ):
    ^
  File "/venv/lib/python3.14/site-packages/mlx_vlm/generate.py", line 384, in generate_step
    mx.eval([c.state for c in prompt_cache])
             ^^^^^^^
  File "/venv/lib/python3.14/site-packages/mlx_lm/models/cache.py", line 345, in state
    if self.offset == self.keys.shape[2]:
                      ^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'shape'
```

PR #2109, opened 2026-08-31
feat(speculative): declare drafter batch lifecycle

Thump604:
## Summary

- declare whether each registered MTP drafter supports continuous batching
- distinguish stateful drafters that require verified-token reconciliation from stateless assistants
- bind the public capability contract for Qwen3.8/DeepSeek V4 MTP and Gemma 4 assistants

This lets serving integrations reject unknown batch behavior instead of assuming it is safe.

## Validation

- full mlx-vlm suite: `2956 passed, 6 skipped, 75 subtests passed`
- Qwen3.8 Flash-Next composed serving qualification: exact and ragged continuous batching at 64K
- 128K MTP replay: 129,998 prompt tokens; 55 attempted, 42 accepted, 0 errors
- full Runtime promotion and installed-source verification passed

### DIFF
--- a/mlx_vlm/speculative/drafters/deepseek_v4_mtp/deepseek_v4_mtp.py
+++ b/mlx_vlm/speculative/drafters/deepseek_v4_mtp/deepseek_v4_mtp.py
@@ -43,6 +43,7 @@ class DeepseekV4MTPDraftModel(nn.Module):
     supports_greedy_draft_argmax = True
     prefer_requested_block_size = True
     requires_uniform_batch_acceptance = True
+    requires_verified_token_reconciliation = True
 
     def __init__(self, config: DeepseekV4MTPConfig):
         super().__init__()
--- a/mlx_vlm/speculative/drafters/gemma4_assistant/gemma4_assistant.py
+++ b/mlx_vlm/speculative/drafters/gemma4_assistant/gemma4_assistant.py
@@ -30,6 +30,7 @@ class Gemma4AssistantDraftModel(nn.Module):
     supports_greedy_draft_argmax = True
     # The drafter owns no cache; batched target KV is normalized per call.
     supports_continuous_batching = True
+    requires_verified_token_reconciliation = False
 
     def __init__(self, config: Gemma4AssistantConfig):
         super().__init__()
--- a/mlx_vlm/speculative/drafters/gemma4_assistant/gemma4_assistant.py
+++ b/mlx_vlm/speculative/drafters/gemma4_assistant/gemma4_assistant.py
@@ -28,6 +28,8 @@ class _DraftInner(nn.Module):
 
 class Gemma4AssistantDraftModel(nn.Module):
     supports_greedy_draft_argmax = True
+    # The drafter owns no cache; batched target KV is normalized per call.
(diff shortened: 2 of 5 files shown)


lyonsno (commented):
Read this against current main. Three things, in order of how much they block:

**1. Conflicts with main.** The drafter hierarchy moved under this after it opened: `Qwen4ExpMTPDraftModel` and `DeepseekV4MTPDraftModel` now both inherit `AutoregressiveMTPDraftModel` from `speculative/drafters/mtp_base.py`, which already carries `filter_batch`, `accept_verified_tokens_batch`, and `requires_uniform_batch_acceptance = True` at the base. Needs a rebase, and the rebase changes where the flags belong (3).

**2. The flags have no reader.** The body says this lets serving integrations reject unknown batch behavior, but nothing consumes `supports_continuous_batching` or `requires_verified_token_reconciliation` — the tests assert the attributes exist, not that anything acts on them. Every existing capability flag has a reader: `supports_greedy_draft_argmax` (mtp.py:786, eagle3.py:93), `prefer_requested_block_size` (mtp.py:550, dflash.py:35), `requires_uniform_batch_acceptance` via `_requires_uniform_batch_acceptance(draft_model, target_model)` (mtp.py:445, :1025). The rejection this PR promises would be a sibling helper in `speculative/common.py` checked at the top of `_mtp_rounds_batch` (mtp.py:871): a drafter entering the batched loop without `supports_continuous_batching = True` raises naming the drafter class. Without that these are documentation, and `requires_verified_token_reconciliation` documents a contract the batch loop already enforces by calling `accept_verified_tokens_batch` and `filter_batch` unconditionally when present.

**3. Put the truth on the base.** Since `AutoregressiveMTPDraftModel` owns the batch lifecycle hooks, `supports_continuous_batching = True` and `requires_verified_token_reconciliation = True` belong there, not on Qwen4 alone. That also fixes the asymmetry in the current diff where DeepSeek gets reconciliation but not batching support — if DeepSeek genuinely isn't batch-safe, an explicit `supports_continuous_batching = False` override is a clearer statement than silence. `Gemma4AssistantDraftModel` stays the real stateless case, and the `callable(...)` assertions in `test_qwen4_mtp_declares_continuous_batching_support` now pass trivially by inheritance, so they read better as a base-class test.

The replay runs from the PR body would be a great addition to the test suite since you've already got them — a small ragged-batch version on `_tiny_text_config()` (two rows of different prompt lengths through `_mtp_rounds_batch`, drop one with `filter_batch`, check the survivor's positions and cache stay aligned) would carry the same claim in-repo. What do you think about pinning that?

lyonsno:
Correction on (2): the reader exists — it's on the vllm-mlx side. #755 reads `supports_continuous_batching` through `explicit_bool(getattr(drafter, ...))` and #754 selects reconciliation off `requires_verified_token_reconciliation` with the hook pair, so "serving integrations" in your summary meant exactly that and I was looking in the wrong repo. What's left of (2) is smaller: since mlx-vlm is now the declaring side of a cross-repo contract, the flags probably want a line in `drafters/README.md` so the base class carries them as public surface rather than as attributes a downstream happens to read. (1) and (3) unchanged.

PR #2110, opened 2026-08-31
fix: require compatible mlx-audio for VoiceChat

Thump604:
## Summary

- raise the `mlx-audio` floor to `0.4.8`, the first published release that contains `mlx_audio.codec.models.nemotron_voicechat`
- add a metadata-level regression that binds the dependency floor to the actual Nemotron VoiceChat import

The current `>=0.4.3` declaration permits clean installation followed by an import failure.

## Validation

- dependency-floor regression passes against the corrected declaration and fails against the prior floor
- published `mlx_audio-0.4.8-py3-none-any.whl` SHA-256: `e62859facd7ae8a93c13870921dfab03db483fe3c34b66384cb865b980f8ebe6`
- full mlx-vlm suite: `2954 passed, 6 skipped, 75 subtests passed`


mlx_vlm/tests/test_nemotron_voicechat_dependency_floor.py
```diff
@@ -0,0 +1,57 @@
+"""Keep the Nemotron VoiceChat audio dependency floor in sync with its import."""
```
Lazarus-931: I don't this this is needed, a regression for a specific model on specific mlx-audio version.

lucasnewman (approved):

Issue #2041, opened 2026-08-26
qwen4_exp: Qwen3.8-Flash-Next MLX 4-bit checkpoint loads cleanly on main but generates garbage


yobuce:
`Vontra/Qwen3.8-Flash-Next-MLX-4bit` loads cleanly on `main` — every weight maps, no warnings — but generation is deterministic garbage. The same checkpoint files produce coherent text on the superseded #2028 branch, so I don't think it's the checkpoint.

Raising it because the failure is silent: nothing in the load path complains, and `mlx_vlm/tests/test_qwen4_exp.py` passes 8/8.

## Repro

```sh
pip install "git+https://github.com/Blaizzy/mlx-vlm@main"   # c2a67b63
hf download Vontra/Qwen3.8-Flash-Next-MLX-4bit --local-dir Q38FN-4bit

python -m mlx_vlm.generate --model Q38FN-4bit \
  --prompt "Write a short paragraph explaining what Gated DeltaNet is and why it helps long-context inference." \
  --max-tokens 128 --temperature 0.0 --verbose
```

Output:

```
  
4<|im_start|>
###

Prompt: 31 tokens, 1.070 tokens-per-sec
Generation: 7 tokens, 33.636 tokens-per-sec
Peak memory: 111.835 GB
```

Seven tokens and it stops. Byte-identical across runs and across `505267ca` / `c2a67b63`, so it's deterministic rather than a sampling artifact.

For contrast, the same directory on `JJJYmmm:add_qwen4exp` (#2028) — plus two small compat patches, see below — gives:

> Gated DeltaNet is an advanced recurrent neural network architecture that integrates the delta rule (a fast, parallelizable update mechanism) with a gating mechanism to manage information flow within the hidden state. It helps long-context inference by enabling efficient, linear-time processing of sequences, avoiding the quadratic computational complexity and memory overhead associated with traditional Transformers' self-attention mechanisms. […]

at 33.9 tok/s decode, 658 tok/s prefill on a 2,585-token prompt.

## What I ruled out

- **Checkpoint integrity** — 22/22 shards, 3,671 tensors, on-disk bytes exactly equal `metadata.total_size` (111,578,339,151).
- **Mixed precision hiding in the checkpoint** — all 998 quantized modules have `weight_cols / scales_cols == 4.0`, i.e. uniformly 4-bit at group 32, matching `config.json`'s `{"group_size": 32, "bits": 4, "mode": "affine"}`. No per-module exceptions.
[...]

PR #2111, opened 2026-08-31
fix(qwen4_exp): use standard RMSNorm gamma convention

Lazarus-931:
Qwen4ExpRMSNorm scaled by (1 + gamma), but the MLX-converted checkpoints store standard gamma centered at 1.0, so every norm was double-shifted and generation was deterministic garbage (#2041). This makes it a plain rms_norm(x) * gamma and offsets raw-HF checkpoints at load time via should_offset_norm_weight, matching qwen3_5 and the rest of the family. The MTP drafter shares the module and its checkpoints follow the same convention, so its released-equations test is updated to match.

### DIFF
--- a/mlx_vlm/models/qwen4_exp/language.py
+++ b/mlx_vlm/models/qwen4_exp/language.py
@@ -574,7 +574,7 @@ class QSAQuantizedKVCache(QuantizedKVCache):
 
 
 class Qwen4ExpRMSNorm(nn.Module):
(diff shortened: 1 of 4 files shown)


Lazarus-931:
Closing in favor of #2114, which fixes the same issue (#2041) while keeping the upstream `1+w` RMSNorm convention.

Lazarus-931:
closing, duplicate of #2114

PR #2114, opened 2026-08-31
qwen4_exp: load checkpoints with pre-shifted RMSNorm gains

### LINKED ISSUE #2041: qwen4_exp: Qwen3.8-Flash-Next MLX 4-bit checkpoint loads cleanly on main but generates garbage

Lazarus-931:
Fixes #2041.

`Qwen4ExpRMSNorm` applies `1 + w` to gains the released checkpoint stores centered at zero, matching upstream `Qwen4ExpTextRMSNorm`, but MLX conversions published before #2032 folded that `+1` into their saved weights — so loading one applies the offset twice and inflates every gain (~2.6x on the hyper-connection norms), which is the noise reported in the issue. Because such a checkpoint is otherwise identical to a correct one in keys, shapes and dtypes, `sanitize()` now tells them apart by value — released gains average ~+0.22, pre-shifted ones ~+1.15 — and subtracts the offset back out with a warning pointing at re-conversion. The same correction is applied to the MTP draft head, which shares `Qwen4ExpRMSNorm` and had the same defect, and `text_config.preshifted_norm_weights` overrides the detection either way.

Verified end-to-end on `Vontra/Qwen3.8-Flash-Next-MLX-4bit` (gain center +1.22, 148 tensors corrected): `main` stops after 7 tokens of noise, matching the issue exactly, the fix generates coherent text to the full 128-token limit, and forcing the detection off reproduces the 7-token failure.

Surveying the Hub, 8 of 11 published MLX conversions carry the doubled offset, and there was no canonical conversion to point people at — so I made one:

link to correct model: https://huggingface.co/AlazarM/Qwen3.8-Flash-Next-4bit

Converted from `Qwen/Qwen3.8-Flash-Next` with this repo's own converter on `main`, with norm gains verified bit-identical to the source (148/148) and a byte-exact source integrity check.

### DIFF
--- a/mlx_vlm/models/qwen4_exp/README.md
+++ b/mlx_vlm/models/qwen4_exp/README.md
@@ -137,3 +137,10 @@ python -m mlx_vlm.split_mtp \
   requesting both raises an explicit error to preserve the QSA indexer state.
 - Long image or video prompts may benefit from a smaller
   `--prefill-step-size` to reduce peak memory.
+- MLX conversions published before #2032 folded `Qwen4ExpRMSNorm`'s `1 + w`
+  offset into their saved norm gains, so loading one would apply the offset
+  twice and generate noise. Such a checkpoint is now detected and the offset
+  subtracted back out, with a warning. Re-converting from
+  `Qwen/Qwen3.8-Flash-Next` avoids relying on that fixup, which cannot restore
+  gains the conversion rounded away. Set `text_config.preshifted_norm_weights`
+  to `true` or `false` to override the detection.
--- a/mlx_vlm/models/qwen4_exp/config.py
+++ b/mlx_vlm/models/qwen4_exp/config.py
@@ -54,6 +54,7 @@ class TextConfig(BaseModelConfig):
     split_ngram_parts: int = 128
     seed: int = 1234
     ple_storage: Optional[Dict] = None
+    preshifted_norm_weights: Optional[bool] = None
     indexer_n_heads: int = 4
     indexer_kv_heads: int = 1
     indexer_head_dim: int = 128
--- a/mlx_vlm/models/qwen4_exp/qwen4_exp.py
+++ b/mlx_vlm/models/qwen4_exp/qwen4_exp.py
@@ -1,5 +1,7 @@
+import logging
 import re
 
+import mlx.core as mx
 import mlx.nn as nn
 
 from ..qwen3_5 import Model as Qwen3_5Model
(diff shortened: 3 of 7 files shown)


lucasnewman:
Seems like we should just upload correct conversions to `mlx-community`? Supporting every busted conversion with model-specific code seems like a mistake.

Lazarus-931:
> Seems like we should just upload correct conversions to `mlx-community`? Supporting every busted conversion with model-specific code seems like a mistake.

yeah I think I'll do a proper conversion, mlx-community doesn't seem to have qwen3.8-next

PR #2116, opened 2026-08-31
Add compact proposal-only heads for Qwen MTP

tjansn:
## Summary

Add an opt-in compact, quantized proposal-only LM head for Qwen 3.5 MTP drafters.

- loads a local `config.json` + `compact_head.safetensors` sidecar
- maps compact-row argmax results back to real vocabulary IDs
- keeps the full target model/head as the sole commit authority
- uses the compact head only for greedy drafting
- falls back to the original full proposal head and sampler for stochastic drafting, preserving the independent drafter RNG stream
- exposes the feature through `--draft-compact-head`

No model tensors, vocabulary mapping, or checkpoint-specific artifact is included in this PR.

## Why

Large-vocabulary proposal projection can be a material fraction of MTP draft cost. A proposal-only row subset can reduce that cost without changing target verification. A missing compact row can reduce acceptance, but cannot change committed output under exact verification.

The compact head is deliberately opaque runtime state rather than an `nn.Module`, so loading MTP checkpoint weights cannot traverse or mutate the separate sidecar.

## Correctness

Tests cover:

- sidecar shape and vocabulary-map validation
- exact affine quantized projection/argmax remapping
- greedy compact proposal dispatch
- stochastic fallback to the full head/sampler without invoking the compact head
- server loading and binding
- CLI environment propagation

Local real-checkpoint screens on current `main` (`09a987a`) used an affine-8/group-64 Qwen 3.8 27B target and Qwen MTP drafter:

- greedy compact MTP: 128/128 emitted IDs matched serial
- seeded `temperature=1`, `top_k=40` compact-configured MTP: 128/128 emitted IDs matched serial through full-head stochastic fallback

Both arm orders were exercised across the two cases. These are correctness screens, not benchmark claims.

## Operational notes

The feature is off by default. Existing drafters and generation paths are unchanged unless a compact sidecar is explicitly configured.

### DIFF
--- a/mlx_vlm/server/app.py
+++ b/mlx_vlm/server/app.py
@@ -893,6 +893,7 @@ def get_cached_model(
         apc_manager=runtime.apc_manager,
         draft_model_path=cfg.spec_draft_model,
         draft_kind=cfg.spec_draft_kind,
+        draft_compact_head_path=cfg.spec_draft_compact_head,
     )
     try:
         model, processor, config = response_generator.wait_until_ready()
--- a/mlx_vlm/server/cli.py
+++ b/mlx_vlm/server/cli.py
@@ -18,7 +18,8 @@ from .generation import (
     get_server_thinking_end_token,
     get_server_thinking_start_token,
 )
-from .runtime import MODEL_DISCOVERY_ENV, MODEL_DISCOVERY_MODES
+from .runtime import MODEL_DISCOVERY_ENV, MODEL_DISCOVERY_MODES, runtime
+from .runtime_config import RuntimeConfig
 
 DEFAULT_SERVER_HOST = "0.0.0.0"
 DEFAULT_SERVER_PORT = 8080
@@ -237,6 +238,12 @@ def main():
         ),
     )
     parser.add_argument(
+        "--draft-compact-head",
+        type=str,
+        default=None,
+        help="Optional local compact proposal-only head for an MTP drafter.",
+    )
+    parser.add_argument(
         "--draft-kind",
         type=str,
         default=None,
@@ -314,6 +321,8 @@ def main():
     os.environ["MLX_VLM_VISION_CACHE_SIZE"] = str(args.vision_cache_size)
     if args.draft_model:
         os.environ["MLX_VLM_DRAFT_MODEL"] = args.draft_model
+    if args.draft_compact_head:
+        os.environ["MLX_VLM_DRAFT_COMPACT_HEAD"] = args.draft_compact_head
     if args.draft_kind is not None:
         os.environ["MLX_VLM_DRAFT_KIND"] = args.draft_kind
     if args.draft_block_size is not None:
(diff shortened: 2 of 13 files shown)


lucasnewman:
@tjansn Where does the sidecar some from? This also PR also needs performance comparisons to be considered.

lucasnewman (changes requested):
See comments.

tjansn:
Thanks, you’re right. I left out where the sidecar comes from and did not include a clean performance comparison.

The sidecar is a smaller copy of the model’s output layer. It is not trained separately and does not come from the MTP checkpoint. I generated it locally from the 4-bit Orca model ( https://huggingface.co/orcarouter/Qwen3.8-27B-Uncensored-MLX/tree/main ) by copying 98,336 selected vocabulary rows.

 It is only used to suggest draft tokens during greedy generation. The full model still checks every suggestion and decides which token is emitted. For sampled generation, the code uses the original full output layer.

The idea was influenced by https://github.com/Layr-Labs/qwen-3.8-mtp-challenge/pull/59 

I did not use its weights or exact vocabulary mapping.

On an M5 Max, using the same model, prompts, settings, and MTP block size:
 
| Proposal head      | Throughput  | Difference |
|--------------------|-------------|------------|
| Full 8-bit head    | 27.14 tok/s | baseline   |
| Smaller 8-bit head | 28.12 tok/s | +3.58%     |
| Smaller 4-bit head | 29.27 tok/s | +7.82%     |


The 4-bit version was faster on all four prompts. All generated token IDs matched the full-model reference. It added about 0.26 GiB of memory because the original output layer must remain loaded.

These results are encouraging, but the test used an earlier version of the code and was not a fully controlled benchmark. Before requesting another review, I will:

- add a tool and instructions for creating the sidecar;
- document its source and file format;
- add checks that it matches the selected model;
- repeat the comparison on the current PR branch with everything else held constant.

No model weights or checkpoint-specific mapping will be added to the PR.

Blaizzy:
Hi @tjansn, appreciate the effort! 
We are going to decline this PR, while the performance gains are interesting, the non official Qwen/HF transformers changes and extra 1K lines also seems like a significant trade-off (~7%). I'd be happy to discuss alternative approaches that stay within the official model ecosystem if you're interested.

tjansn:
Thanks, understood. I agree that roughly 1.4K lines is too much maintenance surface for this gain when no official Qwen checkpoint ships the additional proposal head. I will leave the PR closed.

For completeness, the controlled run finished before I saw the closure:

- full 8-bit proposal head: 32.87 tok/s pooled
- compact 4-bit proposal head: 34.98 tok/s pooled
- pooled gain: 6.42%; median per-case gain: 5.82%
- faster in all 8 cases; worst case: +4.94%
- 4 fixed prompts, 2 repeats, opposite arm orders, 512 emitted tokens per case
- all 4,096 emitted IDs per arm and finish semantics matched the serial reference
- draft acceptance: 52.54% full vs 51.62% compact
- peak MLX memory unchanged at 30.17 GiB
- all process and thermal gates passed; swap stayed flat

The measured runtime source was commit `1eee9bad58a7e18a66ff1b44d7a97bccd370fac4`. The follow-up branch commit only fixes the four failed test fixtures and records these results.

This confirms the mechanism, but it does not change the upstream maintenance trade-off. I pushed the fixes to my fork for a clean record and will keep this downstream. If a smaller version tied only to an official model-owned proposal head becomes useful later, I would be happy to narrow it to that surface without the builder and custom sidecar workflow.

PR #2117, opened 2026-08-31
Add opt-in exact 8-bit Qwen verifier fusion

tjansn:
## Summary

Allow Qwen exact speculative verification to opt into the existing adjacent-linear fused QMV kernel for affine 8-bit linears.

- adds `set_target_verify_8bit_linear_fusion(...)`
- keeps the feature disabled by default
- requires every adjacent linear to use the same bit width/group size/mode
- limits the 8-bit path to verification lengths below 6; longer 8-bit blocks safely retain the existing path
- exposes the feature through `--target-verify-fuse-8bit-linears`
- resets the process-global selection explicitly during server model initialization

## Why

Qwen verification repeatedly applies adjacent QKV, linear-attention projection, and MLP gate/up linears to the same hidden states. The fused kernel already supports parameterized affine bit widths and preserves per-timestep accumulation order. Dispatch had previously been restricted to 4-bit linears.

## Correctness

Tests cover:

- exact fused-vs-timewise equality for 8-bit adjacent groups at verification lengths 3 and 4
- existing 4-bit fused coverage at lengths 3, 6, and 8
- server/CLI opt-in propagation

Additional local checks on Apple Silicon:

- synthetic affine-8/group-64 groups with 2, 3, and 4 adjacent linears matched bit-for-bit at lengths 3 and 4
- lengths 6 and 8 were confirmed to decline the 8-bit fused path and retain the safe fallback
- a current-`main` real-checkpoint screen on an affine-8/group-64 Qwen 3.8 27B target matched serial for 128/128 emitted IDs

Balanced local end-to-end screens on the same model improved pooled MTP throughput by 8.28% on four fixed prompts and 5.07% on eight held-out prompts, with identical emitted IDs, acceptance, and round counts. These figures are hardware/workload-specific and are not presented as a general benchmark.

## Operational notes

The feature is off by default. Existing 4-bit dispatch and all non-opted-in 8-bit paths remain unchanged.


Lazarus-931:
hi @tjansn, thanks for the pr!

would this work only for specific qwen quantizations?

if i may, can you please provide e2e sweeps for several context sizes to see perf as well as a gist to replicate?

tjansn:
Thanks @Lazarus-931!

The added path is specifically for [eligible affine 8-bit linears](https://github.com/Blaizzy/mlx-vlm/blob/0245fce9a3294ab42689078d69add89709b244e2/mlx_vlm/models/qwen3_5/speculative_verifier.py#L1120-L1140) in the Qwen exact verifier. It is not tied to a checkpoint name or fixed group size: adjacent linears must share bits/group size/mode and pass the existing shape/dtype checks. Existing 4-bit dispatch is unchanged, and unsupported groups retain the existing path. The `<6` limit refers to verification-block length, not accumulated context length.

I have now run the context sweep on this PR's exact commit, `0245fce9a3294ab42689078d69add89709b244e2`:

- M5 Max, 40-core GPU, 128 GiB; MLX 0.32.1
- `orcarouter/Qwen3.8-27B-Uncensored-MLX`, affine-8/group-64 target, original BF16 MTP drafter
- two public synthetic length-scaling tasks, two separate sessions with opposite arm orders
- 256 emitted tokens per case, greedy decoding, batch size 1, MTP block total 4
- same loaded target/drafter across serial/off/on; fresh KV state per request; no compact proposal head or downstream geometry changes

Pooled results across all four cases per context:

| Input tokens | MTP off E2E tok/s | MTP on E2E tok/s | E2E gain | Decode gain |
|---:|---:|---:|---:|---:|
| 512 | 34.97 | 36.68 | +4.87% | +5.34% |
| 2,048 | 28.23 | 29.25 | +3.62% | +4.88% |
| 8,192 | 15.19 | 15.57 | +2.47% | +5.44% |
| 32,768 | 4.57 | 4.60 | +0.60% | +3.51% |

E2E here is the complete in-process generation call, including prefill and synchronized drain, but excluding model loading, tokenization, detokenization and HTTP. At 32k, about 47 of 56 seconds are TTFT/prefill work, so the total-call gain is small with this 256-token output budget.

All 4,096 emitted IDs per arm and finish semantics matched serial and repeated across sessions. Fusion off/on had identical acceptance histories and round counts. All pre-arm CPU-quiet and ≤44°C gates passed; swap did not grow. Separate synthetic checks passed bit-for-bit for group sizes 32/64/128, FP16/BF16 and verification lengths 2–5. Real-checkpoint performance coverage remains limited to affine-8/group-64 on this machine.

[Reproduction gist with scripts, pinned checkpoint/source/package versions, raw emitted IDs, serial references, TTFT/total times, per-task results, acceptance counts and repeat variability](https://gist.github.com/tjansn/fafe5830e661d89333df840cd0ee0c88).

Lazarus-931 (approved):

PR #2118, opened 2026-08-31
muse_glimmer: load flattened OptiQ checkpoints (#2108)

### LINKED ISSUE #2108: muse_glimmer: sanitize() strips `model.` prefix instead of remapping to `language_model.model.`
# muse_glimmer: sanitize() strips `model.` prefix instead of remapping to `language_model.model.`, so the mlx-community checkpoint fails to load

## Summary

`mlx-community/Muse-Glimmer-30B-OptiQ-4bit` (the only mlx conversion of this
model, revision `b4a74fa6001f1eca3b23eeeb702ffad2773a218f`) fails at load:

```text
ValueError: Received 1460 parameters not in model:
embed_tokens.biases,
embed_tokens.scales,
embed_tokens.weight,
layers.0.input_layernorm.weight,
...
norm.weight
```

Every one of the 1460 rejected parameters is a text-model weight; the vision
tower, adapter, and projection load fine.

## Root cause

The checkpoint stores its text weights under plain HF-style keys -- the
top-level prefixes in `model.safetensors.index.json` are:

```text
lm_head, model, vision_adapter, vision_projection, vision_tower
```

with the text stack at `model.embed_tokens.*`, `model.layers.*`,
`model.norm.*`.

`MuseGlimmerModel.sanitize()` (`mlx_vlm/models/muse_glimmer/muse_glimmer.py`)
handles `model.language_model.*` and `lm_head.*`, but its fallback branch
just strips the `model.` prefix:

```python
if key.startswith("model.language_model."):
    key = key.replace("model.language_model.", "language_model.model.", 1)
elif key.startswith("model."):
    key = key[len("model.") :]          # -> embed_tokens.*, layers.*, norm.*
elif key.startswith("lm_head."):
    key = "language_model." + key
```

That produces top-level `embed_tokens.*` / `layers.*` / `norm.*` keys, but
the module houses the text model at `self.language_model` (a `LanguageModel`
whose inner module is `.model`), so the expected parameter paths are
`language_model.model.embed_tokens.*` etc. `load_weights(strict=True)` then
rejects exactly those keys ‚Äî the `embed_tokens, layers, norm` families in
the error are the strip's output verbatim.

The `elif key.startswith("model."):` branch appears to have been written for
a conversion layout that nested text weights as `model.language_model.*`;
the published mlx-community checkpoint does not.

## Suggested fix

Map the bare `model.` prefix to the language model instead of stripping it:

```python
elif key.startswith("model."):
    key = "language_model." + key       # model.embed_tokens -> language_model.model.embed_tokens
```

(i.e. `model.X` ‚Üí `language_model.model.X`, keeping the existing
`model.language_model.*` branch first for layouts that already nest.)

With that mapping the 1460 keys align with the module tree; `lm_head.*` and
the three vision prefixes already resolve.

## Reproduction

```bash
python -m mlx_vlm.generate --model mlx-community/Muse-Glimmer-30B-OptiQ-4bit \
  --image any-local-image.jpg --prompt x --max-tokens 8 --temperature 0.0 \
  --revision b4a74fa6001f1eca3b23eeeb702ffad2773a218f --trust-remote-code
```

Observed on current git main (editable install reporting 0.7.0rc0); the
sanitize shown above is the code at main today.

## Environment

| Component       | Value                        |
|-----------------|------------------------------|
| mlx-vlm         | 0.7.0rc0 (also git main)     |
| mlx             | 0.32.3.dev20260830+37c26e575 |
| transformers    | 5.16.1                       |
| Python          | 3.14.7                       |
| macOS           | 26.6.2 (Apple M5 Max)        |

Full harness evidence (complete traceback, all 1460 rejected keys, memory
and provenance facts):
<https://github.com/jrp2014/check_models/blob/main/src/output/issues/issue_mlx-community_Muse-Glimmer-30B-OptiQ-4bit.md>

Lazarus-931:
Fixes #2108: `muse_glimmer` sanitize maps flattened OptiQ checkpoints' bare `model.` keys to `language_model.model.`. The strip is kept for the nested vision prefixes, which is what raw HF checkpoints use (`model.vision_tower.*` etc.).

### DIFF
--- a/mlx_vlm/models/muse_glimmer/muse_glimmer.py
+++ b/mlx_vlm/models/muse_glimmer/muse_glimmer.py
@@ -8,6 +8,13 @@ from .config import ModelConfig
 from .language import LanguageModel, RMSNormNoScale
 from .vision import VisionModel, gelu
 
+_NESTED_VISION_PREFIXES = (
+    "model.vision_tower.",
+    "model.vision_adapter.",
+    "model.vision_projection.",
+    "model.perception_emb_norm.",
+)
+
 
 def masked_scatter(input_tensor: mx.array, mask: mx.array, source: mx.array):
     shape = input_tensor.shape
@@ -126,9 +133,9 @@ class Model(nn.Module):
                 continue
             if key.startswith("model.language_model."):
                 key = key.replace("model.language_model.", "language_model.model.", 1)
-            elif key.startswith("model."):
+            elif key.startswith(_NESTED_VISION_PREFIXES):
                 key = key[len("model.") :]
-            elif key.startswith("lm_head."):
+            elif key.startswith(("model.", "lm_head.")):
                 key = "language_model." + key
             sanitized[key] = value
         return sanitized
--- a/mlx_vlm/tests/test_models.py
+++ b/mlx_vlm/tests/test_models.py
@@ -9911,6 +9911,66 @@ class TestModels(unittest.TestCase):
         self.assertIsInstance(cache[0], RotatingKVCache)
         self.assertIsInstance(cache[1], KVCache)
 
+    def test_muse_glimmer_sanitize_weight_layouts(self):
+        from mlx_vlm.models import muse_glimmer
+
+        model = muse_glimmer.Model(self._muse_glimmer_config())
+
+        # Raw HF nests both stacks under ``model.``; MLX conversions ship the
+        # already-sanitized paths; OptiQ/flattened quants store LM tensors under
+        # a bare ``model.`` with the vision stack and lm_head already at top
+        # level. All three must land on the same target module paths.
+        raw_hf = {
+            "model.language_model.embed_tokens.weight": mx.zeros((64, 32)),
+            "model.vision_tower.ln_pre.weight": mx.zeros((8,)),
+            "model.vision_adapter.fc1.weight": mx.zeros((16, 32)),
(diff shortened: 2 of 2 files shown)


mlx_vlm/models/muse_glimmer/muse_glimmer.py
```diff
@@ -126,9 +133,9 @@ def sanitize(self, weights):
                 continue
             if key.startswith("model.language_model."):
                 key = key.replace("model.language_model.", "language_model.model.", 1)
-            elif key.startswith("model."):
+            elif key.startswith(_NESTED_VISION_PREFIXES):
                 key = key[len("model.") :]
-            elif key.startswith("lm_head."):
+            elif key.startswith(("model.", "lm_head.")):
                 key = "language_model." + key
```
lucasnewman, line 139: This clause doesn't need to trim the "model." prefix? It looks like it would end up with `language_model.model.<key>`, is that correct?

lucasnewman (approved):

mlx_vlm/models/muse_glimmer/muse_glimmer.py
Lazarus-931: yep! `LanguageModel` holds the decoder at `self.model `and the head at `self.lm_head`

PR #2119, opened 2026-08-31
fix(generate): chunk prefill for prompts shorter than prefill_step_size

Lazarus-931:
Prefill only chunked when the prompt was **longer** than `prefill_step_size`. The chunk loop discards its return value, so MLX never evaluates the `[B, N, vocab]` logits. The unchunked path hands the whole prompt to `_step`, which reads `logits[:, -1, :]` — materializing every row to keep one.

So a shorter prompt costs more than a longer one:

| prompt | chunked | peak (Qwen3.5-0.8B) |
|---|---|---|
| 2047 | no | **2.896 GB** |
| 2049 | yes | **1.965 GB** |

Ordinary chat turns land on the expensive side. This chunks whenever more than one prompt token remains. The same gate existed in `PromptProcessingBatch.needs_processing`, the batch path the server uses.

Greedy decode, output unchanged. Savings scale with `prompt_tokens x vocab_size`:

| model | prompt | before | after |
|---|---|---|---|
| Qwen3.5-0.8B-4bit | 2047 | 2.898 GB | 1.973 GB |
| Qwen3.8-27B-4bit | 1229 | 17.505 GB | 17.013 GB |

Swept 56 local checkpoints across 24 families: 54 byte-identical. `diffusion_gemma` is nondeterministic on unpatched `main` and already opts out of chunking.

**PaliGemma** builds a prefix-LM mask over the whole prompt, so causal chunking changes its output. It gains a `chunked_prefill_policy` opting out while bidirectional attention is on, using the hook `gemma4` and `lfm2` already use. Already latent: its prompts over `prefill_step_size` chunk on `main` today. Other whole-prompt-mask models already opt out or are unaffected.

Two regression tests, both failing on `main`. Suite green: 533 passed.

Downstream: Blaizzy/nativ#427.

### DIFF
--- a/mlx_vlm/generate/ar.py
+++ b/mlx_vlm/generate/ar.py
@@ -465,8 +465,13 @@ def generate_step(
             else None
         )
         checkpoint_done = False
+        # Chunk whenever there is more than one prompt token left to process.
+        # The chunk loop discards its output, so the [B, N, vocab] logits are
+        # never evaluated; the unchunked path feeds the whole prompt to _step,
+        # which reads logits[:, -1, :] and so materializes every row. Gating on
+        # prefill_step_size made short prompts peak higher than long ones.
         should_chunk = (
-            prefill_step_size is not None and inputs_embeds.shape[1] > prefill_step_size
+            prefill_step_size is not None and inputs_embeds.shape[1] > 1
         ) or (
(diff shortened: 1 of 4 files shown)


lucasnewman (approved):
👍

lucasnewman:
@Lazarus-931 Failing tests.

lucasnewman (approved):

PR #2124, opened 2026-09-01
Accelerate exact Qwen affine Q3 verification

YuhuaBillChen:
## Summary

- extend Qwen's exact affine quantized matrix-vector kernels to 3-bit weights
- preserve MLX's packed 3-bit reduction order for short speculative blocks
- fuse related Q/K/V, Gated DeltaNet and MLP projections
- reduce 3-bit LM heads directly to exact argmax IDs, including packed token masks

The ordinary autoregressive greedy path remains unchanged; the new fast path is limited to exact speculative verification.

## Why

`Qwen3_5ExactSpeculativeVerifier` has optimized exact paths for affine 4-, 5- and 8-bit projections. Affine 3-bit targets instead fall back to invoking each projection once per verification token, leaving Q3 deployments unable to use the same short-block acceleration.

MLX packs eight 3-bit values across three bytes. The new codec branch decodes that layout directly while preserving the singleton `QuantizedLinear` accumulation and rounding order.

For Q3 argmax kernels, normalized packed-field extraction avoids input-side rescaling and is faster, while projection kernels retain the unshifted formulation to preserve byte-identical singleton outputs.

## Correctness

- exact equality against singleton affine 3-bit calls for BF16 and FP16 inputs,  group sizes 32 and 64, verification lengths 2 through 4, input widths 512 and 6656, and batch sizes 1 and 2
- exact equality for fused two-, three- and four-projection paths
- exact LM-head argmax equality, including packed token masks spanning multiple  mask words
- the Q3 LM-head argmax path uses equivalent shifted 3-bit fields to reduce exact argmax time from 5.65 ms to 4.46 ms on the real checkpoint
- a real Qwen3.8-27B Q3 checkpoint comparison covered the LM head, attention,  Gated DeltaNet and MLP projections with zero numerical difference
- a full target-verification A/B captured 160 prefix-cache arrays and 706 verification outputs/cache states; every array was byte-identical

## Results

Apple Silicon, fresh server process per arm, Qwen3.8-27B affine Q3 target with its Q8 vision tower loaded, MTP block 3, temperature 0, forced equal output, APC disabled and a 20.5 GiB external watchdog. The multimodal arm used a fixed 2048x1542 image; the AIME arm was text-only.

Two workloads show why draft acceptance matters. The first deliberately forces an image-description trajectory whose acceptance is only about 58%; the second is a reasoning workload with 77.4% acceptance.

| Workload | Current main MTP | This PR MTP | This PR AR | PR vs main | MTP vs AR |
| 7,196 prompt + 512 image-description output | 13.09 tok/s | 15.57 tok/s | 16.06 tok/s | +18.9% | -3.1% |
| AIME reasoning + 1,024 output | 15.31 tok/s | 19.43 tok/s | 16.63 tok/s | +26.9% | +16.8% |

For the AIME workload, this PR and plain AR produced the same output SHA-256. Current `main` produced a different trajectory. The patched MTP arm accepted 622 of 804 draft tokens over 402 rounds. All arms completed under the watchdog, and the patched MTP arm reported a 15.173 GB MLX peak.

## Tests

- `pre-commit run --files mlx_vlm/models/qwen3_5/speculative_verifier.py mlx_vlm/tests/test_speculative.py`
- focused affine 3-bit verifier tests: 149 passed
- full suite excluding smoke tests: 3,198 passed, 7 skipped, 80 subtests passed


mlx_vlm/models/qwen3_5/speculative_verifier.py
```diff
@@ -28,17 +28,28 @@ def _use_target_verify_dense(linear, x: mx.array) -> bool:
 
 
 def _target_verify_qlinear_header(
-    bits: int, group_size: int, results_per_simdgroup: int = 4
+    bits: int,
+    group_size: int,
+    results_per_simdgroup: int = 4,
+    *,
+    q3_shifted_fields: bool = False,
```
Blaizzy, line 35: Do we need this argument?

Blaizzy (commented):
Mostly LGTM, just have one question!

mlx_vlm/models/qwen3_5/speculative_verifier.py
YuhuaBillChen: Great question @Blaizzy , thanks for pointing out. I didn't explain well in the code. TL;DR: Yes, unless we split the shared Metal header into separate projection and argmax variants. This is a compile-time code-generation switch, not a runtime option or branch.

Q3 needs two algebraically equivalent extraction forms because they have different correctness/performance properties:

 - Projection kernels keep q3_shifted_fields=False. They operate on the packed bit fields directly and compensate by scaling the corresponding input values. This preserves the same floating-point reduction order as the singleton QuantizedLinear path, which is required for byte-identical projection outputs and cache states.
 - The unmasked and masked argmax kernels set it to True. They shift each packed field into its normalized 3-bit value and leave the inputs mostly unscaled. This avoids repeated input-side rescaling and reduced real-checkpoint Q3 LM-head argmax latency from 5.65 ms to 4.46 ms, while preserving the exact argmax token IDs in our parity tests.

 Making either form unconditional would lose one of those properties:

 - always shifted: faster argmax, but projection outputs are no longer byte-identical;
 - always unshifted: exact projections, but gives up the Q3 argmax speedup.

The flag defaults to False and is enabled only at the two Q3 argmax call sites, so Q3 projections and all Q4/Q5/Q8 paths remain unchanged. Since its value is substituted while generating the Metal source, the compiler sees a constant and there is no runtime dispatch cost.

The alternative would be to duplicate the shared header or add separate projection/argmax wrappers. I used the explicit keyword argument as the smallest way to specialize the shared Q3 decoder without duplicating the kernel code, but I’m happy to split the wrappers if you prefer that API shape.

YuhuaBillChen: @Blaizzy Friendly ping. Any updates on this PR? Thanks!

YuhuaBillChen:
Rebased onto main. #2106 touched the same file, so this had gone
conflicted after it landed; it's clean again and CI is green.

The only overlap was `can_quantized_head`. #2106 added the mxfp4 head
path, and this PR excludes 3-bit from the affine head — it widens
`_can_target_verify_quantized_head` to accept `bits == 3` for the
projection and argmax kernels, but the head itself must stay out.
Resolved as the union of both:

return _can_target_verify_mxfp4_head(linear) or (
        _can_target_verify_quantized_head(linear) and linear.bits != 3
    )
    
@Blaizzy Let me know if you need anything from me.

YuhuaBillChen:
Superseded by #2209, which ports this change to the generalized quantized verifier introduced in #2127.

@Blaizzy, #2209 also addresses your question above about

q3_shifted_fields`: it is still needed, and the new PR splits its use as you suggested. Projection kernels retain the unshifted arithmetic for byte-identical outputs, while only the masked and unmasked argmax kernels enable shifted-field extraction

PR #2128, opened 2026-09-01
sam3: fix pred_boxes inflation and offset (RoPE scale, RPB grid, ref_point_head)

### LINKED ISSUE #2103: SAM 3 / SAM 3.1: `pred_boxes` systematically inflated + offset, masks are pixel-accurate
Hello, I just discovered a bug in SAM 3/3.1 while developing one of my projects.

Boxes from `Sam3Predictor.predict` are consistently larger than the detected object and
shifted (down/right bias), while the returned masks are exact. Meta's reference
implementation returns tight boxes for the same inputs.

Repro (mlx-vlm 0.6.17, `mlx-community/sam3.1-bf16`): draw a black square at known pixel
coords on white, prompt "a black square", compare:

| canvas | ground truth (xyxy) | pred box error (x1,y1,x2,y2) | mask-bbox error |
|---|---|---|---|
| 800×600 | (300,200,500,400) | −10.5, −1.8, +19.0, +10.6 px | 0, 0, 0, 0 |
| 600×800 | (200,300,400,500) | −10.2, −0.1, +16.4, +12.1 px | 0, +1, 0, −1 |
| 1000×1400 | (350,500,650,900) | −15.8, −2.2, +25.6, +22.1 px | +1, +1, 0, −1 |

So the error is in the box path, not preprocessing (masks resized to image res are ≤1px
off). Workaround i'm using: tight bbox from `result.masks`.

An example of the bug in one of my projects.

Before:
Image

After with the workaround:
Image

I also compared with the bbox of an interence service and it matches the workaround result.

Lazarus-931:
Fixes #2103.

Three spots in the SAM3 vision/decoder path diverged from the reference: `compute_axial_cis` had no `rotary_scale` so global-attention blocks advanced RoPE positions 3x too fast, the box RPB coordinate grid added a half-cell offset where the reference uses cell corners, and `ref_point_head` applied a ReLU to its output where the reference 2-layer MLP activates only between layers, clamping `query_pos` and inflating every box.

Verified against `transformers` on identical weights: `pred_boxes` error on the issue's synthetic cases drops from -17..+29 px to ~1 px, matching HF to within 0.1 px, and objects that were previously missed entirely are detected again.

### DIFF
--- a/mlx_vlm/models/sam3/decoder.py
+++ b/mlx_vlm/models/sam3/decoder.py
@@ -354,9 +354,8 @@ class DETRDecoder(nn.Module):
         y2 = cy + h / 2
         boxes_xyxy = mx.stack([x1, y1, x2, y2], axis=-1)  # (B, Q, 4)
 
-        # Coordinate grids normalized to [0, 1]
-        coords_h = (mx.arange(height).astype(mx.float32) + 0.5) / height
-        coords_w = (mx.arange(width).astype(mx.float32) + 0.5) / width
+        coords_h = mx.arange(height).astype(mx.float32) / height
+        coords_w = mx.arange(width).astype(mx.float32) / width
 
         # Compute deltas: coords - box boundaries
         # y deltas: (B*Q, HW, 2) = coords_h - [y1, y2]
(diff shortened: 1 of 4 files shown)


Lazarus-931:
@lucasnewman tagging you for review since you looked at the other qwen4_exp thread — this one is a straight reference-parity fix, no compensation code.

lucasnewman (approved):
Looks good!

PR #2132, opened 2026-09-01
Support multiple tool call invocations at once

lucasnewman:
Many models emit a list of tool calls to be executed consecutively, and we were dropping all tool calls when the list didn't contain exactly one element. This updates the tool call parser to resolve all of the calls.


Blaizzy (approved):
LGTM, thanks! 🚀

PR #2127, opened 2026-09-01
Add GLM-5 Next support

Blaizzy:
## Summary

- add GLM-5 Next text and multimodal model support with a native processor
- generalize FP8 handling and optimize sparse prefill and decode
- add MTP speculative decoding support
- align configuration, preprocessing, sparse attention, vision, KDA masking, tied embeddings, video metadata, and frame sampling with Transformers 5.16.1

## Validation

- GLM-5 test suite: 580 passed, 53 subtests passed
- sparse scorer recall: 100% at 4K, 16K, and 64K contexts
- dense and MoE parity max error: 2.38e-7
- vision parity max error: 1.24e-10
- KDA valid-token max error: 1.34e-7; padding behavior matches exactly
- video frame sampling parity: 1000/1000 cases
- checkpoint configuration parity: 59/59 fields
- Black, isort, autoflake, and diff checks pass

PR #2141, opened 2026-09-02
glm5_next: speed up multi-token steps in MLA and the sparse indexer

stooit:
Two related performance fixes (specific to glm5_next):

**Absorbed MLA path** — same change as ml-explore/mlx-lm#1817, scoped to `glm5_next`, the only model here with an absorbed path. It was gated on `L == 1`, so wider steps expanded the whole cached latent instead. The crossover is a fixed query count per model, `kv_lora_rank * (nope + value) / (2 * kv_lora_rank - nope - value)`, which is 512 here. `models/mla.py` gains `max_absorbed_queries()`. This model is sparse (`index_topk = 2048`), so the win is below that budget only: 3.4× on the verify step at cache 512, 0.99–1.02× above it. The dense-MLA models in the linked PR are where it matters. The sparse top-k gather stays at `L == 1` — correctness, not speed.

**Indexer pooling** — incremental pooling was gated on `S == 1`, so every verify fell back to full O(T) pooling. The suffix recompute is exact for any S: pools completed before `s0` cannot change, as `s0` is a multiple of `index_kpool` and tokens only append. Cost per sparse layer, stock → patched (medians of 3 runs, M3 Max):

| context | S=1 (control) | S=2 | S=4 |
|---:|---:|---:|---:|
| 8,192 | 0.692 / 0.679 | 1.215 → 0.830 | 1.215 → 0.780 |
| 32,768 | 0.685 / 0.696 | 1.993 → 0.861 | 1.980 → 0.857 |
| 131,072 | 1.007 / 1.006 | 5.729 → **1.117** | 5.906 → **1.390** |

Stock grows with context, patched is flat — 5.1× at 128K for S=2, or ~63 ms → ~12 ms across this model's 11 sparse layers. S=1 unchanged.

Reuse needs two checks that `S == 1` was masking: the pool must have been built at the current cache length minus this step's tokens, or a speculative rejection that trims the cache leaves pools describing rejected tokens (without it, 0.97 error on a scale of 2.5; with it, 2.4e-07); and padding on new tokens is checked directly, since `_no_pad` only records the last full rebuild.

Verified a wide step pools identically to the same tokens one at a time, for S = 2/3/4/8 at aligned and misaligned prefills — indices, cache length and top-k match exactly, pooled keys to a few float32 ULP. `pytest ./tests` gives 2699 passed / 5 skipped on both arms.

### DIFF
--- a/mlx_vlm/models/glm5_next/language.py
+++ b/mlx_vlm/models/glm5_next/language.py
@@ -347,20 +347,30 @@ class Glm5NextIndexer(nn.Module):
         kv_len = T
         kv_pos = mx.arange(T)
 
-        # Incremental pooling at decode: complete pools are stable across steps, so
-        # recompute only the suffix (last partial pool + any new pool) and reuse the
-        # cached complete pools -- turns the per-step pool cost from O(T) to O(kpool).
-        # Exact; falls back to full pooling on prefill, when padding is present, or when
-        # the cached pool's batch axis no longer matches the current batch. That last
-        # guard matters under continuous batching: BatchGenerator grows/shrinks the
-        # batch (extend/filter) on the batch axis but does not carry this per-cache
-        # _pool along, so a stale _pool must be discarded and rebuilt for one step.
+        # Incremental pooling: complete pools are stable across steps, so recompute
+        # only the suffix (last partial pool + the new tokens) and reuse the cached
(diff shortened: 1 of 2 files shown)


stooit:
Closing, superseded by #2127

PR #2142, opened 2026-09-02
qwen3_vl: render per-group video timestamps like the reference processor

Micha0827:
## Problem
The reference Qwen3-VL processor renders a video as one `<t.t seconds>` marker followed by its own `<|vision_start|>…<|vision_end|>` block **per temporal patch group**. mlx-vlm rendered a single block of `<|video_pad|>` tokens with no time information, so the model saw a timeless stack of frames.

Symptom (Qwen3.8-27B-MLX-4bit, synthetic 8 s clip: red square moves left→right, blue phase, then the text "42"): the model described the moving square as *several simultaneous squares in different areas* and never mentioned the "42" — the same frames passed as separate images were described correctly.

## Fix
- `processing_qwen3_vl.py`: expand each video into `grid_t` timestamped blocks. Timestamp = mean of the first and last frame of the group (reference formula), from the per-video `fps` that `prepare_inputs` already passes; falls back to the video processor's configured sampling rate (the reference falls back to a fixed fps as well).
- `get_rope_index` in `qwen3_5`, `qwen3_vl`, `qwen3_vl_moe` (used by `qwen3_5_moe` and `qwen4_exp` through inheritance): expand each video grid `(t, h, w)` into `t` rows of `(1, h, w)` — the split the reference performs — so position ids line up with the per-group blocks.

## Validation
- New `Qwen3VLVideoTimestampTests` in `mlx_vlm/tests/test_processors.py` (marker count, timestamps, token count, fps fallback).
- Real-world, same clip and model: prompt grows from 1792 to 1854 tokens (8 markers) and the description is now correct and ordered: "square appears → moves right → background turns blue → white again → the number 42 appears".

### DIFF
--- a/mlx_vlm/models/qwen3_5/language.py
+++ b/mlx_vlm/models/qwen3_5/language.py
@@ -1744,6 +1744,14 @@ class LanguageModel(nn.Module):
         video_token_id = self.config.video_token_id
         vision_start_token_id = self.config.vision_start_token_id
         mrope_position_deltas = []
+        # The processor emits one vision block per temporal patch group, so
+        # expand each video grid (t, h, w) into t rows of (1, h, w) — the same
+        # split the reference implementation performs before indexing.
+        if video_grid_thw is not None:
+            rows = []
+            for thw in video_grid_thw.tolist():
+                rows.extend([[1, thw[1], thw[2]]] * int(thw[0]))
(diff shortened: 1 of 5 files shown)


Blaizzy (approved):
LGTM, thanks!

PR #2150, opened 2026-09-03
mla: use the absorbed path for multi-token steps

stooit:
The absorbed MLA path is gated on `L == 1`. Any step with more than one query materialises K and V across the whole cached latent instead, and materialising costs O(T) in the cached length regardless of how many queries you have, while the absorbed path costs O(L). The crossover is therefore a fixed query count per model:

```
kv_lora_rank * (nope + v) / (2 * kv_lora_rank - nope - v)
```

For DeepSeek V3 that is 170 queries and for GLM-4 MoE Lite it is 398, against a gate that admits exactly one. So every speculative verify step, and every short continuation on a warm cache, has been taking the expensive branch.

This adds `max_absorbed_queries()` to `models/mla.py` and applies it in the ten models with an absorbed path: `deepseek_v3`, `deepseek_v32`, `kimi_k3`, `kimi_linear`, `longcat_flash`, `longcat_flash_sparse`, `glm4_moe_lite`, `glm_moe_dsa`, `glm5_next` and `youtu_vl`. The helper is byte-identical to the one in ml-explore/mlx-lm#1817, including the attribute name, so the two runtimes pick the same width for a shared model.

It is the same reasoning as my mlx-lm PR ml-explore/mlx-lm#1817, extended here to four models that do not exist in mlx-lm.

## Multi-turn chat, which needs no drafter

`DEFAULT_PREFILL_STEP_SIZE` is 2048, so a long prompt chunks above the crossover and is unaffected. A follow-up turn on a warm cache is not chunked though, it is one step of however many tokens the turn holds, and that is where most people will see this. Moonlight-16B-A3B-Instruct-4-bit at 32K context, full forward pass, M3 Max, medians of 5:

| follow-up turn | stock | patched | speedup |
|---:|---:|---:|---:|
| 16 tokens | 497 ms | 85 ms | 5.8x |
| 32 tokens | 515 ms | 161 ms | 3.2x |
| 64 tokens | 543 ms | 271 ms | 2.0x |

## Speculative verify steps

Same model and setup, at the draft lengths the in-tree MTP drafters use:

| context | L=1 | L=2 | L=4 | L=8 |
|---:|---:|---:|---:|---:|
| 32768 stock | 17.3 ms | 380.8 ms | 394.6 ms | 413.0 ms |
| 32768 patched | 17.3 ms | 22.1 ms | 28.7 ms | 47.8 ms |
| speedup | 1.00x | 17.3x | 13.8x | 8.6x |

At 16384 the same three steps are 10.7x, 8.9x and 5.9x, and at 4096 they are 4.2x, 3.4x and 2.5x. L=1 is unchanged because the gate already picked the absorbed path there, which makes that row a control on the rest of the table.

The stock row is nearly flat across L because the cost is dominated by materialising K and V over the cache, which does not depend on L at all, so one extra token costs 21x more than the first one does at 32K.

## Memory

Materialising also expands the latent into per-head K and V. Peak memory for a single attention call at 32768 context drops from 2478 MB to 462 MB at L=4, and the absorbed path stays lower at every L up to the threshold (2742 MB against 3602 MB at L=144).

## The sparse indexer gate is left alone

`deepseek_v32`, `longcat_flash_sparse`, `glm5_next` and `glm_moe_dsa` have a second `L == 1` gate on the sparse top-k gather. That one selects with `topk_indices[:, :, 0, :]`, the first query's top-k, which is only correct for a single query, so widening it would silently apply one query's selection to all of them. It stays at `L == 1` and there is a test asserting it stays there.

## Numerics

In float32 the two forms agree to a relative 3.4e-06 across L = 1 to 170 at batch 1, 2 and 4, because they are the same linear algebra associated differently. At L = 1 the patch is a no-op and output is bit-identical.

I want to be clear about the quantised case. On 4-bit weights the two forms reorder quantised matmuls, so a wide step is not bit-identical to the other form. Teacher-forced over 1024 positions with token-by-token prefill as the reference, the stock wide chunk agrees on 98.54% of argmaxes and the patched wide chunk on 98.34%. Every disagreement sits at a near-tie, with top-2 margins between 0.00 and 0.30 against a median margin of 0.664, on a logit scale of
23.4. Drift of this size is already present without the patch, because the stock
wide chunk diverges from sequential prefill at the same rate, and in free-running greedy decode at the same token.

## Testing

Unmodified main gives `3131 passed` and this branch gives `3137 passed`, the difference being the six tests added here, with per-file results otherwise identical, so the change moves no existing test either way. `test_smoke.py` fails to collect on this branch and on unmodified main alike, which is unrelated to this change.

`mlx_vlm/tests/test_absorbed_mla.py` covers the threshold formula, wide-step parity against the materialised form, and the preserved indexer gate.

## The glm5_next hunk overlaps #2127

`glm5_next` is one of the ten models changed here, and #2127 adds that file at full length, so this is the one place the two PRs touch the same lines. That overlap is what superseded my earlier #2141, which only changed `glm5_next`. I am happy to wait on #2127 and re-apply this hunk afterwards, or to drop `glm5_next` from this PR entirely if that is easier to review, since the other nine models stand on their own either way. Let me know which you prefer.

## What I did not measure

Moonlight-16B is the only one of the ten models I have local weights for, so the end-to-end numbers above are all from `deepseek_v3` and the other nine are covered by the unit tests and by the block being identical. I also have no cached model that pairs a dense-MLA target with a working MTP drafter, so the speculative numbers are per-step costs rather than a full generation loop with measured acceptance.

### DIFF
--- a/mlx_vlm/models/mla.py
+++ b/mlx_vlm/models/mla.py
@@ -1,7 +1,44 @@
 import math
+from typing import Optional
 
 import mlx.core as mx
 import mlx.nn as nn
+
+
+def max_absorbed_queries(
+    kv_lora_rank: int,
+    qk_nope_head_dim: int,
+    v_head_dim: int,
+    cache_len: Optional[int] = None,
+) -> int:
+    """Query count below which folding into the query beats materializing K/V.
+
+    Per head, with ``r = kv_lora_rank`` and ``d = qk_nope_head_dim +
+    v_head_dim``, the absorbed path costs ``L*r*d + 2*L*T*r`` and materializing
+    costs ``T*r*d + L*T*d``, where ``T`` is the post-update cache length. The
+    absorbed path wins while::
+
+        L < r*d / (r*d/T + 2*r - d)
+
+    With ``cache_len`` omitted this returns the ``T -> inf`` limit,
+    ``r*d / (2*r - d)``, which is the right answer once the cache is much
+    larger than the latent. Pass ``cache_len`` for the exact bound: on a cold
+    cache, where ``T == L``, it correctly falls below ``L``, because
+    materializing K and V is cheaper there for every current model.
+    """
+    d = qk_nope_head_dim + v_head_dim
+    denom = 2 * kv_lora_rank - d
+    if cache_len is not None and cache_len > 0:
+        denom += kv_lora_rank * d / cache_len
+    if denom <= 0:
+        return 1
+    return max(1, int(kv_lora_rank * d / denom))
+
+
+def latent_length(kv_latent) -> int:
+    """Length of the cached latent, tolerating the quantized 3-tuple form."""
+    arr = kv_latent[0] if isinstance(kv_latent, tuple) else kv_latent
+    return arr.shape[-2]
 
 
 class MultiLinear(nn.Module):
--- a/mlx_vlm/tests/test_absorbed_mla.py
+++ b/mlx_vlm/tests/test_absorbed_mla.py
@@ -1,0 +1,180 @@
+import unittest
+
+import mlx.core as mx
+
+from mlx_vlm.models.cache import KVCache
+from mlx_vlm.models.mla import latent_length, max_absorbed_queries
+
+# Dims chosen only to steer the gate; they do not affect the attention maths.
+FORCE_MATERIALIZED = (1, 1, 1)  # threshold 1
+FORCE_ABSORBED = (512, 511, 511)  # threshold ~261k
+
+
+def _dense_attentions():
+    """Every MLA attention whose __call__ is (x, mask, cache)."""
+    from mlx_vlm.models.deepseek_v3.config import ModelConfig as DSV3Config
+    from mlx_vlm.models.deepseek_v3.language import DeepseekV3Attention
+    from mlx_vlm.models.glm4_moe_lite.config import ModelConfig as GlmConfig
+    from mlx_vlm.models.glm4_moe_lite.language import Glm4MoeLiteAttention
+
+    common = dict(
+        vocab_size=128,
+        hidden_size=256,
+        intermediate_size=256,
+        moe_intermediate_size=128,
+        num_hidden_layers=1,
+        num_attention_heads=4,
+        num_key_value_heads=4,
+        n_shared_experts=1,
+        n_routed_experts=4,
+        num_experts_per_tok=2,
+        max_position_embeddings=512,
+        rms_norm_eps=1e-6,
+    )
+    yield "deepseek_v3", DeepseekV3Attention(
+        DSV3Config(model_type="deepseek_v3", **common)
+    )
+    yield "glm4_moe_lite", Glm4MoeLiteAttention(
+        GlmConfig(model_type="glm4_moe_lite", **common)
+    )
+
+
+class TestMaxAbsorbedQueries(unittest.TestCase):
+    def test_asymptotic_matches_mlx_lm(self):
+        # cache_len omitted -> the T -> inf limit, as in ml-explore/mlx-lm#1817
+        self.assertEqual(max_absorbed_queries(512, 128, 128), 170)
+        self.assertEqual(max_absorbed_queries(512, 192, 256), 398)
+
+    def test_cold_cache_rejects_the_absorbed_path(self):
+        # At T == L materializing is cheaper for every current model, so the
+        # threshold must fall below L rather than admit it.
+        for r, n, v in ((512, 128, 128), (512, 192, 256)):
+            for L in (2, 32, 64, 169, 398):
+                self.assertLess(
+                    max_absorbed_queries(r, n, v, cache_len=L),
+                    L,
+                    f"cold cache admitted L={L} for dims {(r, n, v)}",
+                )
+
+    def test_warm_cache_approaches_the_asymptote(self):
+        r, n, v = 512, 128, 128
+        limit = max_absorbed_queries(r, n, v)
+        self.assertEqual(max_absorbed_queries(r, n, v, cache_len=32768), limit - 1)
+        self.assertLess(max_absorbed_queries(r, n, v, cache_len=1024), limit)
+        # monotonic in cache length
+        seq = [max_absorbed_queries(r, n, v, cache_len=t) for t in (256, 1024, 8192)]
+        self.assertEqual(seq, sorted(seq))
+
+    def test_degenerate_dims_keep_the_decode_path(self):
+        self.assertEqual(max_absorbed_queries(64, 128, 128), 1)
+        self.assertGreaterEqual(max_absorbed_queries(1, 1, 1), 1)
+
+
+class TestLatentLength(unittest.TestCase):
+    def test_plain_array(self):
+        self.assertEqual(latent_length(mx.zeros((1, 1, 37, 8))), 37)
+
+    def test_quantized_tuple(self):
+        # A quantized KV cache hands back a 3-tuple rather than an array.
+        q = (mx.zeros((1, 1, 37, 2)), mx.zeros((1, 1, 37, 1)), mx.zeros((1, 1, 37, 1)))
+        self.assertEqual(latent_length(q), 37)
+
+
+class TestGateWiring(unittest.TestCase):
+    """Run the real __call__ and force each branch, so the gates are executed."""
+
+    def _both_branches(self, attn, L, cache_len):
+        mx.random.seed(0)
+        outs = []
+        for dims in (FORCE_MATERIALIZED, FORCE_ABSORBED):
+            mx.random.seed(0)
+            x = mx.random.normal((1, L, 256))
+            cache = KVCache()
+            if cache_len:
+                warm = mx.random.normal((1, L, 256))
+                mx.random.seed(0)
+                attn._absorbed_dims = dims
+                attn(warm, cache=cache)
+            attn._absorbed_dims = dims
+            outs.append(attn(x, cache=cache))
+        mx.eval(outs)
+        return outs
+
+    def test_branches_agree_through_call(self):
+        for name, attn in _dense_attentions():
+            mx.eval(attn.parameters())
+            with self.subTest(model=name):
+                materialized, absorbed = self._both_branches(attn, L=4, cache_len=8)
+                self.assertTrue(
+                    mx.allclose(materialized, absorbed, atol=1e-4, rtol=1e-4),
+                    f"{name}: absorbed and materialized disagree through __call__",
+                )
+
+
+class TestGatePairing(unittest.TestCase):
+    """The two gates must never disagree: one boolean drives both."""
+
+    MODELS = [
+        "deepseek_v3",
+        "deepseek_v32",
+        "kimi_k3",
+        "kimi_linear",
+        "longcat_flash",
+        "longcat_flash_sparse",
+        "glm4_moe_lite",
+        "glm_moe_dsa",
+        "glm5_next",
+        "youtu_vl",
+    ]
+
+    def test_one_boolean_two_uses(self):
+        import pathlib
+
+        root = pathlib.Path(__file__).resolve().parents[1] / "models"
+        for m in self.MODELS:
+            src = (root / m / "language.py").read_text()
+            with self.subTest(model=m):
+                self.assertEqual(
+                    src.count("absorbed = L == 1 or L <= max_absorbed_queries("),
+                    1,
+                    f"{m}: expected exactly one gate decision",
+                )
+                self.assertEqual(
+                    src.count("if absorbed:"), 2, f"{m}: expected both gates to use it"
+                )
+
+
+class TestIndexerGateUnchanged(unittest.TestCase):
+    """The sparse top-k gather stays at L == 1.
+
+    It selects with ``topk_indices[:, :, 0, :]``, the first query's top-k, so
+    widening it would apply one query's selection to all of them.
+    """
+
+    SPARSE = [
+        "deepseek_v32",
+        "longcat_flash_sparse",
+        "glm5_next",
+        "glm_moe_dsa",
+    ]
+
+    def test_first_query_gather_is_still_gated_on_one_query(self):
+        import pathlib
+        import re
+
+        root = pathlib.Path(__file__).resolve().parents[1] / "models"
+        gate = re.compile(
+            r"if L == 1:\s*\n\s*(?:clamped = mx\.clip\(topk_indices|idx = topk_indices)"
+        )
+        for m in self.SPARSE:
+            src = (root / m / "language.py").read_text()
+            with self.subTest(model=m):
+                self.assertRegex(
+                    src,
+                    gate,
+                    f"{m}: the first-query top-k gather is no longer gated on L == 1",
+                )
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/mlx_vlm/models/deepseek_v3/language.py
+++ b/mlx_vlm/models/deepseek_v3/language.py
@@ -10,7 +10,7 @@ from ..base import (
     create_attention_mask,
     scaled_dot_product_attention,
 )
-from ..mla import MultiLinear
+from ..mla import MultiLinear, latent_length, max_absorbed_queries
 from ..mlp import DeepseekMLP as DeepseekV3MLP
 from ..pipeline import PipelineMixin
 from ..rope_utils import initialize_rope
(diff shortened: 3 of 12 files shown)


mlx_vlm/tests/test_absorbed_mla.py
```diff
@@ -0,0 +1,100 @@
...
+    """The sparse top-k gather reads topk_indices[:, :, 0, :] and stays at L == 1."""
+
+    def test_sparse_models_keep_the_indexer_gate(self):
+        import inspect
+
+        from mlx_vlm.models import deepseek_v32, glm5_next, longcat_flash_sparse
+
+        for mod in (deepseek_v32, glm5_next, longcat_flash_sparse):
+            src = inspect.getsource(mod.language)
+            self.assertIn(
```
Blaizzy: This is the whole CI failure. `isort` (5.13.2, `--profile=black`, per `.pre-commit-config.yaml`) collapses this to one line — it fits in 88 columns — and the `test` job fails on `git diff --quiet` afterwards. Reproduced locally:

```diff
-        from mlx_vlm.models import (
-            deepseek_v32,
-            glm5_next,
-            longcat_flash_sparse,
-        )
+        from mlx_vlm.models import deepseek_v32, glm5_next, longcat_flash_sparse
```

`black` and `autoflake` both pass, so that one hunk is the only thing red.

---
_Generated by [Claude Code](https://claude.ai/code)_

mlx_vlm/models/mla.py
```diff
@@ -4,6 +4,20 @@
...
+) -> int:
+    """Query count below which folding into the query beats materializing K/V.
+
+    Absorbed cost scales with the number of new queries; materializing scales
+    with the cached length, so the crossover does not depend on context size.
+    """
+    denom = 2 * kv_lora_rank - qk_nope_head_dim - v_head_dim
+    if denom <= 0:
+        return 1
+    return max(1, int(kv_lora_rank * (qk_nope_head_dim + v_head_dim) / denom))
```
Blaizzy: I worked the crossover through and the formula is right, but "the crossover does not depend on context size" is only true in the large-`T` limit, and the case where it isn't true is reachable.

Per head, with `r = kv_lora_rank`, `d = qk_nope_head_dim + v_head_dim`, `T` the post-update cached length:

- absorbed: `L·r·d` (embed_q + unembed_out) + `2·L·T·r` (scores + attn@latent)
- materialized: `T·r·d` (building K and V) + `L·T·d` (scores + attn@v)

Dropping the `L·r·d` term and dividing by `T` gives exactly `L < r·d / (2r − d)`, so the helper is the `T → ∞` answer. Keeping the term, the real condition is `L < r·d / (r·d/T + 2r − d)`.

That matters at the other end. At `T = L` — a cold cache — it reduces to `2r < d`, false for all ten models here, so the absorbed path is *always* the wrong choice on a cold cache. And a cold-cache wide step is not hypothetical: `generate/ar.py:494` chunks at `n_to_process = min(prefill_step_size, len − 1)`, so a prompt shorter than `prefill_step_size` is one step with `T = L`. For DeepSeek V3 dims at `L = T = 169` that's `169·512·256 + 2·169²·512 = 51.9M` vs `169·512·256 + 169²·256 = 29.6M` — about 1.75x more attention FLOPs than before.

In absolute terms a sub-170-token cold prompt is cheap, so I don't think this blocks anything. But `offset` is already in scope at all ten call sites (`offset = cache.offset` / `cache[0].offset`), so `L <= max_absorbed and L <= f(offset)` would be nearly free and would make the gate right at both ends. At minimum I'd soften the docstring — as written it reads as unconditional, and someone will later reuse the helper somewhere the `T ≫ r` assumption doesn't hold.

---
_Generated by [Claude Code](https://claude.ai/code)_

mlx_vlm/tests/test_absorbed_mla.py
```diff
@@ -0,0 +1,100 @@
...
+                "if L == 1:",
+                src,
+                f"{mod.__name__} lost its sparse-indexer L == 1 gate",
+            )
+
+
+if __name__ == "__main__":
+    unittest.main()
```
Blaizzy: This is the test guarding the invariant the PR description leans on hardest, and as a source-text grep it doesn't actually guard it.

`assertIn("if L == 1:", src)` passes if *any* `if L == 1:` survives anywhere in the file. It would pass on a revert of the attention gate (that restores two of them per file). It would fail on a harmless `if L < 2:` refactor of the indexer gate. And it can't tell the two gates apart, which is the whole distinction being protected.

Two concrete gaps:

- **`glm_moe_dsa` is missing.** It has the same first-query gather at `mlx_vlm/models/glm_moe_dsa/language.py:76` (`topk_indices[:, :, 0, :, None]`), and it's one of the ten files this PR edits, so it's exactly as much at risk as the three listed.
- A behavioural assertion would hold. With a stub indexer returning fixed `topk_indices`, call the attention at `L=1` and `L=2` and assert on the shape that comes back from the gather — `kv_latent.shape[2] == topk` at `L=1` vs unchanged at `L=2`. That fails on a real widening and survives refactors.

---
_Generated by [Claude Code](https://claude.ai/code)_

mlx_vlm/tests/test_absorbed_mla.py
```diff
@@ -0,0 +1,100 @@
...
+                attn.unembed_out(kv),
+                cache=None,
+                scale=attn.scale,
+                mask=pe,
+            )
+            mx.eval(absorbed, materialized)
+            self.assertTrue(
+                mx.allclose(absorbed, materialized, atol=1e-4, rtol=1e-4),
+                f"absorbed and materialized MLA disagree at L={L}",
+            )
```
Blaizzy: This hand-rolls both forms outside `DeepseekV3Attention.__call__`, so it asserts the algebra — which is true by construction, the two forms are the same product associated differently — but never runs any of the 20 gates the PR actually changes.

It would still pass if a file applied `embed_q` under the widened gate but left `unembed_out` under `L == 1`, which is the realistic failure mode for a ten-file mechanical edit with two gates each. (I checked all ten by hand and they're consistent — `grep -c 'L <= self._max_absorbed'` is exactly 2 in every file — but nothing in the suite keeps them that way.)

Calling the module instead would cover the wiring:

```python
attn._max_absorbed = 1          # force materialized
wide = attn(x, mask, cache=make_cache())
attn._max_absorbed = 64         # force absorbed
absorbed = attn(x, mask, cache=make_cache())
assert mx.allclose(wide, absorbed, atol=1e-4)
```

and it parameterizes over the ten attention classes for free, which would also cover the four models that don't exist in mlx-lm and therefore aren't covered by ml-explore/mlx-lm#1817's tests either.

---
_Generated by [Claude Code](https://claude.ai/code)_

mlx_vlm/tests/test_absorbed_mla.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def test_always_at_least_one(self):
+        # A degenerate rank must never disable the decode fast path.
+        self.assertEqual(max_absorbed_queries(64, 128, 128), 1)
+        self.assertGreaterEqual(max_absorbed_queries(1, 1, 1), 1)
+
+    def test_threshold_exceeds_one_for_real_models(self):
+        # Every affected model must widen past the old `L == 1` gate.
+        for rank, nope, v in ((512, 128, 128), (512, 192, 256), (512, 128, 128)):
+            self.assertGreater(max_absorbed_queries(rank, nope, v), 1)
```
Blaizzy: Nit: `(512, 128, 128)` appears twice, so this is two distinct triples, both of which the two tests above already assert exact values for (170 and 398). As written it adds no coverage. If the intent is "every affected model widens", it'd be worth pulling the real dims from the ten configs rather than restating two literals.

---
_Generated by [Claude Code](https://claude.ai/code)_

Blaizzy (commented):
Reviewed the diff and traced all ten call sites. The core reasoning holds up and I re-derived the threshold independently — it's the correct large-`T` crossover, and `L <= 170` is the right floor of `170.67` for DeepSeek V3 dims. The old `L == 1` gate really was leaving a lot on the table.

**What I verified**

- All ten files convert exactly two gates each (`grep -c 'L <= self._max_absorbed'` → 2 everywhere), no copy-paste slip where one of the pair was missed.
- `_max_absorbed` is set before use in every path. `GlmMoeDsaAttention` has its own `__init__` but calls `super().__init__(config)`, so it inherits the assignment from `DeepseekV32Attention` — that's the one place it isn't obvious from the diff.
- `self.kv_lora_rank` / `self.qk_nope_head_dim` / `self.v_head_dim` all exist before the new call in all ten, including `longcat_flash` (set at lines 26–29) and `glm5_next` (447–450), where the surrounding code otherwise reads from `args`/`config` directly.
- Both branches consume the same mask object (`pe_scores`, or `attn_mask` in `glm5_next`), so widening the gate can't change masking — that was my main correctness worry and it's clean.
- The `L == 1` indexer gates are genuinely untouched: `deepseek_v32:176`, `glm5_next:516`, `longcat_flash_sparse:309`, `glm_moe_dsa:76`. In `glm5_next` the `L == 1` branch gathers `kv_latent` down to top-k while the `else` branch keeps it full-length and masks, so shapes stay consistent with the absorbed path in the `1 < L <= max_absorbed` window that's new.
- The helper is byte-identical to ml-explore/mlx-lm#1817 including `self._max_absorbed`, as claimed — I fetched that diff and compared.
- `int()` truncating to a floor is correct given the gate is `<=`. The `denom <= 0` guard preserves `L == 1` behaviour rather than disabling the decode path; the `max(1, ...)` after it is unreachable for any real dims, which is fine as belt-and-braces.

**CI**

The `test` job is red on `isort`, not `black` — the bot comment points at Black and that's a red herring. One import block in the new test file; details inline.

**Findings**

Nothing I'd call a correctness bug. Two things worth acting on:

1. The threshold ignores the cached length, and the docstring's context-independence claim is only the `T → ∞` limit. On a cold cache the absorbed path is strictly worse for every model here, and `generate/ar.py:494` does produce cold-cache wide steps for prompts shorter than `prefill_step_size` (~1.75x more attention FLOPs at `L = T = 169`). Small in absolute terms; the docstring is the part I'd fix regardless. Inline on `mla.py`.
2. Both new behavioural tests assert something adjacent to the change rather than the change. The parity test hand-rolls both forms outside `__call__` so it never touches a gate, and the indexer test is a substring grep over `inspect.getsource` that would pass on a full revert of the attention gate. Neither would catch the mismatched-pair failure mode this kind of ten-file edit invites. Both inline.

**On `glm5_next` / #2127**

Since the block is identical across all ten and the other nine are independent, dropping `glm5_next` here and re-applying after #2127 lands costs you nothing in review value — that seems like the lower-friction option, but it's the maintainers' call.

**One note on the quantised-KV interaction**, which cuts in your favour and isn't in the description: with `--kv-bits`, `update_and_fetch` hands back a quantised 3-tuple, and the materialised branch would do `self.embed_q(kv_latent, transpose=False)` on it — `x @ self.weight` against a tuple. The absorbed branch passes it straight to `quantized_scaled_dot_product_attention`, which handles any `L`. So in the `1 < L <= max_absorbed` window this patch turns a latent `TypeError` into working code. Pre-existing and out of scope (chunked prefill at 2048 still takes the broken branch), just worth knowing it's there.

---
_Generated by [Claude Code](https://claude.ai/code)_

mlx_vlm/tests/test_absorbed_mla.py
stooit: Fixed. Both hunks are gone: the import collapses to one line, and I re-ran all three hooks with the versions pinned in `.pre-commit-config.yaml` over every changed file, not just this one.

Worth noting for anyone hitting this: running `isort` on the file in isolation gives the opposite answer, because outside the repo root it cannot tell `mlx_vlm` is first-party and merges the two import sections. It has to be run from the repo root.

mlx_vlm/models/mla.py
stooit: You are right, and thank you for working it through. I reproduced it before changing anything: forcing both branches at `T == L` on DeepSeek V3 dims gives absorbed/materialized of 0.558/0.436 ms at 32, 0.665/0.479 at 64 and 2.123/1.088 at 169, so 1.28x to 1.95x, tracking your predicted FLOP ratios.

`max_absorbed_queries` now takes `cache_len` and uses the exact condition. Omitted, it still returns the `T -> inf` value, which keeps it consistent with ml-explore/mlx-lm#1817; passed, it correctly falls below `L` on a cold cache. Checked across 64 dimension combinations that `T == L` never admits `L`, that the result is monotonic in cache length and never exceeds the asymptote, and that `cache_len` of `None`, `0`, `-1` and `2**62` are all handled. The docstring now states both limits rather than the unconditional claim.

I took the cache length from the latent rather than `offset`, since `offset` is not in scope at two of the ten sites, and `latent_length()` handles the quantized 3-tuple.

Two small corrections to the numbers: your absorbed figure is 51.4M rather than 51.9M, which leaves the 1.75x intact, and the `min(prefill_step_size, N - 1)` statement is at `ar.py:486`. At whole-model level the regression measured about 2%, 136.5 ms against 133.7 ms at a 169-token prompt. Small, but wrong in the same direction you said, and reachable.

The same fix is now in ml-explore/mlx-lm#1817, which had the identical bug.

mlx_vlm/tests/test_absorbed_mla.py
stooit: Fixed both parts. `glm_moe_dsa` is now covered, so all four sparse models are, and the test matches the gather itself rather than any `if L == 1:` in the file:

```python
gate = re.compile(r"if L == 1:\s*\n\s*(?:clamped = mx\.clip\(topk_indices|idx = topk_indices)")
```

I mutation-tested it rather than assume: widening that gate to `if L <= 4:` fails the suite.

mlx_vlm/tests/test_absorbed_mla.py
stooit: Fair hit, it never touched a gate. Rather than add a second condition I made the structure prevent it: both guards now read one boolean computed once per call, `absorbed = L == 1 or L <= max_absorbed_queries(*self._absorbed_dims, latent_length(kv_latent))`, so a widened `embed_q` with an untouched `unembed_out` cannot be expressed.

The test also runs the real `__call__` now, forcing each branch through `_absorbed_dims`, and there is a structural test that each of the ten files makes exactly one gate decision with two uses.

Mutation-tested against the failure mode you described: unpairing the gates, reverting one file entirely, widening an indexer gate, and dropping the cache-length term all fail the suite.

mlx_vlm/tests/test_absorbed_mla.py
stooit: Removed. That assertion is gone, replaced by the cold-cache and monotonicity properties, which cover behaviour the two exact-value tests do not.

stooit:
Thanks for the review! Everything is addressed and pushed; replies are inline on each thread. Summary:

**Cold-cache regression.** Reproduced it before changing anything, 1.28x to 1.95x on the attention op at `T == L`, tracking your FLOP prediction. `max_absorbed_queries` now takes the cache length and uses the exact condition, falling back to the `T -> inf` value when omitted so it stays consistent with ml-explore/mlx-lm#1817. At whole-model level the regression measured about 2%. The same fix is now on the mlx-lm PR, which had the identical bug.

**Gate wiring.** Both guards now read one boolean computed once per call, so the mismatched-pair failure you described cannot be expressed rather than merely being tested for.

**Tests.** They run the real `__call__` and force each branch, `glm_moe_dsa` is covered, the indexer test matches the gather rather than a substring, and the redundant triple is gone. I mutation-tested the suite against four breakages, including the one you called out, and all four fail it.

**CI.** Fixed, and it was `isort` as you said.

**`glm5_next`.** I have kept it in, because `TestGatePairing` now covers it and asserts one decision with two uses, so if #2127 lands afterwards and takes its side of that file, the test fails rather than the hunk vanishing quietly. Happy to drop it if you would still rather.

Two corrections. Your absorbed FLOP figure is 51.4M rather than 51.9M, which leaves the 1.75x ratio intact, and the `min(prefill_step_size, N - 1)` statement is at `ar.py:486`.

One thing I could not confirm: I verified the first half of your quantized-KV note, the materialized branch does fail on a quantized latent with `TypeError: unsupported operand type(s) for @: tuple and mlx.core.array`. I could not verify that the absorbed branch handles it, because my harness failed on both branches, so I have not claimed this PR fixes that.

Test suite: unmodified main 3131 passed, this branch 3140, the difference being the nine tests added here, per-file otherwise identical.

Blaizzy (approved):
LGTM, thanks!

PR #2151, opened 2026-09-03
Enable speculative decoding for Laguna

Lazarus-931:
The Laguna DFlash drafter is registered but speculative decoding could not run: the target model exposed no `rollback_speculative_cache` and silently dropped `capture_layer_ids`, so the dflash round loop refused to start. The drafter config also rejected the published XS checkpoints over a missing `sliding_windows` field, which is redundant with `sliding_window` and is now derived when absent.

| ctx | baseline | dflash | speedup | mean accept |
| --- | --- | --- | --- | --- |
| 512 | 84.70 | 304.27 | 3.59x | 10.64 |
| 2048 | 83.48 | 313.96 | 3.76x | 11.80 |
| 4096 | 81.32 | 343.29 | 4.22x | 13.22 |

Decode-only tok/s, 128 tokens greedy, target `poolside/Laguna-XS-2.1-NVFP4-mlx` with drafter `poolside/Laguna-XS-2.1-DFlash`. Output is token-identical to base decoding at every context.

### DIFF
--- a/mlx_vlm/models/laguna/language.py
+++ b/mlx_vlm/models/laguna/language.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, List, Optional, Union
 
 import mlx.core as mx
 import mlx.nn as nn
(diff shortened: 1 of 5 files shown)


lucasnewman (approved):
⚡️

PR #2164, opened 2026-09-05
fix(unlimited-ocr): match native completed-task prompt formatting

### LINKED ISSUE #2161: Unlimited-OCR inherits a trailing-space chat template that differs from the native task prompt
TL;DR: `UnlimitedOCRProcessor` inherits a default chat template from `DeepseekOCRProcessor` that appends a literal space, rendering `<image>document parsing. ` instead of the documented `<image>document parsing.`. In the tested tokenizer this adds token 223, and a controlled local ablation shows that removing it changes generation on the tested page.

### Inherited template source

The native prompt format is documented in the [Unlimited-OCR README](https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/unlimited_ocr/README.md). The relevant implementation is the [Unlimited-OCR processor](https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/unlimited_ocr/processing_unlimitedocr.py) and its inherited [default template](https://github.com/Blaizzy/mlx-vlm/blob/main/mlx_vlm/models/deepseekocr/processing_deepseekocr.py).

The inherited property returns these concatenated Python string literals:

```python
template = (
    "{% for message in messages %}"
    "{% if message['role'] == 'user' %}"
    "{% elif message['role'] == 'assistant' %}{% endif %}"
    "{{message['content']}} "
    "{% endfor %}"
    "{% if add_generation_prompt %}{% endif %}"
)
```

The space before the closing quote in `"{{message['content']}} "` is literal output. Python concatenation adds no newlines between these fragments.

The model's own README corroborates the mismatch: its multi-page example checks `prompt.strip() == "<image>Multi page parsing."`, not raw-string equality. This normalizes the assertion only; the example still passes the original `prompt` to `generate`. It is not an inference-time trimming fix.

### Image-free reproduction

Run with mlx-vlm 0.6.3 and its dependencies. No checkpoint, tokenizer download, image, server, or model inference is required. This deliberately isolates the default-template boundary rather than exercising the entire generation path.

```python
from jinja2 import Template
from mlx_vlm.models.unlimited_ocr.processing_unlimitedocr import UnlimitedOCRProcessor

# The inherited property only returns a template string; no initialized
# tokenizer or processor state is needed to inspect and render that default.
processor = object.__new__(UnlimitedOCRProcessor)
rendered = Template(processor.default_chat_template).render(
    messages=[{"role": "user", "content": "<image>document parsing."}],
    add_generation_prompt=True,
)
expected = "<image>document parsing."
print("rendered:", repr(rendered))
print("expected:", repr(expected))
print("matches:", rendered == expected)
```

Observed:

```text
rendered: '<image>document parsing. '
expected: '<image>document parsing.'
matches: False
```

Expected: the model's default completed task prompt matches the documented native format, without an unintended extra suffix token. Explicit custom templates, assistant continuations, and other models need separate handling.

### Template source and precedence

On the tested processor path, `processor.chat_template` supplies the default rendering template. `UnlimitedOCRProcessor.from_pretrained` reads `processor_config.json`, overlays loader kwargs, and creates the tokenizer separately. A `chat_template` key in processor configuration or a loader override can therefore replace the model default.

The checkpoint's `chat_template.jinja` may populate `tokenizer.chat_template`, but that tokenizer template is not consulted by this processor rendering path. Changing it alone leaves rendering unchanged. Setting `processor.chat_template` to `None` raises `ValueError` rather than falling back to the tokenizer. An explicit `chat_template` argument to the processor's `apply_chat_template` call also overrides the default; preserve that existing behavior.

The locally added checkpoint Jinja file happened to contain the same trailing space, but is inert on the tested path. It is not the cause of this rendering mismatch. The inherited processor default is sufficient to reproduce it.

### Separate inference evidence

A controlled local experiment held processor/checkpoint configuration fixed and removed only the final space from the rendered prompt. In the tested tokenizer, that removes token 223 and changes the tiled-mode input length from 1518 to 1517.

For one document page, both Q8 and original BF16 weights hit the 8192-token cap with the spaced prompt. Removing the suffix produced EOS at 1798 tokens for Q8 and 1799 for BF16. Repeated runs were byte-identical within each condition.

This did not recover perfect transcription: normalized CER remained about 39%. A different page still hit the cap in tiled mode after trimming; base mode was a separate intervention. Quantization is therefore not necessary for these observed failures, but neither EOS nor image-cropping code is proven defective.

### Environment

The inherited template and Unlimited-OCR inheritance were rechecked against upstream main at [`d68a25e7`](https://github.com/Blaizzy/mlx-vlm/commit/d68a25e71e842e8924a54bb3d84d3a3b4d4a2ee1) on 2026-09-04. The image-free reproduction was rerun with the installed versions below.

Python 3.12.11, MLX 0.32.2, mlx-vlm 0.6.3, mlx-lm 0.31.3, transformers 5.12.1. Direct-model generation used temperature 0 and no n-gram guard. Original BF16 weights came from `baidu/Unlimited-OCR` revision `84757cb04b4d9b74adc65dc31cdf52fca7dc2a97`.

### Requested direction

Would a `default_chat_template` override on `UnlimitedOCRProcessor` be appropriate? The fix should preserve single-`<image>` multi-page formatting and assistant partial behavior, and should not alter the shared `DeepseekOCRProcessor` template used by other models.

### Scope and non-goals

- **Reproduction boundary:** The image-free snippet proves the default-template mismatch; the page-level generation results come from separate experiments.
- **OCR quality:** Removing the suffix does not fix every runaway or ensure accurate transcription. The remaining CER and tiled-mode failure are reported above; image mode was a separate intervention.
- **Template compatibility:** Preserve explicit overrides and meaningful assistant-continuation whitespace. This is not a global whitespace-stripping proposal or a change to other models' templates.
- **Reference coverage:** BF16/Q8 used the same local processor configuration, not every official loading path. The original PyTorch/SGLang runtime was not executed.

contrapuntal:
### Summary

Override `UnlimitedOCRProcessor.default_chat_template` so a final user task renders without an extra template-added space. For example, the default now produces `<image>document parsing.` instead of `<image>document parsing. `.

The template preserves message content verbatim, existing separators between messages, and final assistant/continuation behavior. Explicit constructor, processor-config, loader, and per-call templates keep their existing precedence. The shared DeepSeek-OCR template is unchanged.

Closes #2161.

### Verification

Added 20 weight-free tests using a locally constructed byte-level tokenizer. Coverage includes exact single-/multi-page prompt rendering through the processor and public prompt helper, both generation-prompt flags, content whitespace, message separators, assistant continuations, template override precedence, tokenization, and the unchanged DeepSeek-OCR default.

Before the implementation, the new suite produced 15 suffix-regression failures and five passing compatibility checks. After the implementation, the following targeted run passed: 278 passed, 2 skipped, plus 7 subtests.

```sh
python -m pytest \
  mlx_vlm/tests/test_unlimited_ocr_processor.py \
  mlx_vlm/tests/test_processors.py \
  mlx_vlm/tests/test_prompt_utils.py \
  mlx_vlm/tests/test_tokenizer_utils.py \
  mlx_vlm/tests/test_generation_defaults.py -q
```

Tested with Python 3.12.11 and Transformers 5.12.1. This is targeted verification, not a full repository suite run.

### Evidence and limits

This fixes the default prompt-format mismatch reported in the issue. The tests do not run model inference or assert OCR quality, multi-page image-processing correctness, or native-runtime parity. The issue's separate Q8/BF16 ablations are motivation, not end-to-end verification of this patch. No oMLX changes are included.

### DIFF
--- a/mlx_vlm/models/unlimited_ocr/processing_unlimitedocr.py
+++ b/mlx_vlm/models/unlimited_ocr/processing_unlimitedocr.py
@@ -15,6 +15,17 @@ from ..deepseekocr.processing_deepseekocr import (
 
 
 class UnlimitedOCRProcessor(DeepseekOCRProcessor):
+    @property
+    def default_chat_template(self):
+        # Completed user tasks must not gain a suffix token. Keep the inherited
+        # separators between messages and after assistant content unchanged.
+        return (
+            "{% for message in messages %}"
+            "{{message['content']}}"
+            "{% if not loop.last or message['role'] != 'user' %} {% endif %}"
+            "{% endfor %}"
+        )
+
     @classmethod
     def from_pretrained(cls, pretrained_model_name_or_path, **kwargs):
         kwargs.pop("trust_remote_code", None)
--- a/mlx_vlm/tests/test_unlimited_ocr_processor.py
+++ b/mlx_vlm/tests/test_unlimited_ocr_processor.py
@@ -3,6 +3,8 @@
 import json
 
 import pytest
+
+pytest.importorskip("jinja2")
 from tokenizers import Tokenizer, decoders, models, pre_tokenizers
 from transformers import PreTrainedTokenizerFast
 
--- a/mlx_vlm/tests/test_unlimited_ocr_processor.py
+++ b/mlx_vlm/tests/test_unlimited_ocr_processor.py
@@ -1,0 +1,168 @@
+"""Unlimited-OCR prompt boundaries, without model weights or downloads."""
+
+import json
+
+import pytest
+from tokenizers import Tokenizer, decoders, models, pre_tokenizers
+from transformers import PreTrainedTokenizerFast
+
+from mlx_vlm.models.deepseekocr.processing_deepseekocr import DeepseekOCRProcessor
+from mlx_vlm.models.unlimited_ocr.processing_unlimitedocr import UnlimitedOCRProcessor
+from mlx_vlm.prompt_utils import apply_chat_template
+
+
+@pytest.fixture
+def tokenizer():
+    vocab = {"[UNK]": 0, "[PAD]": 1, "[BOS]": 2, "[EOS]": 3, "<image>": 4}
+    for char in sorted(pre_tokenizers.ByteLevel.alphabet()):
+        vocab[char] = len(vocab)
+    backend = Tokenizer(models.BPE(vocab=vocab, merges=[], unk_token="[UNK]"))
+    backend.pre_tokenizer = pre_tokenizers.ByteLevel(add_prefix_space=False)
+    backend.decoder = decoders.ByteLevel()
+    return PreTrainedTokenizerFast(
+        tokenizer_object=backend,
+        unk_token="[UNK]",
+        pad_token="[PAD]",
+        bos_token="[BOS]",
+        eos_token="[EOS]",
+        additional_special_tokens=["<image>"],
+    )
+
+
+@pytest.fixture
+def processor(tokenizer):
+    return UnlimitedOCRProcessor(tokenizer=tokenizer)
+
+
+def render(processor, messages, **kwargs):
+    return processor.apply_chat_template(messages, tokenize=False, **kwargs)
+
+
+@pytest.mark.parametrize("task", ["document parsing.", "Multi page parsing."])
+@pytest.mark.parametrize("add_generation_prompt", [False, True])
+def test_completed_task_has_no_template_suffix(processor, task, add_generation_prompt):
+    prompt = "<image>" + task
+    assert (
+        render(
+            processor,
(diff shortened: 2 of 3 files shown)


contrapuntal:
This overlaps with #2169. One difference: this PR removes the template-added suffix only after a final user message, preserving existing final-assistant formatting; #2169 removes it after any final message. This PR also includes processor-level compatibility tests. Happy to consolidate into whichever PR you prefer.

contrapuntal:
Closing as superseded by #2169, which addresses the completed-user-task suffix mismatch. Upstream also now declares Jinja2. I’ll retain the broader regression coverage for a possible focused follow-up against the merged behavior.

PR #2171, opened 2026-09-05
Add MoGe-3 monocular geometry estimation model

eklipse2k8:
Port MoGe-3 (moge.model.v3) to MLX: DINOv2 encoder with output projections, convolutional neck + points/mask/normal/scale heads, and the sparse 3D UNet refiner.

The refiner normally depends on FlexGEMM (Triton, CUDA-only); here the submanifold convolution, mean pooling, and nearest upsample are re-implemented in pure MLX with sort + binary-search + gather, so the model runs on Apple silicon. Transposed convolutions are computed as exact per-pixel GEMMs followed by a 2x2 pixel shuffle, and all resizes match torch F.interpolate (including antialiased bilinear and the scale-factor bicubic position-embedding path).

Adds moge3 model package (config, vision, conv_stack, sparse, refiner, geometry, generate, processor, checkpoint converter, README) and a TestMoge3 test class.

Converted checkpoints uploaded to mlx-community/moge-3-vitl-mlx-fp32 and mlx-community/moge-3-vitg-mlx-fp32. Validated against the torch reference (CPU, fp32, pure-torch FlexGEMM shim): identical valid-pixel masks, <0.1% median relative depth error at default settings, ~2.1s per 800x547 image on GPU.

### DIFF
--- a/mlx_vlm/models/dinov2/README.md
+++ b/mlx_vlm/models/dinov2/README.md
@@ -1,0 +1,15 @@
+# DINOv2 (MLX)
+
+Channel-last MLX port of the [DINOv2](https://github.com/facebookresearch/dinov2)
+vision transformer. This package currently holds only the shared backbone and
+its architecture presets. The standalone model is not ported yet: there is no
+`Model` class, so `facebook/dinov2-*` checkpoints cannot be loaded through
+`mlx_vlm.load`, and register tokens, masking and the DINO heads are not
+implemented.
+
+## Used by
+
+| Model | Backbones |
+| --- | --- |
+| `video_depth_anything` | ViT-S/14, ViT-B/14, ViT-L/14 |
+| `moge3` | ViT-L/14, ViT-g/14 |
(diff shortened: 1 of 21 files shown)


Lazarus-931:
hi @eklipse2k8, thanks for the pr!

lgtm!

Lazarus-931 (approved):

mlx_vlm/models/dinov2_backbone.py
```diff
@@ -1,49 +1,33 @@
-"""DINOv2 vision backbone for Video Depth Anything (channel-last MLX port)."""
+"""Shared DINOv2 vision transformer (channel-last MLX port).
```
lucasnewman: I think this should stay in the original location or be moved to a `dinov2` directory -- we typically don't have free-floating model classes in the `models` root.

lucasnewman (changes requested):
@eklipse2k8 This looks really cool! Take a look at my comment about moving the backbone into its own directory -- if we resolve that we can merge.

I think it might be worth considering how we could make a unified interface for depth prediction models if you're planning to continue adding models in this family. Not required for this PR but just something to think about.

eklipse2k8:
Okay, I started a port of dinov2 and moved over the beginning bits. It has a huge family of vision models based on it, so I think it's worth the separation. I looked at the official repo for it, and it's large enough that it could use its own full port separate from this.

lucasnewman (approved):
Thanks!

PR #2172, opened 2026-09-06
Add native affine4 KV cache for Apple M5

Cianidos:
## Summary

- add a first-class `affine4` KV-cache format beside `uniform` and `turboquant`
- consume packed signed-int4 K/V directly with Metal Performance Primitives TensorOps on Apple M5 GPUs
- keep the same affine4 representation portable through a dequantized fallback on unsupported hardware and attention shapes
- support single-request generation, continuous batching, left padding, causal verification lengths 2–4, APC snapshots/disk restore, and cache migration
- expose `affine4` through generation, chat, and server configuration without changing existing defaults

TurboQuant stores nonlinear codebook indices, so its packed values cannot be interpreted as native signed-int4 operands. Affine4 is therefore an explicit format rather than a hardware-dependent reinterpretation of TurboQuant snapshots.

The native path is implemented with `mx.fast.metal_kernel`; it JIT-compiles MPP `matmul2d` operations over `metal::int4b_format`. It does not require a C++ extension, bundled metallib, or build-system changes.

## Format and routing

Affine4 applies a deterministic orthonormal rotation, computes one FP16 scale per KV vector, and stores signed codes in `[-8, 7]` as two's-complement nibbles. Key and value rotations use different seeds.

Native routing currently requires:

- Apple GPU architecture `applegpu_g17*`
- FP16 or BF16 queries
- head dimension divisible by 32 and at most 512
- grouped-query geometry with `query_heads % kv_heads == 0`
- query length 1–4
- no attention sinks or arbitrary array mask

Unsupported cases dequantize the same affine4 state and use existing attention paths. Failed JIT signatures are cached so they are not retried on every token.

## End-to-end benchmark

Hardware: Apple M5 Pro, 48 GB unified memory.  
Model: `Qwen3.6-35B-A3B-oQ4e-mtp` (`16q / 2kv / D=256`).  
Protocol: direct `generate_step`, batch 1, greedy, no server or speculative decoding, 2048-token prefill chunks, quantization from token 0, 128 generated tokens. Inputs were fixed across schemes. TurboQuant and affine4 use exactly the same cache bytes.

| Context | Affine4 | TurboQuant 4 | FP16 KV | Affine4 vs TQ4 | Affine4 vs FP16 |
|---:|---:|---:|---:|---:|---:|
| 8K | 80.34 tok/s | 75.19 tok/s | 80.57 tok/s | +6.8% | -0.3% |
| 32K | 74.40 tok/s | 58.67 tok/s | 67.58 tok/s | +26.8% | +10.1% |
| 55K | 70.07 tok/s | 48.88 tok/s | 59.19 tok/s | +43.3% | +18.4% |
| 100K | 60.71 tok/s | 36.75 tok/s | 46.72 tok/s | +65.2% | +29.9% |
| 150K | 54.35 tok/s | 28.37 tok/s | 38.30 tok/s | +91.6% | +41.9% |
| 200K | 49.53 tok/s | 23.33 tok/s | 32.83 tok/s | +112.3% | +50.8% |

The 200K comparison was repeated in reverse order:

- TQ4 first / affine4 second: 23.247 / 49.514 tok/s
- affine4 first / TQ4 second: 49.540 / 23.409 tok/s

Median per-token latency at 200K was 20.08 ms for affine4, 42.67 ms for TQ4, and 30.36 ms for FP16. P10–P90 ranges were 19.97–20.40 ms, 42.06–43.43 ms, and 30.29–30.82 ms respectively.

### Prefill

Regular long-query prefill is intentionally still handled by dequantization plus MLX SDPA; native packed attention is used for decode and short causal verification. Affine4 and TQ4 prefill rates remain close:

| Context | Affine4 | TQ4 | FP16 KV |
|---:|---:|---:|---:|
| 100K | 1276 tok/s | 1262 tok/s | 1368 tok/s |
| 150K | 996 tok/s | 996 tok/s | 1076 tok/s |
| 200K | ~824 tok/s | ~830 tok/s | 894 tok/s |

### Cache memory

| Context | Affine4 / TQ4 | FP16 KV |
|---:|---:|---:|
| 100K | 704 MiB | 2021 MiB |
| 150K | 1025 MiB | 2996 MiB |
| 200K | 1346 MiB | 3971 MiB |

At 200K, measured process peak was 26.3–26.7 GB for affine4/TQ4 and 28.17 GB for FP16 KV. The model keeps its final sensitive attention layer in FP16, matching existing quantized-cache policy.

## Numerical checks

At 100K, affine4, TQ4, and FP16 produced the same 128 greedy tokens. At 150K, affine4 matched FP16 while TQ4 diverged. At 200K all three paths crossed different close token boundaries, so a separate teacher-forced comparison was run after true 200K chunked prefill:

| Metric vs FP16, 64 steps | Affine4 | TQ4 |
|---|---:|---:|
| mean cosine | 0.998593 | 0.998454 |
| minimum cosine | 0.994687 | 0.994621 |
| mean KL | 0.003615 | 0.003837 |
| maximum KL | 0.015028 | 0.009978 |
| top-1 agreement | 61/64 | 59/64 |

Native-kernel tests additionally compare against SDPA over the affine-dequantized cache across dimensions 32–512, dense and MoE GQA geometries, padded batches, and causal lengths 1–4.

## Tests

```text
3216 passed, 8 skipped, 104 subtests passed
```

Full test command excluded three Nemotron voicechat modules because the local optional `mlx-audio` installation does not provide their codec package. Affine4, TurboQuant, generation, batching, speculative KV, APC, server, and model tests all ran in the remaining suite.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1605,6 +1605,32 @@ TurboQuant automatically quantizes `KVCache` layers (global attention). Models w
 
 TurboQuant is supported in both single-request generation and continuous batching on the server. In continuous batching mode, KV states are stored in TurboQuant's compressed format and dequantized at attention time (custom Metal kernels are not yet batch-aware).
 
+### Native Affine4 on M5
+
+Apple M5 GPUs can execute signed 4-bit matrix operands directly through Metal
+Performance Primitives TensorOps. The `affine4` KV scheme uses a native-friendly
+signed-int4 cache format instead of TurboQuant's nonlinear codebook:
+
+```sh
+mlx_vlm generate \
+  --model mlx-community/Qwen3.5-4B-4bit \
+  --kv-bits 4 \
+  --kv-quant-scheme affine4 \
+  --prompt "Your long prompt here..."
+```
+
+Affine4 applies a deterministic orthonormal rotation, stores one FP16 scale per
+KV vector, and keeps two signed values in each byte. On M5, decode attention
+consumes packed K/V directly with `int4b_format` TensorOps. Standard attention
+geometries whose head dimension is divisible by 32 and at most 512 use the
+native path, including grouped-query attention, short causal speculative
+verification, and left-padded batches. Other hardware and unsupported attention
+shapes use the portable affine4 implementation without changing cache semantics.
+
+`affine4` and `turboquant` are distinct formats. Affine4 requires exactly four
+bits for both keys and values; fractional widths and mixed affine4 schemes are
+rejected. Existing `uniform` and `turboquant` defaults are unchanged.
+
 ## Distributed Inference
 
 mlx-vlm supports distributed inference across multiple computers. It works by sharding the language model (not the vision tower), because the LLM is much larger and vision embeddings only need to be computed once.
--- a/mlx_vlm/affine4.py
+++ b/mlx_vlm/affine4.py
@@ -1,0 +1,978 @@
+from __future__ import annotations
+
+import logging
+import math
+from functools import cache, lru_cache
+
+import mlx.core as mx
+
+from .kv_quant import AFFINE4_SCHEME
+from .models.cache import create_causal_mask
+from .turboquant import (
+    DEFAULT_TURBOQUANT_SEED,
+    BatchTurboQuantKVCache,
+    TurboQuantKVCache,
+    TurboQuantMSEState,
+    _pack_lowbit,
+    _rht_forward,
+    _rht_inverse,
+    _rotation_matrix,
+    _slice_state,
+    _state_length,
+    _unpack_lowbit,
+)
+
+logger = logging.getLogger("mlx_vlm.affine4")
+
+__all__ = [
+    "AFFINE4_SCHEME",
+    "Affine4Codec",
+    "Affine4KVCache",
+    "BatchAffine4KVCache",
+]
+
+_BITS = 4
+_MIN_NATIVE_TOKENS = 256
+_MAX_PADDED_ROWS = 32
+_NATIVE_LAUNCHABLE = {}
+_FUSED_QUANTIZE_LAUNCHABLE = {}
+
+
+class Affine4Codec:
+    """Signed int4 codec with deterministic orthonormal rotation."""
+
+    bits = _BITS
+
+    def __init__(self, dim: int, seed: int):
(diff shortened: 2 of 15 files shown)


Lazarus-931:
hi @Cianidos, thanks for the pr!

I totally like this, but unsure if repo wise, this is the best long term solution.

I think the goal is to do drive the best perf for any hardware x model, but feel like this can get out of hand quickly.

I say we park this for now, and revisit :)

Cianidos:
Thanks — the long-term maintenance concern makes sense. Happy to park this while we find the right layer for it.

One possible direction would be to treat `affine4` as an alternative codec/backend under the existing quantized KV-cache infrastructure, rather than as another model-specific optimization. The attention and cache plumbing is already geometry-driven rather than Qwen-specific; the M5 implementation could remain isolated and lazily loaded, while unsupported hardware uses the portable affine4 path.

I would keep `affine4` as an explicit scheme rather than silently selecting it for `turboquant`, because TurboQuant’s nonlinear codebook indices are not signed int4 values. Switching based on hardware would change cache semantics and make snapshots non-portable. But both schemes could sit behind the same generic cache/backend interface.

I think the optimization is worth finding an appropriate home because the gain grows substantially with context while retaining the same cache size as TQ4:

- 32K: 74.4 vs 58.7 tok/s
- 100K: 60.7 vs 36.8 tok/s
- 200K: 49.5 vs 23.3 tok/s

At 200K it is also 50.8% faster than FP16 KV, with roughly one-third of the KV-cache memory.

Would either of these shapes fit the repository better?

1. Keep only a generic quantized-cache codec/backend interface plus portable affine4 support here, with the M5 kernel maintained downstream.
2. Move the native signed-int4 attention primitive into MLX/MLX-LM first, then keep the `mlx-vlm` integration small.

If an explicit affine4 variant under the existing quantized KV-cache/TurboQuant infrastructure would be acceptable, I can reshape and reduce this PR around that boundary. Otherwise, I’m fine leaving it parked.

Lazarus-931:
Hi @Cianidos, this seems like something to visit later on.

I think an issue/request for specific kernel support in mlx-vlm could be a great start. 

We can circle back and address this.

PR #2177, opened 2026-09-07
fix(deepseek_vl_v2): convert spatial crop scalars to int

inureyes:
Follow-up to #1982, which fixed 12 files but not this one. `deepseek-vl2` was not among the models swept in #1981, so the same defect went unobserved here.

`images_spatial_crop[idx, jdx]` unpacks into 0-d `mx.array` values that reach `mx.reshape` inside a shape tuple, and since the nanobind 2.15 bump those no longer convert implicitly. Every image fails:

```
TypeError: reshape(): incompatible function arguments.
  1. reshape(a: array, /, shape: Sequence[int], ...)
Invoked with types: mlx.core.array, tuple
```

at `deepseek_vl_v2.py:236` in `process_image_features`. Instrumenting the call confirms the tuple is `['array', 'array', 'int', 'int', 'int']` with values `[1, 1, 14, 14, 2048]`.

The fix matches the `lfm2_vl` change in #1982. The `.tolist()` on the tile count is dropped because its operands are now ints.

Verified with `mlx-community/deepseek-vl2-small-4bit`, mlx 0.32.2, mlx-vlm at `64f44c0`:

| | result |
|---|---|
| before | `TypeError` at warmup |
| after | 436 prompt tokens, generates normally |
| revert the patch | `TypeError` again |

436 is what the same checkpoint produces on mlx 0.31.1, where the implicit conversion still worked, so this restores the original behaviour rather than masking the error. Reproduced on M5 Max and M1 Ultra. `pre-commit run --files` passes.


Lazarus-931 (approved):
great, thanks @inureyes!

PR #2184, opened 2026-09-07
Add standalone DINOv2 model with register tokens and mask support

eklipse2k8:
Finish the DINOv2 port: a Model class that loads the HF facebook/dinov2-* and facebook/dinov2-with-registers-* checkpoints directly (sanitize maps HF keys, fuses qkv, transposes the patch conv), register tokens, patch masking (bool_masked_pos), forward_features, and a ModelConfig following the HF dinov2/dinov2_with_registers fields.

Pos-embed interpolation gains the HF size-based branch with antialiased bicubic (matches ATen _upsample_bicubic2d_aa); the original repo scale-factor + offset path is kept for the MoGe-3/VDA backbones.

Validated against transformers on facebook/dinov2-small and facebook/dinov2-with-registers-small at several resolutions (~1e-4 fp32 noise), including masked inputs and the SwiGLU FFN.

### DIFF
--- a/mlx_vlm/models/dinov2/README.md
+++ b/mlx_vlm/models/dinov2/README.md
@@ -1,11 +1,39 @@
 # DINOv2 (MLX)
 
 Channel-last MLX port of the [DINOv2](https://github.com/facebookresearch/dinov2)
-vision transformer. This package currently holds only the shared backbone and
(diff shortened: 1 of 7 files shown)


Lazarus-931 (approved):

Lazarus-931:
thanks @eklipse2k8!

PR #2186, opened 2026-09-07
fix(qwen3_5): transpose Qwen3.8 Conv3d patch embed to NDHWC

Taderich73:
## Summary

Qwen3.8-27B (which declares `model_type: qwen3_5` / `Qwen3_5ForConditionalGeneration`) currently cannot be loaded or converted. It ships `visual.patch_embed.proj.weight` in PyTorch NCDHW order while MLX convolutions expect NDHWC, and `sanitize()` — which already rewrites `conv1d` weights — leaves this tensor untouched. `load_model` uses `strict=True`, so the load aborts:

```
ValueError: Expected shape (1152, 2, 16, 16, 3) but received shape (1152, 3, 2, 16, 16)
             for parameter patch_embed.proj.weight
```

This is the *only* blocker. Diffing the instantiated model's parameter tree against all 1199 checkpoint tensors post-`sanitize`:

```
model params 1184 | ckpt 1199 | after sanitize 1184
MISSING 0 | EXTRA 0 | SHAPE MISMATCH 1   <- patch_embed.proj.weight
```

Every linear-attention, full-attention, MLP, embedding and vision-block tensor already binds correctly. (The 15 unbound checkpoint tensors are the `mtp.*` draft head, dropped deliberately at `qwen3_5.py` sanitize.)

## Change

One transpose in `Model.sanitize`, alongside the existing `conv1d` case. The guard compares the trailing axis against `vision_config.in_channels`, so re-sanitizing weights already in NDHWC is a no-op rather than a second transpose.

## Verification

With the fix, the same structural diff reports `MISSING 0 | EXTRA 0 | SHAPE MISMATCH 0`, and Qwen3.8-27B converts and runs. Image-conditioned generation from the converted model returns an accurate description of the input, so the axis order is right and not merely shape-compatible.

Three regression tests added in `mlx_vlm/tests/test_qwen3_5_patch_embed.py`, built on a tiny synthetic config (no checkpoint download). Two fail without the fix; the idempotence test passes either way by design, guarding against a future over-eager transpose.

## Test plan

- [x] `pytest mlx_vlm/tests/test_qwen3_5_patch_embed.py` — 3 passed
- [x] Same file against unpatched `sanitize` — 2 failed, confirming the tests bite
- [x] `pytest mlx_vlm/tests/ -k qwen3_5` — 72 passed, no regressions
- [x] `pre-commit run --files ...` — black, isort, autoflake all pass
- [x] End-to-end: converted Qwen3.8-27B at mixed precision and generated from an image

### DIFF
--- a/mlx_vlm/models/qwen3_5/qwen3_5.py
+++ b/mlx_vlm/models/qwen3_5/qwen3_5.py
@@ -159,6 +159,13 @@ class Model(Qwen3VLModel):
 
             if "conv1d.weight" in key and value.shape[-1] != 1:
                 value = value.moveaxis(2, 1)
+            # Qwen3.8 ships the Conv3d patch embed in PyTorch NCDHW order,
+            # while MLX convolutions expect NDHWC. Guard on the trailing axis
+            # so re-sanitizing already-converted weights is a no-op.
+            if key.endswith("patch_embed.proj.weight") and value.ndim == 5:
+                in_ch = self.config.vision_config.in_channels
+                if value.shape[-1] != in_ch and value.shape[1] == in_ch:
+                    value = value.transpose(0, 2, 3, 4, 1)
             if any(key.endswith(sfx) for sfx in NORM_WEIGHT_SUFFIXES):
                 if value.ndim == 1 and should_offset_norm_weight(
                     original_key, shift_norm_weights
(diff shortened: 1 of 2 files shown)


Lazarus-931 (approved):
thanks for the catch @Taderich73, lgtm :)

PR #2189, opened 2026-09-08
Return OpenAI segments for NeMo-alignment STT models (Parakeet/Canary)

### LINKED ISSUE #2183: Parakeet (and other NeMo-alignment STT models) never return "segments" via /v1/audio/transcriptions - AlignedResult exposes .sentences, not .segments
## Description
The OpenAI-compatible transcription route in `mlx_vlm/server/audio.py` only forwards a handful of attributes from the STT model's result object into the HTTP response:

```python
for attr in ("text", "language", "segments"):
    if hasattr(item, attr):
        data[attr] = getattr(item, attr)
```

Parakeet's `generate()` (`mlx_audio/stt/models/parakeet/parakeet.py`) returns:

```python
result = sentences_to_result(tokens_to_sentences(all_tokens))
```

and `sentences_to_result` (`mlx_audio/stt/models/nemo/alignment.py`) returns an `AlignedResult`:

```python
def sentences_to_result(sentences: list[AlignedSentence]) -> AlignedResult:
    return AlignedResult("".join(sentence.text for sentence in sentences), sentences)
```

`AlignedResult` exposes `.text` and `.sentences` - **not** `.segments`. So `hasattr(item, "segments")` is always `False` for Parakeet, and the response never includes timing information, regardless of `response_format=verbose_json`, `word_timestamps`, or `timestamp_granularities`. `word_timestamps` isn't even a recognized parameter of `Parakeet.generate()` (it falls into an unused `**kwargs`), so passing it has no effect either.

The base `STTOutput` dataclass (`mlx_audio/stt/models/base.py`) does define a `segments: List[dict]` field, which suggests other STT models (Whisper family) populate it correctly and get proper `segments` through this same route - Parakeet (and presumably Canary and anything else built on `sentences_to_result`/`AlignedResult`) is the one falling through this gap.

## Reproduction
Server running with `mlx-community/parakeet-tdt-0.6b-v3` loaded as the STT model:

```bash
curl -X POST "http://localhost:8080/v1/audio/transcriptions" \
  -F model=mlx-community/parakeet-tdt-0.6b-v3 \
  -F file=@audio.wav \
  -F response_format=verbose_json \
  -F word_timestamps=true
```

Response is always just:
```json
{"text":"..."}
```
No `segments`, no `language`. Tried with and without `word_timestamps=true`, with `timestamp_granularities[]=segment`/`word`, and with `verbose=true` - identical flat `{"text": ...}` response in every case.

## Suggested fix
In the STT response-building helper in `mlx_vlm/server/audio.py`, also check for `.sentences` and convert `AlignedSentence` entries (with their `.tokens[].start`/`.end`) into the same `segments` shape used for Whisper-family models, e.g.:

```python
if hasattr(item, "segments"):
    data["segments"] = item.segments
elif hasattr(item, "sentences"):
    data["segments"] = [
        {"text": s.text, "start": s.tokens[0].start, "end": s.tokens[-1].end}
        for s in item.sentences
    ]
```

## Environment
- mlx-vlm server (bundled via Blaizzy/nativ), `/v1/audio/transcriptions` endpoint
- Model: `mlx-community/parakeet-tdt-0.6b-v3`
- Confirmed via direct `curl` against the server, independent of any client library

Lazarus-931:
Fixes #2183 — NeMo-alignment STT models (Parakeet, Canary) never returned `segments`, so transcription timing was lost.

`_stt_item_to_dict` now maps `sentences` → OpenAI `segments` (`{id, start, end, text}`) when a result has `sentences` and no `segments`. This yields conformant `segments` that also flow through `srt`/`vtt`. Whisper-family results (already carry `segments`) and streaming results (no `sentences`) are untouched.

- before: `{"text": "hello world"}`
- after: `... "segments": [{"id": 0, "start": 0.0, "end": 0.9, "text": "hello world"}]`

### DIFF
--- a/mlx_vlm/server/audio.py
+++ b/mlx_vlm/server/audio.py
@@ -615,6 +615,36 @@ def _iter_stt_items(result):
 
 
 def _stt_item_to_dict(item) -> Dict[str, Any]:
+    data = _stt_item_base_dict(item)
+    # NeMo-alignment STT models (Parakeet, Canary) report timing as ``sentences``
+    # rather than ``segments``; map it across so it survives the ``segments``-keyed
+    # reassembly in ``_transcription_result_from_chunks``.
+    if isinstance(data, dict) and not data.get("segments") and data.get("sentences"):
+        data["segments"] = _sentences_to_segments(data["sentences"])
+    return data
+
+
+def _sentences_to_segments(sentences: Any) -> List[Dict[str, Any]]:
+    segments: List[Dict[str, Any]] = []
+    for index, sentence in enumerate(sentences):
+        if not isinstance(sentence, dict):
+            continue
+        start, end = sentence.get("start"), sentence.get("end")
+        if start is None or end is None:
+            continue
+        text = sentence.get("text")
+        segments.append(
+            {
+                "id": index,
+                "start": start,
+                "end": end,
+                "text": text.strip() if isinstance(text, str) else text,
+            }
+        )
+    return segments
+
+
+def _stt_item_base_dict(item) -> Dict[str, Any]:
     if isinstance(item, bytes):
         return {"text": item.decode("utf-8", errors="replace")}
     if isinstance(item, str):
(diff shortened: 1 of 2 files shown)


lucasnewman (approved):
🙏

PR #2194, opened 2026-09-08
fix recurrent kernel dispatch for unsupported dimensions

Lazarus-931:
fall back to the reference recurrence when ssm or gated-delta dimensions cannot be handled by the metal kernels. this prevents truncated state updates or invalid kernel launches when the state or key dimension is not a multiple of 32.

### DIFF
--- a/mlx_vlm/models/gated_delta.py
+++ b/mlx_vlm/models/gated_delta.py
@@ -185,6 +185,8 @@ def gated_delta_kernel(
     mask: Optional[mx.array] = None,
 ) -> Tuple[mx.array, mx.array]:
     B, T, Hk, Dk = k.shape
+    if Dk < 32 or Dk % 32:
+        return gated_delta_ops(q, k, v, g, beta, state, mask)
     Hv, Dv = v.shape[2:]
     input_type = q.dtype
     state_type = state.dtype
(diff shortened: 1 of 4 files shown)


mlx_vlm/models/gated_delta.py
```diff
@@ -289,12 +291,6 @@ def gated_delta_update(
...
         state = mx.zeros((B, Hv, Dv, Dk), dtype=mx.float32)
 
-    if (
-        not use_kernel
-        or mx.default_device() != mx.gpu
-        or not mx.metal.is_available()
-        or k.shape[-1] < 32
-        or k.shape[-1] % 32 != 0
-    ):
+    if not use_kernel or mx.default_device() != mx.gpu or not mx.metal.is_available():
```
lucasnewman, line 294: @Lazarus-931 What's the thinking around removing the shape check here? Isn't it still valid?

Blaizzy:
> fall back to the reference recurrence when ssm or gated-delta dimensions cannot be handled by the metal kernels. this prevents truncated state updates or invalid kernel launches when the state or key dimension is not a multiple of 32.

Could you share a model that is faces this issue?

mlx_vlm/models/gated_delta.py
Lazarus-931: hm, I gated it in line 188, so removing it here was cleaner, can put back though!

Lazarus-931:
> > fall back to the reference recurrence when ssm or gated-delta dimensions cannot be handled by the metal kernels. this prevents truncated state updates or invalid kernel launches when the state or key dimension is not a multiple of 32.
> 
> Could you share a model that is faces this issue?

this was more of a safety precaution for dims  <32 or not ÷32, in retrospect i think this can be addressed if any do exist. 

closing

PR #2195, opened 2026-09-08
honor cache limits across single and batched generation

Lazarus-931:
replaces #1995 by making each cache type explicitly support or reject max_kv_size and applying the limit to both single and batched generation. preserves fixed prefixes and prevents reuse of caches built with a different limit.


Blaizzy:
What's the use case?

Lazarus-931:
> What's the use case?

well, when `--max-kv-size` is passed, i'm unable to pass it for 97 of 126 models. this was a better solution based of your feedback from #1995

Blaizzy:
Closing it in favour of an issue with reproducible examples and expected outcomes.

PR #2196, opened 2026-09-08
honor gemma 3 rope_theta and report unused config keys

Lazarus-931:
honor gemma 3’s rope_theta config while preserving legacy values and rejecting conflicting aliases. warn when shared config parsing drops unknown keys, with a dedicated warning category that callers can filter or treat as an error.

### DIFF
--- a/mlx_vlm/models/base.py
+++ b/mlx_vlm/models/base.py
@@ -1,5 +1,6 @@
 import inspect
 import math
+import warnings
 from abc import abstractmethod
 from dataclasses import dataclass
 from typing import Dict, List, Optional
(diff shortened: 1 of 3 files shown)


lucasnewman:
@Lazarus-931 Is there a repo config that has conflicting values here? What motivated this change?

Lazarus-931:
> @Lazarus-931 Is there a repo config that has conflicting values here? What motivated this change?

i was testing out EssentialAI/rnj-1, and the rope base of 1,000,000(for gemma 3) differs from rnj-1's 10,000 from it's config

Blaizzy:
Closing it in favour of an issue with reproducible examples.

Issue #1962, opened 2026-08-19
Per-row tail zeroing in speculative rollback leaves phantom KV attended


pierre427:
## Summary

When a batched speculative round produces different accept counts per row,
rollback trims the cache to `max(accepted)` and zeroes the rejected tail of the
shorter rows. Zeroing does not remove those positions from attention: a zero key
scores `q · 0 = 0` and keeps softmax weight `e^0 / Z`. Its value contributes
nothing to the numerator but still enlarges the denominator, so the attention
output of every affected row is scaled down.

Rows that accepted the maximum count are unaffected. Rows that accepted fewer
are silently wrong.

## Affected models

Five models roll back ragged accepts this way:

- `models/gemma4/language.py`
- `models/deepseek_v4/language.py`
- `models/muse_glimmer/language.py`
- `models/qwen3_5/language.py`
- `models/minimax_m3_vl/language.py`

`models/inkling/language.py` is the exception: its drafter declares
`requires_uniform_batch_acceptance = True`
(`speculative/drafters/inkling_mtp/inkling_mtp.py:93`), so it never takes the
ragged path.

This overlaps #1432, which added the `zero_row_tail` hook so TurboQuant caches
could zero per-row tails like dense caches do. That PR treats dense-cache
zeroing as correct. The reproduction below indicates it is not — the problem is
not how the tail is zeroed, but that zeroing is used at all where removal is
required.

## Reproduction

The reproduction script (inlined at the end) drives `BatchKVCache` and
`BatchRotatingKVCache` — the classes `generate/ar.py:800-835` wraps model caches
in for batched serving — through each model's `rollback_speculative_cache` with
a ragged accept array, then compares every row's next decode step against a
reference cache that only ever received that row's accepted tokens.

B=2, prefix 12, block 5, `accepted = [3, 1]`:

```
muse_glimmer / BatchKVCache          row 0 (accepted 3): 0.00%   row 1 (accepted 1): 11.84%
muse_glimmer / BatchRotatingKVCache  row 0: 0.00%                row 1: 11.84%
gemma4       / BatchKVCache          row 0: 0.00%                row 1: 11.84%
deepseek_v4  / BatchKVCache          row 0: 0.00%                row 1: 11.84%
```

(mean relative error on the row's attention output; max |diff| 0.040938)

A second check isolates the mechanism in a single SDPA call: the relative error
equals the softmax mass on the zeroed slots exactly.

Both tests use small shapes so the effect is visible. With
[...]

PR #2197, opened 2026-09-09
Fix qwen3_5 ragged speculative rollback on BatchQSAKVCache

gaineyllc:
## What

Fix `Qwen3_5` batched speculative rollback so a **ragged per-row accepted count** works on `BatchQSAKVCache`.

In `LanguageModel.rollback_speculative_cache`, the batched right-padding trim is gated on:

```python
if c.keys is not None and callable(prepare) and callable(finalize):
```

`BatchQSAKVCache` keeps its KV inside a wrapped `BatchKVCache` and exposes **no `keys` attribute**, so `c.keys` raises `AttributeError` *before* this branch can run. As a result, a ragged rollback (rows accepting different numbers of drafts) over a `BatchQSAKVCache` crashes — even though that cache implements a **correct** ragged trim via its own `prepare(right_padding=...)` / `finalize()` (which uses `dynamic_roll`, not the KV-zeroing fallback that #1962 warns leaves phantom keys attended).

## Fix

Reorder the guard to check `prepare`/`finalize` first and read `keys` via `getattr`:

```python
if callable(prepare) and callable(finalize) and getattr(c, "keys", True) is not None:
```

- `BatchQSAKVCache` (no `keys` attr) → `getattr` returns the default → takes its own correct `prepare`/`finalize` ragged-trim path, and sets `right_trimmed=True` so it does **not** reach the `#1962` uniform-acceptance `RuntimeError` below.
- Caches with a real `keys` tensor still gate on it (the `is not None` check is preserved).

Only the `is_batch` ragged path is affected; uniform-acceptance and single-row rollback are unchanged.

## Why it matters

This unblocks **fused / batched multi-token speculative decode** where concurrent sequences accept different draft lengths (e.g. native MTP on Qwen3.8-Flash-Next under continuous batching). Without it, any ragged batched rollback on the continuous-batch cache aborts.

## Testing

- Verified live end-to-end: 6–16 concurrent streams, temperature 0, coherent output on every row, no cache corruption across 200-token generations (the `dynamic_roll` trim does not exhibit the phantom-key behavior of the zero-tail fallback).
- `python -m py_compile mlx_vlm/models/qwen3_5/language.py` passes.

15 lines changed (14 insertions, 1 deletion), scoped to the one guard.

### DIFF
--- a/mlx_vlm/models/qwen3_5/language.py
+++ b/mlx_vlm/models/qwen3_5/language.py
@@ -1497,7 +1497,20 @@ class LanguageModel(nn.Module):
                 if any(extra_trim_list):
                     prepare = getattr(c, "prepare", None)
                     finalize = getattr(c, "finalize", None)
-                    if c.keys is not None and callable(prepare) and callable(finalize):
+                    # BatchQSAKVCache keeps its KV inside a wrapped BatchKVCache and
+                    # exposes no ``keys`` attribute, so a direct ``c.keys`` here raises
+                    # AttributeError before this branch can run -- even though its
+                    # ``prepare``/``finalize`` implement a correct ragged right-padding
+                    # trim (dynamic_roll, not a zero-fill, so it is not affected by the
+                    # phantom-key issue #1962 that the KV-zeroing fallback below guards
+                    # against). Check prepare/finalize first and read ``keys`` via
+                    # getattr so caches with real ``keys`` still gate on it (None-check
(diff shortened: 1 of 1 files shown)


lucasnewman:
@gaineyllc Thanks! Can you [sign](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) your commits and force push so we can clear the merge requirements?

PR #2202, opened 2026-09-09
Add video understanding support for Mage-VL

lucasnewman:
Fill in the video understanding support for Mage-VL, which was missing in the original implementation. The model uses timestamps to identify where the frames come from, so we need to forward video metadata to the model, hence the `utils.py` changes. Supersedes https://github.com/Blaizzy/mlx-vlm/pull/2069

Tested against real model weights and verified working:

```
“A silver car collides with a black car.”
“A group of four men are discussing a soccer match.”
```


lucasnewman:
Resolves https://github.com/Blaizzy/mlx-vlm/issues/1766

Lazarus-931 (approved):
nice!

PR #2206, opened 2026-09-09
Rebuild speculative decoding around request caches and shared prefill

Blaizzy:
Speculative decoding currently spreads request state and rollback across model-specific loops and verifiers. This PR rebuilds it around a model-independent request-cache contract and the ordinary prefill pipeline. The decoding loop, cache transactions, sampling, and shared operators are designed for reuse across models. GLM-5.3-Flash FP8 and Qwen3.5 dense/MoE provide the initial native MTP adapters and validation coverage.

The target streams prompt features into MTP during ordinary prefill. During decode, MTP proposes tokens, the target verifies a block, and the cache commits the accepted prefix plus a target correction or bonus. Rejected cache entries are discarded without replaying the target model. Single requests and ragged batches follow this same loop.

- **Models implement one common contract.** Each adapter supplies the required target features and a stateless draft forward. Its cache types participate in the shared transaction contract. Adding a model does not require a separate generation loop, prefix manager, or model-specific rollback path.
- **Request state belongs to the cache.** `SpeculativeCache` owns target and draft caches, positions, token history, the live MTP prediction/hidden-state seed, and per-request acceptance counters. Native cache transactions handle commit and rollback. Drafter models own weights and expose stateless forwards. Cancellation commits only tokens already delivered to the caller.
- **APC stores target caches only.** AR and MTP share normal prefix keys, `APCCoordinator`, and `PromptProcessingBatch`, including mixed warm/cold rows and chunked prefill. Prefix hits create a fresh MTP cache from the uncached suffix; the last suffix token's target feature seeds drafting. No cached prefix is replayed, and APC holds no draft cache, seed, hidden features, or rollback state. Target prefixes are reusable across AR and different drafters. APC implementation and schema match main. Conversation cache reuse also retains only target state and does not implicitly create an APC manager.
- **Generation preserves ordinary target behavior.** Greedy and sampled decoding support logit bias and repetition/presence/frequency penalties. `logprobs=True` returns verified target logprobs for every emitted token, including corrections and bonuses; the server supports requested top logprobs. Thinking budgets constrain the target distribution at sampling time, and rejected drafts cannot advance the reasoning budget.
- **Optimizations use shared operators.** Positioned target sampling is batched when processors allow it. Short dense and embedding projections use the existing GEMV kernel; MXFP8 MoE gate/up projections are fused and reuse expert weights across adjacent tokens. Dispatch depends on shape and quantization format, with native fallbacks. No model-specific Metal kernel is introduced. Qwen prefill preserves padding metadata, and GLM sparse attention excludes stale validity flags in compacted padding.

Normal prefix lookup and prefill remain separate from speculative decode, as in [vLLM's KV cache coordinator](https://github.com/vllm-project/vllm/blob/dc36fcce902a63eab06c1b93a5c4a5ee178a0c56/vllm/v1/core/kv_cache_manager.py#L265) and [SGLang's target/draft prefill path](https://github.com/sgl-project/sglang/blob/2f5cc8e33e9717f8284ed48ffc7f012e7b4a4197/python/sglang/srt/speculative/eagle_worker_v2.py#L1263). The cache policy differs: vLLM coordinates cache groups and reserves a matched tail for recomputation; this implementation reuses target state and starts fresh draft context from the suffix. Reduced draft context can change proposals and acceptance rates. Target verification and sampling still govern every emitted token.

This removes the legacy DFlash, EAGLE, DSpark, other MTP adapters, model-specific speculative verifiers, and associated extraction paths and tests. Native quantized KV remains supported independently of model-weight quantization. Rotating/TurboQuant caches are unsupported. Processors requiring an external update after each token use one-token target steps. Qwen batch-two MTP remains slower than AR in the measured small-model workload.

Validation:

- Full suite: **3,594 passed, 9 skipped, and 128 subtests passed**. Coverage includes AR/MTP prefix interoperability, one-token suffix initialization without prefix replay, mixed warm/cold batches, chunks of 1/2/32, target-only disk serialization, tenant isolation, transaction rollback, and cancellation with or without APC.
- Real GLM FP8: **twelve comparisons of 64 output tokens match AR tokens and selected target logprobs exactly**, at temperatures 0/0.8. Cases include cold requests, mixed warm/cold batches in both row orders, and shared prefixes created by AR and reused by MTP. Warm drafting starts with a fresh cache from a one-token suffix.
- Additional real-model checks cover FP8 logprobs and thinking budgets, plus twelve Qwen3.5-0.8B comparisons across batch sizes, temperatures, and draft-block sizes. Formatting, undefined-name checks, and `git diff --check` pass.

[Architecture, usage, validation settings, and performance results](https://github.com/Blaizzy/mlx-vlm/blob/pc/glm53-flash-mtp-rebuild/docs/speculative-decoding.md) are documented separately. The recorded throughput measurements predate the final target-only APC design; they are not new timings of the final implementation.


raullenchai:
I qualified this branch against the Q4 GLM-5.3 configuration used by Rapid-MLX, since the PR's published GLM measurements currently cover FP8.

Environment and composition:

- M3 Ultra, 60 GPU cores, 256 GB; macOS 26.5.2
- MLX 0.32.2; single stream; temperature 0; thinking enabled
- target: `Vontra/GLM-5.3-Flash-MLX-4bit-MTP` at `76add2a`
- target-matched 3.9 GB affine-Q4 MTP sidecar; block-total 2
- this PR plus #2231's strict quantized-lm-head load fix and #2234's gate/up storage fusion
- six-task product gate: executable coding, atomic knowledge, multi-step math, exact instruction following, constrained creative writing, and a 3,126-token contract retrieval prompt

Results:

- Two consecutive MTP runs passed 12/12 tasks and produced byte-identical reasoning and final responses between runs.
- A same-branch AR control passed 6/6. MTP matched every AR reasoning/final byte.
- Paired MTP/AR completion-throughput ratios were `1.253x`, `1.306x`, `1.188x`, `1.297x`, `1.179x`, and `1.082x`; median `1.221x`.
- The two-run MTP per-task medians were `36.674`, `35.025`, `32.781`, `33.815`, `33.494`, and `11.534 tok/s`; median of category medians `33.655 tok/s`.
- Against the previously qualified positioned-MTP + #2234 runtime, the first run moved median client throughput from `31.390` to `33.821 tok/s` (`1.077x`) and median server decode from `35.461` to `38.564 tok/s` (`1.087x`).
- This also clears the same-width oMLX control (`31.813 tok/s`) by 6.3%, while retaining equivalence to its own AR control.

The short-task text differs from the older positioned-MTP artifacts because this branch closes the thinking budget through the target distribution. The same-branch AR equality confirms that the difference is budget-policy semantics, not speculative drift.

Focused GLM/MTP tests on the combined tree passed 37/37. The production source changes from #2234 merged cleanly; only its legacy speculative-test hunk conflicted because this PR intentionally removed that test surface.

raullenchai:
I resolved this PR's current `origin/main` conflicts on a standalone fork branch so the maintainer branch can reuse or compare the exact resolution:

- merge commit: https://github.com/raullenchai/mlx-vlm/commit/c63d88f550b19d86c015aa124d9e5c37fa6a6d97
- branch: `raullenchai:experiment/glm53-cache-owned-rebase`
- parents: this PR head `b9e6897c` + current main `45d6e125`

Resolution details:

1. `speculative/ops/linear.py -> models/quantized_ops.py`: kept this PR's model-owned relocation, then wired main's optimized MXFP4 linear/argmax/fused-linear dispatch ahead of the affine/NVFP4 fallbacks.
2. `tests/test_speculative.py`: kept this PR's intentional retirement of the legacy monolithic test surface, and moved main's three new MXFP4/Q3 exactness tests into `tests/test_quantized_ops.py` instead of restoring the deleted 5K-line file.
3. Fixed the existing ambiguous one-letter test variable caught by current Ruff.

Validation on M3 Ultra / MLX 0.32.2:

- Ruff format/check: pass
- `test_quantized_ops.py`, `test_speculative.py`, `test_speculative_transactions.py`, `test_speculative_serving.py`, `test_speculative_reporting.py`: **551 passed**

I did not open a duplicate PR or modify the maintainer branch. This is only a ready conflict-resolution reference for #2206.

raullenchai:
Follow-up broad validation on the same merge commit: **3,556 passed, 7 skipped, 149 subtests passed** in 78.48s. The local environment lacks the current optional `mlx_audio.codec.models.nemotron_voicechat` and Inkling dependency surfaces, so I excluded only `test_nemotron_voicechat*`, `test_server_realtime.py`, and `test_inkling.py`; these collection failures reproduce from files present in both parents and are unrelated to the merge resolution.

raullenchai:
Post-#2234 follow-up, using the conflict-resolution branch plus #2231 but **without** gate/up fusion:

- Same Q4 target/sidecar and six-task gate: fixed block-total 2 remained AR-exact, passed 6/6, and moved the category median from `26.490` to `32.887 tok/s`; paired median `1.214x`.
- Fixed block-total 3 also stayed exact and passed 6/6. It reached `34.694 tok/s`, but creative writing regressed 4.5% versus block-total 2 while knowledge/instruction gained 8.5%/7.3%.
- A request-local rolling-acceptance spike kept two drafts for coding/knowledge/math/instruction/long-context and fell back to one draft only for creative. Two identical-policy runs passed 12/12 with byte-identical reasoning/final output and identical depth/acceptance counts. Category medians were `34.552` and `34.610 tok/s`; every per-task throughput repeated within 0.23%.
- Versus same-branch AR, the adaptive run delivered a `1.298x` paired median gain. Its `34.552 tok/s` category median is 8.6% above the previously measured same-width oMLX control (`31.813 tok/s`).

Spike branch and reproducible note: https://github.com/raullenchai/mlx-vlm/tree/experiment/glm53-cache-owned-adaptive-k23

I am deliberately **not** opening this as a generic controller PR yet: the 65%/64-round crossover is qualified only for this GLM Q4 target, it currently gates singleton greedy decode, and a production controller should learn per-depth cost rather than ship a model-specific magic number. The useful upstream signal is that #2206 makes block-total 3 materially profitable on five of six real workloads, and the low-acceptance exception can be detected online without changing output.

raullenchai:
I found a small, independent pipeline win on top of the cache-owned MTP + strict quantized `lm_head` composition: launch the replay-head seed with `mx.async_eval` before yielding the verified block's final token. This overlaps replay with caller-side detokenization / stop checks / stream framing, without adding a synchronization or changing selection.

Isolated branch: https://github.com/raullenchai/mlx-vlm/tree/perf/glm53-async-mtp-replay  
Code commit: https://github.com/raullenchai/mlx-vlm/commit/146954bf

Studio M3 Ultra, cached Vontra GLM-5.3-Flash uniform-Q4 target + matched Q4 MTP sidecar, fixed block-total 3, temperature 0, six graded real tasks:

| | coding | knowledge | math | instruction | creative | long context |
|---|---:|---:|---:|---:|---:|---:|
| existing path | 37.14 | 37.25 | 33.51 | 35.88 | 31.74 | 11.95 |
| async replay r1 | 37.50 | 37.97 | 33.46 | 36.43 | 32.14 | 12.02 |
| async replay r2 | 37.44 | 37.68 | 33.99 | 35.90 | 31.78 | 11.96 |

Paired median: **1.011x** and **1.005x**. Both runs passed 6/6; complete reasoning and final outputs were byte-identical to control and to each other. Relevant speculative + server tests: **405 passed**; Ruff and diff check clean. Full methodology and caveat are committed in `docs/benchmarks/glm53_async_mtp_replay.md`.

This is intentionally isolated from the adaptive-depth experiment so the low-risk scheduling change can be reviewed/cherry-picked on its own once #2206's conflict is resolved.

raullenchai:
Follow-up: this is now a clean stacked PR against #2206's own head: #2241. The final review also avoids launching replay after a terminal block and adds a regression test for that case. Exact head `b762dba55ba939cff71536647dc16d0abc7f8031`; 406 relevant tests pass.

raullenchai:
Additional real-server qualification of this PR's request-owned speculative checkpoint path (M3 Ultra, cached Vontra GLM-5.3 uniform-Q4 target + matched Q4 MTP, memory APC on, disk APC off, two resident checkpoint entries):

- 3,324-token English contract prompt: cold/warm user-visible TTFT `12.397s -> 0.090s` (**137.9x**); total request `13.572s -> 1.192s` (**11.39x**); warm usage reported `3,323` cached tokens.
- 1,614-token variant: cold/warm user-visible TTFT `20.952s -> 1.481s`; total `21.143s -> 1.677s`; warm usage reported `1,613` cached tokens. Server-side warm prefill was `0.035s`.
- Complete streamed-output SHA-256 matched cold versus warm for both fixtures.
- Boundary: a 9,352-token variant did not restore under the same two-entry automatic memory budget (`cached_tokens=0`; repeated TTFT about `32.4s`). I am treating 3.3K as the qualified resident result, not claiming unbounded replay.

This is a materially larger user-facing win than the decode-only gain for repeated document/code conversations, and it exercises the paired target + MTP checkpoint rather than target KV alone.

raullenchai:
Concurrent-server follow-up on the same Q4 checkpoint:

- Four simultaneous warm-prefix MTP requests (256 tokens each) completed in `15.416s`: **66.43 aggregate tok/s** versus `40.04 tok/s` singleton (**1.66x aggregate scaling**). All four concurrent rows had the same complete reasoning/output hash.
- The batched MTP trajectory was not byte-identical to singleton at temperature 0. A no-drafter AR control reproduces the boundary: on a 128-token fixture it reached `57.95` aggregate tok/s versus `27.43` singleton, but dynamic admission produced two hashes in the batch (3 rows shared one, 1 row differed). This points to GLM backbone batch-size/dynamic-admission numerics, not speculative acceptance/rollback.
- Forcing all B>1/T=1 affine projections through `singleton_quantized_linear` did not restore equality and reduced AR aggregate throughput to `50.07 tok/s` (-13.6%), so I rejected that fallback.

I would treat cross-batch byte invariance as a separate P1 and avoid claiming it for this model today. The semantic fixture remained sound; singleton AR/MTP equivalence from the earlier six-task gate is unaffected.

raullenchai:
The conflict-resolution reference is now available as #2243, targeting this feature branch directly. Its signed head is `da45d928`; GitHub reports the branch comparison clean. Merging #2243 updates only `pc/glm53-flash-mtp-rebuild` onto current main, after which this original PR can proceed normally.

raullenchai:
Rapid’s immutable GLM-5.3 Q4 MTP sidecar is now public at `rapid-mlx/GLM-5.3-Flash-MTP-4bit@e9d62773d3e5272fb298830e8e06fadc4137ae2c`, and our release activation (exact target/sidecar pins, CLI/Desktop default, sampled-request AR fallback) is ready. We are preparing a Rapid release today; #2206 plus #2231 in a tagged mlx-vlm release is the remaining dependency gate. No pressure if that timing is not workable—we will keep the capability fail-closed rather than pin an untagged commit.

lyonsno:
At `d09666d2`, cancelling a batched MTP request after prefill but before its first output saves an APC checkpoint with the wrong prefix length.

`_completed_cache_checkpoint()` always uses `tokens[:-1]`. Before `_sent_first`, that list contains only the prompt: the target has cached all N prompt tokens, but the checkpoint is keyed by N−1. A subsequent hit reprocesses the last prompt token against a cache that already contains it.

Reproduced with the existing tiny Qwen3.5 fixture and two requests: step until prefill creates the generation batch, remove one UID before the next `next()`, then look up its prompt plus a continuation. With a 17-token prompt and the default APC minimum, lookup reports 16 cached tokens while the restored KV offset is 17. Warm versus fresh next-token logits differ by up to 0.0875. Cancelling after the first output instead gives 17/17 and agrees within 4.8e−7. The other request remains active. This timing is reachable because the server drains cancellations between generation steps.

Could this callback preserve the full prompt key before first output, while retaining the pending-token exclusion after delivery?

PR #2208, opened 2026-09-09
Extract tool-calling into a dedicated tools/ package

Lazarus-931:
## What

Give tool-calling a single home. Parser selection, extraction, `tool_choice`
policy, the neutral result type, **and the parser modules themselves** move into
one `mlx_vlm/tools/` package. The old top-level `mlx_vlm/tool_parsers/` directory
is removed (its modules relocate to `tools/parsers/`).

```
mlx_vlm/tools/
  registry.py   ParserSpec + SPECS (declarative) + selection/override
  extractor.py  process_tool_calls(...) -> ParseResult
  policy.py     tool_choice semantics (moved out of server/openai.py)
  types.py      frozen ParseResult dataclass
  base.py       ToolParser protocol (the parser-module contract)
  parsers/      the per-format parser modules (moved from tool_parsers/)
    qwen3_coder.py  json_tools.py  glm47.py  mistral.py  atem.py  ...
```

## Why this shape

- **Routing is declarative, not position-dependent.** Each format is a
  `ParserSpec(name, template_markers, priority)`. `json_tools` is the generic
  `<tool_call>` JSON fallback at `priority=0`, so format-specific parsers outrank
  it by **priority rather than list order** — specs can be reordered/inserted
  without changing routing.
- **Selection is diagnosable and overridable** (unmatched template logs instead
  of silently returning `None`; an explicit name bypasses inference).
- **`ParseResult` is a real frozen dataclass**, not a `dict` subclass; call sites
  use attribute access and the `responses_state` re-export shim is deleted.

## Refactor is behavior-preserving

Routing is **byte-identical to `main`** across every chat template in a local HF
cache — **152 checkpoints / 191 template samples / 0 disagreements**, identical
SHA256 — and matches on the live `transformers` path for 21 real tokenizers. The
server dependency-injection wiring is unchanged.

## Parser fixes (verified on real weights)

Each parser was exercised end-to-end on a real checkpoint (load → generate a tool
call → extract). Two real, pre-existing gaps were found and fixed (extraction
only — routing/SPECS unchanged):

- **mistral**: handled only the v11 `name[ARGS]{json}` shape and raised on the
  classic v3 JSON-array `[{"name":…,"arguments":…}]` that Ministral-8B / Mixtral /
  Mistral-7B-v0.3 emit. Now handles both. Verified on real Ministral-8B-2410.
- **glm47** (`#2207`, co-authored @Dev-next-gen): the tool name kept a trailing
  newline before the schema lookup, so declared string args were mistyped. Fixed
  with a strip. Verified on real GLM/Spark checkpoints.
- **Laguna** (raised in `#1802`): Poolside Laguna shares GLM's exact
  `<tool_call>/<arg_key>/<arg_value>` markers, so a dedicated marker-keyed parser
  can't distinguish it from GLM. With the glm47 fix above, Laguna parses correctly
  through `glm47` — verified on real `poolside/Laguna-XS-2.1` weights — so this
  adds a regression test rather than a redundant parser.
- **Multi-variant template routing**: tokenizers that expose several named
  templates present `chat_template` as a dict `{name: template}` (Cohere
  command-r, Mistral-7B-v0.3), which hit the `isinstance(str)` guard and got **no
  tool parser at runtime**. The registry now resolves the authoritative
  `tool_use` variant, restoring routing (verified: c4ai-command-r7b →
  cohere2_moe, Mistral-7B-v0.3 → mistral, both previously `None`). The string
  path is unchanged (routing stays byte-identical).

## New capability — parser override

A `tool_parser` request field (chat-completions, responses, Anthropic) forces a
specific parser, bypassing chat-template inference — the escape hatch for models
that mis-route or route to no parser. Validated against the known parser names
(unknown → 422); omitting it leaves inference unchanged.

## Cleanup

Removed the dead `function_gemma` parser — the buggy predecessor of `gemma4`
(its `call:(\w+)` name regex can't handle hyphenated/namespaced names, which
`gemma4` fixes) that routes on `<start_function_call>`, a marker no current model
emits (0 of 1137 cached checkpoints). Routing is unchanged (A/B: 0 disagreements).

## Tests

```bash
pytest mlx_vlm/tests/test_atem_tool_parser.py \
       mlx_vlm/tests/test_gemma4_tool_parser.py \
       mlx_vlm/tests/test_pythonic_tool_parser.py \
       mlx_vlm/tests/test_cohere_tool_parser.py \
       mlx_vlm/tests/test_glm47_tool_parser.py \
       mlx_vlm/tests/test_mistral_tool_parser.py \
       mlx_vlm/tests/test_responses_state.py
pytest mlx_vlm/tests/test_server.py \
       -k "tool_call or infer or minicpm5 or function_call or tool_use or process_tool"
```


mlx_vlm/tools/extractor.py
```diff
@@ -0,0 +1,66 @@
...
+import logging
+import re
+import uuid
+
+from .types import ParseResult
+
+logger = logging.getLogger("mlx_vlm.server")
+
+
+def process_tool_calls(model_output: str, tool_module, tools) -> ParseResult:
```
lucasnewman: Minor but it would be nice to have `tool_module` typed here so things like `.tool_call_start` et al are clearly defined as part of the API contract.

lucasnewman (approved):
LGTM!

mlx_vlm/tools/extractor.py
Lazarus-931: done!

PR #2209, opened 2026-09-10
Restore optimized MXFP4 and Q3 verification

YuhuaBillChen:
MXFP4 speculative verification currently falls back to singleton `gather_qmm`
after the verifier generalization removed the specialized kernels from #2106.
This restores the optimized MXFP4 path in the shared verifier dispatch and adds
affine Q3 support to the optimized affine kernels.

### Changes

- Restore the exact MXFP4 QMV, fused QMV, and masked/unmasked argmax kernels
  removed when the verifier was generalized.
- Dispatch supported MXFP4 projections through those kernels before the generic
  `gather_qmm` singleton fallback.
- Extend the generalized affine verifier to 3-bit weights, including exact
  fused projections and shifted-field masked/unmasked argmax.
- Preserve all generic fallbacks for unsupported shapes, devices, and formats.

### Correctness

- Specialized outputs are compared against the verifier's true B-by-T singleton
  reference, not a wide batched qmm reduction.
- Tests cover BF16 and FP16, MXFP4 batch size 2, T=2/4; Q3 B=1/2/4, T=3/4, affine Q3 group sizes
  32/64, fused projections, and masked/unmasked argmax.
- Real Qwen3.8-27B Q3 checkpoint: attention Q/K/V, Gated DeltaNet projections,
  and MLP gate/up at T=4 were byte-identical (`max_abs_diff=0`). LM-head
  projections and argmax tokens matched at T=1 and T=4.
- `mlx_vlm/tests/test_speculative.py`: **756 passed** on Apple M4 Pro with
  MLX/MLX-Metal 0.32.2.
- Pre-commit: Black, isort, and autoflake passed.

### Performance

Measured against `8f5dc3dd` on Apple M4 Pro with MLX/MLX-Metal 0.32.2:
Qwen3.8-27B target, MTP with draft block size 3, 1,024 tokens of prompt text
before chat templating, and 1,024 forced output tokens. Each arm uses a fresh
server; results are medians from baseline/patched/patched/baseline order.

| Target format | Current main | Patched | Speedup | Output / MTP stats |
|---|---:|---:|---:|---|
| MXFP4 | 13.022 tok/s | 24.609 tok/s | **1.890x** | identical |
| affine Q3 (updated tile) | 13.705 tok/s | 24.186 tok/s | **1.765x** | identical |

MXFP4 T=4 real-checkpoint microbenchmarks were also restored to the earlier
specialized-kernel range: fused Q/K/V 0.469 ms and LM-head argmax 5.119 ms, with
exact outputs.

### Implementation notes

- The MXFP4 path is limited to Metal GPU, BF16/FP16, 4-bit group-size 32,
  bias-free quantization parameters, aligned dimensions, and verifier length at
  most 4. Everything else retains the current generic fallback.
- Q3 projection kernels keep unshifted packed-field arithmetic for byte-identical
  output. Argmax uses shifted extraction, which changes floating-point operation
  order; token IDs matched in the tests and benchmark runs above.

Performance results are specific to these checkpoints and this workload.

### Expanded quantization and batch sweep (September 12)

The broader sweep exposed a Q3 projection regression at T=3/4, which this update fixes by reducing the output tile from four to two results per SIMD group. The projection retains its unshifted arithmetic and singleton reduction order.

I tested affine 3/4/5/8 (group size 32), MXFP4/MXFP8 (group size 32), and NVFP4 (group size 16), with B=1/2/4 and T=1/2/3/4 on an Apple M4 Pro, MLX/MLX-Metal 0.32.2. Each format uses fresh-process ABBA order against base `8f5dc3dd`, five warmups and 30 synchronized wall-clock samples per case. Ratios below are base latency / patched latency; higher is faster.

These are **synthetic-weight component microbenchmarks**, using Qwen3.8-27B dimensions: hidden size 5120, gated Q output 12288, K/V output 1024 each, and full vocabulary head output 248320. They are not seven full-model throughput measurements.

All 252 paired cases passed the checks in all four arms: projection/QKV outputs against native singleton calls, and argmax IDs against the singleton verifier. The speculative test suite also passed (756 tests), including FP16/BF16 and Q3 group sizes 32/64.

| Format | Geomean speedup across 36 cases | Min–max |
|---|---:|---:|
| affine3 | 1.401× | 0.811–1.819× |
| affine4 | 0.998× | 0.977–1.063× |
| affine5 | 0.997× | 0.967–1.010× |
| affine8 | 1.001× | 0.957–1.027× |
| mxfp4 | 1.568× | 0.970–2.190× |
| mxfp8 | 0.998× | 0.967–1.012× |
| nvfp4 | 0.999× | 0.967–1.027× |

T=4 detail:

| Format | Operation | B1 | B2 | B4 |
|---|---|---:|---:|---:|
| affine3 | q_projection | 1.577× | 1.819× | 1.731× |
| affine3 | qkv | 1.643× | 1.796× | 1.776× |
| affine3 | head_argmax | 1.558× | 1.571× | 1.581× |
| mxfp4 | q_projection | 1.741× | 1.967× | 1.956× |
| mxfp4 | qkv | 1.751× | 1.863× | 1.886× |
| mxfp4 | head_argmax | 2.143× | 2.172× | 2.190× |

The unchanged formats stay close to parity overall; individual cases vary by up to about 4% slower. Q3 is not faster at every shape: its T=1 head path remains slower in this sweep. The gains above should not be read as universal speedups across devices or workloads.

For the Q3 regression investigation, I varied only output tile size and loop unrolling while preserving the arithmetic. At B2/T4, K=5120/N=6144, the four-result kernel took 0.743 ms; the two-result variant took 0.424 ms, versus 0.594 ms for singleton gather. All matched exactly. This isolates a kernel tiling/code-generation sensitivity; without GPU register counters, I cannot attribute it specifically to spills. The larger, actual gated-Q shape in the table above independently confirms the fix.

Real Qwen3.8-27B affine-Q3 MTP confirmation (draft block 3, 1024 prompt-text tokens before chat templating, 1024 forced output tokens, fresh-server ABBA): base **13.705 tok/s**, patched **24.186 tok/s**, **1.765×**. Output hashes identical: **True**.
MTP rounds, draft-token count and accepted-token count identical across all four arms: **True**.

### DIFF
--- a/mlx_vlm/models/mxfp4_verifier.py
+++ b/mlx_vlm/models/mxfp4_verifier.py
@@ -1,0 +1,440 @@
+"""Optimized exact MXFP4 operations for speculative verification."""
+
+from functools import lru_cache
+from typing import Optional
+
+import mlx.core as mx
+import mlx.nn as nn
+
+# MLX's wide MXFP4 matmul uses a different reduction order from singleton
+# ``QuantizedLinear`` calls.  Keep that singleton QMV order for each output row
+# while reusing the loaded inputs across the short speculative time dimension.
+_TARGET_VERIFY_MXFP4_HEADER = r"""
+    using namespace metal;
+
+    constant constexpr int SIMD_SIZE = 32;
+    constant constexpr int VERIFY_RESULTS_PER_SIMDGROUP = 4;
+    constant constexpr int VERIFY_NUM_SIMDGROUPS = 2;
+    constant constexpr int VERIFY_ROWS_PER_THREADGROUP =
+        VERIFY_RESULTS_PER_SIMDGROUP * VERIFY_NUM_SIMDGROUPS;
+    constant constexpr int MXFP4_GROUP_SIZE = 32;
+    constant constexpr int MXFP4_PACK_FACTOR = 8;
+    constant constexpr int MXFP4_BYTES_PER_PACK = 4;
+    constant constexpr int MXFP4_PACKS_PER_THREAD = 2;
+    constant constexpr int MXFP4_VALUES_PER_THREAD =
+        MXFP4_PACK_FACTOR * MXFP4_PACKS_PER_THREAD;
+    constant constexpr int MXFP4_BLOCK_SIZE =
+        MXFP4_VALUES_PER_THREAD * SIMD_SIZE;
+    constant constexpr int MXFP4_SCALE_STEP_PER_THREAD =
+        MXFP4_GROUP_SIZE / MXFP4_VALUES_PER_THREAD;
+
+    inline float decode_mxfp4_value(uint code) {
+      half converted = as_type<half>(ushort((code & 7) << 9));
+      converted *= 16384.0;
+      float value = float(converted);
+      return code & 8 ? -value : value;
+    }
+
+    inline float decode_mxfp4_scale(uint8_t encoded) {
+      uint bits = encoded == 0 ? 0x00400000u : uint(encoded) << 23;
+      return as_type<float>(bits);
+    }
+
+    template <typename T>
+    inline void load_mxfp4_vector_exact(
+        const device T* x,
+        thread float* x_thread) {
+      for (int i = 0; i < MXFP4_VALUES_PER_THREAD; ++i) {
+        x_thread[i] = float(x[i]);
+      }
+    }
+
+    inline float mxfp4_qdot_exact(
+        const device uint8_t* w,
+        const thread float* x_thread,
+        float scale) {
+      float accum = 0.0f;
+      const device uint16_t* ws = (const device uint16_t*)w;
+      for (int i = 0; i < (MXFP4_VALUES_PER_THREAD / 4); ++i) {
+        uint packed = ws[i];
+        accum +=
+            (x_thread[4 * i] * decode_mxfp4_value(packed) +
+             x_thread[4 * i + 1] * decode_mxfp4_value(packed >> 4) +
+             x_thread[4 * i + 2] * decode_mxfp4_value(packed >> 8) +
+             x_thread[4 * i + 3] * decode_mxfp4_value(packed >> 12));
+      }
+      return scale * accum;
+    }
+"""
+
+
+_TARGET_VERIFY_MXFP4_QMV_SOURCE = r"""
+    uint n_tile = threadgroup_position_in_grid.y;
+    uint b_idx = threadgroup_position_in_grid.z;
+    uint simd_gid = simdgroup_index_in_threadgroup;
+    uint simd_lid = thread_index_in_simdgroup;
+
+    int out_row =
+        int(n_tile) * VERIFY_ROWS_PER_THREADGROUP +
+        int(simd_gid) * VERIFY_RESULTS_PER_SIMDGROUP;
+    int in_vec_size_w =
+        K_SIZE * MXFP4_BYTES_PER_PACK / MXFP4_PACK_FACTOR;
+    int in_vec_size_g = K_SIZE / MXFP4_GROUP_SIZE;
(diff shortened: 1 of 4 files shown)


Lazarus-931:
hi @YuhuaBillChen, if possible, can you measure more sweeps across quantization types and preferably batches as well?

mlx_vlm/models/mxfp4_verifier.py
```diff
@@ -0,0 +1,440 @@
...
+        or linear.biases is not None
+    ):
+        return False
+
+    K = linear.weight.shape[1] * 32 // linear.bits
+    N = linear.weight.shape[0]
+    return K % 512 == 0 and N % 8 == 0
+
+
+def _can_optimized_mxfp4(linear, x: mx.array) -> bool:
```
Lazarus-931, line 309: i was thinking, this is gated for small T, but can we see if higher T values can be included as well ( can open up all Dflash classes! ) so dflash doesn't go through standard gather_gmm!

Lazarus-931 (approved):
thanks, this perf gain holds(1.4x) with qwen3.6 mtp as well, but if we can extend to dflash with higher T for mxfp4 path, would be amazing!

can open up more models and their drafter counterparts(glm for example)

mlx_vlm/models/mxfp4_verifier.py
YuhuaBillChen: Agreed — higher-T MXFP4 would be useful for DFlash and models such as GLM. The current kernel keeps per-token state proportional to T, so simply removing the `T <= 4` gate risks making the larger blocks slower. I’ll handle that as a follow-up with a time-tiled implementation and correctness/performance sweeps at T=6/8/14/16, including a real DFlash workload once the checkpoint is available.

YuhuaBillChen:
The requested quantization/batch sweep and the Q3 T=3/4 tiling fix are now included in `2917c7c2`. CI is green, all 252 paired microbenchmark cases matched the singleton reference, and the real Qwen3.8-27B Q3 ABBA run improved from 13.705 to 24.186 tok/s with identical output and MTP statistics.

@Blaizzy, Lazarus has approved this and it should be ready to merge when convenient. I’ll keep the higher-T MXFP4/DFlash work in a separate follow-up.

YuhuaBillChen:
Hi @Lazarus-931, I just ran some experiments and updated the PR content. Would you like to merge when you have a chance? Thanks!

Lazarus-931:
> The requested quantization/batch sweep and the Q3 T=3/4 tiling fix are now included in `2917c7c2`. CI is green, all 252 paired microbenchmark cases matched the singleton reference, and the real Qwen3.8-27B Q3 ABBA run improved from 13.705 to 24.186 tok/s with identical output and MTP statistics.
> 
> @Blaizzy, Lazarus has approved this and it should be ready to merge when convenient. I’ll keep the higher-T MXFP4/DFlash work in a separate follow-up.

looking forward to it :)

YuhuaBillChen:
Hi @Blaizzy, sorry but I just noticed #2209 was reverted in #2251 before the new version 0.7.1. I'd rather ask than guess: was there a specific correctness or compatibility regression that motivated the revert, or was it a release-timing / maintainability decision?

If there is a failing model, device, MLX version, or configuration, I'd be happy to reproduce it and follow up with a fix and regression test. If the concern is primarily the scope or maintenance cost of the specialized kernel path, I'd also appreciate knowing whether a smaller, split follow-up would be welcome.

For context, the PR had green Cl, the requested quantization/batch sweeps, and @Lazarus-931 's independent Qwen3.6 MTP confirmation and approval. 

Thanks a ton, I just want to understand the right path forward.

PR #2213, opened 2026-09-10
fix(moe_offload): match nested Experts.switch_glu path (Gemma 4)

typemild:
## Problem

Gemma 4 26B-A4B's MoE layer nests `SwitchGLU` inside an `Experts` wrapper
module, so parameter names look like
`...layers.N.experts.switch_glu.gate_proj.weight` instead of the expected
`...layers.N.experts.gate_proj.weight`.

`STACKED_RE` / `STACKED_FUSED_RE` didn't account for this extra segment, so
`plan()` never matched any of these tensors. As a result, `repack()`
silently finished, reporting `experts: 0 MoE layers`, and the offload dir
appeared valid until `patch_model()` failed at load time with "No expert
files found".

## Fix

Allow an optional `.switch_glu` segment between the `experts`/`switch_mlp`
anchor and the projection name in both regexes.

## Testing

- Added `test_plan_partitions_nested_switch_glu_experts` to
  `TestMoEOffload` in `mlx_vlm/tests/test_models.py`, following the
  existing `test_plan_partitions_*` convention — verifies `plan()`
  correctly classifies the nested `experts.switch_glu.*` names as
  `STACK` entries under the right layer.
- Full `TestMoEOffload` suite (17 tests) passes locally.
- End-to-end verification against `mlx-community/gemma-4-26B-A4B-it-8bit`:
  `repack` now reports `experts: N MoE layers x 128 experts` (previously
  0), and the resulting offload dir loads and serves correctly through
  `mlx_vlm.server`.

**Test environment:** MacBook Pro 14" (M4 Pro, 14-core CPU / 20-core GPU),
48GB unified memory, macOS Tahoe 26.6.2.


Lazarus-931 (commented):
lgtm, but can we close the  silent-success hole of `written_layer` something like 

```python
if not written_layers:
   raise ValueError(...)
```
for consistency and error handling downstream to patch_model

something like "no MoE expert tensors matched..."

typemild:
Fixed — added a `written_layers` check right before `offload_index.json` is written, raising ValueError for both the fast path and the sanitize() fallback. All 17 tests in `TestMoEOffload` still pass.

Also re-verified as a negative control: reverting STACKED_RE/STACKED_FUSED_RE to the pre-fix version and re-running repack() against gemma-4-26B-A4B-it-8bit now fails fast with the new ValueError, instead of silently finishing with 0 MoE layers.

Lazarus-931 (approved):

mlx_vlm/moe_offload.py
```diff
@@ -394,6 +395,17 @@ def load_names(names):
 
     import shutil
 
+    if not written_layers:
```
Lazarus-931: seems a bit too verbose, maybe something like:

moe_offload.repack: {build!r} produced no MoE expert tensors(written_layers is empty). Check STACKED_RE/STACKED_FUSED_RE against this model's actual expert tensor naming

Lazarus-931:
tiny nit and good to merge :)

typemild:
Thanks so much for reviewing

Lazarus-931:
> Thanks so much for reviewing

of course, feel free to fix the error nit and can merge

PR #2219, opened 2026-09-11
Preserve original Chat history when replaying through Responses

Lazarus-931:
Share prompt preparation across Chat Completions, Responses, and Responses token counting while preserving each API’s existing message, media, and tool semantics, establishing the input boundary needed for compaction. 

(1st of of several prs for compression)

I made sure to test this extensively, 4068 tests + 448 subtests, and a 122-snapshot sweep.

### DIFF
--- a/mlx_vlm/apc.py
+++ b/mlx_vlm/apc.py
@@ -4672,6 +4672,11 @@ def apc_lookup_plan(
         return None
 
     matched, prefix_len = manager.lookup_prefix(ids_list, extra_hash=extra_hash)
+    if prefix_len == n:
+        # Recompute the final block to obtain logits for the first output token.
+        manager.release(matched[-1:])
+        matched = matched[:-1]
+        prefix_len -= manager.block_size
     if prefix_len > 0 and prefix_has_media(prefix_len):
         manager.release(matched)
         matched = []
--- a/mlx_vlm/apc_adapters.py
+++ b/mlx_vlm/apc_adapters.py
@@ -222,7 +222,12 @@ def register_default_capabilities() -> None:
         c.SimpleKVCache,
     ):
         register_capability(cls, Capability.PAGEABLE)
-    for cls in (c.RotatingKVCache, c.BatchRotatingKVCache, c.ChunkedKVCache):
+    for cls in (
+        c.RotatingKVCache,
+        c.BatchRotatingKVCache,
+        c.ChunkedKVCache,
+        c.BatchChunkedKVCache,
+    ):
         register_capability(cls, Capability.WINDOWED)
     for cls in (
         c.ArraysCache,
@@ -427,7 +432,7 @@ class ChunkedKVCacheCloneAdapter:
     def merge_rows(self, caches, prefix_lens):
         from .models import cache as lm
 
-        return lm.BatchKVCache.merge(caches)
+        return lm.BatchChunkedKVCache.merge(caches)
 
 
 class ArraysCacheCloneAdapter:
@@ -575,6 +580,8 @@ def clone_cache_entry(c, *, min_capacity_tokens, eval_targets):
         if not c.is_single_row():
             return None
         if c.empty():
+            if isinstance(c, lm.BatchChunkedKVCache):
+                return c.extract(0)
             if isinstance(c, lm.BatchRotatingKVCache):
                 return lm.RotatingKVCache(max_size=int(c.max_size))
             return lm.KVCache()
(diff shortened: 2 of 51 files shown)


lucasnewman:
@Lazarus-931 This patch is huge even without the tests so it's difficult to fully review myself. It might be worth breaking it up into model changes and server-side changes unless they are inextricably linked.

In the meantime I had Codex take a look at it and it found these issues, can you take a look?

```
[P1] Harmony parsing can discard the entire answer.
The parser starts in header mode unless generation begins inside analysis. But LLM-jp-VL’s template already completes the final-channel header in the prompt. Generated Hello world<|return|> therefore produces empty content, in both streaming and nonstreaming parsing. Initialize the parser from the actual prompt’s channel/header state. 

[P2] Responses continuation moves historical images into the newest user turn.
Stored assistant output becomes an untyped Chat message, which switches the whole history to Chat normalization. That strips per-message image placement; prompt rendering then assigns images to the last user turn. Reproduction: an image accompanying “Describe this image” moves to the later “What color is it?” turn. Preserve native media ownership when mixing stored assistant turns with Responses items.

[P2] Replaying native apply_patch calls erases their patch bodies.
The server stores these calls with raw patch text. The new replay path passes that text through Chat’s JSON-argument normalization, where decoding fails and replaces it with {}. Subsequent generations lose the edit associated with the tool result. Convert native patch calls to structured arguments before Chat normalization.
```

Lazarus-931:
> @Lazarus-931 This patch is huge even without the tests so it's difficult to fully review myself. It might be worth breaking it up into model changes and server-side changes unless they are inextricably linked.
> 
> In the meantime I had Codex take a look at it and it found these issues, can you take a look?
> 
> ```
> [P1] Harmony parsing can discard the entire answer.
> The parser starts in header mode unless generation begins inside analysis. But LLM-jp-VL’s template already completes the final-channel header in the prompt. Generated Hello world<|return|> therefore produces empty content, in both streaming and nonstreaming parsing. Initialize the parser from the actual prompt’s channel/header state. 
> 
> [P2] Responses continuation moves historical images into the newest user turn.
> Stored assistant output becomes an untyped Chat message, which switches the whole history to Chat normalization. That strips per-message image placement; prompt rendering then assigns images to the last user turn. Reproduction: an image accompanying “Describe this image” moves to the later “What color is it?” turn. Preserve native media ownership when mixing stored assistant turns with Responses items.
> 
> [P2] Replaying native apply_patch calls erases their patch bodies.
> The server stores these calls with raw patch text. The new replay path passes that text through Chat’s JSON-argument normalization, where decoding fails and replaces it with {}. Subsequent generations lose the edit associated with the tool result. Convert native patch calls to structured arguments before Chat normalization.
> ```

hm, i was thinking as well, was hesitant from spamming model specific prs, but i'll most likely break it apart in server, model and maybe per family so it's easier

Blaizzy:
I wonder why you took this approach all together. 

What's the blocker? What does this change solve? Is there a reproducible example? 

Based on the spec of compression, it's pretty straightforward implementation that wouldn't need such change.

Lazarus-931:
> I wonder why you took this approach all together.
> 
> What's the blocker? What does this change solve? Is there a reproducible example?
> 
> Based on the spec of compression, it's pretty straightforward implementation that wouldn't need such change.

compaction replays the prior context through a response-shaped endpoint, and summarizes.

if you reproduce on main, most if not all model's outputs differ when doing replay on them

which i think is important to fix, since cache prefix therefore won't work and we have bad summary model generates

did u have a different idea in mind?

Blaizzy:
> compaction replays the prior context through a response-shaped endpoint, and summarizes.


Please show me a reproducible example.



> which i think is important to fix, since cache prefix therefore won't work and we have bad summary model generates


It should work. Why isn't it working? Let me see the example and the models you tested 


> did u have a different idea in mind?

Yes.

Lazarus-931:
> > compaction replays the prior context through a response-shaped endpoint, and summarizes.
> 
> Please show me a reproducible example.
> 
> > which i think is important to fix, since cache prefix therefore won't work and we have bad summary model generates
> 
> It should work. Why isn't it working? Let me see the example and the models you tested
> 
> > did u have a different idea in mind?
> 
> Yes.

Start a server on main and one on this branch, you should see a difference, running this: https://gist.github.com/Lazarus-931/6adcdb3270511823e7511dccf83095ec

if u wanna take it diff route, happy to as well

Blaizzy:
Thanks Alazar for catching the issue! 

Closing this PR in favour of #2260

cc: @lucasnewman

Issue #2201, opened 2026-09-09
Qwen assistant reasoning_content is dropped during prompt normalization


bstpierr:
Incoming assistant `reasoning_content` is lost when `apply_chat_template()` normalizes ordinary Qwen assistant dictionaries. The server accepts and forwards the field, but prompt normalization rebuilds these messages from role/content. Tool-bearing assistant messages already retain it, so adding tool metadata changes whether the same reasoning reaches the template.

### Reproducer

Verified against upstream commit `23f513829724365b0cd8884ecaa10e0a1577dba0`, with Python 3.10.20 and MLX 0.32.2. This uses synthetic content and does not load a model or run inference:

```python
import mlx.core as mx
mx.set_default_device(mx.cpu)

from mlx_vlm.prompt_utils import apply_chat_template

message = {
    "role": "assistant",
    "content": "Done",
    "reasoning_content": "Synthetic note",
}
for model_type in ("qwen3_5", "qwen3_5_moe"):
    for prompt in (message, [message]):
        normalized = apply_chat_template(
            None, {"model_type": model_type}, prompt, return_messages=True
        )
        print(model_type, normalized[0].get("reasoning_content"))
```

**Actual:** all four results are `None`; the normalized messages have no `reasoning_content` field.

**Expected:** all four results are `Synthetic note`. Preserve the supplied field through normalization so the checkpoint's chat template can decide whether to render it. Existing content normalization and caller inputs should remain unchanged.

### Affected scope

The failure was reproduced with the pinned official templates for:

- [Qwen3.5-4B](https://huggingface.co/Qwen/Qwen3.5-4B/tree/851bf6e806efd8d0a36b00ddf55e13ccb7b8cd0a), routed as `qwen3_5`.
- [Qwen3.5-35B-A3B](https://huggingface.co/Qwen/Qwen3.5-35B-A3B/tree/59d61f3ce65a6d9863b86d2e96597125219dc754), routed as `qwen3_5_moe`, with the same template bytes.
- [Qwen3.8-27B](https://huggingface.co/Qwen/Qwen3.8-27B/tree/1d4bf0f2ff6012fd82039f2fa52739d0dd7c60c0), also routed as `qwen3_5`.

This report covers assistant dictionaries, including entries in message lists. It does not establish the same scope for other Qwen3.8 MoE routes or raw message objects. History-rendering policy varies by template revision; preserving the input field does not require changing that policy.

### Related work

[...]

PR #2223, opened 2026-09-12
Preserve message metadata through dictionary formatting

bstpierr:
Ordinary dictionary messages lose fields such as `reasoning_content` and `name` when a model formatter rebuilds role/content. This keeps incoming metadata at both dictionary normalization sites, with formatter-produced fields taking precedence.

```python

message = {"role": "assistant", "content": "answer", "reasoning_content": "thought"}
normalized = apply_chat_template(
    None, {"model_type": "qwen3_5"}, [message], return_messages=True
)
assert normalized[0]["reasoning_content"] == "thought"  # KeyError before this fix
```

The shared helper guards both input and output shapes and performs a shallow merge. Retained nested values remain shared with the caller. Absent, null, empty and populated metadata retain their meaning; formatted role/content win. Normalization does not mutate caller inputs. Messages carrying `tool_calls` or `tool_call_id` still enter the existing tool normalizer before the ordinary formatting sites, including null and empty values. String-returning formatters, message-like objects, special model paths and media allocation retain their existing behavior. One rule makes an architecture-specific retention patch unnecessary for each new field.

Templates still own rendering, validation and reasoning-history policy. Retained metadata can intentionally change prompt text and length, available context, cache reuse, or expose template validation errors. For the pinned Qwen3.8 template, echoed historical `reasoning_content` is now visible and retained by default, so those prompts grow. This changes neither reasoning defaults nor the separate `preserve_thinking` API. HTTP coverage demonstrates fields the endpoint already forwards; it does not add arbitrary-extra-field forwarding at the API boundary.

Validation:

- 10 new test functions / 26 cases; on recorded base `d2a1434a03e4c9975b0d505e7178e0cfc4082a83`, 14 targeted failures and 12 passing controls.
- 95 CPU checks pass on both Transformers 5.14.0 (the supported minimum) and 5.17.0: all prompt regressions plus selected HTTP/schema/tool-choice controls and a mocked HTTP request using real normalization/rendering.
- Pinned, licensed Qwen3.5, Qwen3.8, Gemma 4 and Qwen2.5-VL templates run through the Transformers Jinja renderer offline. They cover structured/inline precedence, historical turns, active continuations, null/empty reasoning, and an unaffected rendering control. Qwen3.5 and Qwen3.8 use distinct fixtures; this is representative evidence, not universal checkpoint support.
- Separate offline checks with the pinned official Qwen vocabularies matched native rendered token IDs in 12/12 cases; the base differed in 6/12. Vocabulary downloads are not part of CI.
- Reused #2200’s 20 option-validation tests in a temporary combined checkout: all pass. Its API and tests are not included here.
- `pre-commit run --all-files` and `git diff --check` pass. Python 3.10.20, Transformers 5.14.0/5.17.0, MLX 0.32.2; CPU selected before package import. Two dependency deprecation warnings remain. The full upstream test suite and hardware inference were not run; no weights or GPU workloads were used.

Related work: this is a separate replacement for the narrow fix in #2199 and addresses the normalization issue in #2201. The existing #2199 and #2219 are left unchanged so maintainers can choose which implementation to land. Thanks to @wmlanglois for independently proposing the original-dictionary/formatter-override rule in [this comment](https://github.com/Blaizzy/mlx-vlm/issues/2201#issuecomment-5641455514). #2219 includes an overlapping list-message retention hunk inside a broader request-preparation change; this contribution also covers standalone dictionary input. #2200 remains the separate history-control API work.

AI assistance: implemented and tested with OpenAI Codex, with advisory plan and commit reviews by Anthropic Claude Fable 5.1 at Medium. Fixture sources, revisions, hashes and Apache-2.0 licensing are recorded beside the fixtures; contribution code uses the project’s MIT license. No inference quality or performance improvement is claimed.

### DIFF
--- a/mlx_vlm/prompt_utils.py
+++ b/mlx_vlm/prompt_utils.py
@@ -179,6 +179,15 @@ def _get_role_content(item: Any) -> Union[tuple[str, Any], None]:
     return None
 
 
+def _preserve_message_metadata(
+    original: Any, formatted: Union[str, Dict[str, Any]]
+) -> Union[str, Dict[str, Any]]:
+    """Preserve dictionary fields while letting formatter-owned fields win."""
+    if isinstance(original, dict) and isinstance(formatted, dict):
+        return {**original, **formatted}
+    return formatted
+
+
 def _content_media_count(content: Any, media_types: tuple[str, ...]) -> int:
     if not isinstance(content, list):
         return 0
@@ -928,13 +937,16 @@ def apply_chat_template(
         else:
             content = extract_text_from_content(prompt["content"])
             messages.append(
-                get_message_json(
-                    model_type,
-                    content,
-                    role,
-                    num_images=num_images,
-                    num_audios=num_audios,
-                    **kwargs,
+                _preserve_message_metadata(
+                    prompt,
+                    get_message_json(
+                        model_type,
+                        content,
+                        role,
+                        num_images=num_images,
+                        num_audios=num_audios,
+                        **kwargs,
+                    ),
                 )
             )
     elif isinstance(prompt, list):
@@ -1002,17 +1014,20 @@ def apply_chat_template(
                     # Handle multimodal content: extract only text, skip image/audio URLs
                     content = extract_text_from_content(content)
                     messages.append(
-                        get_message_json(
-                            model_type,
-                            content,
-                            role,
-                            skip_image_token=image_counts[i] == 0
-                            or role in ["system", "assistant"],
-                            skip_audio_token=audio_counts[i] == 0
-                            or role in ["system", "assistant"],
-                            num_images=image_counts[i],
-                            num_audios=audio_counts[i],
-                            **kwargs,
+                        _preserve_message_metadata(
+                            p,
+                            get_message_json(
+                                model_type,
+                                content,
+                                role,
+                                skip_image_token=image_counts[i] == 0
+                                or role in ["system", "assistant"],
+                                skip_audio_token=audio_counts[i] == 0
+                                or role in ["system", "assistant"],
+                                num_images=image_counts[i],
+                                num_audios=audio_counts[i],
+                                **kwargs,
+                            ),
                         )
                     )
 
--- a/mlx_vlm/prompt_utils.py
+++ b/mlx_vlm/prompt_utils.py
@@ -182,7 +182,10 @@ def _get_role_content(item: Any) -> Union[tuple[str, Any], None]:
 def _preserve_message_metadata(
     original: Any, formatted: Union[str, Dict[str, Any]]
 ) -> Union[str, Dict[str, Any]]:
-    """Preserve dictionary fields while letting formatter-owned fields win."""
+    """Shallow-copy dictionary fields, with formatter-owned fields taking priority.
+
+    Other input/output shapes retain their existing formatter behavior.
+    """
     if isinstance(original, dict) and isinstance(formatted, dict):
         return {**original, **formatted}
     return formatted
@@ -856,6 +859,10 @@ def apply_chat_template(
 ) -> Union[List[Dict[str, Any]], str, Any]:
     """
     Apply chat template to prompts.
+
+    Ordinary dictionary messages retain fields not replaced by their dictionary
+    formatter. Values are not coerced; rendering and reasoning-history policy
+    belong to the template. Retained metadata can affect prompts and validation.
 
     Args:
         processor: The processor with chat template functionality
--- a/mlx_vlm/tests/conftest.py
+++ b/mlx_vlm/tests/conftest.py
@@ -1,9 +1,34 @@
 import os
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
 
 # float32 matmul runs at TF32 precision on hardware with matrix units, which is
 # looser than the float32 references these tests compare against. Set rather than
 # setdefault, so an inherited MLX_ENABLE_TF32=1 doesn't leak into the test run.
 os.environ["MLX_ENABLE_TF32"] = "0"
+
+
+@pytest.fixture
+def template_processor():
+    """Actual pinned templates and the Transformers renderer; no model or vocab."""
+    from transformers.utils.chat_template_utils import render_jinja_template
+
+    def load(name):
+        template = (
+            Path(__file__).parent / "fixtures" / "chat_templates" / f"{name}.jinja"
+        ).read_text()
+
+        def render(messages, *, tokenize=False, **kwargs):
+            assert tokenize is False
+            return render_jinja_template(
+                [messages], chat_template=template, bos_token="<bos>", **kwargs
+            )[0][0]
+
+        return SimpleNamespace(chat_template=template, apply_chat_template=render)
+
+    return load
 
 
 def pytest_sessionfinish(session, exitstatus):
--- a/mlx_vlm/tests/fixtures/chat_templates/LICENSE-APACHE-2.0.txt
+++ b/mlx_vlm/tests/fixtures/chat_templates/LICENSE-APACHE-2.0.txt
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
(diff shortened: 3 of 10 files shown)


wmlanglois:
[mlxvlm_metadata_harness_captures.zip](https://github.com/user-attachments/files/32135439/mlxvlm_metadata_harness_captures.zip)


Thanks for opening #2223 - that matches what I had locally, and I've now re-run my harness against your base rather than my older checkout. Hope this helps... spent long enough trying to fix it, thought I would share. By fix, I mean I just keep telling my agents they are wrong and there should be a way to make it work.

### Reproduced on d2a1434a

Same result as I saw on v0.6.13:

| | |
|---|---|
| cases compared | 119 |
| model types | 17 |
| regressions | 0 |
| behaviour changes on error paths | 0 |
| fields newly preserved | `reasoning_content` 28 messages, `name` 14 messages |

`role` and `content` are byte-identical on every case; the diff is purely additive.

### What the harness does

It calls `apply_chat_template(..., return_messages=True)` over a matrix of model types and message shapes, dumps the normalized messages to JSON, and diffs two runs. The rule it enforces is one-directional: **every key/value the baseline produced must survive unchanged, and any added key must have come from the caller's message.**

Because `return_messages=True` returns before `get_chat_template`, it needs no processor, no tokenizer and no weights, so the image and multimodal formatter branches all execute on CPU in about a second. That is the tradeoff versus your tests: mine covers formatter breadth but stops before rendering, yours covers rendering and token-ID equality on real vocabularies. They check different halves, so running both seems worthwhile.

Message shapes per model type: user text, user with image, assistant with `reasoning_content` and no tools, assistant with `reasoning_content` and tool calls, tool-result message with `tool_call_id` and `name`, system + history with `name`, and a multimodal history mixing an image turn with a reasoning turn.

### Coverage

One representative model type per distinct `MessageFormat`, 16 of them, covering 83 registered model types, plus one unregistered type to exercise the non-`MODEL_CONFIG` text-only branch. That is the 17.

```
IMAGE_PATCH_TOKEN            IMAGE_TOKEN                  IMAGE_TOKEN_NEWLINE
IMAGE_TOKEN_PIPE             IMAGE_TOKEN_WRAPPED          LIST_WITH_IMAGE
LIST_WITH_IMAGE_FIRST        LIST_WITH_IMAGE_TYPE         LIST_WITH_IMAGE_TYPE_TEXT
LIST_WITH_IMAGE_URL_FIRST    NUMBERED_IMAGE_TOKENS        PROMPT_ONLY
PROMPT_WITH_IMAGE_TOKEN      SINGLE_IMAGE_TOKEN           START_IMAGE_TOKEN
TEXT_ONLY                    (+ unregistered model_type)
```

### Coordinating with #2219

I fetched both branches and merged them locally. **`git merge` is clean**, and the harness shows **0 regressions across the 112 shared cases**, with no additional fields preserved beyond what #2223 already adds.

The only difference is the case set, and it is a consequence of #2219's `MODEL_CONFIG` edits rather than a regression: moving `paligemma` from `PROMPT_WITH_IMAGE_TOKEN` to `IMAGE_TOKEN` leaves `PROMPT_WITH_IMAGE_TOKEN` with no registered model types, so it drops out of a representative-per-format selection. Worth noting it then joins `LIST_WITH_IMAGE_TYPE_TEXT_IMAGE_LAST`, `PROMPT_WITH_START_IMAGE_TOKEN` and `VIDEO_WITH_TEXT`, which already have no registered models on main - so this looks expected rather than a problem, just a small loss of formatter coverage.

That did require a harness fix: the comparator originally asserted identical case sets, which aborts whenever a change edits `MODEL_CONFIG`. It now compares the intersection and reports set changes separately. Anyone diffing across two PRs that touch the registry will hit the same thing.

### One thing I looked for and did not find

`_coalesce_leading_system_text` in #2219 gates on `set(message) != {"role", "content"}`, so I expected a leading system message carrying a preserved `name` to silently skip coalescing once #2223 retains it. I tested it on the merged tree and **it did not reproduce** through `apply_chat_template` - both the plain and the `name`-carrying case returned the same message count. So it presumably sits on the server prompt-prep path instead. Flagging it only as something worth a glance during review, not as a defect.

### Files

The three captured runs are in the attached zip (`main_base.json` at d2a1434a, `main_pr2223.json`, `main_merged.json` for 2219+2223), along with both scripts. Reproduce with:

```bash
PYTHONPATH=. python capture_matrix.py base.json      # on the base commit
PYTHONPATH=. python capture_matrix.py candidate.json # on the PR branch
python compare_matrix.py base.json candidate.json
```

<details>
<summary><code>capture_matrix.py</code></summary>

```python
"""Capture apply_chat_template(return_messages=True) across a model-type and
message-shape matrix, so stock and patched output can be compared exactly.

return_messages=True stops before get_chat_template, so no processor or model
weights are needed. That lets the image/VLM formatter paths be exercised
without downloading a vision model.
"""
import json
import sys

import mlx_vlm.prompt_utils as pu
from mlx_vlm.prompt_utils import MODEL_CONFIG

OUT = sys.argv[1]

# One representative model type per distinct MessageFormat, so every formatter
# branch is covered rather than just the one Qwen3.8 uses.
by_format = {}
for name, fmt in MODEL_CONFIG.items():
    by_format.setdefault(str(fmt), name)
MODEL_TYPES = sorted(by_format.values())
MODEL_TYPES.append("not_a_registered_vlm")  # text-only branch

IMG = {"type": "image_url", "image_url": {"url": "http://example.com/a.png"}}

CASES = {
    "user_text": [
        {"role": "user", "content": "hello"},
    ],
    "user_with_image": [
        {"role": "user", "content": [IMG, {"type": "text", "text": "describe"}]},
    ],
    "assistant_reasoning_no_tools": [
        {"role": "user", "content": "q"},
        {"role": "assistant", "content": "a", "reasoning_content": "MARK-R"},
        {"role": "user", "content": "q2"},
    ],
    "assistant_reasoning_with_tools": [
        {"role": "user", "content": "q"},
        {"role": "assistant", "content": "a", "reasoning_content": "MARK-R",
         "tool_calls": [{"id": "c1", "type": "function",
                         "function": {"name": "grep", "arguments": {"p": 1}}}]},
        {"role": "user", "content": "q2"},
    ],
    "tool_result_message": [
        {"role": "user", "content": "q"},
        {"role": "assistant", "content": "",
         "tool_calls": [{"id": "c1", "type": "function",
                         "function": {"name": "grep", "arguments": {"p": 1}}}]},
        {"role": "tool", "content": "hit", "tool_call_id": "c1", "name": "grep"},
        {"role": "user", "content": "q2"},
    ],
    "system_history": [
        {"role": "system", "content": "sys"},
        {"role": "user", "content": "q"},
        {"role": "assistant", "content": "a", "name": "bot"},
    ],
    "multimodal_history": [
        {"role": "user", "content": [IMG, {"type": "text", "text": "what"}]},
        {"role": "assistant", "content": "a dog", "reasoning_content": "MARK-R"},
        {"role": "user", "content": "and now?"},
    ],
}

results = {}
for mt in MODEL_TYPES:
    cfg = {"model_type": mt}
    for case, msgs in CASES.items():
        key = f"{mt}::{case}"
        n_img = sum(
            1 for m in msgs if isinstance(m.get("content"), list)
            for c in m["content"] if isinstance(c, dict) and c.get("type") == "image_url"
        )
        try:
            out = pu.apply_chat_template(None, cfg, msgs, add_generation_prompt=True,
                                         return_messages=True, num_images=n_img)
            results[key] = {"ok": True, "messages": out}
        except Exception as e:
            results[key] = {"ok": False, "error": f"{type(e).__name__}: {e}"}

with open(OUT, "w", encoding="utf-8") as fh:
    json.dump(results, fh, indent=1, default=str)
print(f"captured {len(results)} cases across {len(MODEL_TYPES)} model types -> {OUT}")
```

</details>

<details>
<summary><code>compare_matrix.py</code></summary>

```python
"""Compare stock vs patched formatter output.

Regression rule: every key/value the STOCK formatter produced must survive
unchanged in the patched output. The patch may only ADD keys, and every added
key must have come from the caller's original message.
"""
import json
import sys

stock = json.load(open(sys.argv[1], encoding="utf-8"))
patched = json.load(open(sys.argv[2], encoding="utf-8"))

regressions, additions, errors_changed = [], [], []

# Case sets can legitimately differ when a change edits MODEL_CONFIG, since
# cases are keyed by one representative model type per distinct MessageFormat.
# Report that separately and compare the shared cases rather than aborting.
only_stock = sorted(set(stock) - set(patched))
only_patched = sorted(set(patched) - set(stock))
shared = sorted(set(stock) & set(patched))

for key in shared:
    s, p = stock[key], patched[key]
    if s["ok"] != p["ok"]:
        errors_changed.append((key, s.get("error"), p.get("error")))
        continue
    if not s["ok"]:
        if s.get("error") != p.get("error"):
            errors_changed.append((key, s.get("error"), p.get("error")))
        continue
    sm, pm = s["messages"], p["messages"]
    if len(sm) != len(pm):
        regressions.append(f"{key}: message count {len(sm)} -> {len(pm)}")
        continue
    for i, (a, b) in enumerate(zip(sm, pm)):
        if not isinstance(a, dict) or not isinstance(b, dict):
            if a != b:
                regressions.append(f"{key}[{i}]: non-dict changed {a!r} -> {b!r}")
            continue
        for k, v in a.items():
            if k not in b:
                regressions.append(f"{key}[{i}]: key {k!r} disappeared")
            elif b[k] != v:
                regressions.append(f"{key}[{i}]: key {k!r} changed {v!r} -> {b[k]!r}")
        for k in b:
            if k not in a:
                additions.append(f"{key}[{i}]: +{k}")

print("=" * 78)
print(f"cases compared          : {len(shared)}")
if only_stock or only_patched:
    print(f"cases only in baseline  : {len(only_stock)}")
    for k in only_stock[:6]:
        print(f"   -{k}")
    print(f"cases only in candidate : {len(only_patched)}")
    for k in only_patched[:6]:
        print(f"   +{k}")
    print("   (case-set changes mean MODEL_CONFIG was edited, not a regression)")
print(f"REGRESSIONS             : {len(regressions)}")
for r in regressions[:25]:
    print(f"   {r}")
print(f"errors changed          : {len(errors_changed)}")
for k, a, b in errors_changed[:10]:
    print(f"   {k}: {a} -> {b}")
print(f"fields newly preserved  : {len(additions)}")
seen = {}
for a in additions:
    field = a.split("+")[-1]
    seen[field] = seen.get(field, 0) + 1
for field, n in sorted(seen.items(), key=lambda x: -x[1]):
    print(f"   {field}: {n} messages")
print("=" * 78)
print("PASS" if not regressions and not errors_changed else "FAIL")
```

</details>

Happy to fold these cases into the PR's test file in whatever form suits, or to leave the harness as an external check. One caveat on my end: I have not run the full upstream suite or any model inference, so this is formatter-level evidence only.

PR #2224, opened 2026-09-12
Compact a conversation without re-reading it

Lazarus-931:
Stacked on #2219


Lazarus-931:
Opened against the wrong base — reopening as a proper stacked PR on top of #2219.

PR #2226, opened 2026-09-12
Compact a conversation without re-reading it

Lazarus-931:
Summarize an active conversation into a replacement for its own history through `POST /v1/responses/compact`, appending the summarization instruction as a final turn so the preceding request's prefix cache is reused instead of the conversation being re-read. Validated with 4103 passing tests and a real-checkpoint run that reused 2269 of the preceding request's 2270 prompt tokens.

(2nd of several prs for compression, following #2260)


Lazarus-931:
Closing this. A plain `/v1/responses` call with the summarization instruction appended as the final turn reuses 943 of 944 prompt tokens — identical to this endpoint — so the prefix-cache argument for a dedicated endpoint doesn't hold. The endpoint is also 62% verbatim duplication of `responses_endpoint`, which drifts silently: a missing `type: "message"` on the appended item made the converter drop the instruction entirely and the endpoint still returned 200 with the model simply continuing the conversation.

The part worth keeping is the three refusals — empty summary, summary that stopped at the output limit, and summary that can't carry the conversation forward — since compaction replaces the history and a truncated summary is silent corruption. Those belong on the existing Responses path rather than a parallel copy.

Will revisit in that shape if compaction lands server-side.

PR #2231, opened 2026-09-13
fix(glm5-next): remap quantized lm head parameters

raullenchai:
## Summary

GLM-5-Next checkpoints with a quantized output head carry `lm_head.weight`, `lm_head.scales`, and `lm_head.biases`. The multimodal sanitizer previously remapped only the weight, leaving scales and biases at the checkpoint root; strict model loading then failed with `Received 2 parameters not in model: lm_head.biases, lm_head.scales`.

This remaps the complete `lm_head.*` parameter family under `language_model.*`, while preserving already-remapped keys on repeated sanitizer passes.

## Verification

- `pytest mlx_vlm/tests/test_glm5_next_sanitize.py -q` — 2 passed
- `ruff check` and `ruff format --check` on changed files
- Real-model load on Apple M3 Ultra, macOS 26.5.2, MLX 0.32.2: `Vontra/GLM-5.3-Flash-MLX-4bit-MTP` (184 GB target) now completes strict loading; before this patch it failed on the two orphaned quantization tensors.
- The loaded model plus a 3.9 GB target-matched MTP drafter completed all 6/6 coding, knowledge, math, instruction-following, creative-writing, and long-context qualification tasks.


mlx_vlm/tests/test_glm5_next_sanitize.py
```diff
@@ -0,0 +1,46 @@
+from types import SimpleNamespace
```
Lazarus-931, line 1: can we squeeze this into another test file?

raullenchai:
The sole commit has been re-signed and force-updated as `2720675d`; the GitHub commit-signature check now passes. The code diff is unchanged.

Lazarus-931 (changes requested):
see comment

PR #2232, opened 2026-09-13
fix(server): safely support thinking budgets with a drafter

raullenchai:
## Summary

The server currently rejects every `thinking_budget` request when a speculative drafter is configured. Simply removing that guard is unsafe: a speculative round may have already verified and committed multiple target and drafter tokens past the point where `ThinkingBudgetCriteria` needs to force a token.

This adds a correctness-first, request-scoped fallback:

- budgeted requests use the existing autoregressive generation path, which can apply forced tokens exactly
- unbudgeted requests continue to use speculative decoding
- budgeted and unbudgeted rows are partitioned into compatible cohorts so `GenerationBatch` and `SpeculativeGenerationBatch` are never mixed
- an active fallback/speculative cohort drains before the scheduler switches batch type
- fallback requests do not report process-wide draft-token metrics

This deliberately does not claim speculative acceleration for budgeted requests. A transactional target/drafter boundary transition can build on this later.

## Verification

- `pytest mlx_vlm/tests/test_generate.py mlx_vlm/tests/test_server.py -q` — 511 passed, 1 skipped
- `pytest mlx_vlm/tests/test_speculative.py mlx_vlm/tests/test_speculative_transactions.py mlx_vlm/tests/test_qwen4_mtp.py -q` — 857 passed
- Real-model Apple M3 Ultra dogfood with a 184 GB `Vontra/GLM-5.3-Flash-MLX-4bit-MTP` target and 3.9 GB target-matched MTP drafter, block size 2
- Six qualification categories passed: coding 8/8 hidden tests, knowledge 5/5 facts, math exact, instruction exact, constrained creative writing, and 3,126-token contract retrieval
- Budget-fallback and budgeted AR baseline produced byte-identical reasoning and final output on all 6/6 tasks at temperature 0
- Median client throughput: 26.79 vs 26.32 tok/s (+1.8%, treated as measurement noise, not a speed claim)
- Peak Metal: 188.32 GB on the long-context task; no additional model download was needed

The real-model load also needs the independent quantized `lm_head.*` sanitizer fix in #2231.


raullenchai:
Maintainer PR #2206 now supersedes this narrow positioned-budget transaction if that rebuild lands.

I qualified #2206 on the real affine-Q4 GLM-5.3 target and Q4 sidecar, combined with #2231 and #2234: two MTP runs passed 12/12 and were deterministic; a same-branch AR control passed 6/6, with every reasoning/final byte identical to MTP. Median paired MTP gain was 1.221x, and median category throughput reached 33.655 tok/s. Full details are in the qualification comment on #2206.

I recommend merging #2206 rather than stacking this PR on top of it. I am leaving this open only as the smaller fallback until the maintainer chooses the speculative architecture.

Blaizzy:
Thanks! 

But I'm closing this PR for now.

PR #2233, opened 2026-09-13
perf(speculative): keep greedy MTP through thinking budgets

raullenchai:
## Why

`ThinkingBudgetCriteria` currently cannot safely run inside a speculative round: the verifier can sample several future target slots before it knows where the drafter will be rejected. PR #2232 therefore adds a fail-safe AR fallback. This follow-up preserves that guardrail for sampling and batched requests, while making the common batch-one greedy GLM MTP path both correct and speculative again.

Depends on #2232. The real GLM checkpoint also needs the quantized `lm_head` load fix in #2231.

## What

- checkpoint thinking-budget state by absolute output position
- restore and discard future checkpoints after a draft rejection
- run forced `\n</think>` tokens through positioned target verification
- project the singleton target block once and sample its raw logits positionally
- keep sampled requests and multi-row budget cohorts on the conservative AR fallback
- clear a pending forced token when resetting criteria between generations

## Full-model dogfood

M3 Ultra 256 GB, macOS 26.5.2, batch one, temperature 0, thinking enabled, per-task budgets, max 1,024 completion tokens. Target: `Vontra/GLM-5.3-Flash-MLX-4bit-MTP` revision `76add2a341a1cd90ad0e86bb69839ea9c35827c6` (184 GB uniform Q4). Drafter: target-matched Q4 MTP sidecar. Server used `--draft-kind mtp --draft-block-size 2`.

The six-task Rapid-MLX real-task gate covers executable coding (8 hidden tests), atomic knowledge, multi-step math, exact instruction following, constrained creative writing, and long-context contract retrieval. Each MTP run used the same prompts and budgets as the budgeted AR baseline (256/128/192/96/192/128 tokens respectively).

Three independent warm runs all passed 6/6. All 18 task results had byte-identical reasoning and final output versus AR.

| Category | MTP median tok/s | AR tok/s | Speedup |
| --- | ---: | ---: | ---: |
| Coding | 29.702 | 27.788 | 1.069x |
| Knowledge | 31.746 | 26.572 | 1.195x |
| Math | 30.399 | 26.072 | 1.166x |
| Instruction following | 30.617 | 24.996 | 1.225x |
| Creative writing | 29.865 | 27.227 | 1.097x |
| Long context | 11.491 | 10.025 | 1.146x |

Median across categories: **30.132 tok/s vs 26.322 tok/s (+14.5%)**. Every category improved; range +6.9% to +22.5%. Peak Metal memory was 188.674 GB.

Reproduction used the Rapid-MLX `benchmark_glm53_real_tasks.py` harness from PR raullenchai/Rapid-MLX#3380 against a clean local `mlx_vlm.server` process. No model downloads or cache relocation were performed.

## Verification

- `python -m pytest -q mlx_vlm/tests/test_generate.py mlx_vlm/tests/test_speculative.py mlx_vlm/tests/test_speculative_transactions.py mlx_vlm/tests/test_server.py mlx_vlm/tests/test_qwen4_mtp.py`: 1,374 passed, 1 skipped
- tiny randomly initialized GLM target + MTP drafter: exact AR token sequence across rejection and forced thinking close
- explicit coverage for sampled fallback, multi-row fallback, state rollback/reset, and one-shot singleton target projection
- `git diff --check`

## Scope

This deliberately does not enable speculative thinking budgets for non-greedy sampling or multi-row cohorts; those need RNG and per-row transactional state, not only criteria rollback.


raullenchai:
Maintainer PR #2206 now supersedes this narrow positioned-budget transaction if that rebuild lands.

I qualified #2206 on the real affine-Q4 GLM-5.3 target and Q4 sidecar, combined with #2231 and #2234: two MTP runs passed 12/12 and were deterministic; a same-branch AR control passed 6/6, with every reasoning/final byte identical to MTP. Median paired MTP gain was 1.221x, and median category throughput reached 33.655 tok/s. Full details are in the qualification comment on #2206.

I recommend merging #2206 rather than stacking this PR on top of it. I am leaving this open only as the smaller fallback until the maintainer chooses the speculative architecture.

PR #2234, opened 2026-09-13
Optimize GLM-5.3 MoE gate-up storage

raullenchai:
## Summary

- store routed MoE gate/up projections in one `QuantizedSwitchLinear`
- use one gathered QMM for both projections while retaining the exact short-block down path
- enable the fused representation for GLM-5.3-Next target and native MTP layers
- migrate both raw per-expert checkpoints and already-stacked MTP sidecars during sanitize
- generalize the existing HY-V4 fused module instead of introducing a second implementation

This follows the storage-fusion direction explored in oMLX PR #3410, but keeps bit-exact output at every tested verifier width. No checkpoint conversion or duplicate persistent weights are required.

## Why

GLM-5.3-Next routes eight of 288 experts through 42 sparse layers. The gate and up projections consume the same hidden vector and expert indices, but the split representation issues two gathered QMMs. A gate-first fused storage layout turns those into one dispatch without changing the affine Q4 values or activation order.

The first implementation used per-expert concatenations and exposed a real startup-memory failure on the 184 GB checkpoint. The final loader uses one flat stack plus a reshape, eliminating those intermediate allocations. The complete qualification stayed at the existing 188.674 GB peak Metal watermark.

## Apple Silicon measurements

M3 Ultra 256 GB, macOS 26.5.2, Q4 affine/group-size 64, BF16 activations, temperature 0. Target: `Vontra/GLM-5.3-Flash-MLX-4bit-MTP` revision `76add2a`; target-matched Q4 MTP sidecar; speculative block size 2.

The included production-geometry microbenchmark (H=4096, I=2048, 288 experts, top-8) measured the complete routed SwitchGLU call:

| path | median |
|---|---:|
| split gate/up | 0.583 ms |
| fused gate/up | 0.532 ms |
| speedup | **1.086x** |

All 80 paired outputs were bit-exact. Projection parity was also bit-exact at T=1/2/4/8/16/64.

A real checkpoint sparse layer, reconstructed from the original independent gate/up tensors, measured:

| tokens | speedup | parity |
|---:|---:|---:|
| 1 | 1.082x | bit-exact |
| 2 | 1.057x | bit-exact |
| 4 | 1.042x | bit-exact |
| 8 | 1.067x | bit-exact |
| 16 | 1.018x | bit-exact |

Three full warm real-task runs covered coding, knowledge, math, instruction following, creative writing, and 3K-token long context. All **18/18** tasks passed, and every reasoning/final string was byte-identical to the split-storage baseline.

| category | split decode tok/s | fused decode tok/s | gain |
|---|---:|---:|---:|
| coding | 34.676 | 35.661 | +2.8% |
| knowledge | 34.960 | 36.229 | +3.6% |
| math | 33.618 | 34.452 | +2.5% |
| instruction | 36.372 | 37.172 | +2.2% |
| creative | 31.178 | 32.015 | +2.7% |
| long context | 34.769 | 35.277 | +1.5% |

The median per-category decode gain is **+2.65%**. End-to-end client throughput's median paired category gain is **+2.49%**; the median of category medians moved from 30.132 to 31.156 tok/s (**+3.40%**). Peak Metal memory was 188.674 GB, unchanged from the qualified baseline suite.

## Reproduction

```bash
PYTHONPATH=. python scripts/benchmark_glm53_q4_gate_up_storage.py \
  --experts 288 --runs 80 \
  --output /private/tmp/glm53-q4-gate-up-storage.json
```

The full E2E run used the same target/prompts/budgets as the GLM-5.3 positioned-MTP qualification, with #2231 and #2233 applied for strict quantized-head loading and thinking-budget MTP respectively.

## Validation

- `82 passed` focused GLM-5.3, MTP, fused-SwitchGLU, and shared-MoE tests
- `3999 passed, 7 skipped, 149 subtests passed` across the locally runnable suite
- five unrelated Inkling audio tests fail because this host's installed `mlx-audio` lacks the newer `mel_filters(..., precise=...)` API
- three unrelated Nemotron voicechat files cannot collect because the installed `mlx-audio` lacks its new codec module
- real target lazy load resolves every sparse layer to fused Q4 storage
- existing 3.9 GB pre-stacked MTP sidecar loads with `strict=True`
- three full real-task E2E runs: 18/18 pass, byte-identical reasoning and final output


Lazarus-931:
hi @raullenchai thanks for pr

i don't see why 300+ loc change justifies 1.08x improvements, or am i missing something?

raullenchai:
Thanks for challenging the tradeoff — the headline diff is misleading, and I should have made the split clearer.

This is not 300+ production LOC for a 1.08x end-to-end gain:

- **Production:** 138 additions / 55 deletions (**+83 net**). This includes moving HY4's existing private `FusedSwitchGLU` into the shared switch layer instead of adding a second implementation, plus backward-compatible loading of raw-expert and prior fused checkpoints.
- **Tests:** 73 additions / 6 deletions.
- **Standalone reproducible benchmark:** 184 LOC. I can drop this script from the PR if you prefer benchmark evidence in the description only.

The `1.08x` number is the isolated complete routed SwitchGLU operation at production Q4 geometry, not the full-model claim. On the 184 GB target with real prompts, the full decode improvement was:

- six-task median server decode: **+2.65%**
- six-task median client throughput: **+3.40%**
- individual server categories: **+1.5% to +3.6%**
- real checkpoint sparse layer: **+1.8% to +8.2%**, depending on routed token count

All 18/18 repeated real-task outputs were byte-identical, and peak Metal stayed effectively unchanged (188.674 vs 188.674 GB). The change benefits both the 45-layer target and the separate MTP head; the storage migration is also what lets the 3.9 GB Q4 sidecar load strictly without materializing thousands of expert concatenate expressions (the first per-expert implementation was killed during materialization).

So the product trade is roughly +3% sustained decode for +83 net production LOC, while deleting the HY4-specific duplicate. I agree the 184-line benchmark makes review look much larger than the runtime change. If the production/test split still feels too large, I can narrow compatibility to the current GLM checkpoint layout or remove the benchmark script before further review.

Lazarus-931:
hi @raullenchai, thanks for pr, closing this.

Issue #2229, opened 2026-09-13
0.7.0 @ b5379f69 sweep: Llama-3.2-11B-Vision (mllama) mask off-by-one crash, InternVL 3.5 'internvl' model_type unsupported, aya_vision leaks <|START_RESPONSE|> markers


jrp2014:
### Summary

A 41-model sweep of mlx-community checkpoints on 0.7.0 at `b5379f69` (the current tip at the time bar #2209) turned up three things, each reproduced with a single native command below: a regression that crashes `mllama` (Llama 3.2 Vision) before the first token on any image, InternVL 3.5 conversions that cannot load because their `model_type` is `internvl`, and `aya_vision` returning its response markers in the text. The other 38 models completed and reproduced a previous run's output token for token under `--seed 0` (retained artifacts linked at the end).

### Environment

- mlx-vlm 0.7.0, editable checkout at `b5379f69` (clean)
- mlx 0.32.3.dev20260912+229f5b430 (local build)
- transformers 5.17.0, tokenizers 0.23.2, Python 3.14.7
- macOS 26.6.2, Apple M5 Max, 128 GB unified memory
- Install method: editable checkout (`pip install -e .`)
- Commands run from the repository root with `examples/images/cats.jpg`; trust remote code: no

---

### 1. `mllama`: cross-attention mask is one token longer than the hidden states, so Llama-3.2-11B-Vision crashes before the first token

- `mlx-community/Llama-3.2-11B-Vision-Instruct-4bit`, revision `82f31be9840fa0d4c7e99257fe2e28b59a46df97`

```bash
python -m mlx_vlm.generate \
  --model mlx-community/Llama-3.2-11B-Vision-Instruct-4bit \
  --image examples/images/cats.jpg \
  --prompt "Describe this image in one sentence." \
  --max-tokens 32 --no-verbose
```

```text
  File "mlx_vlm/generate/dispatch.py", line 1081, in stream_generate
  File "mlx_vlm/generate/ar.py", line 497, in generate_step
  File "mlx_vlm/models/mllama/language.py", line 360, in __call__
  File "mlx_vlm/models/mllama/language.py", line 321, in __call__
  File "mlx_vlm/models/mllama/language.py", line 251, in __call__
  File "mlx_vlm/models/mllama/language.py", line 80, in __call__
  File "mlx_vlm/models/base.py", line 418, in scaled_dot_product_attention
[...]

PR #2244, opened 2026-09-13
fix: align Mllama cross-attention masks during chunked prefill

pentaoa:
Mllama image prompts currently fail before the first generated token: `generate_step` chunks the text embeddings but passes the full `(B, 1, T, V)` cross-attention mask into each shorter forward. A seven-token prompt with the default prefill size reproduces the same broadcast error as the Mllama report in #2229.

This slices both cross-attention masks with the text, including the last prefill token, and teaches the batch padding/slicing helpers that their query dimension is axis 2. Image features remain intact. Two subsequent failures exposed by the regression tests are also addressed: reading empty self-attention-style cache entries allocated for cross-attention layers, and Mllama passing the entire cache list instead of its first entry to `create_attention_mask`, which discarded batch padding information.

Validation: `python -m pytest mlx_vlm/tests/test_mllama_prefill.py mlx_vlm/tests/test_generate.py mlx_vlm/tests/test_cache.py -q` → **262 passed** on Apple Silicon, MLX 0.32.2. The nine new tests use tiny random Mllama weights and synthetic pixels, with nonzero cross-attention gates. They compare token/log-probability outputs for one/two images across prefill sizes 1, 2 and 2048, plus unequal-length batched prefill against individual prompts. All six single-prompt cases reproduced the original crash before the fix. Black, isort and autoflake pass for all changed files.

No pretrained weights or benchmark datasets were downloaded, and no large-model performance claims are made. This addresses only the Mllama portion of #2229; the InternVL/Aya reports in that issue remain separate.


Blaizzy:
Thanks so much, @pentaoa! 🙌

This PR has been superseded by #2249, so I'll go ahead and close this one.

I'd love to see you contribute again, feel free to browse the issue tracker and grab something that interests you!

Issue #2242, opened 2026-09-13
bug(glm5-next): temperature-zero output changes with batch size


raullenchai:
## Summary

GLM-5.3-Flash-Next produces different temperature-zero trajectories between singleton and continuously batched serving. A four-request batch can also split into multiple trajectories when requests are admitted dynamically. This reproduces in plain AR, so it is not caused by speculative acceptance or MTP cache rollback.

## Environment

- Apple M3 Ultra, 60 GPU cores, 256 GB; macOS 26.5.2
- MLX 0.32.2
- target: `Vontra/GLM-5.3-Flash-MLX-4bit-MTP` revision `76add2a` (affine Q4)
- source composition: current main conflict resolution for #2206 plus #2231; APC memory checkpoint, no disk
- OpenAI chat server, temperature 0, seed 123, identical prompt, max-num-seqs 4

Prompt: `Produce the integers from 1 through 300 in ascending order, separated only by commas. Do not skip any integer and do not add commentary.`

## Reproduction

1. Prewarm the exact prompt.
2. Send one 128-token AR request.
3. Submit four identical 128-token AR requests simultaneously.
4. Hash `reasoning_content + content`.

Singleton chose `I need to be careful to include all numbers.` The dominant batch trajectory chose `...include every number.` Three batch rows shared that hash; one dynamically admitted row produced a third hash. Repeats preserved the singleton hash and the dominant coalesced-batch hash.

Performance was `27.43 tok/s` singleton and `57.95 aggregate tok/s` for four requests, so serializing would discard a material throughput win. The corresponding MTP run also changed singleton versus batch output, while all four coalesced MTP rows agreed; this is consistent with a backbone batch-numerics boundary.

## Rejected workaround

I forced every B>1/T=1 affine projection through `singleton_quantized_linear`. It did not restore output equality and reduced AR aggregate throughput from `57.95` to `50.07 tok/s` (-13.6%). The remaining candidates include recurrent, MoE, and dynamic cache/batch-state paths.

## Expected

Identical greedy requests should retain the singleton token trajectory when batch size changes, without disabling continuous batching. At minimum, the server should document/telemetry this boundary rather than imply cross-batch byte determinism.

Related qualification context: #2206 (the same divergence is recorded there, with MTP and no-drafter AR controls).

raullenchai:
[...]

PR #2245, opened 2026-09-13
Fix GLM-5.3 cached decode batch invariance

raullenchai:
## Summary

- keep affine-Q4 linear and per-head MultiLinear reduction geometry fixed across B=1 and B>1 short decode blocks using row-parallel Metal kernels
- use the existing fused fixed-row hyperconnection normalization for all supported short-decode batches
- let fused indexed sparse attention read a physical KV-cache backing buffer while honoring its logical length, avoiding a selected-K/V materialization
- preserve native fallbacks for unsupported formats, shapes, devices, and prefill

This is the parallel implementation following the correctness oracle in #2242. It is stacked on #2206. Full-model qualification used #2231 as the target model's quantized lm_head loading dependency.

## Correctness

Environment: Apple M3 Ultra, 60 GPU cores, 256 GB; macOS 26.5.2; MLX 0.32.2; Vontra/GLM-5.3-Flash-MLX-4bit-MTP revision 76add2a (affine Q4).

Starting from one prefetched B=1 cache copied exactly to B=4:

- stock first divergence: layer 0, max_abs 0.000213623046875
- stock first sparse-attention layer: max_abs 0.00390625
- candidate: exact at 45/45 layers, every sparse top-k result, final norm, and all 154,880 logits
- candidate final max_abs: 0; both paths selected token 43799

The diagnostic deliberately separates cached decode arithmetic from prefill and scheduler admission. Long prompts still use native prefill arithmetic, so this does not claim complete singleton-versus-batched-prefill byte identity.

Real-task dogfood with MTP K2/K3 and a 128-token thinking budget:

- coding: correctly repaired the sliding-window invariant and returned an O(n) implementation
- knowledge: accurately qualified the Westphalian-sovereignty myth within the requested word limit
- creative: produced exactly four restrained sentences as requested

## Performance

Target-shape operator medians:

- affine-Q4 MultiLinear, B=4: native 0.290 ms; parallel 0.259 ms; row-loop oracle 0.308 ms
- sparse attention, 8,192 cached positions / 2,048 selected: gather+SDPA 1.060 ms; direct indexed 0.726 ms

End-to-end qualification, reporting the smaller real effect:

- 62-token prompt, 128 output tokens, warm singleton: 3.99 s -> 3.87 s (3.1%)
- same prompt, four concurrent requests: 53.63 -> 53.39 aggregate tok/s (neutral)
- 14,476-token prompt, 64 output tokens: output byte-identical; decode 38.2 -> 38.4 tok/s (noise-level)

The operator microbenchmarks are not presented as whole-model speedups. The production conclusion is cached-decode determinism without a measured concurrency regression.

## Verification

- 376 focused tests passed: test_quantized_ops.py and test_sparse_attention.py
- stacked qualification with #2231: 402 tests passed including GLM-5.3 sanitize tests
- targeted Ruff checks passed
- diff whitespace check passed
- 45-layer real-model cached-decode diagnostic passed with max_abs 0

Known base-branch issue: broader speculative toy-model tests currently fail before this patch's paths because the existing exact_speculative_verify Metal kernel receives a constant-address-space input but declares a device pointer. The identical failures reproduce at the unmodified #2206 baseline and are not introduced here.


Blaizzy:
Thanks @raullenchai 

I'm fixing all of this, it will take a bit of time so please bear with me. But after I'm done will be much easier.

Issue #2227, opened 2026-09-12
Mage-VL-OptiQ-4bit (`mage_vl`) fails to load on 0.7.0: flattened OptiQ `model.*` keys are not remapped by `sanitize` (904 parameters not in model)


jrp2014:
### Summary

`mlx-community/Mage-VL-OptiQ-4bit` fails to load on 0.7.0 (`b5379f69`) with `ValueError: Received 904 parameters not in model`. The checkpoint stores its language-model weights already flattened as `model.layers.*` / `model.embed_tokens.*` / `model.norm.*`, but `Model.sanitize` in `mlx_vlm/models/mage_vl/mage_vl.py` only remaps the hub layout (`model.language_model.*`, `model.visual.*`, `lm_head.*`), so every `model.*` key is passed through unchanged and rejected by `load_weights`. This is the same flattened-OptiQ layout that #2092 (`lfm2_vl`) and #2118 (`muse_glimmer`) added a remap for.

### Environment

- mlx-vlm 0.7.0, editable checkout at `b5379f69` (clean)
- mlx 0.32.3.dev20260912+229f5b430 (local build)
- transformers 5.17.0, tokenizers 0.23.2, Python 3.14.7
- macOS 26.6.2, Apple M5 Max, 128 GB unified memory
- Install method: editable checkout (`pip install -e .`)

### Model

- `mlx-community/Mage-VL-OptiQ-4bit`, revision `bde6c9c7146acff6af09e203245014f19306c5c5`, complete snapshot in the Hugging Face cache (`model.safetensors` + index, `optiq_metadata.json`)
- `config.json` `model_type`: `mage_vl`; `quantization`: OptiQ mixed 4/8-bit affine, group size 64, `model.embed_tokens` and `lm_head` at 8 bits
- Trust remote code: yes (flag passed; same result without it)

### Reproduction

From the mlx-vlm repository root:

```bash
python -m mlx_vlm.generate \
  --model mlx-community/Mage-VL-OptiQ-4bit \
  --revision bde6c9c7146acff6af09e203245014f19306c5c5 \
  --image examples/images/cats.jpg \
  --prompt "Describe this image." \
  --max-tokens 64 --temperature 0.0
```

### Expected behavior

The model loads and describes the image, as the non-OptiQ Mage-VL checkpoints do through the hub-layout remap in `sanitize`.

### Actual behavior

```text
Traceback (most recent call last):
[...]

PR #2247, opened 2026-09-13
mage_vl: remap flattened OptiQ model.* keys in sanitize

Lazarus-931:
Adds the missing `model.*` fallback remap to `mage_vl.sanitize` so flattened OptiQ checkpoints load, matching what `lfm2_vl` (#2092) and `muse_glimmer` (#2118) already do (fixes the first issue of #2227).

regression test as well


lucasnewman (approved):
LGTM

PR #2250, opened 2026-09-14
Add OpenAI Responses terminal bridge

themanamong-stack:
## Why

Provide a small, local terminal bridge so MLX-LM/MLX-VLM workflows can call the OpenAI Responses API without embedding credentials or changing existing serving APIs.

## What changed

- Added `mlx_vlm.openai_bridge`, a stdlib-only client with:
  - `OPENAI_API_KEY` configuration and optional model, base URL, and timeout environment variables.
  - Parameterized JSON request construction for `/v1/responses`.
  - Prompt input from `--prompt` or stdin, plus `--json` for the complete response.
  - Explicit configuration, network, HTTP, and response-shape errors.
- Registered the bridge as `python -m mlx_vlm openai_bridge` and a package entry point.
- Added focused request/configuration tests.
- Documented localhost usage and a loopback-only SSH tunnel for a remote OpenAI-compatible proxy. SSH is not needed for a localhost-only workflow.

## Validation

- Python 3.11 compilation checks passed.
- Dependency-free request/configuration smoke checks passed, including `Authorization: Bearer <OPENAI_API_KEY>`.
- `git diff --check` passed.
- The targeted pytest file could not run because `pytest` is not installed in the environment; full package CLI execution also requires the missing `mlx` dependency.


Blaizzy:
Thanks for putting this together, @themanamong-stack, this is a cool idea, could you share more about the use case you have?

After reviewing it, I think this fits better as part of the existing `generate` CLI rather than as a standalone top-level command. Keeping the command surface small and routing prompt-based inference through `generate` keeps the entry points consistent for users.

So I'm going to close this PR in favor of folding that functionality into `generate`. That said, I'd love to see you contribute again, feel free to open an issue describing the OpenAI-bridge behavior you'd like, or pick up another open issue.

Issue #2254, opened 2026-09-14
glm5_next: remaining KDA decode kernel-launch overhead after #2127 (spec + measurements for a fused/compiled step)


freddyhaddad:
> **Update (2026-09-15):** #2255 is now **bit-exact** against eager decode. The only op that rounded differently under `mx.compile` was `mx.sigmoid`; details and the per-op table are in [this comment](https://github.com/Blaizzy/mlx-vlm/issues/2254#issuecomment-5670004915). The "Exactness" section below describes the first version and is kept for the record.

## Summary

After #2127, a cached GLM-5-Next KDA decode step still launches its post-projection chain as many small, separately dispatched kernels in each of the 34 linear-attention layers. Compiling the whole cached S=1 step with `mx.compile` measurably speeds up end-to-end decode on GLM-5.3-Flash (**+1.7% to +2.1% at B=1 on uncontended runs, +1.6% aggregate at B=4**), and it works with batched, left-padded and ragged continuous-batch caches.

The gain is modest and **not bit-exact** (details below), so I'd like maintainer input on where this belongs before anyone invests further. This follows @Lazarus-931's invitation on #2105 to open an issue with a spec. A working opt-in implementation is in #2255 so there's something concrete to review, measure or reject.

## Background

- #2105 (@avlp12) folded the KDA decode chain into one `mx.fast.metal_kernel` per layer: bit-exact, +7.8% end-to-end decode on GLM-5.3-Flash at the time. It was closed because it would conflict with the shared kernel components introduced in #2127.
- #2127 (merged 2026-09-09) refactored `Glm5NextLinearAttention` onto shared pieces: `mx.fast.rms_norm` for q/k, the shared `gated_delta_update`/`gated_delta_kernel`, and a cache that owns the conv window and recurrent state through `update_window`/`update_recurrent`. That refactor already recovered a large part of what #2105 targeted.
- #2107 (indexer fast path) is about the DSA layers, not KDA; it is orthogonal to this.

## What is left after #2127

[...]

PR #2255, opened 2026-09-14
glm5_next: opt-in compiled cached KDA decode step (bit-exact, +2% decode)

freddyhaddad:
Refs #2254 (design discussion, full measurements, and alternatives)

## What

This adds an opt-in env flag, `MLX_VLM_GLM5_COMPILED_KDA_DECODE=1` (default **off**). When it is set, a cached single-token GLM-5-Next KDA step runs as one `mx.compile` graph instead of about 30 separately launched kernels per layer. The compiled step is **bit-identical to eager decode**: outputs, conv window and recurrent state.

With the flag unset, nothing changes: the module-level check short-circuits before any new code runs.

## Why

After #2127, the cached S=1 step in each of the 34 KDA layers still launches its projections, short conv, norms, gates, delta update and gated output norm as separate kernels. Compiling that chain gives, on GLM-5.3-Flash (M3 Ultra, mlx 0.32.2, alternating eager/compiled per prompt, 3 rounds, greedy):

| workload | eager | compiled | speedup |
|---|---|---|---|
| B=1 prose (256 tok) | 28.7 tok/s | 29.3 | +2.1% |
| B=1 code (256 tok) | 28.6 | 29.2 | +2.0% |
| B=1 Lebanese Arabic (200 tok) | 28.8 | 29.4 | +2.0% |
| B=1 short assistant answer (120 tok) | 28.9 | 29.5 | +2.2% |
| B=4 aggregate (4 × 200 tok) | 61.6 | 62.1 | +0.9% (+1.6% in an earlier run) |

Every one of the 15 alternating pairs was faster compiled. **Generated tokens were identical in every pair**, all 832 B=1 tokens and all four B=4 texts. Method and caveats are in the issue.

## How

- **`Glm5NextLinearAttention._decode_step`** is a pure mirror of the cached S=1 path through `__call__`, `ShortConv1d.__call__`, `ArraysCache.update_window` and `ArraysCache.update_recurrent`, in the same order. It takes `(x, conv_state, recurrent_state, mask, lengths)` and returns `(output, new_conv_state, new_recurrent_state)` instead of writing into the cache, so it can be compiled.
- **`_compiled_decode`** keeps one compiled function per `(mask is None, lengths is None)` variant, the only ways the traced graph structure can change. `mx.compile` then retraces only when a new batch size appears, not every step. It writes the returned states back into `cache.cache` and calls `cache.advance(1)`, the same as the eager path.
- **The compiled step engages only when all of these hold:** the flag is set, `S == 1`, a cache is present with both states initialised, the module is not training, and `cache.history_capacity == 0`. It declines during prefill, on the first token before the cache holds state, and while a speculative transaction is recording history. MTP and DFlash verify keep the eager path, which records intermediate states.
- **Batches are supported,** not only B=1:
  - left-padded batches: the mask is passed as a compiled input
  - ragged continuous batches: `cache.lengths` is passed, and the window update uses the same `take_along_axis` selection as `update_window`

## Exactness

The first version of this PR was not bit-exact. The cause turned out to be a single op. Under `mx.compile`, `mx.sigmoid` lowers its `exp` to Metal's fast approximation, while the prebuilt sigmoid kernel uses the precise one; every other op in the chain (`exp`, `rsqrt`, division, multiply-add, `where`/`abs`, the L2-norm chain, and the already-compiled `compute_g_safe` and `nn.silu`) is bit-exact when compiled. `mx.exp` stays precise under compilation, so `_sigmoid_exact` spells out the prebuilt kernel's formula

```python
y = 1 / (1 + mx.exp(mx.abs(x)))
return mx.where(x < 0, y, 1 - y)
```

and matches the prebuilt kernel bit for bit in fp32 and bf16 (0 of 2^20 random values differ, for either dtype). The compiled step uses it for the output gate, and passes `beta=_sigmoid_exact(b)` into `gated_delta_update` through a new optional `beta` argument (default `None`, in which case the function computes `mx.sigmoid(b)` as before, so eager callers are untouched).

With that, the compiled step reproduces eager decode exactly: over 32 cached steps, outputs, conv window and fp32 recurrent state are identical in fp32 and bf16, quantised and unquantised, for B=1, left-padded B=3 and ragged B=3. On the real model, greedy tokens are identical for all prompts above.

## Tests

- **New: `test_glm5_next_compiled_kda_decode_matches_eager`.** It runs a quantised random KDA layer for 8 cached decode steps after a 6-token prefill, in three configurations: B=1 with no mask, B=3 left-padded, and B=3 ragged with `cache.lengths` changing per step. It asserts `array_equal` on outputs, conv window and recurrent state, and that exactly three compiled variants were created, which proves the compiled path actually ran.
- **Existing suites pass both with the flag unset and with `MLX_VLM_GLM5_COMPILED_KDA_DECODE=1`:**
  - `test_models.py -k glm5`: 79 passed
  - `test_speculative.py`, `test_speculative_transactions.py`, `test_cache.py`: 989 passed
- **Formatting:** black 26.3.1 and isort 5.13.2 (profile black) clean on the touched files.

## Benchmark script

<details>
<summary>decode_ab.py (alternating eager/compiled A/B on a loaded model, token comparison)</summary>

```python
"""Real GLM-5.3-Flash decode A/B on mlx-vlm main: eager vs compiled KDA decode.

One process, one loaded model; COMPILED_KDA_DECODE is read at call time, so the
two arms alternate per prompt (order flipped each round) under the same load.
B=1 via stream_generate, B=4 via batch_generate. Greedy; tokens compared.
"""
import os, statistics

import mlx.core as mx
import mlx_vlm.models.glm5_next.language as L
from mlx_vlm import load
from mlx_vlm.generate.ar import batch_generate
from mlx_vlm.generate.dispatch import stream_generate

model, processor = load(os.environ["MODEL"])
PROMPTS = {
    "prose": ("Explain in about 200 words why memory bandwidth limits token generation speed.", 256),
    "code": ("Write a Python function that merges two sorted lists without using sorted(). Include a docstring.", 256),
    "arabic": ("اكتبلي بالعربي اللبناني فقرة قصيرة لصاحبي بتقلو إني رح إتأخر نص ساعة عالعشا وليش.", 200),
    "assistant": ("In two sentences, is there a conflict between a 3pm supplier call and a 3:30pm school pickup?", 120),
}
fmt = lambda p: apply_chat_template(processor, model.config, p, num_images=0, enable_thinking=False)

def one(prompt, n):
    toks, last = [], None
    for r in stream_generate(model, processor, fmt(prompt), max_tokens=n, temperature=0.0):
        if r.token is not None:
            toks.append(int(r.token))
        last = r
    return last.generation_tps, toks

def batch(n):
    r = batch_generate(model, processor, prompts=[fmt(p) for p, _ in PROMPTS.values()], max_tokens=n)
    return r.stats.generation_tps, r.texts

def arm(compiled, fn, *a):
    L.COMPILED_KDA_DECODE = compiled
    return fn(*a)

for compiled in (False, True):  # warm both arms
    arm(compiled, one, PROMPTS["assistant"][0], 16); arm(compiled, batch, 16)

res = {k: {"off": [], "on": []} for k in [*PROMPTS, "batch4"]}
for rnd in range(3):
    order = (False, True) if rnd % 2 == 0 else (True, False)
    for name, (prompt, n) in PROMPTS.items():
        for c in order:
            res[name]["on" if c else "off"].append(arm(c, one, prompt, n))
    for c in order:
        res["batch4"]["on" if c else "off"].append(arm(c, batch, 200))
for name, r in res.items():
    off, on = statistics.median(t for t, _ in r["off"]), statistics.median(t for t, _ in r["on"])
    same = all(a == b for (_, a), (_, b) in zip(r["off"], r["on"]))
    print(f"{name:9} {off:6.1f} {on:6.1f} {on / off:.3f}x  identical output: {same}")
```

</details>

## Scope and non-goals

- Only `glm5_next`, plus the optional `beta` argument on the shared `gated_delta_update`. `qwen3_next`, `kimi_linear` and others could use the same approach, but I haven't measured them and didn't want to widen the change before hearing whether this direction is wanted at all.
- No change to prefill, the DSA or MLA layers, the indexer (see #2107), or the shared `gated_delta` kernel itself.
- If folding this into the shared kernel from #2127 is preferred (benefits every model that uses it), I'm happy to close this in favour of that.


freddyhaddad:
Pushed a second commit that makes the compiled step **bit-exact** with eager decode; the PR description is updated to match.

The non-exactness had one cause: `mx.sigmoid` under `mx.compile` uses Metal's fast `exp`, while the prebuilt sigmoid kernel is precise. Everything else in the chain is already bit-exact when compiled (I checked `exp`, `rsqrt`, division, fused multiply-add, `where`/`abs`, the L2-norm chain on 2^20 values each: 0 differing). Since `mx.exp` stays precise under compilation, `_sigmoid_exact` spells out the prebuilt kernel's formula and matches it bit for bit in fp32 and bf16.

Changes in the commit:
- `_sigmoid_exact` in `glm5_next/language.py`, used for the output gate in the compiled step
- optional `beta` argument on `gated_delta_update` (default `None` → `mx.sigmoid(b)` as before) so the compiled step can pass the exact beta
- the test now asserts `array_equal` on outputs as well as states

Re-measured on the real model with the exact step: +2.0–2.2% at B=1, +0.9% at B=4, and the generated tokens are identical to eager for every prompt (832 B=1 tokens, four B=4 texts). Suites still pass with the flag off and on (79 + 989).

PR #2256, opened 2026-09-14
Fix Qwen3-Omni deepstack injection during chunked prefill

### LINKED ISSUE #2099: Qwen3-Omni: GPU Address Fault when audio and video are passed together (each works alone)
Passing **both** `--audio` and `--video` to `Qwen3-Omni-30B-A3B-Instruct-4bit` crashes with a Metal page fault. Either modality alone works.

## Repro

```bash
python -m mlx_vlm.generate \
  --model mlx-community/Qwen3-Omni-30B-A3B-Instruct-4bit \
  --video clip.mov --audio clip.wav \
  --prompt "What instrument is played and how does it sound?" --max-tokens 40
```

```
File "mlx_vlm/models/qwen3_omni_moe/language.py", line 651, in __call__
    out = self.model(...)
File "mlx_vlm/models/qwen3_omni_moe/language.py", line 266, in __call__
    mx.eval(h)
RuntimeError: [METAL] Command buffer execution failed: Caused GPU Address Fault Error
(0000000b:kIOGPUCommandBufferCallbackErrorPageFault)
```

The clip is an 8.17s 2604x2160 HEVC `.mov` from an iPhone; the audio is its own track extracted to 16 kHz mono WAV (`ffmpeg -i clip.mov -vn -ac 1 -ar 16000 clip.wav`). Nothing about it seems unusual, and the same crash occurs with the audio passed as a separate file or left inside the video.

## What works and what doesn't

| input | result |
|---|---|
| text only | works |
| audio only | works |
| video only (audio track stripped) | works |
| **audio + video** | **GPU Address Fault** |

`line 266` is `mx.eval(h)` inside the decoder loop, so that is where the lazy graph is forced rather than where the bad index is built.

## The processor builds correct inputs

Calling the processor directly with both modalities produces what looks like a consistent set, which suggests the fault is downstream of input preparation:

```python
out = processor(text=[text], videos=[frames], audio=[audio],
                return_tensors="pt", padding=True)
# input_features        (1, 128, 813)
# feature_attention_mask (1, 130107)
# pixel_values_videos   (24768, 1536)
# video_grid_thw        (1, 3)
# audio placeholders in input_ids: 106
# video placeholders in input_ids: 6192
```

Both feature sets are present and both placeholder counts are non-zero, so `get_placeholder_mask` has something to scatter into. I did not bisect further into the model to find which index goes out of bounds.

## Two related observations

**`use_audio_in_video=True` raises `StopIteration`.** The flag that is supposed to govern interleaving fails before reaching the model:

```python
processor(text=[text], videos=[frames], audio=[audio], use_audio_in_video=True, ...)
# StopIteration
```

**`prepare_omni_inputs` appears to be unreachable and untested.** `mlx_vlm/models/qwen3_omni_moe/omni_utils.py:25` is the helper for exactly this case, but `grep -rn prepare_omni_inputs` finds only its own definition — nothing in the package calls it. It also passes `videos=` the raw *paths* from the conversation, while the processor expects `(T, C, H, W)` arrays:

```
ValueError: Expected video as (T, C, H, W), got shape ().
```

So it cannot work as written. If it is the intended entry point for Omni A/V, wiring it up would need the video loaded first; if it is dead code, removing it would save the next person the detour.

## Environment

- mlx-vlm 0.6.15
- mlx 0.32.1
- Python 3.14.7
- macOS 26.5.2 (25F84), arm64, 64 GB
- `mlx-community/Qwen3-Omni-30B-A3B-Instruct-4bit`

Happy to run further probes on this machine — I have the model loaded and can bisect inside the model if it would help to know which tensor is out of range.

kiarina:
## Summary

Qwen3-Omni aborts with a Metal GPU address fault, or decodes `!!!!…`, once an image/video prompt spans more than one prefill chunk — for example a 28-frame video (~3.6k prompt tokens). Fixes #2099.

## Cause

Since the deepstack embeds reach the decoder (#1635), chunked prefill in `generate()` passes the full-prompt `visual_pos_masks` and `deepstack_visual_embeds` to every chunk and to the final one-token step. `Qwen3VLMoEModel._deepstack_process` takes `np.where(visual_pos_masks)` over the whole prompt and scatters into the current chunk's hidden states, so the indices run past the end of the chunk.

Logging every `_deepstack_process` call for a 3611-token video prompt on `main`:

| hidden length | mask length | visual tokens | embed rows |
|---:|---:|---:|---:|
| 2048 | 3611 | 3584 | 3584 |
| 1562 | 3611 | 3584 | 3584 |
| 1 | 3611 | 3584 | 3584 |

Short image prompts only overrun by one row (hidden 83 vs mask 84), which is why they looked fine.

## Fix

Window both inputs to the current chunk in `LanguageModel.__call__`, right after the existing `position_ids` windowing, the same way the Qwen3-VL language model already does:

- slice `visual_pos_masks` to `[offset, offset + len)`
- take the embed rows for the visual tokens inside that window (skipping the ones consumed by earlier chunks)
- per-row cache offsets in batch mode are handled; chunks and decode steps without visual tokens skip deepstack

## Testing

- New `test_chunked_prefill_windows_deepstack_inputs`: a two-chunk prefill split inside the image span must match a single prefill. It fails on `main` and passes with this change.
- `pytest mlx_vlm/tests/test_qwen3_omni_moe.py mlx_vlm/tests/test_models.py -k "omni or qwen3_vl"`: 17 passed.
- On an M4 Max (128 GB), `mlx-community/Qwen3-Omni-30B-A3B-Instruct-4bit`, mlx 0.32.2, default prefill step size:

| input | prompt tokens | before | after |
|---|---:|---|---|
| video only | 3603 | `!!!!…` / GPU address fault | "The video starts with a close-up of a pink pixelated bear, then zooms out to reveal the bear walking along a river in a colorful village." |
| image only | 81 | OK | "This is a pixel art image of a pink, cat-like character with large eyes, wearing a red bow tie." |
| video + audio (#2099) | 4004 | GPU address fault | quotes the song's lyrics and describes the video |
| image + video | 3676 | GPU address fault | "Yes, the pink bear character appears in the video." |

## Not in this PR

Two separate issues showed up while testing on mlx 0.32.2; the device runs above worked around them in the test script only:

- The image + video joint path in `qwen3_omni_moe/thinker.py` calls the one-argument `mx.where(mask)[0]` and `mx.scatter`, which mlx 0.32 does not provide (`TypeError: where(): incompatible function arguments`).
- `load_audio` resamples stereo input along the channel axis before downmixing, so a 48 kHz stereo WAV reaches the audio tower un-resampled and the transcript comes out garbled. Downmixing to mono first avoids it.

Happy to send follow-ups for either.


lucasnewman:
@kiarina Thanks for the contribution! This won't work properly for padded batches as-is -- I put up a fix to resolve it in https://github.com/Blaizzy/mlx-vlm/pull/2265.

PR #2257, opened 2026-09-14
Fix Qwen3-Omni image+video deepstack join on current mlx

kiarina:
## Summary

A Qwen3-Omni prompt with both an image and a video raises

```
TypeError: where(): incompatible function arguments
```

in `Thinker.get_input_embeddings`, and even where it ran it placed the video deepstack rows at the wrong positions.

## Cause

The block that joins the image and video deepstack embeds (`qwen3_omni_moe/thinker.py`) had two problems:

1. It calls the one-argument `mx.where(mask)[0]` and a free-function `mx.scatter`, neither of which mlx provides (tested on mlx 0.32.2).
2. `image_indices` / `video_indices` are positions inside the joint visual sequence, but they were also used to `mx.take` rows from `img_embed` / `vid_embed`, which only hold that modality's tokens. For a prompt with the image first, the video rows were read from offsets past the end of `vid_embed`.

## Fix

Compute the joint positions with `np.where` (as `masked_scatter` in the same file already does) and assign each modality's whole embeds at its own positions, matching the reference implementation:

```python
embed_joint[image_joint] = img_embed
embed_joint[video_joint] = vid_embed
```

## Testing

- New `test_image_and_video_deepstack_embeds_are_joined_by_position`: with one image and one video in the prompt, the joint deepstack embeds must equal the image-only rows at the image positions and the video-only rows at the video positions. It fails on `main` with the `TypeError` and passes with this change.
- `pytest mlx_vlm/tests/test_qwen3_omni_moe.py mlx_vlm/tests/test_models.py -k "omni or qwen3_vl"`: 17 passed.
- On an M4 Max (128 GB), `mlx-community/Qwen3-Omni-30B-A3B-Instruct-4bit`, mlx 0.32.2, an image plus a 28-frame video (3676 prompt tokens) now answers "Yes, the pink pixelated bear character appears in the video." Image-only and video-only outputs are unchanged.

A prompt this long also needs #2256 (deepstack windowing during chunked prefill); the device run above had both applied. The two changes touch different files and are independent otherwise.


lucasnewman:
@kiarina Thanks for the contribution! I'm fixing this in a slightly different way here: https://github.com/Blaizzy/mlx-vlm/pull/2287

PR #2258, opened 2026-09-14
Downmix stereo audio before resampling in load_audio

kiarina:
## Summary

`load_audio` returns audio at the wrong sample rate for any stereo file whose rate differs from the target. A 48 kHz stereo WAV loaded with `sr=16000` comes back as 48,000 samples per second of audio instead of 16,000, so models receive audio three times too long and slowed down.

## Cause

`read_audio` returns `(samples, channels)`. `load_audio` passes that straight to `mlx_audio.utils.resample_audio`, which resamples along `axis=-1` — the channel axis — and only downmixes afterwards:

```python
>>> resample_audio(np.zeros((48000, 2)), 48000, 16000).shape
(48000, 1)
```

Mono input takes the correct path, which is why this goes unnoticed; the existing test uses a stereo array already at the target rate, so it never resamples.

## Fix

Downmix to mono first, then resample. Mono input and same-rate input behave exactly as before.

## Testing

- New `test_load_audio_downmixes_stereo_before_resampling`: a fake 48 kHz stereo read with `sr=16000` must return 16,000 samples equal to resampling the mono downmix. It fails on `main` (`(48000,) == (16000,)`) and passes with this change.
- `pytest mlx_vlm/tests/test_audio_utils.py mlx_vlm/tests/test_processors.py mlx_vlm/tests/test_server_audio.py`: 185 passed, 2 skipped.
- On an M4 Max (128 GB), `mlx-community/Qwen3-Omni-30B-A3B-Instruct-4bit`, a 30 s 48 kHz stereo song passed by path and asked to quote a lyric:
  - before: "Shh, don't you shh" (168.6 s)
  - after: 「加速する世界の中から君の声が聞こえてくる」, the actual first line (1.2 s)


lucasnewman (approved):
Good catch, thank you!

lucasnewman (approved):

PR #2266, opened 2026-09-16
fix: correct LFM2-VL/LFM2.5-VL vision tower and preprocessing

ykhrustalev:
**Problem**
The LFM2-VL vision tower dropped `pixel_attention_mask` and resized SigLIP2 position embeddings with bicubic instead of bilinear + antialias. Measured against the reference `transformers` implementation in fp32, logits were off by 19–86% relative error and next-token top-1 agreement was 40–65%.

**Solution**
- Masks padded patch rows in the vision tower from `pixel_attention_mask`, the way SigLIP2 does
- Resizes position embeddings with bilinear + antialias instead of bicubic
- Honors the checkpoint's `resample` filter, derives `max_num_patches`, and pins `do_resize` on, matching the official image processor
- Merges `preprocessor_config.json` and `processor_config.json`, since neither is a superset of the other
- Adds `encode_images` and a top-level `make_cache`, repairing the inert vision-feature cache and the conv/attention cache split
- Honors `rope_parameters` and untied `lm_head`, and splits a flat image list across batched prompts

**Testing**
- fp32 parity against `transformers` for 450M/1.6B/3B of both generations: byte-identical `input_ids`, logits within `1e-4`, ≥99.6% top-1 agreement (was 40–65%)
- 40 of the 41 public LFM2-VL/LFM2.5-VL safetensors repos under `LiquidAI` and `mlx-community` load and generate (the 41st, `mlx-community/LFM2-VL-450M-6bit`, is an empty repo); `mlx-community/LFM2.5-VL-450M-{6,8}bit` previously crashed on `max_num_patches`
- 12 new regression tests


Blaizzy (approved):
LGTM, thanks!

PR #2268, opened 2026-09-16
fix(speculative): allow chunked GLM MTP prefill

raullenchai:
## Why

GLM-5.3 MTP requests capture target hidden states but do not use shared-KV capture. The current policy requires both flags, so MTP disables chunked prefill and materializes a long prompt in one target forward. A 32K prompt on a 256 GB M3 Ultra consequently fails with Metal out-of-memory.

## Change

For GLM MTP, permit chunked prefill whenever `return_hidden` is requested. Keep the no-hidden case blocked and leave other draft kinds unchanged.

## Evidence

- Focused policy test passes.
- Before: a 32K-token GLM MTP request takes the unchunked path and fails with `kIOGPUCommandBufferCallbackErrorOutOfMemory`.
- After policy change: the request enters the chunk loop; downstream integration still needs to retain the full prompt token IDs alongside accumulated hidden chunks, which is handled in the serving integration rather than this model policy.

## Scope

This PR only corrects the model-owned chunked-prefill eligibility decision. It does not change prefill chunk size, cache layout, or speculative decode math.


Lazarus-931:
deferring until #2206

PR #2271, opened 2026-09-16
Bound unused cache graphs during single-request decoding

### LINKED ISSUE #2042: DeepSeek-V4 retains unused rotating-cache value graphs during decode

Blaizzy:
Long DeepSeek-V4 generations retain a growing graph in the rotating cache's unused zero-width values, even after logits are evaluated. The batch scheduler already periodically evaluates cache state, but standard single-request decoding lacked that protection.

Evaluate prompt-cache state every 50 generated tokens in `generate_step`, sharing `DEFAULT_CACHE_EVAL_INTERVAL` with the batch scheduler. This bounds unused cache graphs without changing generated outputs. The existing batch environment override and allocator-cache clearing cadence are preserved.

This change covers standard single-request decoding; direct cache/model calls and speculative decoding loops remain outside its scope.

## Validation

- **356 tests passed** across `test_generate.py`, `test_generation_defaults.py`, and `test_cache.py`.
- A focused `generate_step` regression check covers 49, 50, and 100 generated tokens: no periodic cache evaluation before the threshold, then one and two cache-state evaluations, respectively.
- Black, isort, and `git diff --check` passed.

## Retained values graph edges

Measured after 512 generated tokens from a 32-token prompt using a small DeepSeek-V4 model on Apple M5 Max with MLX 0.32.2. Counts refer to the unused rotating-cache values graph in each attention path.

| Attention path | Before | After |
|---|---:|---:|
| Local attention | 3,598 | 84 |
| Compressed attention | 3,598 | 84 |
| Sparse-compressed attention | 3,598 | 84 |

Periodic evaluation clears the graph every 50 generated tokens. This snapshot was taken 12 tokens after the last evaluation, so the remaining 84 edges represent growth within one interval. All eight before/after pairs produced these counts and identical generated tokens.

Closes #2042


lucasnewman (approved):

Lazarus-931 (approved):
i tested this with the fix for dsv4, and it well!

i guess a follow up would be actually make values optional so v4 never creates it

but that means we'd have to add a bunch of `None`'s and touch on a lot of sensitive surfaces

Blaizzy:
> i guess a follow up would be actually make values optional so v4 never creates it
> but that means we'd have to add a bunch of None's and touch on a lot of sensitive surfaces

Thanks Alazar! 

This fix should be more than enough, it's actually a bug that we didn't evaluate the single request path the same way we did batch.

PR #2277, opened 2026-09-16
deepseek_v4: restore five weights dropped during checkpoint key normalization

Lazarus-931:
**What's wrong.** DeepSeek-V4 returns the same token — `,` — whatever you prompt it with. Nothing errors and the speed looks fine, so it reads like a quality problem rather than a broken load.

**Scope.** Only checkpoints written in the Transformers namespace are affected, which is what the MLX conversions publish; the official `deepseek-ai/DeepSeek-V4-Flash-0731` spells the hyper-connection tensors with underscores and carries no quantisation scales, so it loads correctly and hid this.

**How this surfaced.** While benchmarking V4 against a separate C implementation of the same architecture, the other engine predicted the text's actual next token where mlx-vlm returned `,` at every context and batch size; comparing loaded parameters against the safetensors files showed five tensors never arriving.

**Reproducing on main.** Load `mlx-community/DeepSeek-V4-Flash-4bit`, prefill any prompt, and the argmax is `14` (`,`) at every context from 512 to 8192 and every batch from 1 to 16. Loading with `strict=True` fails and names the five tensors; the default non-strict path drops them without a word.

**The issue.** `normalize_checkpoint_key` strips the leading `model.` from every key and restores it only for `embed_tokens.weight`, while `top_remap` recognises only the older underscore spellings such as `hc_head_base` — so `embed_tokens.scales`, `embed_tokens.biases` and `hc_head.{base,fn,scale}` are lost. `hc_head` collapses the hyper-connection copies into the final hidden state, so zeroing it leaves the output mathematically unable to depend on the input.

**The fix.** Five further `top_remap` entries mapping the Transformers spellings to the model's parameter names; the loop only fires for keys that are present, so checkpoints that already loaded are untouched.

**Verification.** All five parameters now equal the checkpoint (`hc_head.base` std 0 → 2.272, `embed_tokens.scales` std 0.0049 → 0.0366), and predictions go from 0 of 4 sampled positions correct with the target never in the top five, to 2 of 4 exact with the target in the top five at 3 of 4 — the top five finally varying by context instead of repeating the same punctuation. The added test covers both spellings: its Transformers subtest fails without this change, its official subtest passes either way. Suite: 4145 passed, 9 skipped.


Blaizzy (approved):
LGTM, thanks!

mlx_vlm/models/deepseek_v4/language.py
```diff
@@ -1343,6 +1343,17 @@ def sanitize(self, weights: Dict[str, mx.array]) -> Dict[str, mx.array]:
             "hc_head_fn": "model.hc_head.fn",
             "hc_head_base": "model.hc_head.base",
             "hc_head_scale": "model.hc_head.scale",
+            # `normalize_checkpoint_key` drops the leading `model.` from every
+            # key and only restores it for the embedding weight, so a
+            # checkpoint already written in the Transformers namespace arrives
+            # here with these five stripped and nothing puts them back.
```
Blaizzy, line 1349: Follow up PR removing comments please :)

PR #2280, opened 2026-09-16
lfm2_vl: inherit the text LFM2 language model to enable speculative decoding

tugot17:
## What

Enables speculative decoding (DFlash / DSpark drafters) for `lfm2_vl` targets. One file, net −32 lines.

The LFM2-VL language tower wraps the same shared `Lfm2Model` as the text `lfm2` model, but through its own thin `LanguageModel` that swallowed `capture_layer_ids` / `speculative_verify` in `**kwargs` and returned no `hidden_states` — so drafters validated against text LFM2 targets but could not pair with the VL model. This PR makes the VL `LanguageModel` subclass the text one, inheriting hidden-state tap capture, exact speculative verification, and conv-state cache rollback. The only remaining override is `sanitize` (VL checkpoints carry the `language_model.` weight prefix).

No changes to the speculative loop, drafters, or any other model: `validate_dflash_target` and `bind()` were already VL-shape-aware, and the generate paths call `model.language_model(...)` with keyword args, so the old `mask` positional is safely absorbed.

## Validation

- lfm2/DSpark speculative tests pass (`test_speculative.py -k "lfm2 or dspark"`: 21 passed on this branch).
- Tiny random `lfm2_vl` model: captured taps have the right shapes; `speculative_verify=True` logits match one-token-at-a-time decode exactly (max |diff| = 0.0); rollback after partial acceptance leaves the cache usable; `validate_dflash_target` + `bind()` accept the VL target; the plain text-only forward is unchanged.
- End to end with `LFM2.5-VL-3B` + a 4-layer LFM2.5-VL DSpark drafter (M4 Max, greedy): outputs byte-identical with and without the drafter, at both the default verify width and `--draft-block-size 10`.
- Full MMSpec `test` split (600 samples / 723 turns, image-first multi-turn protocol, greedy, no prompt cache): accept length 3.74, decode 163.0 tok/s vs 63.9 baseline (2.55× decode, 1.77× end-to-end), identical token counts per subset across arms, zero errors.

The VL drafter used for validation is available gated at `tugot17/mango-9` (request access) ahead of its official release; happy to add a README example alongside the existing DSpark section once the official checkpoint is up.


Blaizzy:
Thanks Piotr, could you please sign the commit and we can merge right after

tugot17:
Signed — the branch is now a single verified commit rebased on current main, tests pass. Ready to merge.

PR #2289, opened 2026-09-17
Respect generation config sampling defaults

### LINKED ISSUE #2228: mlx_vlm.generate CLI and generate() ignore the checkpoint's generation_config.json sampling defaults that the server honours since #1337
### Summary

#1331 / #1337 made the server apply a checkpoint's declared `temperature`, `top_p`, `top_k` and `do_sample` when a request leaves them unset. The values are loaded onto `model.config` for every entry point (`mlx_vlm/utils.py`, `GENERATION_CONFIG_DEFAULT_KEYS` and `apply_generation_config_defaults`), but only `mlx_vlm/server/request_normalization.py` reads them back. `python -m mlx_vlm.generate` fills its parser defaults from `DEFAULT_TEMPERATURE = 0.0` / `DEFAULT_TOP_P = 1.0` (`mlx_vlm/generate/dispatch.py`, `mlx_vlm/generate/common.py`), and `generate()` / `stream_generate()` default their `temperature` argument the same way (`mlx_vlm/generate/ar.py`). So the same checkpoint samples at its authors' settings through the server and decodes greedily from the CLI, with no flag passed in either case.

### Environment

- mlx-vlm 0.7.0, editable checkout at `b5379f69` (clean)
- mlx 0.32.3.dev20260912+229f5b430 (local build)
- transformers 5.17.0, tokenizers 0.23.2, Python 3.14.7
- macOS 26.6.2, Apple M5 Max, 128 GB unified memory
- Install method: editable checkout (`pip install -e .`)

### Model

- `mlx-community/gemma-4-e4b-it-4bit`, revision `475b9088d29754a3379866cf5aeb6b41acd313c2`, Hugging Face cache
- `generation_config.json` declares `"do_sample": true, "temperature": 1.0, "top_k": 64, "top_p": 0.95` (the same family as the #1331 repro)
- Trust remote code: no

### Reproduction

The fields are loaded:

```python
from mlx_vlm import load
model, _ = load("mlx-community/gemma-4-e4b-it-4bit")
print({k: getattr(model.config, k) for k in ("do_sample", "temperature", "top_p", "top_k")})
# {'do_sample': True, 'temperature': 1.0, 'top_p': 0.95, 'top_k': 64}
```

The CLI does not use them. From the repository root, no sampling flags, two seeds:

```bash
for s in 1 2; do
  python -m mlx_vlm.generate \
    --model mlx-community/gemma-4-e4b-it-4bit \
    --image examples/images/cats.jpg \
    --prompt "Describe this image in one sentence." \
    --max-tokens 48 --seed $s --no-verbose
done
```

Output, identical for both seeds (greedy):

```text
Two tabby cats are relaxing on a bright pink surface, with a remote control nearby.
Two tabby cats are relaxing on a bright pink surface, with a remote control nearby.
```

Passing the declared values explicitly (`--temperature 1.0 --top-p 0.95 --top-k 64`) with the same two seeds gives two different sentences, so sampling would have applied had the defaults been honoured:

```text
Two cats, one striped and one tabby, are relaxing on a bright pink bedspread.
Two tabby cats are resting on a bright pink surface, accompanied by a remote control.
```

### Expected behavior

With no sampling flag on the command line (or no `temperature` / `top_p` / `top_k` kwarg to `generate()`), the CLI and API use the checkpoint's declared values, falling back to `DEFAULT_TEMPERATURE` / `DEFAULT_TOP_P` / `0` only where the checkpoint declares nothing, and `do_sample: false` forces greedy: the precedence `request_normalization.py` already implements for the server.

### Actual behavior

The CLI and API always decode greedily unless a flag is passed, whatever the checkpoint declares. `model.config.temperature` and friends are populated and never read outside the server.

### Why it matters

In a sweep of 38 cached mlx-community checkpoints, 17 declare `do_sample: true` with temperatures from 0.15 to 1.0 (Gemma 4, Qwen3-VL / 3.5 / 3.8, GLM-4.6V, MiniCPM, Kimi-VL, Devstral, ERNIE, Muse Glimmer, Ornith, North-Micro). Greedy versus declared decoding changed the outcome in both directions: one model that loops greedily completes cleanly at its declared temperature, another that completes greedily degenerates at its declared temperature. Users comparing CLI output with server output, or with the model card's reference behaviour, currently see different decoders without having asked for either.

### Proposed fix

Resolve the four fields once, after `load()`, in the CLI (`dispatch.py`, by making the parser defaults `None` and filling from `model.config`) and in `generate()` / `stream_generate()` (`ar.py`, `temperature: float | None = None` resolved the same way), reusing the server's `_model_config_field_or_default` precedence: explicit value, else `model.config` field, else the existing default; `do_sample: false` forces `temperature = 0.0`. A test in the shape of the #1337 `test_utils.py` additions, loading a config that declares sampling and asserting the resolved CLI arguments, would pin it. Happy to open the PR if that is welcome.

Found by a local sweep harness; everything above was reproduced with the native CLI as shown.

JakeATX:
## Summary

Fixes #2228 by making the CLI and high-level generation API honor sampling defaults declared by a checkpoint's `generation_config.json`.

## Changes

- Resolve omitted `temperature`, `top_p`, and `top_k` values from `model.config`.
- Preserve explicit overrides, including zero-valued greedy/top-k/top-p settings.
- Treat `do_sample: false` as greedy decoding when temperature is omitted.
- Share the resolver with server request normalization so the entry points use the same precedence rules.
- Keep diffusion generation's existing defaults and explicit override behavior.
- Add regression coverage for CLI deferral, model-config fallback, explicit overrides, high-level autoregressive forwarding, and diffusion compatibility.

## Validation

- `python3 -m compileall -q` on all changed Python modules and tests
- Black, isort, autoflake, and `git diff --check`
- The repository pytest suite could not run in this Linux checkout because MLX and pytest are not installed; the added tests are included for the project's normal macOS test environment.


Blaizzy:
Thanks @JakeATX for you awesome contribution!

This is a real problem that's on the backlog, I just think the approach could be better and more general.

We already have the model generation config which usually contains this info.

Let me make a PR building on this one and add you there as a co-author.

PR #2299, opened 2026-09-18
qwen3_vl_moe: feed video pixels through the vision tower

### LINKED ISSUE #2297: qwen3_vl_moe: video understanding still broken after #2142 (empty output / degenerate repetition)
## Environment
- mlx-vlm: 0.7.1 (also reproduced a related-but-different symptom on 0.6.17, pre-#2142)
- mlx: 0.32.2 / mlx-metal: 0.32.2
- Model: `mlx-community/Qwen3-VL-30B-A3B-Instruct-4bit` (`qwen3_vl_moe`)
- Hardware: Apple M5 Max, macOS
- Server: `mlx_vlm.server`, `backend=continuous_batching`

## Problem
After upgrading past #2142 (per-group video timestamps), the `qwen3_vl_moe` (MoE) variant still cannot correctly understand video input, even though the same install's single-image path and Qwen3-Omni's video path both work correctly. Three distinct failure modes observed, one per test:

1. **mlx-vlm 0.6.17 (pre-#2142), real ~2 min video (1080x1920, h264)**: model claims "there is nothing in the frame, no one is speaking" (`画面中没有出现任何场景和物体，也没有人说话或有旁白。`), despite `prompt_tokens` confirming ~11k vision tokens were actually produced from the video. The **same content**, sent as a single still frame extracted from the video (`image_url`), is described correctly and in rich detail by the same model/server.
2. **mlx-vlm 0.7.1 (post-#2142), same real video**: generation now stops after exactly 1 token (empty completion, `finish_reason=stop`). Prefill completes normally (12156 prompt tokens, 1783 tok/s, frame sampling log shows `sampled 242 of 3655 frames at 1.99 fps`).
3. **mlx-vlm 0.7.1, small synthetic 6s video** (180 frames, 320x240, a red square sliding left→right, with the text `42` appearing in the last second — minimal repro below): the model correctly identifies the square's movement direction, then degenerates into an infinite repetition loop (`788888888888...`) until hitting `max_tokens`, never mentioning the `42` text that appears later in the clip.

**Control tests that work correctly on the same 0.7.1 install:**
- Single-image understanding (`image_url`) with this same `qwen3_vl_moe` model: fully correct and detailed.
- `Qwen3-Omni-30B-A3B-Instruct-4bit` (`qwen3_omni_moe`) given the **same real video** via `video_url`: fully correct, detailed, accurate description — both before and after the mlx-vlm upgrade.

This suggests #2142's fix was validated against the dense `qwen3_5`/`qwen3_vl` checkpoint (per the PR description's own validation model) but leaves a residual issue specific to the `qwen3_vl_moe` variant. The degenerate-repetition symptom (#3) looks structurally similar to the previously-fixed #1619 ("Qwen3-Omni: decode steps lose the multimodal RoPE delta — generation degenerates instantly"), which makes me suspect a related positional/RoPE-delta tracking gap during decode that #2142 didn't cover for the MoE routing path.

## Minimal reproduction (synthetic, no external files needed)

Generate a tiny test clip:
```python
from PIL import Image, ImageDraw
import os
os.makedirs("frames", exist_ok=True)
W, H = 320, 240
for i in range(180):  # 6s @ 30fps
    img = Image.new("RGB", (W, H), (240, 240, 240))
    d = ImageDraw.Draw(img)
    x = int(20 + i / 180 * (W - 60))
    d.rectangle([x, H // 2 - 20, x + 40, H // 2 + 20], fill=(220, 30, 30))
    if i > 150:
        d.text((W // 2 - 15, H - 30), "42", fill=(0, 0, 0))
    img.save(f"frames/f{i:04d}.png")
```
```bash
ffmpeg -y -framerate 30 -i frames/f%04d.png -c:v libx264 -pix_fmt yuv420p synthetic_test.mp4
mlx_vlm.server --model mlx-community/Qwen3-VL-30B-A3B-Instruct-4bit --port 18002
```

Request:
```python
import json, urllib.request
body = {
    "model": "mlx-community/Qwen3-VL-30B-A3B-Instruct-4bit",
    "messages": [{"role": "user", "content": [
        {"type": "text", "text": "Describe what happens in this video: how does the square move, and what text (if any) appears at the end?"},
        {"type": "video_url", "video_url": {"url": "file:///absolute/path/to/synthetic_test.mp4"}},
    ]}],
    "max_tokens": 300, "temperature": 0.2, "stream": False,
}
req = urllib.request.Request("http://localhost:18002/v1/chat/completions", data=json.dumps(body).encode(),
    headers={"Content-Type": "application/json"})
print(json.load(urllib.request.urlopen(req, timeout=120))["choices"][0]["message"]["content"])
```

**Observed output (0.7.1):**
> 视频中，一个白色的方块在黑色的背景上从左向右移动。方块移动时，其右侧出现了一行白色的数字"78888888888888888888..."（一直重复到 max_tokens）
>
> *(Translation: "In the video, a white square moves left to right on a black background. As it moves, a line of white digits appears on its right: '788888...' (repeats until max_tokens)." — direction of movement is correct, but color/background are misread and it never reaches the "42" text due to the repetition loop.)*

## Additional regression noticed on the same 0.7.1 install
`qwen3_vl_moe`'s APC self-check now fails at model load (worked as `mode=exact` on 0.6.17):
```
APC self-check failed mode=None schema=v4 layers=0 unsupported=['model has no make_cache']
```
Not the main focus of this issue, but flagging since it appeared in the same upgrade and may share a root cause (an interface change for this model class that both APC's cache-introspection and video decode depend on).

## Versions
- mlx-vlm 0.7.1
- mlx 0.32.2 / mlx-metal 0.32.2
- macOS, Apple M5 Max, Python 3.12

Lazarus-931:
qwen3_vl_moe couldn't see video inputs since it never read `pixel_values_videos`, and in this pr, i aliases `pixel_values_videos` onto `pixel_values` such that videos go through vision tower.

this pr also ports the dense model's column-indexed DeepStack slicing, since the MoE's contiguous slice mismatched on multi-row batches

i verified by running the issue's repro on the real 30B: broken output flipped to correct, with image/text/dense controls byte-identical.

closes #2297


lucasnewman:
@Lazarus-931 We need to clear `pixel_values_videos` for prefix cache reuse in this case, like we do for `pixel_values`.

Lazarus-931:
> @Lazarus-931 We need to clear `pixel_values_videos` for prefix cache reuse in this case, like we do for `pixel_values`.

done!

lucasnewman (approved):
Looks good!

PR #2301, opened 2026-09-18
Support current moondream2 checkpoints and moondream3 mlx quants

Lazarus-931:
## What

`vikhyatk/moondream2`'s current checkpoints (and moondream3 mlx quants) didn't load in mlx-vlm. This makes both work as proper ports.

### moondream2
The checkpoints self-report `model_type: "moondream1"` (the remap fixes routing), but loading also needed:
- **Weight layout** — the current checkpoints use the `model.{text,vision,region}.*` layout (same architecture family as moondream3, not the old Phi-based one). `sanitize` now maps it onto the port (region head dropped).
- **Tokenizer** — the checkpoint bundles a stale GPT-2 `tokenizer.json` the model doesn't actually use; its own code loads `moondream/starmie-v1`. We force that tokenizer by reusing the moondream3 processor.
- **bf16** — the prefix attention mask is cast to the compute dtype so the image path runs under bf16.

### moondream3
`sanitize` wasn't idempotent, so already-converted mlx quants (e.g. `beshkenadze/moondream3-preview-mlx-4bit`, an `mlx_vlm.convert` output) doubled the `text.model` / `vision.encoder` prefixes and failed to load. Now guarded so both raw and pre-converted checkpoints load.

## Verification

I verified it with the images in examples/

```bash
# moondream2 (current default) — caption + VQA
python -m mlx_vlm.generate --model vikhyatk/moondream2 \
  --image examples/images/cats.jpg --prompt "Describe this image." --max-tokens 80
# -> Two cats are sleeping on a pink blanket on a couch. ... Both cats have striped fur ...

python -m mlx_vlm.generate --model vikhyatk/moondream2 \
  --image examples/images/cats.jpg \
  --prompt "How many cats are in this image and what color is the blanket?"
# -> There are two cats in this image, and the blanket is pink.

# moondream3 (mlx 4-bit quant)
python -m mlx_vlm.generate --model beshkenadze/moondream3-preview-mlx-4bit \
  --image examples/images/cats.jpg --prompt "Describe this image."
# -> Two tabby cats lie curled up on a pink couch. ... Two remote controls rest ...
```

Tests: `pytest mlx_vlm/tests/test_models.py -k Moondream` (sanitize tests for both) and the full `test_models.py` pass.


lucasnewman (approved):
Nice!

PR #2314, opened 2026-09-19
Define make_cache for qwen2 and qwen3 so APC engages for dense Qwen models

### LINKED ISSUE #2312: Qwen2/Qwen3 (dense) don't define `make_cache`, so APC silently self-disables and re-prefills the whole prompt every turn
## Description

The APC self-check needs the loaded model (or its `language_model`) to expose a `make_cache` method. When it's missing, `model_apc_mode()` returns `None` and APC skips prefix caching for that model, with nothing logged beyond the self-check line.

The dense Qwen families don't have `make_cache`: neither `Model` nor `LanguageModel` in `mlx_vlm/models/qwen2/language.py` or `mlx_vlm/models/qwen3/language.py` defines it. So APC never engages for Qwen2/Qwen2.5 and Qwen3 dense checkpoints. The self-check reports `model has no make_cache` and `mode=None`, and every turn of a multi-turn conversation re-prefills from scratch, even when it shares a long prefix with the previous turn.

These are ordinary full-attention decoders, the same shape as `llama` (they update a standard KV cache through `cache.update_and_fetch` / `cache.offset`). Their MoE and newer siblings (`qwen2_moe`, `qwen3_5`, `qwen3_moe`, `qwen3_next`) do define `make_cache`.

## Location in Source

*(line numbers from `mlx-vlm==0.6.17`; unchanged on `main`.)*

The gate, in `mlx_vlm/apc.py` (`self_check_model_apc()` / `model_apc_mode()`):

```python
language_model = model.language_model if hasattr(model, "language_model") else model
if not hasattr(language_model, "make_cache"):
    # -> APCSelfCheckResult(ok=False, apc_mode=None, ... "model has no make_cache")
```

`mlx_vlm/models/llama/language.py` is the working reference: `Model.make_cache` builds one cache per layer, and `LanguageModel.make_cache` delegates to it.

In `mlx_vlm/models/qwen2/language.py` and `mlx_vlm/models/qwen3/language.py`, `LanguageModel` delegates `sanitize` and `shard` to `Model` but has no `make_cache`, and `Model` doesn't define one either.

## Reproduction

Load any Qwen2.5/Qwen3 dense checkpoint through the server (for example `mlx-community/Qwen3-0.6B-4bit`) with `APC_ENABLED=1 APC_TRACE=1`:

```
APC self-check failed mode=None schema=v1 layers=0 unsupported=['model has no make_cache']
APC_TRACE self_check ok=False mode=None schema_version=1 layers=0 unsupported=0
```

Or directly:

```python
from mlx_vlm import load
model, _ = load("mlx-community/Qwen3-0.6B-4bit")
lm = model.language_model
print(type(lm).__module__, hasattr(lm, "make_cache"))
# mlx_vlm.models.qwen3.language False
```

For contrast, `mlx-community/Llama-3.2-1B-Instruct-4bit` logs `APC self-check ok mode=block ... layers=16 types=[KVCache, ...]`.

## Environment

- `mlx-vlm` 0.6.17 (also present on `main` / 0.7.1)
- `mlx` 0.32.2
- Apple M4, macOS

rayanhanader:
Fixes #2312.

## Problem

Dense Qwen2/Qwen2.5 and Qwen3 define no `make_cache`, in either `Model` or `LanguageModel`. The APC self-check resolves to `mode=None` for them, so prefix caching never engages and every turn re-prefills the whole prompt. Their MoE and newer siblings (`qwen2_moe`, `qwen3_5`, `qwen3_moe`, `qwen3_next`) already define it.

## Fix

`qwen2` and `qwen3` are full-attention decoders with the same shape as `llama`, and neither uses a sliding window, so a plain `KVCache` per layer is the right block-mode layout. Following `llama`, `Model.make_cache` returns one `KVCache` per layer and `LanguageModel.make_cache` delegates to it.

## Tests

`qwen2` and `qwen3` are added to the `dense` cases in `model_cases.json`. `test_dense_model` runs `forward_cache`, which calls `make_cache` and decodes a token against the returned cache, so the method is exercised and a regression fails the suite. No dense Qwen module was covered by a `forward_cache` case before, which is how the gap went unnoticed.

Before the fix:

```
FAILED mlx_vlm/tests/test_models.py::test_dense_model[qwen2] - AttributeError: 'LanguageModel' object has no attribute 'make_cache'
```

After:

```
$ python -m pytest -q mlx_vlm/tests/test_models.py
93 passed, 2 subtests passed
```

## Verification on a real model

On `main` with this change, `mlx-community/Qwen3-0.6B-4bit` goes from `APC self-check failed mode=None ... 'model has no make_cache'` to `APC self-check ok mode=block layers=28 types=[KVCache, ...]`.

Prefix reuse then behaves like an already-supported model on the same build: sending the same 85-token prompt twice, `cached_tokens` goes from 0 to 80 on the second request (`Llama-3.2-1B-Instruct-4bit` goes from 0 to 96 on a 104-token prompt).


rayanhanader:
Withdrawing: the problem this addressed (#2312) is not a bug. APC already engages for dense Qwen2/Qwen3 via the runtime fallback (`model_apc_mode` treats an absent `make_cache` as block-eligible and the per-request builder makes a per-layer KV cache), so this change is a no-op for prefix caching. Sorry for the noise.

PR #2318, opened 2026-09-20
apc: report an unreachable prefill reserve instead of failing silently

peterplaykkk:
## Summary

`APCManager._make_room` required the advisory prefill reserve verbatim. When that
reserve exceeds `memory_max_bytes` plus the device headroom, no amount of eviction can
satisfy it, so `_make_room` returns `False` for every allocation and all store/restore
paths are skipped. Declining admission there is the intended policy — going quiet is
not: `memory_skips` keeps climbing while `exact_stores` and `disk_writes` stay at zero
and nothing is logged, so the prefix cache looks broken with no signal to point at.

This keeps the default policy intact and makes the state visible:

- `_make_room` is split into `_make_room_for(required, retain_bytes)` plus a guard, so
  the unsatisfiable case never starts eviction work it cannot satisfy.
- Unsatisfiable reserves are counted (`reserve_unreachable`, `last_reserve_bytes`) with
  a one-time warning that names the escape hatch.
- `APC_RELAX_UNREACHABLE_RESERVE=1` lets operators bill allocations against the real
  requirement instead (`reserve_relaxed`).

## How it showed up

27B hybrid model (`qwen3_5`), 48 GB Mac, `APC_ENABLED=1` with the disk tier:

- `_prefill_reserve_bytes` was left at **173 GB** by a stale per-token estimate.
- Every 21.8k-token request re-prefilled from scratch: ttft 33.3 s, prefill 657 tok/s,
  repeated on identical prompts.
- `GET /v1/cache/stats` at that moment: `memory_skips=22`, `exact_stores=0`,
  `disk_writes=0`, `disk_hits=1` — stores and restores were skipped, no error anywhere.

After a restart (reserve back to ~1.8 GB) the identical prompt hit the cache in 0.52 s,
which pins the reserve as the cause. Note the estimator that produced that particular
number lives in the pre-`PrefillMemoryPlan` code and is already replaced on `main`; this
PR is about the residual case — a legitimately large reserve (long-context request) or a
stale fallback estimate can still exceed budget + headroom, and today that state is
indistinguishable from a broken cache.

## Testing

- `cd mlx_vlm && pytest -s ./tests/test_cache.py` — 319 passed.
- Three new tests: reporting (`test_unreachable_prefill_reserve_is_reported`), opt-in
  relaxation (`test_unreachable_prefill_reserve_can_be_relaxed`) and the verbatim path
  (`test_achievable_prefill_reserve_is_kept_verbatim`); all three fail on the previous
  code.
- `pre-commit run --all-files` — clean.


Blaizzy:
Thanks @peterplaykkk!

Please open a GH issue with reproducible example!

PR #2320, opened 2026-09-20
fix(server): return 400 when a chat/omni model is sent to /v1/audio/*

### LINKED ISSUE #2319: /v1/audio/transcriptions crashes with cryptic TypeError when pointed at a chat/omni model (e.g. Qwen3-Omni)
## Summary

`POST /v1/audio/transcriptions` (and presumably `/v1/audio/translations`) crashes with a confusing `TypeError` when the `model` field points at a chat/omni multimodal model this server already loads natively for `/v1/chat/completions` (e.g. a Qwen3-Omni checkpoint) — instead of a dedicated STT-only model.

## Reproduction

Start the server with a Qwen3-Omni checkpoint (any variant; I used `mlx-community/Qwen3-Omni-30B-A3B-Instruct-4bit`), then POST a real WAV file to `/v1/audio/transcriptions` with `model` set to that same checkpoint's path — the same path that works fine as the `model` in `/v1/chat/completions` requests to this same server.

Result: HTTP 500, with a detail message shaped like:

> Failed to load audio model: ModelConfig.__init__() missing 11 required positional arguments: 'hidden_size', 'num_hidden_layers', 'intermediate_size', 'num_attention_heads', 'rms_norm_eps', 'vocab_size', 'num_key_value_heads', 'max_position_embeddings', 'rope_theta', 'head_dim', and 'tie_word_embeddings'

Reproduces on the latest `main` (I tested against `e79b0e0`), not just the last PyPI release.

## Root cause

`load_audio_model()` (in `mlx_vlm/server/app.py`) unconditionally hands whatever `model` path it's given to `mlx_audio.utils.load_model()`. That function auto-detects a model *category* (tts/stt/sts/lid/vad) from the checkpoint's `config.json` and then constructs a category-specific, flat `ModelConfig` dataclass directly from that top-level dict.

For a Qwen3-Omni-style checkpoint, the actual transformer hyperparameters (`hidden_size`, `num_hidden_layers`, etc.) live nested under `thinker_config.text_config`, not at the top level of `config.json`. Whatever category `mlx_audio` guesses for this config, its `ModelConfig(**config)`-style construction can't find those keys at the top level, so the constructor call fails with the "missing N required positional arguments" `TypeError` — which then gets wrapped into a generic 500 by the server's catch-all exception handler, with no indication of what actually went wrong.

This isn't really a "wrong values" bug so much as a **model-category mismatch**: `mlx_audio`'s loader was never meant to understand a Qwen3-Omni-shaped config in the first place, and nothing upstream of it checks that before attempting the load.

## Suggested fix (verified locally, description only — no diff attached per request)

Add a short guard at the top of `load_audio_model()`, before it delegates to `mlx_audio`: resolve the given `model_path` the same way the rest of the server already does (via the existing `get_model_path` / `load_config` helpers in `mlx_vlm/utils.py`), read `model_type` out of the resulting config dict, and check whether that `model_type` corresponds to one of the model families this server *already* implements natively under `mlx_vlm/models/` (a plain "does a subdirectory with this name exist" check is enough — no new registry needed). If it does, raise the same kind of `ValueError` the function already implicitly relies on elsewhere in this file to get mapped to a clean `400 Unsupported audio model` response (the server's existing `except (FileNotFoundError, ValueError)` branch around this call already does that mapping) — with a message telling the caller this is a chat/multimodal model with native audio support and pointing them at `/v1/chat/completions` with an `input_audio` content part instead of this endpoint.

I implemented and tested exactly this locally: the same request that used to 500 with the cryptic `TypeError` now returns a `400` with a one-sentence explanation and the correct redirect, and a follow-up sanity check confirmed the normal `/v1/chat/completions` + image path on the same server instance is completely unaffected.

## Possible follow-up (not attempted, just flagging as a direction)

The above only makes the failure legible; it doesn't make transcription actually work for an omni-capable model through this endpoint. A further improvement worth considering: when `/v1/audio/transcriptions` is pointed at a model the server recognizes as natively audio-capable through the chat path, transparently synthesize an internal `/v1/chat/completions` request (audio content part + a transcription-style prompt) instead of rejecting outright, so callers get one consistent Whisper-style API regardless of whether the backing model is a dedicated STT model or a chat-capable omni model. That's a materially bigger change (needs to build the right prompt, map the chat response back into the transcription response shape, decide how `response_format`/timestamps degrade, etc.), so I'd treat it as a separate, optional enhancement rather than part of this bug fix.

## Environment

- mlx-vlm: tested on both the last PyPI release (0.7.1) and `main` @ `e79b0e0`
- Model: `mlx-community/Qwen3-Omni-30B-A3B-Instruct-4bit`
- macOS, Apple Silicon

Lazarus-931:
fixes #2319 

where, mlx_audio.load_model wasn't validating that the model it's given is an audio model, it just tries to route it to one of its five sub-loaders (tts/stt/sts/lid/vad).

because of this, user described the following errors:

```
Failed to load audio model: ModelConfig.__init__() missing 11 required
positional arguments: 'hidden_size', 'num_hidden_layers', ...
```

this pr adds a gate at the top of load_audio_model. Before delegating to mlx_audio, decide whether this model even belongs on an audio endpoint

Verified across all 220 native model types (217 now redirect cleanly, 3 backbone-name overlaps pass through, 0 crashes), 27 real audio types (all pass through), 6 real downloaded configs, edge cases, and the full server test suite (224 passing)


lucasnewman (approved):
Looks good!

PR #2322, opened 2026-09-20
fix: count only emitted speculative drafts in draft_n_accepted

ykhrustalev:
**Problem**
Every speculative round loop recorded the matched-draft count before the consumer applied the stop token or `max_tokens` budget, so drafts that were never emitted still counted toward `draft_n_accepted`. `predicted_n - draft_n_accepted` (the llama.cpp `timings` forward-count convention) went negative on short replies and understated forwards on ordinary ones.

**Solution**
- Adds `_emitted_speculative_round`, which cuts a round at the first `stop_check`/`eos_token_ids` hit or the budget and caps the recorded accepted count
- Routes all six round loops (dflash, eagle3, mtp; single-row and batch) through it for the lifetime counter
- Adds a `remaining_tokens` callback so batch loops honour the consumer's per-row budget and rows it already finished
- Averages the eagle3/mtp batch count over live rows only, so a finished row cannot dilute a running request
- Keeps `accept_lens`, cache commits and emission on the matched count, so text output and the adaptive block-size controllers are unchanged

**Testing**
- New unit tests in `mlx_vlm/tests/test_speculative.py` cover stop-in-drafts, stop-as-bonus, budget cut, first-draft rejection, per-row batch stops, per-row budgets and prefill-finished rows
- Streamed prompts against a target/draft pair at `--draft-block-size 8`: 15/16 violated `draft_n_accepted <= predicted_n - 1` before, 0 after, with identical text and unchanged `predicted_n`/`draft_rounds`/`draft_n`


ykhrustalev:
Closing this. The problem it addresses is real: `draft_n_accepted` counts verifier matches, including drafts discarded at the stop token, so it can exceed `predicted_n - 1`. But the change is the wrong size for it. `timings.draft_rounds` already gives consumers the exact verification count, and this PR corrects `draft_n_accepted` without trimming `draft_n` for the final round. That moves the bias in `accepted / drafted` from slightly high to slightly low, and it does not remove it. I will follow up with a short README note that documents what the three draft counters mean and points consumers at `draft_rounds` for accept length.

PR #2325, opened 2026-09-21
Add direct MLX inference for Decider-2b - [WIP]

Lazarus-931:
Adds support for decider-2b[decider-2b](url), which is a system one model that given a state and arguments, returns a calibrated probability in one forward pass.

This re-uses the qwen3_5 for the backbone, with a new api to allow these class of families to run, which are a `choice`, `score`, and `noul` decisions through a dedicated `predict(state, questions)` path


mlx_vlm/tests/test_models.py
```diff
@@ -1546,3 +1546,38 @@ def test_raw_keys_are_remapped(self):
...
+            "lm_head.weight": mx.zeros((2, 2)),
+        }
+        mapped = _mapped_weights(weights)
+        self.assertEqual(len(mapped), 2)
+        np.testing.assert_array_equal(
+            np.asarray(mapped["model.layers.0.input_layernorm.weight"]), [1, 1]
+        )
+        self.assertEqual(
+            mapped["model.layers.0.linear_attn.conv1d.weight"].shape, (2, 4, 3)
+        )
```
Blaizzy: We don't need this.

mlx_vlm/tests/test_models.py
```diff
@@ -1546,3 +1546,38 @@ def test_raw_keys_are_remapped(self):
...
+        question = _render_question(
+            {
+                "type": "choice",
+                "instructions": "Route this request",
+                "criteria": {"billing": "payments", "technical": None},
+            }
+        )
+        self.assertEqual(question["options"], ["billing: payments", "technical"])
+        records = _annotate([{"value": i} for i in range(8)])
+        self.assertEqual(records[7], {"_index": 7, "value": 7})
```
Blaizzy: I would make this a general test for system one models and add this model in the `model_cases.json` with the workflow type name for the test.

Blaizzy (commented):
Left a few comments

mlx_vlm/models/decider2/README.md
```diff
@@ -0,0 +1,26 @@
+# Decider-2b decisions on MLX
+
+Load the original `convaiinnovations/decider-2b` checkpoint directly. It uses the existing Qwen3.5 text backbone and scores only the allowed option tokens at answer slots. No converted checkpoint is required.
+
+```python
+from mlx_vlm.models.decider2 import load
+
+model = load("convaiinnovations/decider-2b")
```
lucasnewman: Could we instead route this through `load_model` so we can just use the standard sanitization instead of a custom weight mapping?

Blaizzy (changes requested):
Use general loader as @lucasnewman suggested

PR #2327, opened 2026-09-21
MiMo-V2.6-Flash-RL support [WIP]

Lazarus-931:
Adds support for [MiMo-V2.6-Flash-RL](https://huggingface.co/XiaomiMiMo/MiMo-V2.6-Flash-RL).

[4bit](https://huggingface.co/nativ-community/MiMo-V2.6-Flash-RL-MLX-4bit)

M3 Ultra 512 GB · 4.361 bpw · 32 decode steps

**prefill tok/s**

| ctx | b1 | b2 | b4 | b8 |
|---|---|---|---|---|
| 512 | 595 | 614 | 607 | 650 |
| 2048 | 565 | 586 | 594 | 591 |
| 4096 | 502 | 502 | 500 | 501 |
| 8192 | 382 | 382 | 374 | 324 |
| 16384 | 249 | 194 | 123 | oom |

**decode tok/s**

| ctx | b1 | b2 | b4 | b8 |
|---|---|---|---|---|
| 512 | 43.6 | 1.4 | 108.6 | 169.1 |
| 2048 | 42.9 | 71.2 | 110.9 | 162.5 |
| 4096 | 42.4 | 71.2 | 107.4 | 157.2 |
| 8192 | 41.2 | 68.3 | 103.2 | 3.9 |
| 16384 | 0.4 | 1.3 | 1.7 | oom |

**peak GB**

| ctx | b1 | b2 | b4 | b8 |
|---|---|---|---|---|
| 512 | 170 | 170 | 171 | 173 |
| 2048 | 171 | 173 | 177 | 184 |
| 4096 | 175 | 181 | 193 | 217 |
| 8192 | 190 | 210 | 251 | 331 |
| 16384 | 244 | 320 | 469 | oom |


mlx_vlm/tests/test_models.py
```diff
@@ -1159,6 +1159,113 @@ def test_patch_embed_is_transposed_from_ncdhw_to_ndhwc():
...
+        key.startswith(("language_model.", "vision_tower.")) for key in sanitized
+    )
+
+    model.load_weights(list(sanitized.items()), strict=True)
+    mx.eval(model.parameters())
+
+    # already-prefixed checkpoints must survive a second pass unchanged
+    assert model.sanitize(dict(sanitized)) == sanitized
+
+
```
Blaizzy: Please revert this.

mlx_vlm/fp8.py
Blaizzy, line 1: Is this really needed for it to work?

mlx_vlm/tests/test_models.py
Lazarus-931: Done in f0863ee — removed the fixture and its helper. Model coverage now comes from the `model_cases.json` case alone.

The one test I kept is much smaller and pins the tensor-parallel qkv layout, which the JSON contract has no way to express. It is the regression guard for the bug this PR fixes, and I verified it fails against the previous contiguous read.

mlx_vlm/fp8.py
Lazarus-931: Yes — without it the conversion aborts:

```
ValueError: Fine-grained FP8 scale shape does not match its weight:
weight=(13568, 4096), scales=(108, 32), expected=(106, 32).
```

`transform_fp8_weights()` rewrites every `.weight_scale_inv` in the checkpoint, which assumes the whole checkpoint is block-FP8. MiMo-V2.6 declares `quant_method: "fp8"` but only its dense projections are; the routed experts (98% of the parameters) are MXFP4, flagged by `store_dtype`.

The 108-vs-106 mismatch is because the fused qkv ships as four tensor-parallel shards, each with its own grid padded to whole 128-row blocks (4 x 27 = 108), so it cannot be read as one grid at all. The model handles that in its own `sanitize`, which is what `fp8.py`'s docstring says unsupported layouts should do.

Checkpoints without `store_dtype` are unaffected — the default keeps them on the existing path.

Lazarus-931: yep!

lucasnewman (approved):

PR #2328, opened 2026-09-22
Release generation batch arrays without waiting for cyclic GC

### LINKED ISSUE #2310: GenerationBatch._eval_pending_state leaks whole KV caches via a self-referencing closure (64 GB peak at batch 32 instead of 11 GB)
## Summary

`GenerationBatch._eval_pending_state` (`mlx_vlm/generate/ar.py`, 0.7.1 and current `main`) builds a nested function `append_arrays` that calls itself recursively. Because the function refers to itself, its closure has a cell for `append_arrays` **and** a cell for `targets`, so `function → cell → function` is a reference cycle, and `targets` (a list holding every KV-cache array of the batch) hangs off that cycle. Refcounting cannot free it; the whole batch's KV cache stays allocated until CPython's cyclic GC happens to run.

`_eval_pending_state` is called from `extend()` (every time new sequences are merged into a running batch) and from `filter()` (every time a sequence finishes), so under batched generation the stale caches pile up between collections.

## Impact (real workload)

Batched OCR with `BatchGenerator` on a DeepSeek-OCR-derived model (jina-ocr-v1, the same setup as #2300), M5 Max 128 GB, `mx.get_peak_memory()`:

| batch | prefill_batch | peak GB (0.7.1 as shipped) | peak GB (with `gc.collect(0)` on batch-membership change) |
|---|---|---|---|
| 8 | 8 | 17.1 | 10.6 |
| 8 | 4 | 30.9 | 11.2 |
| 16 | 16 | 32.9 | 9.5 |
| 16 | 1 | 52.0 | 11.0 |
| 32 | 32 | 64.1 | 11.2 |

The real KV cache for these runs was at most ~2 GB (61 KB/token). A per-step trace showed active memory climbing to ~41 GB and then dropping abruptly to ~7 GB whenever the cyclic GC fired. Smaller `prefill_batch_size` (continuous refill) makes it *worse* because `extend()` is called more often.

## Model-free reproduction

Only `mlx` and `mlx-vlm` are needed; it calls the real method on a minimal stand-in object.

```python
import gc, types, weakref
import mlx.core as mx

class FakeCache:
    def __init__(self, k, v): self.state = (k, v)

class FakeBatch:  # only the attributes _eval_pending_state touches
    _current_tokens = _current_lps = _next_tokens = _next_lps = None
    _next_top_idx = _next_top_lp = _rope_deltas = None
    def __init__(self, caches): self.prompt_cache = caches

def make_kv(n_layers=12, batch=16, kv_heads=10, length=4096, head_dim=128):
    caches = []
    for _ in range(n_layers):
        k = mx.zeros((batch, kv_heads, length, head_dim), dtype=mx.bfloat16)
        v = mx.zeros((batch, kv_heads, length, head_dim), dtype=mx.bfloat16)
        mx.eval(k, v); caches.append(FakeCache(k, v))
    return caches

gc.disable()  # an automatic collection must not hide the cycle
gb = lambda: mx.get_active_memory() / 2**30

caches = make_kv()
print(f"after allocating KV cache:              {gb():5.2f} GB active")
GenerationBatch._eval_pending_state(FakeBatch(caches))
del caches
print(f"after _eval_pending_state + del caches:  {gb():5.2f} GB active")

stale = [o for o in gc.get_objects() if isinstance(o, mx.array) and o.ndim == 4]
fn = [f for f in gc.get_objects() if type(f) is types.FunctionType and
      f.__qualname__ == "GenerationBatch._eval_pending_state.<locals>.append_arrays"]
print(f"{len(stale)} KV arrays still reachable; append_arrays closures alive: {len(fn)}, "
      f"free vars: {fn[0].__code__.co_freevars}")
del stale, fn
gc.collect()
print(f"after gc.collect():                      {gb():5.2f} GB active")
```

Output on 0.7.1:

```
after allocating KV cache:               3.75 GB active
after _eval_pending_state + del caches:   3.75 GB active
24 KV arrays still reachable; append_arrays closures alive: 1, free vars: ('append_arrays', 'targets')
after gc.collect():                       0.00 GB active
```

## Fix

Remove the self-reference. Any of these works; the first is the smallest change:

```python
def _eval_pending_state(self):
    """Materialize lazy decode outputs before mutating batch-owned state."""
    targets = []
    stack = [(self._current_tokens, self._current_lps, self._next_tokens, self._next_lps,
              self._next_top_idx, self._next_top_lp, self._rope_deltas)]
    for c in self.prompt_cache:
        try:
            stack.append(c.state)
        except (AttributeError, TypeError):
            pass
    while stack:
        value = stack.pop()
        if isinstance(value, mx.array):
            targets.append(value)
        elif isinstance(value, (list, tuple)):
            stack.extend(value)
    if targets:
        mx.eval(*targets)
```

Alternatives: hoist `append_arrays` to a module-level helper that takes `targets` as an argument, or keep the closure but end the method with `targets.clear()` / `del append_arrays` (fragile). With the iterative version, the same repro prints `0.00 GB active` immediately after `del caches`, with GC still disabled.

`SpeculativeGenerationBatch` shares the method via inheritance, so it is covered by the same change.

## Workaround for users

Call `gc.collect(0)` after each `BatchGenerator.insert()`/`remove()` cycle (i.e. whenever batch membership changes). That is what brought peak memory from 64 GB to 11 GB at batch 32 in the table above.

## Environment

- mlx-vlm 0.7.1 (also present on `main` as of 2026-09-19), mlx 0.32.2
- Python 3.14.6, macOS 27.0, Apple M5 Max 128 GB

Happy to open a PR with the iterative version and a regression test (the repro above, asserting 0 bytes active after `del`) if that is welcome.

aprylewu:
`GenerationBatch._eval_pending_state()` uses a recursive local function whose closure retains the array targets. After batch membership changes, those pending outputs and KV-cache arrays remain alive until cyclic GC runs. Traverse nested cache state with a stack so materializing the arrays does not extend their lifetime.

The regression disables cyclic GC and checks that `filter([])` immediately releases every pending output and nested cache array. A second test checks that all nested arrays still reach `mx.eval`. Both full generation and cache test modules pass: 431 tests on MLX 0.32.2. The lifetime regression fails on the base. No model weights or full-model inference were needed for this validation.

Fixes #2310.

AI assistance: Codex assisted with implementation and regression testing.


lucasnewman (approved):
@aprylewu Good find, thank you!

